Issue #178, opened 2018-11-02
Migration of existing PhysioBank projects


tompollard:
Existing PhysioBank projects will need to be migrated to the new system. We will need to find a way of populating all of the mandatory metadata fields (or alternatively create a project model for historical content and accept that some information is missing). 

List of PhysioBank projects at: https://physionet.org/physiobank/database/DBS

List of PhysioNetWorks projects at: https://physionet.org/works/

tompollard:
To decide how to handle each of the PhysioNetWorks projects, it would be useful to collate the following details:

- Project name
- Date created
- Number of approved collaborators
- Date last accessed by a collaborator
- Date when a collaborator was last added.
- Email address of the person/people responsible for adding collaborators
- Number of users who have requested access and who have not received a response.

Is there a quick way of gathering these details?

elfeto:
So, for the things mentioned above:
- The Project name is easy to get. 
- If the Works project was created after January 15, 2017, then we have the creation and activation date. 
- If the project has the waiting for access, then is easy to get the number of people waiting for access.
- Number of approved collaborators  ->(only collaborators?  or also reviewers ?)
- Not possible ... -> Date last accessed by a collaborator
- Not possible ... -> Date when a collaborator was last added.
- Not possible ... -> Email address of the person/people responsible for adding collaborators

We should ignore the Works for now, and ask the people that have the works projects, to create it as a new project in the new version of physionet. This is because we wont hold closed access to X projects, and everything WILL be published under a DUA or no DUA. 

We should move the PhysioBank and PhysioTools just AS IS, and add a "Legacy" tag.

tompollard:
Thanks, for "Number of approved collaborators", you are right, I mean collaborators and reviewers. 

> We should move the PhysioBank and PhysioTools just AS IS, and add a "Legacy" tag.

Think we have a plan for this now :) I'll add a few issues for points that came up in the conversation.

cx1111:
Going to make a list of dbs to migrate. I'll comment the differences from the [DBS](https://physionet.org/physiobank/database/DBS) file here.

- `gait-maturation-db` has 2 pages to be combined manually. Original data is in `data` subdir.
- `meditation` has 2 pages to be combined manually. Original data is in `data` subdir.
- `synthetic/tns/` should just be `tns` (and added to DBS)
- merge mimicdb and mimicdb/numerics
- skip mimic2cdb for now
- skip mimicwdb for now
[...]

PR #213, opened 2019-02-07
Add code and data for migrating physiobank projects

cx1111:
We have an issue here tracking migration of content: https://github.com/MIT-LCP/physionet-build/issues/178

This branch does:
- Add an intermediate model called LegacyProject to help with migrating pbank databases
- The DBS.tsv file is my edited version of DBS in the physiobank root directory. I got the pulication date of each database if available from the news, otherwise tried to get an approximate date.
- Small changes with displaying projects in publish order
- Add a license and news item to be used in production site
- Added 93 database metadata to LegacyProject objects. Edited and structured them manually. You can see and edit their content from the admin panel. Put the data into fixtures. If you want to load them locally, call `./manage.py loaddata pbank` as the data is in the `pbank.json` file in the project fixtures directory. If you want to see what the published metadata would look like, you can call the `publish` instance method on each LegacyProject model.

The files are not done yet. We will do them soon.

Notes about migration will be commented in the mentioned issue.


physionet-django/project/models.py
```diff
@@ -1,8 +1,10 @@
+from datetime import datetime
 import hashlib
 import os
 import shutil
 import uuid
 import pdb
+import pytz
```
tompollard, line 7: Is `pytz.timezone(timezone.get_default_timezone_name()).localize(dt)` setting the time to the server timezone? Is there a reason why we can't specify the timezone in `settings.py`?

physionet-django/project/pbmigration.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    Load data from a pbank database and create a LegacyProject object
+    # Some pages end with this: <!--#include virtual="/dir-footer.shtml"-->
+    """
+    try:
+        f = urlopen('https://physionet.org/physiobank/database/{}/HEADER.shtml'.format(slug))
+    except:
+        f = urlopen('https://physionet.org/physiobank/database/{}/index.shtml'.format(slug))
+    content = f.read().decode('utf8')
+    f.close()
```
tompollard, line 23: Seems odd that all this is sitting in the module, outside functions/classes. Is this intended? Might be worth moving to functions and adding a main() if it is intended to be run as a script? Will help to avoid strange things happening on import.

cx1111: This module isn't imported or called from anywhere else. I just use its snippets in the shell when appropriate at various stages of the manual content editing.

physionet-django/project/models.py
cx1111: `timezone.get_default_timezone_name()` is reading the timezone from the settings file. pytz.timezone() is then adding that time zone to the dt object.

tompollard:
From reviewing a random set of legacy content, this looks great. The lack of structure in the old articles makes things a little messy in places, but I think this is the best we can do given the inconsistency.

We could consider introducing a short_description/blurb/teaser field for all projects, so this can be listed in places like the "New content" index, instead of the abstract. This would also help to clean things up for legacy content where abstracts are missing. e.g.

```
LegacyProject.objects.get(pk=3).publish()
LegacyProject.objects.get(pk=10).publish()
LegacyProject.objects.get(pk=12).publish()
```

screen shot 2019-02-08 at 11 41 55

I'll add this as a feature request...

tompollard:
Thanks Lucas, there are a couple of tweaks that we should make, but overall I think this is an improvement on the current hardcoded page. @bemoody do you have any objections to merging after we clean up the points below?

### 1. Filtering on latest version

If a project has multiple versions, it will be listed multiple times:

Screen Shot 2020-06-22 at 17 53 05

### 2. Formatting issues due to the HTML abstracts

After loading the legacy content using the steps below:

```py
# load the fixture
python manage.py loaddata project/fixtures/pbank.json 

# publish the projects
x = LegacyProject.objects.all()
for p in x:
    try:
        LegacyProject.objects.get(pk=p.id).publish()
    except:
        pass
```

I notice there are some formatting issues in the content. e.g.

Screen Shot 2020-06-22 at 18 05 57

It might be worth trying to get a plain text version of the abstract (`project.abstract_text_content`?) and using that instead.

### 3. Text in the overview

I think the overview text could do with some slight rewording, and will suggest a minor update before we merge.

### 4. Header tags

Currently the section headings look like they are `<h1>`s. They should be `<h2>` I think, for consistency with the other headers and to avoid breaking `w3c` validation.

PR #219, opened 2019-02-13
Merge main and special files together

cx1111:
The special-files interface has been removed as it was confusing.

Now all published projects have a file root and a main file root. ie
-  `/static/published-projects/PID/` is the file root for that project
-  `/static/published-projects/PID/files/` is where the main files for the project goes, including generated checksum, license, and files list.
-  `/static/published-projects/PID/something.zip` is where the zip file goes

Had to change some views and templates to support these changes. Hopefully everything is cleaner now.


tompollard:
I'm happy for this to be merged. Removing the special file section is an improvement and I didn't hit any bugs when testing. Does "ANNOTATORS.tsv" need to be tab delimited? I think CSVs are generally easier to work with.

cx1111:
Thanks for checking.

Some of the dbs have commas in their names. Jose Vicente from the FDA added some ecg databases of patients after taking long lists of drugs which he wanted in the titles.

tompollard:
CSVs handles commas in fields if properly formatted. See: https://en.wikipedia.org/wiki/Comma-separated_values#RFC_4180_standard

PR #220, opened 2019-02-14
Add REST export

cx1111:
Made a new app for exporting content in accessible formats such as for metadata, references etc later. Started with a json response of the list of dbs and software packages.

### DIFF
--- a/physionet-django/export/admin.py
+++ b/physionet-django/export/admin.py
@@ -1,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
--- a/physionet-django/export/apps.py
+++ b/physionet-django/export/apps.py
@@ -1,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class ExportConfig(AppConfig):
+    name = 'export'
--- a/physionet-django/export/models.py
+++ b/physionet-django/export/models.py
@@ -1,0 +1,3 @@
+from django.db import models
+
+# Create your models here.
--- a/physionet-django/export/serializers.py
+++ b/physionet-django/export/serializers.py
@@ -1,0 +1,10 @@
+from rest_framework import serializers
+
+from project.models import PublishedProject
+
+
+class PublishedProjectSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = PublishedProject
+        fields = ('title', 'version', 'slug', 'abstract', 'main_storage_size',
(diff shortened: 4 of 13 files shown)


tompollard:
Great idea and works for me. The url design needs some planning, but I'm good with merging this as a starting point. Listed some thoughts below:

- Instead of separate endpoints for each type of project (database-list, rest/software-list, etc), it feels like we should have a project endpoint that supports arguments (e.g. `projects?type=database`, `?query=ecg`, `creator=rgmark`, `publicationyear=1978`, etc). 
- Not an issue, but odd `main_storage_size` is smaller than `compressed_storage_size` in at least one of the fixtures. Assume this is just because you entered them randomly?

```{'title': 'eICU Collaborative Research Database',
 'version': '2.0.0',
 'slug': '3Yd8AnIWSd2jmgAErXUZ',
 'abstract': '<p>The eICU Collaborative Research Database is a large multi-center critical care database made available by Philips Healthcare in partnership with the MIT Laboratory for Computational Physiology.</p>',
 'main_storage_size': 81,
 'compressed_storage_size': 3090}
```

- Calling the app "export" suggests we'd only use the service for exporting information. Is it possible we will use it for other tasks (e.g. checking status of users, creating new projects, updating status of projects, etc)? If so, should the name be more general?
- This seems to be a popular spec for JSON APIs: https://jsonapi.org/format/, so perhaps a good reference (e.g "A document MUST contain at least one of the following top-level members: data; errors; meta"). OpenAPI has supplementary recommendations: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md
- Documentation for the DataCite API might also be a useful reference: https://support.datacite.org/docs/api. Use of a subdomain for their API (api.datacite.org) seems nice.

cx1111:
- The multi-optioned querystring makes sense and should be implemented.
- For that really small database, the zip is actually bigger than the individual files.
- I don't know if we want to allow an API for users to create projects. Maybe we can think about this later. The structure of the app/urls probably will have to be reorganized many times.
- I'll fix this and implement with the specifications in the next REST branch.
- Yea the intention is to add apis for datacite and such later.

tompollard:
Thanks, merged! Comment about the DataCite API was just to say that maybe we can learn from their example, because their use case (search data resources) seems similar to ours.

PR #228, opened 2019-02-21
Make menus work using CSS without JavaScript.

bemoody:
The first rule should work for mouse navigation with older browsers
that don't support :focus-within, but it will not work for keyboard
navigation.  The second rule should work for keyboard navigation with
current Firefox (60+), and possibly other browsers.

(In Firefox 52, :focus-within is supported but seems buggy in this
regard.)


tompollard:
Works fine for me in Chrome. @elfeto good with you to merge?

PR #234, opened 2019-02-22
Use python3 interpreter, not python.

bemoody:
We are using python3, right?!


tompollard:
Yep, Python3! We're generally in a virtual environment when runnning manage.py so this shouldn't usually be an issue, but no harm in being explicit...

PR #238, opened 2019-02-25
Add a few info pieces to help with where you are on the site

cx1111:
Responding to @rgmark 's feedback about not knowing where you are.

- Added a 'welcome to physionet' box on the front page
- Added heading in the database/software/content index
- Added order by on those index pages
- Restructured many views/templates, bringing them into the `search` app.
- Made the topics badges the same color as the main PN color so we have fewer colors flashing before us on the front page.

Still have not implemented pagination/a proper search. Those will be done in a future branch.


physionet-django/search/views.py
```diff
@@ -53,3 +52,88 @@ def all_topics(request):
...
+            orderby=orderby, direction=direction)
+    else:
+        projects_authors_topics = get_content(resource_type=None,
+            orderby=orderby, direction=direction)
+
+    return render(request, 'search/content_index.html', {'form':form,
+        'projects_authors_topics':projects_authors_topics})
+
+
+def database_index(request):
```
tompollard, line 105: Why have separate views for `database_index` and `software_index`? Seems like it would be easier to handle the type with an argument passed from urls.py, particularly as we are likely to be adding new content types in the near future.

physionet-django/search/templates/search/database_index.html
```diff
@@ -0,0 +1,27 @@
+{% extends "base.html" %}
```
tompollard: Could we have a single template to handle all data types? Seems unnecessary to have separate templates for database and software.

physionet-django/templates/home.html
```diff
@@ -33,39 +33,22 @@
...
         <br><br>
       </div>
 
       <!--Right Column -->
       <div class="col-md-5">
+
+        <div class="card">
+          <h2 class="card-header">Welcome to PhysioNet</h2>
+          <div class="card-body">
+            <p class="card-text mt-1">PhysioNet is a massive digital library and publisher of physiologic <a href="{% url 'database_index' %}">data</a> and <a href="{% url 'software_index' %}">software</a>, helping to make biomedical research more transparent and reproducible.</p>
```
tompollard: Perhaps stick with language from the current site. e.g. "PhysioNet offers free web access to large collections of recorded physiologic signals and related open-source software. For an overview of the content, see our XXX page."

physionet-django/search/views.py
cx1111: I don't have a strong opinion about this, but Roger wanted that list of data categories at the top (which I we haven't implemented) yet so that would be a rather large difference. For now, there isn't too much boiler code, and is quite clear without needing to pass in a lot of string variables just for display.

physionet-django/templates/home.html
cx1111: Signals? Web access? Are you sure we want to keep exactly as is?

tompollard: My general view is the more the new site looks like the current site, the easier it will be to make the move. We can tweak text later..

physionet-django/search/views.py
tompollard: A single approach for all projects (single template/function) is more future proof. The current approach may seem fine now, but if someone decides to modify each of the functions/templates independently then it becomes tricky to try to merge them back into a single function/template.

tompollard:
Thanks! "Order by" for title doesn't sort as I would expect (WFDB comes before eICU Collaborative...) when set to ascending, but minor issue that can be fixed later.

cx1111:
I'll add this next:

```
from django.db.models.functions import Lower

Model.objects.order_by(Lower('title'))
```

Issue #240, opened 2019-02-25
Highlight active page in the admin console menu


tompollard:
When viewing menu items in the admin console, ideally the drop down menu should remain open with the current page highlighted. e.g. in the screenshot below, when "Ongoing Applications" is selected, the text should be highlighted and the drop down should remain open.

![screen shot 2019-02-25 at 17 13 42](https://user-images.githubusercontent.com/822601/53372511-0c06ff80-3921-11e9-93b3-e4d78f443df1.png)

PR #246, opened 2019-02-26
Minor changes to the console and admin pages

tompollard:
This pull request introduces some minor changes to the console and admin pages. The main ones are:

- Remove icons from the headers of admin pages because they interfere with readability.
- Move projects to a submenu for consistency with the other menu items.
- Removes explanatory text from the sections in /console/editor-home/ based on feedback by Felipe. The headers are clear and we can add explanatory text later if requested.

### DIFF
--- a/physionet-django/console/static/console/css/console.css
+++ b/physionet-django/console/static/console/css/console.css
@@ -501,3 +501,8 @@ footer.sticky-footer {
     width: calc(100% - 55px);
   }
 }
+.fa-fw {
+    text-align: center;
+    width: 1.25em;
+    margin-right: .25rem;
+}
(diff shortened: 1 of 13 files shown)


elfeto:
- The icons should stay at the left and not the right.
- Why is users and news (like the current) a drop down, if there is only one thing there.
- Why is there a add news sub menu, News should be its own thing (not drop down), and should be ok with just the add news button

tompollard:
Thanks for taking a look!

> The icons should stay at the left and not the right.

At the moment they affect readability because the left alignment of the text isn't consistent. I'll take another look to see if I can find a happy alternative (either make fixed width on left or better, just remove them while the bar is open).

> Why is users and news (like the current) a drop down, if there is only one thing there.

I agree it's inconsistent, but this is current functionality. I think Chen probably anticipated more items appearing in these menus in future.

> Why is there a add news sub menu, News should be its own thing (not drop down), and should be ok with just the add news button

This is the current functionality. I agree the "add items" section seems unnecessary, and if it was removed then the menu could be collapsed. If Chen has no objections, I can add a change to this pull request.

tompollard:
The news menu is now collapsed. Looks like the base theme is at: https://startbootstrap.com/themes/sb-admin-2/

tompollard:
Perfect, thanks. Fixes #240. Okay to merge?

PR #247, opened 2019-02-26
Add some charts to show database stats

cx1111:
If you go to `/data/stats/` you'll see charts made from the published databases.

First one shows new number of dbs and volume of storage added each year, second one shows cumulative.

Reads from new REST API at `http://127.0.0.1:8000/rest/database-stats-list/`


tompollard:
Thanks, this looks useful. I tested by loading the PhysioBank data with:

```bash
./manage.py loaddata pbank
```

...then publishing from the shell with:

```python
for n in range(100):
    try:
        LegacyProject.objects.get(pk=n).publish()
    except:
        pass 
```

Let me know if there's an easier way. An example of the plots is below:

screen shot 2019-02-26 at 18 33 27

I've listed a couple of quick thoughts below:
- Are we happy for this information to be public or should it be restricted to admin? Public seems okay, but best to confirm with Roger if we haven't already.
- Same issues with the API path as last time, but I guess this can be cleaned up later.
- In the example plot, the legacy projects don't have a volume. Any idea why?
- The URL for the plots (`/data/stats/`) says `data`, but I assume this captures all projects. If so, perhaps a better URL would be `project/stats/`?
- Same point about `data` applies to the API endpoint and title of the charts.

cx1111:
Thanks for checking.

- I included demo metadata in the fixtures, but the actual files are only on physionet-production so their storage sizes are 0, even when you publish the legacy projects. There is also `from project.pbmigration import publish_all_legacy` which you call to publish them all.
- I put this view only for databases because it's what we've always done in the reports, and people come here primarily for that. We could extend it for software and other content types later, though storage volume for example don't really mean anything for non-database resources.
- I'll fix the all the JSON to conform to standards together. Was just trying this simple format first to make it work.

I haven't asked Roger about making it public, but this page is not linked from the rest of the site so it should be safe.

elfeto:
- Should "pbank" be "pbmigration"?
- Also, ModuleNotFoundError: No module named 'requests'
- I don't understand, this "* the actual files are only on physionet-production so their storage sizes are 0, even when you publish the legacy projects." Why dont we have the sizes? should be easy to get.
- "Total Databases Per Year" is confusing, when I read that I think that is the same as New DB per year. Should be "Total DB" ? 
- Why not call the URL 'data/stats/' -> 'stats/' and have the page "admin" only ? leave the rest api like that but not the actual page
- Why you fetch the data twice? (d3.json)
- The project size (main_storage_size) is in bytes, but here we treat it as GB.

cx1111:
- I've made the REST API and template name generalizable. Can pass in parameters for different project types.
- Refactored js so there is less duplicate code. Also more tidy and understandable. Fetch data once.
- I made this set of stats public because this info is already public. We can talk about other stats for admin panel or more stats in the future, but I figured we'd start with this. Can we hone in on the url and all the stats we want to show another time? Also note that this url isn't linked from the rest of the site at this point.
- I slightly changed the name of the first chart.
- Changed values of storage size into gigabytes.

For the published projects, I didn't write in the fixtures their sizes because I don't know how large they are. If you want demo storage values for the published items, run this:

```
for p in PublishedProject.objects.all():
    p.main_storage_size = 100000000 * p.id
    p.save()

```

tompollard:
Thanks, I'm good for this to be merged. Still not keen on `/data/stats/`, but that's okay :)  Works well in my testing and it's a great starting point.

elfeto:
- Why you import project utility in views? 
- Why create a function that it has no purpuse other than to call another function?
   - If the idea of the "project_stats" is to treat the databases and software as projects, then why have "database_stats" ? 
- How can the "resource_type" can be changed ? (I know is not implemented)
- function "database_stats_list" in the exports views the for loop is faulty. It gets from year 1 to last year. (unless we dont want to show current year stats)
- the fixture that loads the DB lists the eICU as 81 bytes. 
- If its only one database, should it say "1 Databases" or '1 Database'
- If two projects of 500MB are published (total in between 1 GB), the size displayed is 0 because of how its calculated

cx1111:
Addressed all points, except for that plural one.

URL is now at `/charts/` with optional `?resource_type={}`

PR #248, opened 2019-02-27
Update notification emails

tompollard:
Summary of changes below:

- General tidying of the text.
- Got rid of those single quoted docstrings in project models.
- Add `get_first_name` method to the user class. Attempts to get first name from the first_names field. We could update this in future to avoid addressing emails to "Dear Dr" etc where someone has included the prefix in the name field when registering.
- Add header information to emails, called in `email_project_info` function
- Add `email_footer` function for getting information for the email footer. For now this is empty, but in future we can add privacy policy etc if needed.
- Move signature of emails to a `email_signature()` function

### DIFF
--- a/physionet-django/notification/templates/notification/email/accept_submission_notify.html
+++ b/physionet-django/notification/templates/notification/email/accept_submission_notify.html
@@ -3,15 +3,13 @@
 
 Dear {{ name }},
 
-Based on the recommendations of our editors, we are delighted to accept your project entitled {{ project.title }} for publication on PhysioNet. 
+Based on the recommendations of our editors, we are delighted to accept your project entitled "{{ project.title }}" for publication on PhysioNet. 
 
 A summary of our standard set of editorial checks is listed below:
 {% for result in edit_log.quality_assurance_results %}- {{ result }}
 {% endfor %}
 Additional comments from the editorial team are listed below:
-"""
 {{ edit_log.editor_comments }}
-"""
 
 After copyediting of the project and files is complete, we will be in touch again. Then, once all authors approve of the copyedited version at: http://{{ domain }}{% url 'project_submission' project.slug %}, the project will be ready for publication. 
 
--- a/physionet-django/notification/templates/notification/email/accept_submission_notify.html
+++ b/physionet-django/notification/templates/notification/email/accept_submission_notify.html
@@ -1,18 +1,21 @@
 {% load i18n %}{% autoescape off %}
-Dear {{ name}}
+{{ project_info }}
 
-The editor has accepted the submission of your project: {{ project.title }}.
+Dear {{ name }},
 
-The standard quality assurance results are as follows:
+Based on the recommendations of our editors, we are delighted to accept your project entitled {{ project.title }} for publication on PhysioNet. 
+
+A summary of our standard set of editorial checks is listed below:
 {% for result in edit_log.quality_assurance_results %}- {{ result }}
 {% endfor %}
-The following comments were made:
+Additional comments from the editorial team are listed below:
 """
 {{ edit_log.editor_comments }}
 """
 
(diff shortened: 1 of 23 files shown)


bemoody:
On 2/27/19, Tom Pollard <notifications@github.com> wrote:
> - Add `get_first_name` method to the user class. Attempts to get first name
> from the first_names field. We could update this in future to avoid
> addressing emails to "Dear Dr" etc where someone has included the prefix in
> the name field when registering.

Hmm, I'd say for a program to address people by their first name is
dicey at best.  If it's necessary to address them at all, I'd address
them by full name or by their chosen username.

Side note, have you all read this?

https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

cx1111:
`40. People have names`

tompollard:
I have read several discussion about firstnames/lastnames, but hadn't seen that one. Given that we ask for firstname and lastname, I don't think it's a big deal to address someone by their firstname. 

If you strongly prefer "Dear Firstnames Lastname" in the email, I'm happy to switch to that. Just for me, seeing the "Dear Alistair Edward William Johnson" sounds pretty spammy.

physionet-django/notification/templates/notification/email/accept_submission_notify.html
```diff
@@ -1,18 +1,21 @@
 {% load i18n %}{% autoescape off %}
-Dear {{ name}}
+{{ project_info }}
 
-The editor has accepted the submission of your project: {{ project.title }}.
+Dear {{ name }},
 
-The standard quality assurance results are as follows:
+Based on the recommendations of our editors, we are delighted to accept your project entitled {{ project.title }} for publication on PhysioNet.
```
cx1111: entitled "{{ project.title }}"

elfeto:
- I don't like the "{{ project_info }}" in the email, and definitely don't like it before the "Dear person"
- Also, we should remove the """  since is obvious that the comments/changes are bellow the text?
    - The changes made during this process are summarized below
    - Additional comments from the editorial team are listed below:)

cx1111:
I've made minor suggestions to all of the email content. Feel free to address them within this pull or afterwards.

- I only mention the triple quotes thing because you guys brought it up. Keep in mind that this free text field includes newlines.
- Regarding the first name thing, the reason why I put a `first_names` field is that it's difficult for us to understand first and middle names independently (whether those concepts conceptually exist) in all languages. But we do know that in a vast majority of cases, people have a last name, and some other name which we call first names. Overall, I think trying to figure out their real first name is much more trouble than it's worth, and we should just put the full name they gave us in the first place. So my suggestion is to remove the `get_first_name` function and reinstate their full names in the emails. Also I imagine the reason you put the email header/footer is to make us seem like a more official standardized publisher, and being more 'friendly' by putting their first name only isn't supporting that cause.

tompollard:
Thanks for comments! I have pushed some updates that I hope deal with these points and added replies to your feedback below:

> * I don't like the "{{ project_info }}" in the email, and definitely don't like it before the "Dear person"

Yep, fair point and I can see that this isn't to everybody's taste, but it is common in the publication systems that I've used, so would like to go with it! 

> * we should remove the """  since is obvious that the comments/changes are below the text?

I agree they don't look very nice in the email, so have removed them. If we find the emails get confusing because of newlines in comments, we can review/update.

> * my suggestion is to remove the `get_first_name` function and reinstate their full names in the emails.

Dear First Names Lastname is back by popular demand :)

> the reason why I put a `first_names` field is that it's difficult for us to understand first and middle names independently (whether those concepts conceptually exist) in all languages.

That's fine, and it's a good point. We should check that this won't cause issues when submitting metadata to indexing services (e.g. PubMed). We will generally be dealing with American systems, so it may be difficult to avoid forcing names into the standard firstname/lastname convention at times.

PR #249, opened 2019-02-27
User console

elfeto:
Cleaned the user console tab and added a login counter

### DIFF
--- a/physionet-django/console/urls.py
+++ b/physionet-django/console/urls.py
@@ -54,7 +54,7 @@ urlpatterns = [
 
     path('users/all/', views.users, name='user_list'),
     path('users/admin/', views.admin_users, name='user_list_admin'),
-    path('users/inactive/', views.inacvite_users, name='user_list_inactive'),
+    path('users/inactive/', views.inactive_users, name='user_list_inactive'),
     # path('users/lcp/', views.lcp_affiliates, name='lcp_affiliates'),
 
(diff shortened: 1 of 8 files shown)


physionet-django/console/views.py
```diff
@@ -489,6 +489,28 @@ def users(request):
     users = User.objects.all()
     return render(request, 'console/users.html', {'users':users})
 
+@login_required
+@user_passes_test(is_admin)
+def inacvite_users(request):
```
tompollard: inacvite_users -> inactive_users

physionet-django/console/urls.py
```diff
@@ -53,6 +53,8 @@
         name='process_credential_application'),
 
     path('users/all/', views.users, name='user_list'),
+    path('users/admin/', views.admin_users, name='user_list_admin'),
+    path('users/inactive/', views.inacvite_users, name='user_list_inactive'),
```
tompollard: views.inacvite_users -> views.inactive_users

tompollard:
Looks good, thanks! I'm happy to merge after the "inacvite" typo is fixed. Couple of quick thoughts:
- on the /console/users/inactive/ page, what is listed in the "Logins" column? Could the title be changed to make the contents clearer? (e.g. `Logins (90 days)` or `Logins  (total)`)
- Could we have a chat about `update_user_login`? Just like to understand what it is doing.

PR #253, opened 2019-03-01
added legacy credential model

elfeto:
Added the model for credentialing.
Also it checks when a account is created and a email is added to see if the person was credentialed in the past.

### DIFF
--- a/physionet-django/user/templates/user/edit_credentialing.html
+++ b/physionet-django/user/templates/user/edit_credentialing.html
@@ -17,7 +17,7 @@
   <p><i class="fa fa-clock"></i> Your credentialing application was submitted on {{ current_application.application_datetime }}. We will respond shortly.</p>
 {% else %}
   <p>Your account is not credentialed. You may <a href="{% url 'credential_application' %}">apply for access</a>.</p>
-  <p><strong>If you have already been credentialed</strong>, please add the approved email to your profile.</p>
+  <p><strong>If your account on the old PhysioNet site is already credentialed</strong>, please add the email address from the old site to this account. Your credentialed status will be automatically transferred when your email is verified.</p>
 {% endif %}
 
 {% if applications %}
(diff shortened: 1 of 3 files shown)


physionet-django/user/templates/user/edit_credentialing.html
```diff
@@ -17,6 +17,7 @@ <h1>PhysioNet Credentialing</h1>
   <p><i class="fa fa-clock"></i> Your credentialing application was submitted on {{ current_application.application_datetime }}. We will respond shortly.</p>
 {% else %}
   <p>Your account is not credentialed. You may <a href="{% url 'credential_application' %}">apply for access</a>.</p>
+  <p><strong>If you have already been credentialed</strong>, please add the approved email to your profile.</p>
```
cx1111: "If your account on the old PhysioNet site is already credentialed, please add the email address from the old site to this account. Your credentialed status will be automatically transferred when your email is verified."

I think we have to use the word 'old' here. Make it 100% clear.

cx1111:
Great feature. First thing to do is to pull the common code into a `check_legacy_credentials` function and call it from `activate_user` and `verify_email`, not `register`.

physionet-django/user/views.py
```diff
@@ -286,6 +296,17 @@ def verify_email(request, uidb64, token):
             associated_email.verification_date = timezone.now()
             associated_email.is_verified = True
             associated_email.save()
+
```
tompollard: Can you avoid this duplication with something like a `legacycredentialed(email)` function? Might be nice to separate it out.

tompollard:
This is excellent! Thought it was going to be super complicated to work out...big step towards being able to switch to the new system.

elfeto:
- Made the change to avoid the duplication of the credentialing code
- Fixed it so that a person can only use this feature once.

I will do another pull request adding some tests

PR #254, opened 2019-03-01
Update the about pages

tompollard:
Updates content in the about pages. Quick summary below:

- Removed the list of team members for now. Can be added later.
- Removed custom css from all about pages except the timeline
- Instructions for authors: Instead of explaining how the submission system works (which should be self-explanatory during the submission process), the page now gives details on editorial policies, file preparation etc. We can add more detail later.
- Added a content overview to the about page to help address Roger's comment about arriving in Paris. Software isn't covered at the moment, but this can be added later. The isn't intended to stay long-term...just until we improve the search functionality.

There are existing issues with styling. This pull request does not attempt to fix that (aka go easy ;)).


physionet-django/physionet/urls.py
```diff
@@ -26,10 +26,11 @@
     path('', views.home, name='home'),
 
     # about pages
-    path('about/', views.author_guidelines, name='author_guidelines'),
+    path('about/timeline/', views.timeline,
+        name='timeline'),
     path('about/author-guidelines/', views.author_guidelines,
         name='author_guidelines'),
-    path('about/physionet/', views.about_physionet, name='about_physionet'),
+    path('about/', views.physionet, name='physionet'),
```
cx1111: Please rename the view, url name, and template name to 'about_physionet'.

PR #255, opened 2019-03-04
Added test for the credentialing

elfeto:
Added test for credentialing users


cx1111:
Great. Made a few small fixes:

- document tests and remove unnecessary steps
- format the fixtures to keep space consistent
- add timezone to datetime object when carrying over legacy credentials

Issue #250, opened 2019-02-28
Add share button to published projects


tompollard:
Add prominent "share" button to published projects that offers a quick way to share the project on social media (e.g. email, twitter, reddit, etc).

PR #257, opened 2019-03-08
Add share buttons to published projects

cx1111:
Go to any published project and you'll see.

The linkedin and reddit links try to see if the url actually exists so won't work in development. But I've tried it by replacing with valid urls and they work. Can test on staging.


tompollard:
Thanks! Closes #250

PR #260, opened 2019-03-12
Added a basic empty app for lightwave

elfeto:
Empty app for lightwave


elfeto:
All hail the hypnotoad

PR #268, opened 2019-03-15
Altered credentialing to have everything in the same page

elfeto:
I added a one time text box so that the reference can send a message to the credential person.
Also, a one page credentialing.

If in the future we want to switch back to the other page of credential application view, just swap them in the console/urls.py


tompollard:
Awesome, thanks!

PR #271, opened 2019-03-15
Fix local instance django guidelines

lbulgarelli:


elfeto:
Please provide a description on the changes.

lbulgarelli:
I ran the installation process and came up with some issues.

Basically I changed the guidelines to add:

> - Copy `.env.example` file to `.env`.

and changed the order so activating the virtual environment comes before installing the pip requirements:

> - Create python environment with python 3.6.
> - Activate virtual python environment.
> - Install python packages in `requirements.txt`.

tompollard:
Thanks Lucas!

Issue #212, opened 2019-01-31
Add "challenge" as a new project type


tompollard:
It should be possible to submit challenges, in a similar way to data and software. Initially, the project type should be available to physionet staff only.

PR #276, opened 2019-03-17
Add challenge project type. Ref #212

tompollard:
This is a first attempt at adding a challenge project type (#212). I've added some notes below:
- The main section headers are similar to other popular competition sites. 
    - "Objective" gives the challenge objective
    - "Participation" outlines rules and the timeline.
    - "Data description" describes the data used for the challenge. The data may be included in the project or linked to elsewhere.
    - "Evaluation" provides scoring details and information on how to submit. Sample code is also encouraged here.
- There is a fixture based on our BHI & BSN Challenge to demonstrate a published project (https://mimic.physionet.org/events/bhibsn-challenge/).
- There are places where similar functions, templates etc exist for each project type. Most of these could be collapsed to a one function/template. We can review these later and perhaps refactor.
- I dumped a new fixture with the following command. There seem to be a few differences to the original fixture (e.g. we now have `contenttypes.contenttype` data). Was this intentionally omitted before?
    - `./manage.py dumpdata --indent 2 > physionet-django/project/fixtures/demo-project.json`

Couple of questions to discuss:
- It looked to me like licenses can only be attached to a single project type. Is that right? 
- Should the challenge type be open to the public for submission? 
- How do we deal with updates (e.g. adding the winners list) after the challenge has been posted? Perhaps news items or similar could be attached to published projects?


cx1111:
Good stuff.

Issue #229, opened 2019-02-21
Text files served as application/octet-stream


bemoody:
Some text files are served as application/octet-stream, for example:
 - *.hea
 - RECORDS
 - ANNOTATORS

These should either be served as text/plain or should be served
without a content-type at all.

PR #286, opened 2019-03-19
Send the response to nginx without content type

### LINKED ISSUE #274: Uploaded file content type
If I:
- create a project (/projects/string-of-garbage/files/)
- upload a binary file (100s.dat)
- try to download the same file (/projects/string-of-garbage/files/100s.dat)

then the file is served with "Content-Type: text/html; charset=utf-8", despite the fact that the file is neither UTF-8, nor HTML.

elfeto:
Now, is really fixes #274 and #229.

### DIFF
--- a/physionet-django/project/utility.py
+++ b/physionet-django/project/utility.py
@@ -196,15 +196,18 @@ def get_form_errors(form):
     return all_errors
 
 
-def serve_file(request, file_path):
+def serve_file(file_name, file_path):
     """
     Serve a file to download. file_path is the full file path of the
     file on the server
     """
-    if os.path.exists(file_path):
-        with open(file_path, 'rb') as f:
-            response = HttpResponse(f.read())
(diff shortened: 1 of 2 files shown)


physionet-django/project/utility.py
```diff
@@ -196,15 +196,18 @@ def get_form_errors(form):
...
-            response['Content-Disposition'] = 'attachment; filename=' + os.path.basename(file_path)
-            return response
-    else:
+    try:
+        response = HttpResponse()
+        response['Content-Type'] = ''
+        response['Content-Disposition'] = 'attachment; filename=' + os.path.basename(file_name)
+        response['X-Accel-Redirect'] = '/protected/' + file_path
+        return response
+    except:
```
tompollard: @alistairewj will get upset if we let this through with a bare except. Could you add the exception type here? e.g. `except ValueError`

PR #298, opened 2019-03-21
Unify open_files_panel and protected_files_panel templates

bemoody:
Currently there are three templates that are virtually identical, a
fourth that is very similar, and a fifth that isn't used anywhere
AFAICT.

These patches unify "open_files_panel" and "protected_files_panel"
into one.

"preview_files_panel" is different in that the inline script uses a
"csrfmiddlewaretoken".  (I don't know where this comes from, what if
anything is checking it, or why it exists, and I can't think of any
reason that "preview" project requests should be handled differently
than "protected" project requests.  But perhaps I'm missing
something.)  This should be easy to resolve either way, and
"preview_files_panel" should also, presumably, be unified with the
other two.

"project_files_panel" is significantly different, and it may or may
not make sense to unify that with the other three.

I don't know whether "open_special_files_panel" is something that is
no longer used and should be removed, or if it's something that's not
yet used and should be kept. :)


cx1111:
Implemented these changes on top of the new version functionality in PR #311. Thanks.

PR #302, opened 2019-03-21
Restore serve_file functionality in development server

bemoody:
Recently Felipe added the ability for protected files to be served
directly by nginx (rather than reading the file into python and
returning it) which is important for performance and simplicity (let
the web server do what web servers are good at.)

Unfortunately this change also made it difficult to test the server
with 'manage.py runserver', because of course the built-in web server
doesn't understand X-Accel-Redirect.

These patches add a variable (MEDIA_X_ACCEL_ALIAS) which is set in the
'staging' and 'production' configurations, but not in the
'development' configuration.  Internally, files are still referred to
by their actual paths in the local filesystem, but when it is time to
serve the file to the user, we check whether it is located in
MEDIA_ROOT and issue an X-Accel-Redirect if so.  If files are stored
in locations other than MEDIA_ROOT, this may need further enhancement.

In addition to serve_project_file and serve_published_project_file
(which previously used project.utility.serve_file), there are also two
functions profile_photo and training_report which previously used
physionet.utility.serve_file.  (Those two functions were originally
duplicates, but their APIs had then diverged.)

All four functions now use the new function
project.utility.serve_abs_file, and both serve_file functions are
removed.

Please verify this doesn't conflict with Chen's pending changes to
project directory structure before merging.

### DIFF
--- a/physionet-django/physionet/settings/development.py
+++ b/physionet-django/physionet/settings/development.py
@@ -35,6 +35,12 @@ DEMO_FILE_ROOT = os.path.join(os.path.abspath(os.path.join(BASE_DIR, os.pardir))
 
 MEDIA_ROOT = os.path.join(os.path.abspath(os.path.join(BASE_DIR, os.pardir)), 'media')
 
+# If defined, MEDIA_X_ACCEL_ALIAS is the virtual URL path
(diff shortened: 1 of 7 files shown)


cx1111:
Implemented these changes on top of the new version functionality in PR #311. Thanks.

PR #304, opened 2019-03-21
Fix navbar resizing

lbulgarelli:
- Fixed navbar adding some space when window size was reduced.
- Fixed navbar having different spacing between logo and menu in the admin console.


tompollard:
Nice, thanks Lucas

Issue #291, opened 2019-03-19
[Suggestion]: Return button on preview page


alistairewj:
The preview page doesn't have a direct link back to the project (as far as I am aware) - you have to go My Projects -> Find the Project -> Click. Seems like a common workflow would be Preview project -> Back to project -> Preview -> Back to project etc so a button would be useful.

alistairewj:
After discussion with @elfeto I think probably adding the text "Opens in new window" next to the preview button is the best fix since at the time it wasn't clear to me that it would open in a new window.

tompollard:
Thanks, looks like there is W3C guidance for this: https://www.w3.org/TR/WCAG20-TECHS/G201.html.

> G201: Giving users advanced warning when opening a new window
> Opening new windows automatically when a link is activated can be disorienting for people who have difficulty perceiving visual content, and for some people with cognitive disabilities, if they are not warned in advance. Providing a warning allows the user to decide it they want to leave the current window, and the warning will help them find their way back, if they do decide they would like to go to the new window. It will help them understand that the "back" button will not work and that they have to return to the last window they had open, in order to find their previous location.

The webpage suggests two methods for warning the user (HTML text and CSS). We'll add a fix!

PR #306, opened 2019-03-22
Minor changes to styling

tompollard:
Some minor changes to styling:
- Add back the card borders
- Remove the container on the project submissions overview to fix alignment of text at top
- Get rid of glasses image in the submission buttons
-  Move the project invitation section to the top of project home to make it clearer when actions are outstanding. In the longer term we could use this section for all task reminders, not just invitations.
- Fix the black text on dark background issue.
- Other minor text and styling changes.

### DIFF
--- a/physionet-django/project/static/project/css/project-content.css
+++ b/physionet-django/project/static/project/css/project-content.css
@@ -1,9 +1,9 @@
-.citation {
+/*.citation {
   margin-top: 0.25rem !important;
   margin-left: 3rem !important;
   margin-right: 3rem !important;
   background: rgb(230, 230, 230) !important;
-}
+}*/
 
 blockquote {
   display: block;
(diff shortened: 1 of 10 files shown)


tompollard:
Added "Opens in new window" icon to address Alistair's point in https://github.com/MIT-LCP/physionet-build/issues/291.

### Icon alongside links in text:

Screen Shot 2019-03-23 at 10 13 28

### Icon alongside buttons:

Screen Shot 2019-03-23 at 10 13 35

elfeto:
Is there a specific reason you moved the "Project invitations" to the top above active ? 
In most cases one wont have  much invitations or only one, that goes away one is rejected or accepted, why not just display the invitations if you have one?

tompollard:
Chen and I discussed this before this change was added. The top of the page is the most prominent place, so it makes sense to put outstanding actions here. At the moment, if you have more than one or two active or published projects then the invitations are hidden away off-screen. 

The plan is to use the section for much more than just invitations...a place that lists all tasks that need dealing with. We could make the section invisible if there are no outstanding actions if we want to add this later.

PR #311, opened 2019-03-26
Implement multiple project publishing versions

cx1111:
You can now publish multiple versions of a project. If you do, you will be able to see it in the published project page, and also in your project home. In your project home, you will not be able to 'publish new' from the same project multiple times at once.

- Once a project is published for the first time, the slug chosen is given to it forever and cannot be changed. This slug will be used for all future versions too.
- **Big change**: files are now stored in `<static or media root>/published-projects/<slug>/<version>/` and all zips of the same project are stored in  `<static or media root>/published-projects/<slug>/<title-version>.zip`

As a demonstration, if we publish mimic several times, the files/folders will look like this:
- `<media root>/published-projects/mimiciii/1.3.0/`
- `<media root>/published-projects/mimiciii/1.4.0/`
- `<media root>/published-projects/mimiciii/1.5.0/`
- `<media root>/published-projects/mimiciii/mimic-iii-clinical-database-1.3.0.zip`
- `<media root>/published-projects/mimiciii/mimic-iii-clinical-database-1.4.0.zip`
- `<media root>/published-projects/mimiciii/mimic-iii-clinical-database-1.5.0.zip`

Also implemented @bemoody's two pull requests here (#298 and #302)so that there would be no conflicts with the new file structure.
- Reduced the number of files_panel.html boiler templates. Now there is one for viewing files (`files_panel.html`), and one for editing (`edit_files_panel.html`).


cx1111:
After we deploy to the server, we should run this:

```

for p in PublishedProject.objects.all():
    os.rename(os.path.join(p.project_file_root(), 'files'), p.file_root())
```

Since there are no versions currently, all published projects just have one subdirectory called 'files' in their project file roots.

elfeto:
- The latest version is the latest by date, not version number.
- Is confusing to show all the versions in the /data page, there should only be one, and a link to all the versions. 
- The files are not handled, If a person want to just add more files, how would they ? or they should just re-add everything ?
- selinium no tests? all commented out.
- The function "zip_dir" is imported wierdly
`
physionet/utility.py:def zip_dir(zip_name, target_dir, enclosing_folder=''):
project/models.py:    list_items, StorageInfo, get_tree_files, list_files, zip_dir)
project/models.py:        zip_dir(zip_name=fname, target_dir=self.file_root(),
project/utility.py:from physionet.utility import serve_file, zip_dir
`

elfeto:
Fixed the version order by version number
Added the files as hard links to the new versions

physionet-django/project/test_browser.py
```diff
@@ -114,7 +114,7 @@ def send_ck_content(self, outer_id, content):
 
 
 @tag('browser')
-class TestSubmit(TestMixin, BaseSeleniumTest):
+class estSubmit(TestMixin, BaseSeleniumTest):
```
tompollard: "TestSubmit" -> "estSubmit". Assume this was committed by mistake!

physionet-django/project/test_browser.py
```diff
@@ -126,7 +126,7 @@ def setUp(self):
         TestMixin.setUp(self)
         BaseSeleniumTest.setUp(self)
 
-    def test_submit(self):
+    def est_submit(self):
```
tompollard: "TestSubmit" -> "estSubmit".

physionet-django/project/test_browser.py
```diff
@@ -260,7 +260,7 @@ def test_submit(self):
         # self.assertTrue(project.under_submission())
         # self.assertEqual(project.storage_used(), 50)
 
-    def test_publish(self):
+    def est_publish(self):
```
tompollard: "Test_publish" -> "est_publish".

physionet-django/project/test_browser.py
```diff
@@ -208,7 +208,7 @@ def test_submit(self):
         # self.selenium.find_element_by_id('create-folder-button').click()
         # self.selenium.find_element_by_id('id_folder_name').send_keys('subject-1')
         # self.selenium.find_element_by_id('create-folder-button-submit').click()
-        # self.selenium.find_element_by_name('items').click()
+        # self.selenium.find_element_by_name('items').click()BaseSeleniumTest
```
tompollard: Commented code, so doesn't really matter, but assume this wasn't intended.

tompollard:
This is great :) After chatting about the title, we thought that it is best to make this fixed (i.e. it cannot be edited between versions). One issue with editable titles is that people may try to insert versions into the title (e.g. https://zenodo.org/record/2604314#.XKNVHetKhSM) which could end up disagreeing with our semantic versions.

PR #316, opened 2019-03-28
Responsive pages

lbulgarelli:
I've made the following changes:

- Form to submit buttons aligned right so it's more intuitive
- Aligned items in pages and forms
- Footer at bottom of small pages
- The following sections were changed to be responsive:
 - User
 - Projects
 - Search


tompollard:
Great, thanks!

PR #318, opened 2019-03-28
Search

lbulgarelli:
I've combined the topic search and content type pages in one.

Now in the find page the user can:

- See all content (default)
- Filter by content type (database, challenge, software)
- Search title, abstract and tags
- Sort results 
- It is now possible to use all filters together

### DIFF
--- a/physionet-django/search/views.py
+++ b/physionet-django/search/views.py
@@ -84,7 +84,7 @@ def content_index(request, resource_type=None):
     List of all published resources
     """
     LABELS = {0:['Database', 'databases'],
-        1:['Software', 'software'], 2:['Challenge', 'challenge']}
+        1:['Software', 'softwares'], 2:['Challenge', 'challenges']}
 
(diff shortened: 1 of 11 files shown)


physionet-django/search/views.py
```diff
@@ -54,52 +58,89 @@ def all_topics(request):
...
+        published_projects = PublishedProject.objects.all().annotate(relevance=Count('core_project_id'))
+    elif type(resource_type) != list:       
+        resource_type = [resource_type]
+
+    if topic == '' or len(topic) == 0:
         published_projects = PublishedProject.objects.filter(
-            resource_type=resource_type)
+            resource_type__in=resource_type).annotate(relevance=Count('core_project_id'))
+    else:
+        import re
```
tompollard: sure you want the import here and not at the beginning of the module?

tompollard:
Nice :) The weighting of search results still needs work, but definitely an improvement on the current search!

PR #330, opened 2019-04-02
Project name

elfeto:
Made the title of the new versions of a project disabled (JS). Its also handed in the backend.


tompollard:
Looks great, thanks!

PR #331, opened 2019-04-02
Improve search

lbulgarelli:
I've made a couple of changes to improve the search, with case insensitive matching and better relevance sorting.


tompollard:
Nice, much better! Feels like we should just drop the Google search now?

- https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/search/views.py

`def google_custom_search(request):`

`def redirect_google_custom_search(request):`

- https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/search/urls.py

`    path('search/google-custom-search/'`

`path('search/redirect-google-custom-search/'`

PR #336, opened 2019-04-02
Search fix

lbulgarelli:
Fix to work in both sqllite and postgresql.


tompollard:
Interesting bug, we'll have to watch out for database issues like this...

PR #339, opened 2019-04-04
News in new template

lbulgarelli:
Changing the template for the news page to be the same as Home and Find.


tompollard:
Thanks! This looks good. Testing the change made me think we should alter the logic used to generate the items too.

Currently we forward the user to a page listing items from the current year. Instead, perhaps the default news page should be all items (in reverse chronological order, with paging)? We can still link to archives for previous years.

tompollard:
thanks!

PR #340, opened 2019-04-04
Make header consistent

tompollard:
Minor change to the header text for news, to make it consistent with the front page (following up on #339).
- Front page used sentence case (Latest resources) but the news had title case (Latest News), so I've updated the front page for consistency. 
- Removed "PhysioNet" from the news header, because it seems redundant.


lbulgarelli:
Thanks!

PR #341, opened 2019-04-04
Clean user-submitted HTML

bemoody:
We need to ensure that HTML form fields are sanitized against a
whitelist, to ensure that the result is syntactically valid HTML,
won't break the page display (now or in the future), and to prevent
malicious inline scripts and such.

These changes implement server-side HTML whitelisting using the Bleach
library, and also *client*-side whitelisting using CKEditor's
allowedContent feature.

I previously sent around an an earlier version of these changes, which
Felipe was kind enough to review.  The major change in this version is
the addition of client-side whitelisting, which has two advantages:

- Controls for forbidden attributes are automatically hidden from the
  client side GUI.

- Manual editing of the source will also be automatically filtered, so
  that what the user sees in the WYSIWYG view should (approximately)
  match the server-side sanitization.

The same whitelist is used both on the client side (via ckeditor) and
the server side (via bleach); this makes the implementation slightly
more complicated, but I think it's worthwhile to avoid duplication.

Finally, to make up for the lack of custom table styling, I've added
some new CSS rules to produce (IMHO) pretty-looking tables as well as
code boxes.

### DIFF
--- a/physionet-django/notification/models.py
+++ b/physionet-django/notification/models.py
@@ -1,12 +1,13 @@
-from ckeditor.fields import RichTextField
 from django.db import models
+
+from project.models import SafeHTMLField
 
 
 class News(models.Model):
     """
     """
     title = models.CharField(max_length=40)
-    content = RichTextField()
+    content = SafeHTMLField()
     publish_datetime = models.DateTimeField(auto_now_add=True)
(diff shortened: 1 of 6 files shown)


elfeto:
Looks really good :)

PR #342, opened 2019-04-04
Publish pages in new template

lbulgarelli:
All the pages inside the 'Publish' dropdown were group and adapted to the new template.


bemoody:
/about/licenses/ and /about/author-guidelines/ now seem broken.

There's still a link to /about/licenses/ in the "Preparing for
submission" section.  Perhaps change that link to point to the
'Licenses' section on the same page, and remove the two broken
URLs.

The reformatting seems fine to me, but if we were to add a lot
more licenses the page could become unwieldy.  (So... don't add a
lot more licenses? :P)

physionet-django/templates/about/author_guidelines.html
```diff
@@ -1,73 +1,47 @@
...
-        <p>PhysioNet helps you to ensure that your data and software is sufficiently well organized to enable reuse by the community. Before submitting your project for review, please make sure that you have:</p>
-        <ul>
-          <li>Added all co-authors to the system. Co-authors will receive a notification message upon submission and must approve of the submission prior to publication.</li>
-          <li>Provided detailed information about the resource. This information must include a high-level background, relevant methodology on the resource creation, notes for reuse, acknowledgments, and any conflicts of interest.</li>
-          <li>Selected an appropriate <a href="{% url 'licenses' %}">license</a> for releasing your data files.</li>
-          <li>Converted all files to an open, non-proprietary format.</li>
-        </ul>
-  </div>
+<section id="submittal">
+  <h2>Submittal process</h2>
```
tompollard: "Submittal process" should be "Submission process"

tompollard:
Looks great, thanks Lucas. Please could you change "Submittal process" to "Submission process"? Then I'm good to merge!

tompollard:
Thanks! I'm happy with this. @bemoody if you're good that the issues you raised have been dealt with, please could you merge?

bemoody:
Sorry, Lucas, that internal link is still broken (points to
'#lincenses' rather than '#licenses').  Can you fix that please?

PR #343, opened 2019-04-05
About pages in new template

lbulgarelli:
All the pages inside the 'About' dropdown were group and adapted to the new template.


physionet-django/physionet/views.py
```diff
@@ -89,7 +72,19 @@ def contact(request):
...
-    return render(request, 'about/contact.html', {'contact_form':contact_form})
+    return render(request, 'about/about.html', {'contact_form':contact_form})
+
+def timeline(request):
+    """
+    Frequently asked questions
+    """
+    return render(request, 'about/timeline.html')
+
+def faq(request):
```
tompollard: if there's no template, we should remove or comment this

tompollard:
Thanks!

elfeto:
The create project should not go under the about.

Create project should be its own page, and not a cramped space in a button right corner.
The link of "submit" in the about page should redirect you to the login or to that page.

PR #348, opened 2019-04-08
Search database engine

lbulgarelli:
The search wasn't entirely working on different database engines.

### DIFF
--- a/physionet-django/search/views.py
+++ b/physionet-django/search/views.py
@@ -55,9 +55,12 @@ def get_content(resource_type, orderby, direction, topic):
         query = Q(resource_type__in=resource_type)
     else:
         topic = re.split(r"\W",topic);
-        query = reduce(operator.or_, (Q(topics__description__icontains = r'{0}'.format(item)) for item in topic))
-        query = query | reduce(operator.or_, (Q(abstract__icontains = r'{0}'.format(item)) for item in topic))
-        query = query | reduce(operator.or_, (Q(title__icontains = r'{0}'.format(item)) for item in topic))
+        query = reduce(operator.or_, (Q(topics__description__icontains = r'\b{0}\b'.format(item)) for item in topic))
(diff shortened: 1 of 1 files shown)


physionet-django/search/views.py
```diff
@@ -50,14 +51,18 @@ def get_content(resource_type, orderby, direction, topic):
     Helper function to get content shown on a resource listing page
     """
 
+    # Word boundary for different database engines
```
tompollard, line 54: Compound statements like this are discouraged in pep8 (https://pep8.org/)

tompollard:
This looks good to me, thanks Lucas. @bemoody, happy to merge? This is a fix to current functionality, so doesn't directly affect #349.

bemoody:
The changes look reasonable, but I'd appreciate if you would write a
little more about what the changes do and why you're making them.  It
looks like there are three entirely separate changes here:

- adding word-boundary constraints to some database queries that
didn't have them before

(Also, you changed two of these from "icontains" to "iregex", but not
the third; why?)

- changing the syntax from "\b" to "\y" depending on database server

- changing the ordering of results in some cases

lbulgarelli:
Sorry for not explaining the changes further @bemoody .

> adding word-boundary constraints to some database queries that
didn't have them before

Yes, that is right.

> (Also, you changed two of these from "icontains" to "iregex", but not
the third; why?)

This was actually something I missed, thanks for spotting it, I just made a commit to fix it.

> changing the syntax from "\b" to "\y" depending on database server

That is correct, for some reason postgresql uses a different representation for word boundaries.

> changing the ordering of results in some cases

Yes, previously when the search encountered a tie in the relevance sorting it would show the oldest resource first, and I think we want the opposite to happen, so by default the ties show the latest first now.

bemoody:
Looks good to me.

tompollard:
Good spot on the missed iregex!

PR #352, opened 2019-04-10
Google cloud

elfeto:
### Integration with the google cloud platform.

- Here I create altered the ".gitignore" file to exclude the credential file. (I might change the location afterwards.)
- The buckets are created with a special "service account" credentials in GCP. At the moment only the "service account" and the Lab google account can create "physionet.org" buckets. 
   - The buckets have to be lower case and end with "<slug>-<version>physionet.org". 
- Once a person click "I agree" on a DUA, all of their "capable emails" will be added to the bucket, the emails that wont work on google cloud will be rejected with a "Badrequest" error
- A button in the manage published project has been added to create a bucket and do a background task to send the files to GCP. Once the files are sent the model is updated.
- For a background task to run another Django command has to be run called "process_tasks". That command can be always running in the background or executed with a cronjob. 

**If a person wants to try this, they need the credentials file. Let me know**


physionet-django/physionet/settings/development.py
```diff
@@ -17,6 +18,8 @@
 DEFAULT_FROM_EMAIL = 'PhysioNet Automated System <noreply@dev.physionet.org>'
 CONTACT_EMAIL = 'PhysioNet Contact <contact@dev.physionet.org>'
 
+os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = os.path.join(BASE_DIR, 'PhysioNet-Data-credentials.json')
```
tompollard, line 21: When visiting the console and selecting the "Create GCP bucket" button for the demobsn project, I am taken to: http://127.0.0.1:8000/console/published-projects/demobsn/1.0/

The following error is displayed:

```
File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/google/auth/_default.py" in _load_credentials_from_file
  89.             'File {} was not found.'.format(filename))

Exception Type: DefaultCredentialsError at /console/published-projects/demobsn/1.0/
Exception Value: File /Users/tompollard/projects/physionet-build/physionet-django/PhysioNet-Data-credentials.json was not found.
```

It would be good to have a friendlier way of handling this. e.g. gray out button if credentials are not found or valid. It might be good to include a default/template version of the file in the repo if it isn't there already.

physionet-django/physionet/settings/staging.py
```diff
@@ -30,6 +31,7 @@
 DEFAULT_FROM_EMAIL = 'PhysioNet Automated System <noreply@staging.physionet.org>'
 CONTACT_EMAIL = 'PhysioNet Contact <contact@staging.physionet.org>'
 
+os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = os.path.join(BASE_DIR, 'PhysioNet-Data-credentials.json')
```
tompollard, line 34: Is anything done to protect the files on Google Cloud when people are playing around with the button on Staging?  If not, should we do something to prevent creation of junk buckets and overwriting of existing buckets? e.g. inserting "staging-" before the bucket name.

physionet-django/project/models.py
```diff
@@ -979,6 +979,7 @@ def publish(self, doi, slug=None, make_zip=True, title=None):
                     published_project.is_latest_version = False
                     project.save()
             previous_published_projects.update(has_other_versions=True)
+            print("previous_published_projects")
```
tompollard: Intended to be removed?

physionet-django/console/utility.py
```diff
@@ -0,0 +1,130 @@
...
+            else:
+                blob = bucket.blob(file)
+                blob.upload_from_filename(file)
+    if project.compressed_storage_size:
+        zip_name = project.zip_name()
+        chdir(project.project_file_root())
+        blob = bucket.blob(zip_name)
+        blob.upload_from_filename(zip_name)
+
+## Unused functions but usefull information.
```
tompollard: Does this comment apply just to `set_bucket_permissions` or to the following functions too? If it applies to the multiple functions, then best to add next to each function (otherwise its meaning may be unclear when people make changes in the future). Typo in "usefull"!

physionet-django/console/views.py
```diff
@@ -12,14 +14,16 @@
 from django.template import loader
 from django.urls import reverse
 from django.utils import timezone
+from background_task import background
 
 from . import forms
+from .utility import create_bucket, check_bucket, upload_files
```
tompollard: I know that we use relative imports elsewhere, but absolute imports seem clearer and are preferred by pep8 (e.g. here `from console.utility import create_bucket, check_bucket, upload_files`  and `from console import forms`)

physionet-django/notification/templates/notification/email/contact_administrators.html
```diff
@@ -0,0 +1,12 @@
+{% load i18n %}{% autoescape off %}
+
+Physionet Team take my greetings,
```
tompollard: When people reuse the code and don't know the background, this will seem weird...so sadly I think best not to include!

physionet-django/physionet/settings/staging.py
tompollard: Same goes for testers using the development system too, I guess. Would be nice if we could play around without risk of messing things up!

physionet-django/console/utility.py
```diff
@@ -0,0 +1,130 @@
...
+        policy[role].add('allUsers')
+    bucket.set_iam_policy(policy)
+    logger.info("Made bucket {} public".format(bucket.name))
+
+def remove_bucket_permissions(bucket):
+    """
+    Function to remove all permissions from bucket but owner 
+    """
+    policy = bucket.get_iam_policy()
+    To_remove = []
```
tompollard: To_remove -> to_remove

physionet-django/console/views.py
```diff
@@ -506,12 +525,32 @@ def manage_published_project(request, project_slug):
             project.make_zip()
             messages.success(request, 'The zip of the main files has been generated.')
 
+        elif 'bucket' in request.POST:
+            slug = request.POST['bucket'].lower()
+            if not check_bucket(slug, project.version): # Check if the bucket exists
```
tompollard: These comments might be clearer if not inline (pep8 discourages inline comments, I think).

physionet-django/project/templates/project/published_project.html
```diff
@@ -220,6 +220,20 @@ <h5 class="card-header">Versions</h5>
     </div>
     <!-- /.sidebar -->
   </div>
+
+  {% if project.gcp.sent_files %}
+    <h2 id="files">Files on Google Cloud</h2>
+    {% if project.access_policy %}
+      {% if has_access %}
+        <p>You can find the files uploaded at Google Cloud <a href="https://console.cloud.google.com/storage/browser/{{ project.gcp.bucket_name }}/">here</a>. Login with a google account is required.</p>
```
tompollard: Lowercase "google" appears a couple of times - we should switch to "Google".

tompollard:
Thanks Felipe, this is looking good. I added a few minor comments above, though haven't had a chance to play around properly...will need to ask for your help in setting up the credentialing file!

physionet-django/physionet/settings/development.py
elfeto: How about just removing the button if the credential file is not there, and display a message ?

physionet-django/physionet/settings/staging.py
elfeto: I adding a "testing-delete" if the system is not production

physionet-django/project/models.py
elfeto: Removed.

physionet-django/notification/templates/notification/email/contact_administrators.html
elfeto: HAHAHA, i forgot I left that there!

tompollard:
Nice work! Going to raise a few issues for related things to consider, like:

- update deploy and testing instructions to explain use of process_tasks
- should it be possible to activate/deactivate functionality that relies on external services (like this) via settings?
- could we include a report on process_tasks in the admin console? (e.g. a list of tasks that are currently running).

PR #354, opened 2019-04-10
Server config fixes

bemoody:
Felipe already fixed this issue on the live server but it was a serious one: /media was an alias that allowed unrestricted access to all supposedly-protected files.


tompollard:
thanks Benjamin

PR #355, opened 2019-04-10
Git configuration fixes

bemoody:
A couple of minor improvements to the Git configuration.

* Use diff=python so that function names show up properly in diffs

* Ignore various things that should be ignored


tompollard:
thanks!

PR #356, opened 2019-04-11
Fixtures for physiotools pages

cx1111:
- Add fixtures for most of the projects in Physiotools so that we can migrate them.
- Add some redirect urls.


cx1111:
To move the files:
```
import os
from project.models import LegacyProject, PublishedProject

lps = LegacyProject.objects.filter(resource_type=1)

for l in lps:
    l.publish(make_file_roots=False)
    p = PublishedProject.objects.get(slug=l.slug)
    os.mkdir(p.project_file_root())
    os.rename(os.path.join('/data/transfers/tools/{}'.format(p.slug)), p.file_root())
    p.set_storage_info()
```

PR #357, opened 2019-04-11
Landing page

lbulgarelli:
This is a redesign of the landing page:

- Big PhysioNet
- Added `The Research Resource for Complex Physiologic Signals.` as subtitle
- Data, software and challenges button
- Latest and news were moved to bottom of scrolling
- NIH funding added to footer


tompollard:
Nice work Lucas! I like it. Let's share this with Ary and Madalena tomorrow to ask for their thoughts (perhaps through the staging site).

PR #361, opened 2019-04-12
Create project in own page

lbulgarelli:
The `Create Project` section of `Publish` has its own page.


tompollard:
That was quick...thanks Lucas.

PR #364, opened 2019-04-12
Use default MIME types for known file types

bemoody:
For SVG files in particular, the MIME type must be set correctly in order for the file to be displayed as an image.


tompollard:
thanks benjamin

Issue #232, opened 2019-02-21
(Non-JS) browsing of subdirectories from project page


bemoody:
In https://alpha.physionet.org/content/crisdb/, there is a link
for each subdirectory ("e", "f", "m", "old") but these links
don't do anything without JavaScript.

bemoody:
The 'browsable-files' branch should fix this, but has the same severe security issue (issue #233) as the existing code.

bemoody:
I should probably mention that this code is also likely broken if
the subdirectory requires URL encoding or HTML encoding.

We probably want to forbid users from creating such directories
(e.g., with spaces, ampersands, semicolons, apostrophes, quotes,
less-than or greater-than signs), but the code should be designed
to do the right thing anyway.

(The current code is, I assume, already broken for the cases of
quote, apostrophe, ampersand, less-than, greater-than, and/or
backslash.  I assume that there's some way to do the requisite
escaping in Django templates, but I haven't read that far just
yet. :P)

elfeto:
As for now we only allow for folder creating or files the following syntax: letters, numbers, dashes, underscores, and dots. They may not contain adjacent dots, begin with, or end with a dot.

As for the 'browsable-files', I am allowing Nginx to index the published projects
https://alpha.physionet.org/static/published-projects/crisdb/files/

PR #367, opened 2019-04-13
Making subdirectories hyperlinked & browsable again

bemoody:
The fixes for issue #232 seem to have gotten lost at some point.  For example:

https://alpha.physionet.org/content/wctecgdb/1.0.0/ links to
https://alpha.physionet.org/content/wctecgdb/1.0.0/?subdir=patient001#files-panel

which just displays the same thing.

Here is a better fix for the issue: encode subdirectories as... well, subdirectories.  /content/wctecgdb/1.0.0/?subdir=patient001 becomes /content/wctecgdb/1.0.0/patient001/, which is what users expect to begin with.

In the case of published projects, some of the internal and metadata URLs are rewritten to avoid confusion.


tompollard:
Thanks Benjamin. It would be good to try to add some unit tests for this at some point, though not necessarily in the current set of changes. 

Without tests I worry that this kind of fix might get overwritten again, because we're (I'm!) less good at keeping an eye on no-javascript behaviour. Will try to do better...

bemoody:
It's quite understandable that it wasn't noticed, probably got lost
somewhere in the conversion to separate project versions.  I just
happened to notice late yesterday afternoon that it was broken, and
thought, hey, we ought to be able to do this in a more elegant way
than the way I had done it before.

Regarding unit tests, the existing tests mostly (entirely?) rely on
using django's reverse function, which means they're testing the URL
patterns that django knows about (not necessarily the URL patterns
that users are using or that are linked from elsewhere.)  That makes
sense for early development as many things are changing, but now that
things have mostly stabilized, we should probably start writing tests
to verify that the expected URL structure *hasn't* changed.  (Of
course, I say this having just made some changes to that structure...
:P)

PR #372, opened 2019-04-15
Remove restriction on project types for users

tompollard:
Minor change to remove the restriction on project types for users. At the moment this only applies to challenges, which I'd like to open up. We may want to implement a restriction on types in future, so I've left the `if not user.is_admin:` in place.


elfeto:
The challenges type doesn't allow to change the "Access Policy" nor the "License" before submission or at the copy edit part.

cx1111:
I'm doing the dynamic licenses/policies and removing this restriction too in my current branch.

PR #373, opened 2019-04-15
Add model project type

tompollard:
Following discussion around recent project submissions, these changes add model as a distinct project type. Some notes on the changes are below:

- Switched relative imports to absolute imports when I came across them
- Tidied spacing and page width in a couple of places
- Checks for PHI and for being "soundly produced" are now included on all projects.

Some points that led to the decision to treat models as a new project type:

- neither software nor data are an obvious fit for models.
- we will need to provide custom guidance around sharing of models.
- the project type will appeal to certain communities (e.g. machine learning).

Happy to chat about the need for this! The pull request should not be merged before https://github.com/MIT-LCP/physionet-build/pull/372


bemoody:
Thanks.

Since I gave Chen a hard time about this, I'll give you a hard time as
well: you've made a number of style fixes (whitespace,
indentation...), which do improve readability, but since they're not
related to the topic at hand they make it harder to see what you're
actually changing.  Please try to make such changes as a separate
commit.

List of licenses should be limited to software licenses unless
somebody can make a good case why they shouldn't.

I think I would rather see "Model" listed above "Challenge" in places
such as the project creation form or the search page.

Currently, if I try to create a "Model" project, I don't see *any*
options in the "Access Policy" and "License" fields
(/projects/xyzzy/access/).

tompollard:
Thanks for looking. I've responded to the points below:

> you've made a number of style fixes (whitespace, indentation...), which do improve readability, but since they're not related to the topic at hand they make it harder to see what you're actually changing.

Yep, my bad. It's sometimes difficult to resist tidying up bits and pieces, but in future I'll try to make the separation between formatting and functional changes clearer.

> List of licenses should be limited to software licenses unless somebody can make a good case why they shouldn't.

My reason for including data licenses is that many (most) of the people we surveyed consider models to be data rather than software. As a result I feel like we're better leaving the choice for the community, unless we have a strong reason to choose one over the other.

> I think I would rather see "Model" listed above "Challenge" in places such as the project creation form or the search page.

I'll update the order.

> Currently, if I try to create a "Model" project, I don't see *any* options in the "Access Policy" and "License" fields

At the moment, you'll need to run: `./manage.py loaddata project/fixtures/site-data.json`. I'll take a look at incorporating this into `loaddemo`.

bemoody:
(I'm not sure why this pull request still shows a yellow dot on
GitHub.  Looks like it passes all tests anyway.)

> My reason for including data licenses is that many (most) of the people we
> surveyed consider models to be data rather than software. For this reason I
> feel like we're better leaving the choice for the community, unless we have
> a strong reason to choose one over the other.

The strong reason is that, whether you want to call the original work
"software" or not, non-software licenses don't offer users a guarantee
that they can *use* the work as free software.

Software licensing is a well studied field, and the experts are pretty
clear about this: if you want to distribute free software, use an
established free software license.  The rule I've heard, and I think
it's a good one, is that your license should be one that has been
approved by (a) the Free Software Foundation, (b) the Open Source
Initiative, and (c) the Debian project.  If you don't meet those three
conditions, you are putting yourself and users in a state of legal
uncertainty.

Permitting other licenses "just because" doesn't benefit anyone.

tompollard:
> I think I would rather see "Model" listed above "Challenge" in places such as the project creation form or the search page.

Thinking about this a little more, I don't see a clean way of sorting the types to `['database', 'software', 'model', 'challenge']`. 

Unless you can suggest one, it feels like the best solution would be to sort alphabetically, either forward `['challenge', 'database', 'model', 'software']` or reverse `['software', 'model', 'database', 'challenge']`. Do either of these work for you?

tompollard:
I'm not sure I completely agree with your argument on licenses because if we are happy that the current data licenses are suitable for data (which presumably we are because are using them) and if we agree that models are sometimes data not software, then why wouldn't a data license sometimes be appropriate? Either way, it's easier to add later than to remove, so I'm happy to make software licenses the starting point!

The following updates address all of your points, I think, except the ordering issue. I'd prefer to merge now and worry about that later if possible:

- licenses and software languages were removed from the `demo-project.json` fixture as suggested in #429. 
- there is a new step in the `loaddemo` management command to load these fixtures from the `site-data.json` file. 
- licenses for models have been matched to those for software.
- unit tests were updated to include the site-data fixture.

bemoody:
Thanks.  Looks good to me!

PR #376, opened 2019-04-16
Add background image to home page

lbulgarelli:
Adds background image and gradient to home page layout.


tompollard:
Thanks Lucas, we'll see how this goes down!

PR #378, opened 2019-04-16
Added custom 404 and 500 page

elfeto:
### Added custom 404 and 500 error pages
The 404 only suggests to check the archive site.
The 500 will send a email to the admins specified in the settings file. That admin at the moment is the same as the contact person


elfeto:
Note... by Django documentation, for a custom error to be triggered, the server cannot be in debug mode.
That would also mean that the static content will be handled by your web server and not django.

tompollard:
Looks great, thanks Felipe.

tompollard:
Tested on production...works nicely. I triggered a 500 error by viewing a broken project link and received the traceback email as expected.

```
Internal Server Error: /content/challenge-2000/

DoesNotExist at /content/challenge-2000/
PublishedProject matching query does not exist.

Request Method: GET

...
```

Issue #242, opened 2019-02-25
WFDB-compatible directory/URL structure


bemoody:
URLs such as 'https://physionet.org/physiobank/database/mitdb/100.hea'
must be maintained.

Likewise 'https://physionet.org/physiobank/database/wfdbcal',
'https://physionet.org/physiobank/database/DBS' and possibly other
things.

/physiobank/database/ could in theory be a symlink to some other
location if you think that's prettier.  But, to make lightwave (and
future things) work in a sane way, we do actually want to have a
single directory containing

    DBS
    mitdb/RECORDS
    mitdb/100.hea
    mitdb/100.dat
    mimic3wdb/RECORDS
    mimic3wdb/30/3000003/3000003_0008.hea

and so forth...

bemoody:
It's also worth mentioning that for security, the main PhysioBank
directory must never contain data for unpublished projects, even if
restricted by filesystem or webserver permissions.

We want to know that if sandboxed-lightwave (or anything else: rsync
server, sftp server, bittorrent server, ...) chroots into that
directory, it will be incapable of accessing any private data.

I doubt this will ever be an issue but I figured it was worth pointing
out.

Issue #230, opened 2019-02-21
Public resources must be accessible via HTTP


bemoody:
http://alpha.physionet.org/static/published-projects/mitdb/files/233.hea
gives a 301 redirection to https://...

Public resources must be accessible via HTTP.  Although it's
arguably not a great idea for clients not to support HTTPS, it's
often impossible to do so in a way that's future-proof, let alone
in a way that will both provide security for normal users, and
also work correctly in fascist-firewall environments.

And for many (most?) users, encryption of public resources is
nothing but a needless hassle.

Issue #217, opened 2019-02-12
Handling data requests from wfdb packages


tompollard:
Various tools in the wfdb package allow people to download data from the servers. e.g. 

```py
>>> wfdb.dl_files('ahadb', os.getcwd(),
                  ['STAFF-Studies-bibliography-2016.pdf', 'data/001a.hea',
                  'data/001a.dat'])
```

We need to decide how to handle this (e.g. redirect requests to cloud servers and/or add fix to the wfdb packages). https://github.com/MIT-LCP/physionet-build/issues/174 seems related.

cx1111:
It's not wfdb's fault! :(

All joking aside, yea it needs to come with a more clear option to replace the origin. A guy actually requested this feature which I implemented a few months ago. We can do something similar with the new PhysioNet download API.

tompollard:
ha, sorry...fix was the wrong word!

PR #381, opened 2019-04-16
URL compatibility fixes

bemoody:
* HTTP access to public data (issue #230).

(It'd be better to move this configuration to a separate file
rather than duplicating it in both server blocks...)

* WFDB access (/physiobank/database/ alias) (issues #217, #242).

The /physiobank/database/ directory, and DBS file, should really
be managed by the project publication process, but in order to
quickly get things to a usable state, we can create the directory
by hand for now.

* /physiotools/ alias for existing PhysioTools files that need to
be kept where they are.

### DIFF
--- a/deploy/physionet_nginx.conf
+++ b/deploy/physionet_nginx.conf
@@ -26,6 +26,17 @@ server {
         alias /data/pn-static/published-projects;
     }
 
+    location /physiobank/database {
+        autoindex on;
+        alias /data/pn-static/physiobank/database;
+        location = /physiobank/database/ {
+            return 301 /data/;
+        }
+        location ~ /physiobank/database/([-\w]+)/$ {
+            return 301 /content/$1/;
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


cx1111:
In the long term, is using the DBS file the right method?
- We have a relational database now and can expose an API that gives a similar thing, without needing to try to sync with an additional flatfile.
- There are multiple versions of projects.

bemoody:
Chen: I agree - would be good for lightwave and similar applications
to be able to get detailed project/version information from the
database.

Obviously I'm partial to tab separated text files and similar things
that are easy to parse in any programming language - but it doesn't
necessarily have to be that format, or only that format.

Either way, though, ultimately the information needs to be serialized
into a file of some sort, and having it as a real file or files on
disk (rather than the application having to connect to a server) makes
the lightwave sandboxing MASSIVELY simpler.  Also makes it possible to
mirror the file and use the information without having access to the
master server.

PR #403, opened 2019-04-18
Regexp URLs should be anchored

bemoody:
URLs defined using 're_path' are not implicitly anchored; that is to say, re_path('foo', ...) will match any URL containing 'foo', not only the exact URL '/foo'.

In most if not all cases, this isn't what we want; so change all existing uses of re_path to use caret and dollar-sign at the start and end.


bemoody:
Follow-up to address a tangentially related issue: we should only redirect to a project if the project exists; otherwise we should show a 404.

tompollard:
good call, thanks

Issue #407, opened 2019-04-19
Crawler issue - Internal Server Error: /content/<PROJECT>/files-panel/


tompollard:
It looks like a Google crawler (66.249.73.92) is raising a server error:

```Traceback:

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/utils/datastructures.py" in __getitem__
77.             list_ = super().__getitem__(key)

During handling of the above exception ('subdir'), another exception occurred:

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/handlers/exception.py" in inner
34.             response = get_response(request)

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
126.                 response = self.process_exception_by_middleware(e, request)

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
124.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "./project/views.py" in published_files_panel
1107.     subdir = request.GET['subdir']

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/utils/datastructures.py" in __getitem__
79.             raise MultiValueDictKeyError(key)

Exception Type: MultiValueDictKeyError at /content/gmse/files-panel/1.0.0/
Exception Value: 'subdir'
Request information:
USER: AnonymousUser

GET: No GET data

POST: No POST data

FILES: No FILES data

COOKIES: No cookie data

REMOTE_ADDR = '66.249.73.92'
```

PR #408, opened 2019-04-19
files-panel: don't crash when required parameter is missing

bemoody:
This should address issue #407.

As a general matter, it's problematic to use 'request.GET[foo]' without catching the possible exception.

(Most other instances of 'request.GET[foo]' are in functions that are only accessible to logged-in users, so normally shouldn't be triggered by spiders.  Still...)

### DIFF
--- a/physionet-django/project/views.py
+++ b/physionet-django/project/views.py
@@ -1104,7 +1104,9 @@ def published_files_panel(request, project_slug, version):
     else:
         raise Http404()
 
-    subdir = request.GET['subdir']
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks, this seems useful. Should we also add a nofollow on the directories to the robots.txt? Feels like crawlers might not like us for 404s?

bemoody:
There shouldn't be any links (as such) pointing to these URLs.  I'm
guessing Google somehow detects URLs within inline scripts?

PR #409, opened 2019-04-19
Licenses support multiple resource types

cx1111:
Each license used to only be applicable to projects of one resource type. Now they can be assigned to projects of multiple resource types. ie. MIT license can be used for data and challenge projects.

Each license still has 1 access policy.


cx1111:
Btw if this gets merged, let me push to production. Need to fill in the new fields for existing licenses.

physionet-django/project/forms.py
```diff
@@ -686,42 +684,43 @@ class Meta:
...
-            if not include_credentialed:
-                self.fields['access_policy'].choices = ((0, 'Open'),(1, 'Restricted'))
-        # Software projects are all open
-        else:
-            self.fields['access_policy'].choices = ((0, 'Open'),)
+        # Get licenses for this resource type
+        licenses = License.objects.filter(
+            resource_types__icontains=str(self.instance.resource_type))
+        # Set allowed access policies based on license policies
+        available_policies = [a for a in range(3) if licenses.filter(access_policy=a)]
```
tompollard: is there a simple way to avoid the range? might be more flexible..

tompollard:
Looks great and works nicely for me. Only real issue I noticed is:

- The PhysioNet licenses at: http://127.0.0.1:8000/about/publish/ link to a private repo. If Ken and Roger are okay with it, we could make the repo public.

cx1111:
Changed the home pages to physionet. Removed the range.

tompollard:
thanks!

PR #411, opened 2019-04-22
Add the 'derived projects' functionality.

### LINKED ISSUE #347: 'Derived from' field
We can implement a 'derived from' field which should be either a foreign key, or a many to many field, which indicates that a project was derived or depends on another. ie. someone labels a bunch of annotations based on the MITDB signals, someone uploads code that uses mimic, mimic derived data....

cx1111:
Fixes #347

When submitting projects, you now have the ability to select parent project(s) that the current one is derived from. The parent/child projects will show in published projects.

Other changes:
- Show related publication at the top to cite
- Fix comma in css
- Make textbox for 'short_description' field bigger.

### DIFF
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -5,13 +5,14 @@ import re
 
 from django import forms
 from django.contrib.contenttypes.forms import BaseGenericInlineFormSet
+from django.db.models.functions import Lower
 from django.template.defaultfilters import slugify
 from django.utils import timezone
 from django.utils.crypto import get_random_string
 
 from .models import (Affiliation, Author, AuthorInvitation, ActiveProject,
     CoreProject, StorageRequest, ProgrammingLanguage, License, Metadata,
-    Reference, Publication, Topic, exists_project_slug)
+    Reference, Publication, PublishedProject, Topic, exists_project_slug)
 from . import utility
 from . import validators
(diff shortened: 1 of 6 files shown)


cx1111:
Sorry for push into PR. Fixed the server errors from reaching the sign_dua page when users have already signed it. Was missing a variable in redirecting them to the project page.

tompollard:
Thanks Chen, looks good. Before we merge, please could you update https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/templates/project/project_preview.html to match the published project template?

I think the two main things are:
-  move related publication to the header
- the "related projects" card should be added to the side bar

I also think we should also be listing the parent project/s as a suggested citation at the top of projects. I know it will lead to a large number of citations at the top, but it seems fairer on the authors of the original project.

I'm not too keen on the submission form for adding derived projects (it feels like it will be unwieldy when there are lots of projects, and it takes up a fair amount of space for the 1% of projects will use it). If you don't want to deal with this now, we can add raise this as an issue...

cx1111:
Thanks for checking. Just updated the project preview.

Yes the interface for adding parent projects is bad. This is supposed to be temporary.

tompollard:
Thanks! Added a couple of new issues about the points we discussed.

PR #417, opened 2019-04-24
LightWAVE

bemoody:
These patches implement a mostly working version of LightWAVE.

Things that are working:
- viewing public databases that are listed in the DBS file
- viewing active projects
- creating/editing annotations for public databases (saved anonymously to archive.physionet.org)

(Major) things that are not working:
- automatically maintaining the DBS file as new databases are published
- saving annotations in a semi-permanent place where you can easily find them
- running server tests as part of the automated test suite

To test, you'll need to compile and install the lightwave server.  This requires root privileges and a recent version of Linux.  (It should be possible to reconfigure things to use the non-sandboxed server for development, but I haven't done that.)

To install sandboxed-lightwave:
- compile and install WFDB in /usr/local
- install libseccomp-dev
- clone the lightwave repository
- run 'make sandboxed-server CGIDIR=/usr/local/bin'
...should work, let me know if it doesn't. :)

Let me know what you think.


tompollard:
Are we good to merge?

PR #418, opened 2019-04-25
More integrations with lightwave into site

cx1111:
- Replace relative static urls with absolute using django static tag
- And remove lightwave urls/views just for serving static content.

Another thing to consider: we already have jquery in the project. If we can use the same version for both, would save us having two copies. Also the jquery js files can be moved to the general static directory instead of the lightwave one.


cx1111:
My final wave of changes for lightwave:
- Lightwave server now queries all open-access databases for the `has_wfdb` flag to construct its list of available databases.
- Remove truncated file/database names shown in lightwave
- Add links to lightwave in project preview, published projects, project files for projects with WFDB, and the content index.

For a demonstration: the published demobsn project and test mitdb project (owned by rgmark) have test wfdb files and annotations.

cx1111:
After this is pushed to production, run:

```
import os

for p in PublishedProject.all():
    if os.path.isfile(os.path.join(p.file_root(), 'RECORDS')):
        p.has_wfdb = True
        p.save()
```

bemoody:
These changes are good in general, although I have a large number of
small gripes.

But to start with the big one: all the tests are broken!  Your first
commit breaks the tests for 'lightwave_js', and your later commits
break the server tests, as the expected files (wfdbcal and udb) can't
be found in the test environment.

(As I noted in my original comments, we should also include the server
tests in the automatic checks done by GitHub.)

Looking through these changes in order:


1. "replace relative static urls with absolute using django static
tag. Remove lightwave urls/views just for serving static content"

First of all, I appreciate the whitespace fixes, but it would be
better to make those as a separate commit.

The notion here (linking directly to the static resources without a
redirect) is not a bad one.  The redirects are an annoyance.  But I
used the redirects for a reason; namely, to keep things in sync with
the standalone lightwave client.  If you want to make changes here
(home.html and lightwave.js), please try to make changes that can
*also* be applied to the standalone client.

A better way to handle these resources might be:

- for the js and css paths, modify the paths in the HTML file as part
of 'make client'.

- for the *doc* paths, add a hidden link element and read its path
from lightwave.js.


2. "add parameter to track wfdb databases, and use this field in lightwave"

Please try to avoid adding cruft like the commented 'pdb.set_trace' statement.

I'm not sure whether changing PUBLIC_ROOT is a good idea or not.  The
static root should correspond exactly to the files that are publicly
accessible, which is why I set it that way.

(Ultimately, any files within the static root that are world readable
will be, well, *world* readable - anybody can read them using HTTP.
There is a question of whether we might have files in the static root
that are readable by the django user but not by the web server or
other services, but that seems like a bad idea.)

Would it be more secure to limit lightwave to the published-projects
subdirectory?  Maybe.  But it does limit flexibility: it means we
can't create symbolic links to places outside that directory, and any
other resources needed by lightwave in the future would have to be
stored *in* that directory.

Meanwhile, you changed PUBLIC_ROOT here for the development
configuration, but *not* for production.

About construction of dblist at runtime: I'm a little leery of polling
the database for *every single request* to the lightwave server.  This
may be silly of me.  But at the very least, it'd be sensible to do
this only for 'action=dblist' requests.

In any case, the database list should be sorted by project title and version.


3. "turn shortened db name into long"

This seems reasonable.  You broke the indentation in dblist(), though.


4. "add link to lightwave for projects with wfdb files. Add link to
example RECORDS ANNOTATIONS in docs"

You have some unrelated changes in the first two hunks of
project_preview.html.  Otherwise looks fine.


5. "add demo wfdb data"

Looks fine.


6. "change location of wfdbcal"

I guess this fixes one of the issues above (development-production
inconsistency.)

But please avoid stuff like this:

  DBCAL_FILE = os.path.join(settings.STATIC_ROOT if
settings.STATIC_ROOT else settings.STATICFILES_DIRS[0], 'wfdbcal')

(Pretty sure Python's syntax is designed to punish you for trying to
write complicated one-line statements, by forcing you to write them in
an incomprehensible order.  Please just write this as a normal
if...else.  And don't use more than 79 characters per line.)

In any case... wfdbcal needs to be at
http://physionet.org/physiobank/database/wfdbcal.  We might also want
it at /static/published-projects/wfdbcal.  It doesn't seem beneficial
to put it in yet another place.  And in this commit, rather than
moving the file, you've added a second copy of it.


7. "add links to lightwave"

Here you've made a number of unrelated changes (including a change
that would conflict with pull #425) as well as a lot of unrelated
whitespace fixes.

As for the relevant changes: please write the name (in
content_index.html and database_content.html) as "LightWAVE", not
"LightWave".

cx1111:
I'll fix the issues you pointed, except:

1. I don't agree with needing `home.html` to exactly match the original client file at: https://github.com/bemoody/lightwave/tree/master/client

This is a Django project. We have the right to adapt the code to suit it, despite how you set up the original LightWAVE project code. There is no `js` folder in the same directory as the base html file, which you have already noticed and tried to deal with. Inherently the structure is different. The right solution is not to write a bunch of unnecessary urls/views including tests which just make sure the redirection works.

Within this project, which we have used to build the entire site, templates go in their appropriate directories, js/css files go in their appropriate directories, and we embed urls to directly refer to their correct locations.

Using the {% static %} tag in `home.html` is the simplest solution. Another is to put the static js links in `lw-top.html`

2. Why does `wfdbcal` have to be accessible at http://physionet.org/physiobank/database/wfdbcal? The url? The location in the static root? We have the option to specify the file to be whichever one we want, since we are writing the views in this project.

cx1111:
Addressed the comments

1. Sticking with static urls as discussed. Added code in `loaddemo` to symlink to `wfdbcal` file from `/usr/local/database/wfdbcal` as agreed upon. Also a url to redirect physiobank/database/wfdbcal to the static file.
2. Reimplemented the public root and dbpath.
3, Show links to LightWAVE for individual databases.

bemoody:
OK.

There are still some changes here which have nothing to do with LightWAVE:

	--- a/physionet-django/search/templates/search/content_index.html
	+++ b/physionet-django/search/templates/search/content_index.html
	@@ -13,14 +13,8 @@ PhysioNet Index
	 {% block content %}
	 <div class="main">
	   <div class="main-side">
	-    <div class="card">
	-      <h2 class="card-header">PhysioNet Index</h2>
	-      <div class="card-body">
	-        <p class="card-text mt-1">This page contains all of the
databases, software and challenges published on PhysioNet.</p>
	-      </div>
	-    </div>
	     <form method="GET" action="" class="">
	-      <div class="card my-4">
	+      <div class="card">
	         <h2 class="card-header">Search</h2>
	         <div class="card-body no-pd">
	             {% include "inline_form_snippet.html" with form=form_topic %}
	@@ -30,7 +24,7 @@ PhysioNet Index
	         <h2 class="card-header">Resource type</h2>
	         <div class="card-body no-pd resource-search" id="type-check">
	             {% include "inline_form_snippet.html" with form=form_type %}
	-            <a
href="javascript:$('#type-check').find('input').prop('checked',
true);">Select all</a>
	+            <a id="selectall" href="#">Select all</a>
	         </div>
	       </div>
	       <div class="card my-4">
	@@ -81,3 +75,9 @@ PhysioNet Index
	     </div>
	 </div>
	 {% endblock %}
	+
	+{% block local_js_bottom %}
	+<script>
	+  $('#selectall').click(function(){
$('#type-check').find('input').prop('checke
	d', true); return false; });
	+</script>
	+{% endblock %}
	--- a/physionet-django/templates/about/database_content.html
	+++ b/physionet-django/templates/about/database_content.html
	@@ -156,12 +161,6 @@ alphabetically by the name of the database.
Those designat
	ed below as
	     </ul></li>
	 </ul>
	
	-<p><em>* The MIMIC II and MIMIC-III Clinical Databases are not public.
	-These databases are available to qualified investigators under the
	-terms of a data use agreement.
	-See <a href="http://archive.physionet.org/mimic2/mimic2_access.shtml">Getting
Access</a> for
	-details.</em></p>
	-
	 <br>
	 <hr>
	

I am fine with these changes to the search page although I wish you
would make them as a separate pull request.

@tompollard, what do you think about the change to database_content.html?

tompollard:
Thanks for checking @bemoody. I'm good with removing that sentence ("The MIMIC II and MIMIC-III Clinical Databases are not public"). 

It's almost impossible to keep this kind of content up to date (it already applies to many more db than just MIMIC). Next step should be for us to work out a better way of generating the content overview pages!

This reminds me @cx1111....you mentioned that you were in the process of implementing a tool for updating metadata. Think you'll be able to make a pull request for that before you finish?

tompollard:
(The reason I ask is that generating a content overview would be made much simpler if we could quickly scan through and update the project tags)

Issue #419, opened 2019-04-25
Order of author list is not conserved in project lists and suggested citation


tompollard:
The original order of authors should be retained wherever the list appears (for example, when projects are listed in search results and in the suggested citation that appears at the top of projects).

bemoody:
That's not good...

It looks like project authors have a 'display_order' field.  Is that
the order we should be using?  Is that field not being set correctly
in the database, or are views not correctly sorting the list based on
that field?

tompollard:
Haven't looked into the details yet, but pretty sure it's just a rendering issue.

tompollard:
(sorry, and yes, the "display_order" should be used for the sort)

bemoody:
On the database side, shouldn't we have unique_together
('project','display_order') ?

<https://docs.djangoproject.com/en/2.2/ref/models/options/#unique-together>

tompollard:
To ensure that the display order is unique for a given project? Probably, yes! I was just thinking about something similar...we dodged several issues around authorship for simplicity, one of which is shared authorship (e.g. equal contribution of first-authors). See @alistairewj's issue at: https://github.com/MIT-LCP/physionet-build/issues/369

We should think about how we want to handle this. It might be that the display order is still unique, but we have a flag to indicate equal contribution (then display an icon alongside equal contributors when displaying the author list).

PR #422, opened 2019-04-25
Sort authors in various places.

bemoody:
Anywhere that project authors are (or might be) displayed to the user, be sure to display them in the correct order.

In particular, this applies to the citation text (issue #419)


tompollard:
Looks great, thanks!

Issue #369, opened 2019-04-15
Add option for shared authorship


alistairewj:
We should allow for shared authorship of projects. Probably a check mark next to each author during submission for shared first author and another check mark for shared last author. I'm not sure if this needs to go into official DOI metadata - could probably just be displayed on the published page.

bemoody:
Seems a bit complicated.  Is this common?  How would this typically be
shown in a paper, or a citation?

alistairewj:
Fairly common. The author order is fixed and displayed normally in a
citation. On the publication, a symbol is placed next to author names and a
footnote indicates that they are shared authors. Alternatively, some
publications don't add the symbol, but just add the footnote. I like the
latter approach less.

>

tompollard:
We had a discussion about equal contributors and the outcome is that we either: 

1. Split the universe in these cases, so that each author who contributed equally has a reality where they are displayed first
2. Display the list of authors in 3D, so that equal contributors can be stacked (@cx1111's idea)
3. Randomly swap the position of equal contributors when displaying the author list, so that each person gets their special moment to appear first
4. Display a little symbol and footnote

We're leaning towards number 4 at the moment, just to stick with the crowd.

alistairewj:
Hm, not a big fan. Have you considered averaging the ASCII value of each
character across the shared names?

tompollard:
If you vote for number 2 then everyone gets a PhysioNet branded pair of these.

![WhiteFrame_Red-Cyan3D_937px_large_wide_test_2b292ff0-8e64-4ba3-9d71-66d3ebe03d3d_large](https://user-images.githubusercontent.com/822601/56768350-1e17e980-677c-11e9-93ef-0ac416612685.jpg)

PR #423, opened 2019-04-25
Enforce a strict ordering of authors at database level

bemoody:
For each project, there should be an unambiguous order in which authors' names are displayed.  (Notwithstanding issue #369, we need to display them in some order... well, unless we go with averaging the ASCII values, I suppose.)

This should enforce that ordering at the database level.  Making this a separate pull request from #422 because it requires testing & applying a migration.

### DIFF
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -30,7 +30,7 @@ class CorrespondingAuthorForm(forms.Form):
     def __init__(self, project, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.project = project
-        project_authors = project.authors.all()
+        project_authors = project.authors.all().order_by('display_order')
         self.fields['author'].queryset = project_authors
         self.fields['author'].initial = project_authors.get(is_corresponding=True)
         self.fields['author'].empty_label = None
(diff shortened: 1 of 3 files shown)


tompollard:
Both of the constraints seem safe to add:

```sql
-- check ('display_order', 'content_type', 'object_id') 
-- on class Author(BaseAuthor)

select (display_order, content_type_id, object_id), count(*)
from project_author
group by (display_order, content_type_id, object_id)
having count(*) > 1;

-- 0 rows (both staging and production)
```


```sql
-- check ('display_order', 'project')
-- class PublishedAuthor(BaseAuthor)

select (display_order, project_id), count(*)
from project_publishedauthor
group by (display_order, project_id)
having count(*) > 1;

-- 0 rows (both staging and production)
```

tompollard:
Migrations ran without an issue on both servers:

```
$ python manage.py makemigrations
Migrations for 'project':
  project/migrations/0018_auto_20190427_0218.py
    - Alter unique_together for author (2 constraint(s))
    - Alter unique_together for publishedauthor (2 constraint(s))

$ python manage.py migrate
Operations to perform:
  Apply all migrations: admin, auth, background_task, contenttypes, notification, project, sessions, user
Running migrations:
  Applying project.0018_auto_20190427_0218... OK
```

PR #426, opened 2019-04-26
"Decline" an invitation, don't "reject" it

bemoody:
When a user receives an invitation to be a co-author, give them the option to "accept" or "decline" the invitation, not "accept" or "reject".  "Reject" is a very strong word that seems inappropriate in this context.


tompollard:
good point, agree!

Issue #433, opened 2019-04-30
CSS on Lightwave app needs a clean up


tompollard:
On Lightwave (https://alpha.physionet.org/lightwave/), the footer is getting squashed into a scrollable box with a grey outline:

![Screen Shot 2019-04-30 at 14 17 17](https://user-images.githubusercontent.com/822601/56983875-c2ba7280-6b52-11e9-9dcb-4799a2abc919.png)

The footer should also be forced to the bottom of the page, even when the page itself contains no content, to retain the page structure.

bemoody:
Seems to be broken in aa02ef1db1cdc810268b9b055a152502aab64475 even
though it looks fine in both parent commits :/

tompollard:
Hmm...probably because the commit overlapped with the update to the footer style. The following chunk is definitely part of the problem, because the footer lives in a container too:

```
.container { border:2px solid #ccc; width:90%; height: 100px;
             overflow-y: scroll; }
```

bemoody:
Yeah, I see that.  Strange rule in lightwave.css that AFAICT isn't
actually used anywhere.

The other issue, though, is that the *content* is expected to have a
'<div class="container">' wrapped around it.  I'll add this to the
lightwave templates.  But if that's something that every page is
expected to do, then it should be part of 'base.html', not duplicated
in each individual page template.  (lightwave doesn't use base.html,
but I based the lightwave templates off of base.html to begin with.)

(It should perhaps be <main> rather than <div class="container">.)

tompollard:
`container` is one of the bootstrap components, so it makes sense to use it where possible: https://getbootstrap.com/docs/4.3/layout/grid/

bemoody:
Also, that "make the content large enough to fill the entire window"
rule seems overzealous; at least for me it makes the content TOO large
and forces it to always have a scrollbar.

bemoody:
...Except that it's also overloaded with further rules in
physionet.css.  So what exactly is a "container" and when should one
be used?

In the case of LightWAVE, we definitely do not want anything
restricting the width of the content.

PR #434, opened 2019-04-30
Avoid breaking footer CSS in lightwave

bemoody:
This fixes half of issue #433, and avoids putting the footer into a scrolled box.

The other half of issue #433 (stretching content to fill window) needs more careful thought.


tompollard:
thanks!

PR #437, opened 2019-04-30
Better "sticky footer"

bemoody:
Use the CSS3 flex property to allocate unused vertical space to the main content, instead of setting an ad-hoc min-height value.

(See https://philipwalton.com/articles/normalizing-cross-browser-flexbox-bugs/ for more than you wanted to know about the flex property.)

Moreover, set rules on the 'main' element rather than overloading the 'container' class.

This should fix the remainder of issue #433.


tompollard:
Thanks, looks like an improvement to me. There's now a white bar between the left panel and the footer. It would be nice to remove this at some point, but we can continue to work on styling.

Screen Shot 2019-04-30 at 18 10 48

PR #440, opened 2019-04-30
support restricted access licenses on models

tompollard:
Update the license fixture to allow restricted access licenses to be assigned to models.


tompollard:
Thanks, now added on the production and staging dbs.

PR #443, opened 2019-05-01
Add admin functionality to edit published projects and related items.

cx1111:
- The django admin panel (not the console we built) is now configured to be able to edit published projects and published authors correctly.
  - From a published_project, you can edit the relevant metadata, and its related publication, contact info, and set up to several authors. Unfortunately to set the affiliations of these authors, you'll have to go to the author object after you create it.
  - From the PublishedAuthor objects you create from the step above, you can directly set up to 3 affiliations.
- Refactor the code for tagging published projects. Call new functions in `ActiveProject.publish`
- Now in the `manage published project` page, there is a form to set comma separated topics for published projects. This both adds and deletes. To test this, open the published project, and open this in a new tab: http://127.0.0.1:8000/search/all-topics/ to check out the count, as you keep changing that comma delimited string.

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -268,7 +268,7 @@ class TopicForm(forms.Form):
                     'numbers, spaces, underscores, and hyphens only, and '
                     'begin with a letter or number.')
 
-        self.topic_list = [t.lower() for t in topics]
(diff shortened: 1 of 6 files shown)


tompollard:
Look great, thanks Chen! Now to fix up the project index pages...

Issue #435, opened 2019-04-30
Uncaught TypeError: Cannot read property 'getScreenCTM' of null


tompollard:
https://alpha.physionet.org/lightwave/?db=chfdb/1.0.0

```
Uncaught TypeError: Cannot read property 'getScreenCTM' of null
    at resize_lightwave (lightwave.js:1659)
    at dispatch (jquery.min.js:3)
    at elemData.handle (jquery.min.js:2)
```

PR #446, opened 2019-05-01
Fix lightwave resizing error

bemoody:
Fix a (mostly harmless) error in lightwave when the window is resized (issue #435).


tompollard:
looks good to me!

PR #447, opened 2019-05-02
Past news and RSS feed

bemoody:
Several news-related changes:

- Publish an RSS feed, compatible with the old PhysioNet RSS feed.  This requires a database migration to add the 'guid' field.  (After this is done, we may want to perform a second migration in order to make 'guid' the primary key for news items.)

- All of the old PhysioNet news items, converted to JSON format!  Should be possible to import these to the live system, once the above migration is done.

- A couple of minor tweaks to the news page.


physionet-django/notification/views.py
```diff
@@ -5,23 +5,36 @@
...
+                  {'year': 'Latest', 'news_pieces': news_pieces,
+                   'news_years': news_years})
+
 
 def news_year(request, year):
-    news_pieces = News.objects.filter(publish_datetime__year=int(year)).order_by('-publish_datetime')
+    news_pieces = News.objects.filter(publish_datetime__year=int(year)) \
+                              .order_by('-publish_datetime')
     # The year range of all the PN news in existence.
     # Yes, the start is hardcoded.
```
tompollard: Not directly part of the PR, but these should probably go!

tompollard:
This is excellent! Happy for me to load the fixtures on production? http://localhost:8000/console/news/ becomes a bit unwieldy with the large number of items, but this can be dealt with later.

Issue #448, opened 2019-05-02
Title style guidelines


bemoody:
When writing titles, contributors should be encouraged to:

- use standard English title case (most but not all words capitalized)

- not begin with the word "The"

- use a descriptive but brief phrase that can be used as a *name* for their work, rather than a mini-abstract describing precisely what they studied

It would be good if the project creation forms could gently or not-so-gently steer them in the right direction.

Issue #414, opened 2019-04-23
News titles limited to 40 characters


bemoody:
40 characters seems rather short for the maximum length of a news item title (see the one that Tom just posted...)

Also, does the form not complain if the user tries to enter something longer than the maximum?

tompollard:
> 40 characters seems rather short for the maximum length of a news item title (see the one that Tom just posted...)

I agree we should increase the length.

> Also, does the form not complain if the user tries to enter something longer than the maximum?

Not currently, but it would be nice if it did!

Issue #449, opened 2019-05-02
Error loading news fixtures on staging: psycopg2.DataError: value too long for type character varying(40)


tompollard:
Attempting to load the fixtures added in: https://github.com/MIT-LCP/physionet-build/pull/447 raises a `value too long for type character varying(40)` error on the staging server. Loading is fine locally, so this appears to be sqllite vs postgres issue:

```
python manage.py loaddata news-1999-2011.json
```

```
Traceback (most recent call last):
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/db/backends/utils.py", line 85, in _execute
    return self.cursor.execute(sql, params)
psycopg2.DataError: value too long for type character varying(40)


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "manage.py", line 22, in <module>
    execute_from_command_line(sys.argv)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/management/base.py", line 316, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/management/base.py", line 353, in execute
    output = self.handle(*args, **options)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/management/commands/loaddata.py", line 72, in handle
    self.loaddata(fixture_labels)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/management/commands/loaddata.py", line 113, in loaddata
    self.load_label(fixture_label)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/management/commands/loaddata.py", line 177, in load_label
    obj.save(using=self.using)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/serializers/base.py", line 219, in save
    models.Model.save_base(self.object, using=using, raw=True, **kwargs)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/db/models/base.py", line 748, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
[...]

PR #450, opened 2019-05-02
Increase maximum length of news item titles

bemoody:
Fix for issues #414, #448.


bemoody:
Whoops, I meant issue #449, not #448.

PR #451, opened 2019-05-02
Shared authorship

lbulgarelli:

It needs improvemente in the UX, so any suggestions are welcome.


elfeto:
- I don't know si this is part of the intended use, but if you have 4 authors and 2 contributed equally, and the other 2 also contributed equally, then it will jump in numbers 1 to 3.
- If the persons didn't contributed equally, then they don't have a number. That means they didn't worked at all ?
- The contributed equally is tied tot he position, and not the authors.  So, assuming that the most important are the first and last authors, and they contributed equally, it cant be added like that.
- If you have 4 authors, the second "add equal contribution" and "remove equal contribution" button works for the second and third position at the same time.

PR #455, opened 2019-05-02
Remove gaps

lbulgarelli:
This removes blank gaps between the navigation bar/footer and the page content.


bemoody:
Yeah, in pull #437 Tom pointed out that I had screwed up the
top/bottom of some pages.

On the other hand, there are other pages (e.g. /content/demobsn/1.0/)
that now don't have any space between the bottom of the nav bar and
start of the content.  Should we add another class to the main content
div in that case?

lbulgarelli:
@bemoody I've just pushed a commit that adds a bootstrap class for y-axis margin to both lightwave and project content pages.

Also, I've fixes some issues with padding and margin when the project content is displayed in small screens.

Let me know if there is any other place you remember that needs this kind of adjustment.

PR #456, opened 2019-05-02
Add search to news console

lbulgarelli:
Adds a dynamic search bar to the news console.

Also, embeds the url in the title instead of showing it separately.


bemoody:
Slick!  Looks good to me.

PR #459, opened 2019-05-03
Add margin to container

lbulgarelli:
I noticed some pages continued to lack top/bottom margin, and that they were all wrapped in a div with the container class, so I've added margin to that class.


tompollard:
Thanks Lucas

PR #460, opened 2019-05-03
Remove most uses of target=_blank

bemoody:
This started out with me being irritated that when I want to edit a project, and also see a preview of the result in a separate tab:
 - I'm viewing /projects/hgiavnfvklg/metadata/ (tab #1)
 - I middle-click the "Proofread" button, which opens /projects/hgiavnfvklg/proofread/ in tab #2
 - I click the "View Project Preview" button, which opens /projects/hgiavnfvklg/preview/ in tab #3
Tab #2 is completely worthless and only serves to get in my way.

But in any case, target=_blank is an annoyance in general.  Most web browsers give you a way to open a link in a new tab/window if that's what you want to do.  Most browsers *don't* give you a way to open a link in the current tab/window if target=_blank is specified.

I've left alone pages in the console, which users don't see and I don't much care about for now; and I've left alone links on some pages with complex forms, where clicking the link might result in data loss.  (target=_blank is still probably not the best way to deal with this issue.  And note that the prime example - the Metadata page - doesn't follow this pattern.)

### DIFF
--- a/physionet-django/notification/templates/notification/news.html
+++ b/physionet-django/notification/templates/notification/news.html
@@ -28,7 +28,7 @@
           <div class="card-body">
             <ul class="list-unstyled mb-0">
               <li>
-                <a href="https://github.com/MIT-LCP/" target="_blank" style="text-decoration: none; color: #000000;"><i class="fab fa-github"></i> Github</a>
+                <a href="https://github.com/MIT-LCP/" style="text-decoration: none; color: #000000;"><i class="fab fa-github"></i> Github</a>
               </li>
             </ul>
           </div>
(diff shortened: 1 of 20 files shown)


tompollard:
Thanks @bemoody, I agree that it is cleaner to remove most of these new window targets. As discussed, if we drop the new window for project previews it would be nice if the user had a way of getting back to the previous page (without relying on the browser <-).

bemoody:
What do you think of adding a button to the top of the preview page, like this?

elfeto:
Looks fine for me.... @tompollard  ?

tompollard:
There is now the issue that the "Return to project overview" button is broken for admin users that visit the preview via the console (to test, login as admin user, go to a project in http://localhost:8000/console/unsubmitted-projects/, then click on the project). 

This only affects ~8 or so admin users though, so it's a minor bug that could be raised as an issue to deal with later.  Overall, I'm happy to merge...

Issue #453, opened 2019-05-02
Improve user interface for user console


tompollard:
@lbulgarelli the interface for managing users could also do with some improvement! At the moment we just display a loooong unsorted list (https://alpha.physionet.org/console/users/all/). Perhaps add search for this too?

PR #461, opened 2019-05-03
Console changes

lbulgarelli:
Couple of changes to the admin console:

- Add search to user list #453  
- News search only reloads the table content instead of entire table
- Renamed both user and news related files to make easier to navigate when developing


tompollard:
This is excellent, thanks Lucas. I've added a couple of small changes:

- fixed a reference to the renamed `console_news`
- updated `ifequal` syntax in the templates, which will be deprecated: https://docs.djangoproject.com/en/2.2/ref/templates/builtins/

> ```ifequal and ifnotequal```
> ```{% ifequal a b %} ... {% endifequal %} is an obsolete way to write {% if a == b %}```
> ```... {% endif %}. Likewise, {% ifnotequal a b %} ... {% endifnotequal %} is ```
> ```superseded by {% if a != b %} ... {% endif %}.  The ifequal and ifnotequal tags ```
> ```will be deprecated in a future release.```

- fixed active menu highlighting for http://127.0.0.1:8000/console/published-projects/ 

If you are happy with these changes, please could you merge the branch?

We may want to change the user search to look at names, rather than usernames (or as well as usernames), but we can deal with this later!

PR #462, opened 2019-05-06
Fix username and training_report URL in credential console

tompollard:
Fixes a couple of issues with the credentialing form on the admin console:
- correctly display the username of the applicant.
- fix the url of the completion report.

Also reorganizes the template slightly to:
- display headers for the content (status; personal details; training etc).
- moves status information to the top of the card.


elfeto:
Looks good.

Issue #431, opened 2019-04-30
Rotation of uwsgi log on staging


bemoody:
Seems that currently the uwsgi log is not being rotated on the standby server, and is currently 19 GB (!)

PR #464, opened 2019-05-07
Added log rotation config

elfeto:
Added the config to rotate the logs and separated the html requests from the service logs
#431


bemoody:
Looks good.  Thanks.

PR #467, opened 2019-05-08
Revert "Shared authorship"

elfeto:
Reverts MIT-LCP/physionet-build#451

@tompollard @bemoody  any thoughts?


tompollard:
I think it's a good idea to revert for now...could do with a bit more testing.

PR #472, opened 2019-05-08
Shared authorship

lbulgarelli:

- `BaseAuthor` and `PublishedAuthor` now have `shared` field, which stores a number corresponding to a group of authors that contributed equally
- `share` and `unshare` functions were created so the submitting author can add equal contributions in the project submission, more specifically in the author list `author_list.html`.
- Logic for moving authorship order was changed to account for grouped authorship
- Added equal contributions to project preview and published projects
- Fixed the issue in the content list, as in the `Find` page, where it was not showing the authors in the correct order


bemoody:
I haven't been following this, so I have no comments on the substance of the pull request, sorry.

However, it seems that GitHub isn't going to automatically *re-check* this branch now that pull request #479 is merged.

So please do not merge this until migration issues are resolved (easiest way is probably to rebase.)

bemoody:
In d397fd2a1f6a5b01c369afe186b2a55dbc96ad3e you added some junk migrations that don't relate to your branch in any way.  Please try again.  The following commands may help:

    git reset c7d696b9f918aec82d89004456a02d77ae5bfc2e
    rm */migrations/*.py
    git checkout */migrations/
    ./manage.py makemigrations
    git add */migrations/

Also, it's impossible to review this when all you give me to go on is 'Revert "Revert "Shared authorship""'.

Please start over, and explain in detail what changes you're making, and why.  Break up your changes into logical commits, with a detailed description of each.

lbulgarelli:
I've fixed the commits and rebased.

It would be good if you could review this and give feedback if something's wrong.

elfeto:
Any comments on this? good? bad? cancel? proceed?

tompollard:
We should keep this open - it's definitely functionality that we'd like to have - but I'd like to focus on other things for the moment (those things that we need to get done to switch the site live, plus non-PhysioNet tasks!). If Benjamin has time to take a look, it would be good to hear his thoughts.

lbulgarelli:
I don't like this siting here, can you please give feedback so we can launch this feature?

This is actually important to authors.

PR #477, opened 2019-05-09
Fix share page breaking in resized window

lbulgarelli:
Removes unused css that was breaking the page when it was resized.


tompollard:
Looks good, thanks Lucas.

PR #478, opened 2019-05-09
Fix padding and word break in projects

lbulgarelli:
Fixes padding and sets word break to happen only when needed to prevent overflow


tompollard:
Nice, thanks!

PR #485, opened 2019-05-10
Altered credentialing GUI

elfeto:
Altered the page for credentialing.
- removed dropdown for response
- made it similar to the current site


physionet-django/notification/utility.py
```diff
@@ -374,17 +374,21 @@ def mailto_supervisor(request, application):
     return mailto
 
 
-def mailto_process_credential_complete(request, application):
+def mailto_process_credential_complete(request, application, include=True):
```
tompollard: The purpose of the `include` argument isn't clear here. Might be helpful to:
- describe the argument/s in the docstring (styled like https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)
- rename `include` to something that more clearly describes its role.

elfeto: The include is to include the applicant comments in the email mailto function.

tompollard: Looks like it also controls whether or not "Dear NAME" is added to the start of the email template?

elfeto: That is correct, KP has a pre-made email in which he copy pastes there. So One is for acceptance, and the other is for communication. Because of that pre-made email text that he copy pastes, I thought it was easier to just leave it all blank and allow him to be him

tompollard:
Thanks Felipe. Added a couple of thoughts below:

- I worry a bit about how we are handling communication between the person doing the credentialing and the applicant. We are relying on a manual email after clicking "Contact reference" or "Accept" or "Reject". If the email isn't sent for whatever reason, then the system will record the event as happening (e.g. the reference was contacted) even though it did not.
- If the application is accepted, shouldn't we just automatically trigger an "your application has been accepted" email to the applicant? If a tailored email is necessary, it could be sent separately.
- If an application is rejected, I'm concerned that the emails may not be sent, leading to lack of clarity on both sides.

In terms of design/layout there are some improvements, but overall it feels like it would be better to make minor tweaks to the layout without a complete restructure. (e.g. why remove the "Status" header, the "Training" header etc? ).  For reference, both versions are displayed below:

Current:

Screen Shot 2019-05-14 at 09 44 52

Updated:

Screen Shot 2019-05-14 at 09 44 21

@bemoody any thoughts?

physionet-django/notification/utility.py
tompollard: In that case, perhaps better to have as two separate arguments? e.g. something like:

`salutation=True, comments=True`

elfeto: I changed include to comments, but salutation is always there, so there is no need to create a extra variable.

elfeto:
@tompollard  The three points above go to the same point and was fixed in that last commit in which on approval or rejection a email is sent. Also the removal os the things were at request of KP.

bemoody:
On 5/14/19, Tom Pollard <notifications@github.com> wrote:
> - I worry a bit about how we are handling communication between the person
> doing the credentialing and the applicant. We are relying on a manual email
> after clicking "Contact reference" or "Accept" or "Reject". If the email
> isn't sent for whatever reason, then the system will record the event as
> happening (e.g. the reference was contacted) even though it did not.

One thought: we could set up the email template to send a 'Bcc' to
some address like
'credential-reference+7Z0ncQn32kKRPaXdNCz7@alpha.physionet.org', and
set up a script to mark "reference contacted" when and if that mail is
received.  (We could also save the date, message-ID, or even the full
text of the message, if desired.)

We could also, of course, have a way to manually mark as
"contacted"/"not contacted", but this would be a way to improve
automation without interfering with Ken's existing email workflow.

(I don't know offhand how to set up a custom mail address handler like
that, but I'm sure it's not too difficult!)

elfeto:
At the moment, once he clicks to generate the `mailto` email, the applications gets marked as reference contacted, also there is a automatic BCC to `noreply@alpha.physionet.org` that can be replace for what @bemoody suggests. But I encouraged KP to set the content he wanted in the email generated, and it would be copied over to the `mailto`, and then we would have a copy of what he wrote.

Either way, for someone to mark a application as declined, one has to insert a reason, or it wont allow you to.

bemoody:
To me, the page (both versions) looks difficult to read, and if I were
going to use it I'd want to rearrange it (reduce the amount of
boilerplate text, add some padding so the non-boilerplate information
stands out more.)  But (fortunately?) I'm not the one who has to
manage these applications.

It looks like the "state/province" field is required, regardless of
the country; is that intentional?

There's also "My PhysioNet email", which is an odd way to refer to the
applicant's primary email address.  Also worth keeping in mind that
the applicant may *change* their primary address, and it's not
explicitly listed on the application form - perhaps the form should
include a menu to select one of their verified addresses, just so that
they know they might be contacted about the status of their
application.

tompollard:
> To me, the page (both versions) looks difficult to read, and if I were
going to use it I'd want to rearrange it (reduce the amount of
boilerplate text, add some padding so the non-boilerplate information
stands out more.)  But (fortunately?) I'm not the one who has to
manage these applications.

The codebase is public and the goal is for the platform to be reusable by others, so it is good to think about how the system generalizes beyond PhysioNet/LCP.

physionet-django/notification/templates/notification/email/mailto_contact_applicant.html
```diff
@@ -0,0 +1,43 @@
+{% load i18n %}{% autoescape off %}
```
tompollard, line 1: The License/DUA shouldn't really be hardcoded into the email. Makes it tricky to keep the email in sync with the system documents.

tompollard:
thanks! we'll address more dua/license issues in the next pull request.

PR #487, opened 2019-05-11
Consistent more button

lbulgarelli:
The `More resources` button on the home page was not consistent with `More news`.

To fix this I added the button inside both resources boxes.

### DIFF
--- a/physionet-django/search/templates/search/content_list.html
+++ b/physionet-django/search/templates/search/content_list.html
@@ -31,4 +31,3 @@
     <hr>
   </div>
 {% endfor %}
-<br>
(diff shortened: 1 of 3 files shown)


elfeto:
Why have a more resources button for  featured projects that leads to the same content of latest resources ?
I mean... I am not against it, but why ?

elfeto:
There can also be added a page for only featured content (if you want the button), and add it to the search box alongside search by project access type (open, restricted, credentialed)

bemoody:
I wouldn't put a second button under "featured projects"; there aren't any "more" featured projects.

The style fix is a good one.  Can we also get rid of the 'br' at the end of content_list.html?

bemoody:
Looks good, thanks.

PR #489, opened 2019-05-13
Altered error for urls with @ to 404

elfeto:
Altered the error for invalid paths.
If the path contains a `@` it will return a 404 error instead of a 500 Invalid path.

### DIFF
--- a/physionet-django/project/validators.py
+++ b/physionet-django/project/validators.py
@@ -2,7 +2,6 @@ import re
 
 from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext as _
-from django.http import Http404
 
 def validate_filename(value):
     """
@@ -35,8 +34,6 @@ def validate_subdir(value):
     or empty. No consecutive dots or fwd slashes.
     """
     if not re.fullmatch(r'[\w\-\./]*', value) or '..' in value or value.startswith('/') or '//' in value:
-        if '@' in value:
-            raise Http404()
         raise ValidationError('Invalid path')
 
 
--- a/physionet-django/project/validators.py
+++ b/physionet-django/project/validators.py
@@ -2,7 +2,7 @@ import re
 
 from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext as _
-
+from django.http import Http404
 
 def validate_filename(value):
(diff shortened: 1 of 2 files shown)


bemoody:
No, this is terrible.  Why on earth should '@' be any different from any other forbidden character?

What URLs / views are returning 500?  Can you change *those* functions to catch the ValidationError and raise a 404 instead?

elfeto:
So, I undid the first commit, and added a ValidationError to display a 404 page on the function that gets the project file info

bemoody:
Can you give examples of the specific URLs that are currently giving
500 errors and should now give a 404?

elfeto:
For example links like this: 
https://alpha.physionet.org/content/mimiciii/1.4/a@a/ 
I know that link doesn't exist, but for some of the challenge I changed the @ for _at_, this is more for in the future, if we forget to do that challenge assuming that we move the files manually instead of via the site.

bemoody:
Thanks.  Good to document exactly what we're talking about.

But as for the code, this feels like a layering violation and it
bothers me.  get_project_file_info doesn't relate to HTTP; why should
it be raising "HTTP" exceptions?

After thinking about it a bit, I've convinced myself that what we
ought to do here, instead, is to set 'file_error' to 'Invalid path' or
some such.

The reason I made that function return an error message rather than
raising an exception is so that the user interface (esp. the
JavaScript user interface) remains functional and responsive in error
conditions.

For example, try creating a directory called
'static/published-projects/demosn/1.0/abc@def'; then, if you load
/content/demobsn/1.0/, you'll see a link to 'abc@def', which (if
clicked with JS enabled) will do *nothing*.

On the other hand, if you create a directory called 'abc' and set its
permissions to 000, clicking the corresponding link will show an error
message.

And this is the point about layering violations: whereas it makes
sense for '/content/mimiciii/1.4/a@a/' to return a 404,
'/content/mimiciii/files-panel/1.4/?subdir=a@a' needs to return a 200
and some valid HTML.  (Of course, the script that requests files-panel
could, and perhaps should, be made more robust, but that's another
story.)

elfeto:
thanks @bemoody

PR #492, opened 2019-05-14
Console scrolling and footer

lbulgarelli:
This fixes the following in the admin console:

- Footer was hidden by left navigation menu
- Some pages would have multiple scrollbars
- Navigation menu in small screens had bullet points and did not scroll
- Left navigation menu would not scroll x-axis when window was resized


bemoody:
Well, I have to say that having scrollbars (esp. the horizontal
scrollbar) for the left panel is not pretty.  But this is better than
before, and as a non-user-facing page I'm not really concerned about
it being pretty.

PR #496, opened 2019-05-15
redirect from console

### LINKED ISSUE #493: /console/ infinite redirection loop
If logged in as a non-admin user, /console/ redirects to /login/?next=/console/ and /login/?next=/console/ redirects to /console/.

elfeto:
Redirects a non admin user that tries to go to the console

Fixes #493


bemoody:
This is ugly, but still an improvement.

Issue #469, opened 2019-05-08
Clean up relationship between data use agreement and license


tompollard:
When a user chooses to download protected content, they need to sign a data use agreement. The license should reference the data use agreement, rather than restating each of the points in the data use agreement. The data use agreement should be included alongside the license in the file package.

elfeto:
This was addressed right ?

tompollard:
Yep, I think this can be closed.

PR #505, opened 2019-05-20
Display DUA text instead of license text

tompollard:
Displays the DUA text instead of the license.


tompollard:
ref #469. thanks!

Issue #295, opened 2019-03-20
Not clear that authors need to approve final version before publication


tompollard:
Once the project has been copyedited, all authors receive an email to say the project has been accepted. It is not clear from this email that to proceed with the process every author needs to approve the project for publication. We should either: 

1. clarify that all authors need to approve the project, or
2. allow the submitting author to approve the project on behalf of all authors.

> Project title: TITLE
> Submission ID: ID
> Submitting author: NAME
> 
> Dear NAME,
> 
> Based on the recommendations of our editors, we are delighted to accept your project entitled "PROJECT TITLE" for publication on PhysioNet. 
> 
> A summary of our standard set of editorial checks is listed below:
> - Has the data been produced in a sound manner?: Yes
> - Is the data adequately described?: Yes
> - Is the data provided in an open format?: Yes
> - Are the data files machine-readable?: Yes
> - Does the project include everything needed for reuse by the community?: Yes
> - Is the data free of protected health information?: Yes
> - Is the content suitable for PhysioNet?: Yes
> 
> Additional comments from the editorial team are listed below:
> COMMENT.
> 
> After copyediting of the project and files is complete, we will be in touch again. Then, once all authors approve of the copyedited version at: http://THEURL, the project will be ready for publication. 
> 
> Regards,
> 
> The PhysioNet Team,
> MIT Laboratory for Computational Physiology,
> Institute for Medical Engineering and Science,
> 45 Carleton St, Cambridge, MA 02142

Any preference for 1 or 2? Remember that by this point, all authors have already approved the initial submission.

elfeto:
That email that you show there, is before the copy edit phase. 

After the copyedit they receive this email:
In which I think is a bit clear that if they are a author, they have to approve the submission.

> Project title: TITLE
> Submission ID: ID
> Submitting author: NAME
> 
> Dear NAME,
> 
> We are pleased to say that copyediting of your project entitled "PROJECT TITLE" is now complete. We were satisfied with the state of your project and no changes were made.
> 
> When all authors have approved this version of the project using the online system, it will be ready for publication on PhysioNet.
> 
> Regards,
> 
> The PhysioNet Team,
> MIT Laboratory for Computational Physiology,
> Institute for Medical Engineering and Science,
> 45 Carleton St, Cambridge, MA 02142

We can make this even more obvious tho...

tompollard:
[...]

Issue #327, opened 2019-04-01
Extend notifications panel in "Project Home" to include projects awaiting approval


tompollard:
Currently, the notifications panel in Project Home (http://127.0.0.1:8000/projects/) only indicates outstanding invitations to join a project. 

![Screen Shot 2019-04-01 at 18 16 59](https://user-images.githubusercontent.com/822601/55363310-9f949880-54aa-11e9-8c10-398214e04e25.png)

We could extend the panel to include other notifications. For example, it should list projects that are awaiting approval for publication.

PR #506, opened 2019-05-20
Added outstanding Tasks

elfeto:
Merged all todo tasks into one table outstanding tasks


bemoody:
Looks pretty good, but what are the different fields for?

AuthorInvitation already has 'request_datetime' (inherited from
BaseInvitation).  'creation_datetime' and 'invitation_datetime' seem
wholly redundant (and neither field is actually used here.)

Then there's 'reply_datetime', which I guess was meant to record when
the author responded to the invitation, but there's already
'response_datetime' for that, and 'reply_datetime' is never used or
set.

Did you mean to add these fields, and what would they be used for?

elfeto:
So the idea under 'invitation_datetime' was to send another invitation X days after, the initial invitation, assuming the person hasn't answered... (but left unfinished...)

Interesting enough, I used 'request_datetime', and not the variables I added, maybe I was going to add them, and then saw that they existed, and forgot to remove them... 

The only one "needed" is 'creation_datetime' of the author object, that is to be able to have some sort of idea, since when one is author, and might be missing the affiliation. 

I'll fix this in the morning :)

bemoody:
Thanks.  I see another issue, though: in project_home.html line
33, you have '{% if invitation_response_formset|length %}'
preceding the table headings, so the headings are not shown if
you have no invitations (but you do have other tasks.)

Please also remove the commented-out cruft in lines 34-36.

This should be formatted as a `<table>` rather than a bag of divs.
But, fine to leave it the way it is for now and fix that in a
later pull request.

elfeto:
Thanks for that catch... This PL closes #295 #327

bemoody:
Made some style fixes.  Okay with you?

elfeto:
Thanks!

PR #508, opened 2019-05-20
Search button

bemoody:
Felipe, in commit aaa3fb7174a14aeeb5177dad6e926373cb64ebc0 you undid my changes, without giving any explanation or making a pull request.  Please don't do that.

### DIFF
--- a/physionet-django/templates/navbar.html
+++ b/physionet-django/templates/navbar.html
@@ -15,7 +15,7 @@
 		<form class="form-inline" action="{% url 'content_index' %}">
 	    	<input name="topic" class="search-input" type="text" placeholder="Search">
 		    <span class="input-group-btn">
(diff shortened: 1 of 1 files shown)


elfeto:
Sorry for that, but the search bar looked like this. Thought it was a simple fix, but apparently I got the icon too close to the search bar. 
Screen Shot 2019-05-20 at 6 42 12 PM

bemoody:
What browser?  Did you clear cache?

elfeto:
So it seemed fine when i tried it in my local pc, and it looks fine now again. 
Before it was in chrome, and I believe I did, but might have done something wrong.

Lets try again :D and see if its all in my machine

elfeto:
Seems fine, sorry about that.... might not have cleared the cache apparently :(

PR #509, opened 2019-05-20
Return DUA as plain text for emails. Ref #469

tompollard:
Adds a new method to the license model to return the DUA text as a string. This method is needed for displaying the DUA in plain text emails (requested by KP).

Before this change is merged, @elfeto please could you call the method in the emails as discussed earlier (`dua_text_content()`)?

A couple of notes:
- this adds a new dependency on html2text. I tried manually parsing the HTML but it gets pretty hacky, particularly if we want to number the items in the list.
- I removed some junk characters from the HTML in site-data.json, and will update the live server when necessary.


tompollard:
Thanks Felipe, looks good.

Issue #511, opened 2019-05-21
Add process for handling migration files to the contribution guidelines


tompollard:
@bemoody when you have a chance, please could you add a short note about handling migrations to the contribution guidelines at https://github.com/MIT-LCP/physionet-build ?

tompollard:
Dealt with in https://github.com/MIT-LCP/physionet-build/pull/513

PR #513, opened 2019-05-21
Some documentation about how to create migrations

bemoody:
Good advice?  Bad advice?  Let me know what you think.  Addresses issue #511.


tompollard:
Great advice

PR #514, opened 2019-05-21
changed inactive account login message

### LINKED ISSUE #507: "This account is inactive" message displayed to new registrants who have not validated their email address
It sounds like new users see a "This account is inactive" message if they have registered on the system but not yet validated their email address. The message should be clearer, explicitly asking the user to check their email.

elfeto:
Changed inactive account login message 
Closes #507


tompollard:
thanks!

PR #517, opened 2019-05-22
fixed order of applications in Ongoing Applications

elfeto:
The application order was fixed in the "Management" but not in "Ongoing Applications"

### DIFF
--- a/physionet-django/console/templates/console/complete_credential_applications.html
+++ b/physionet-django/console/templates/console/complete_credential_applications.html
@@ -6,70 +6,70 @@
 
 {% block content %}
 <div class="card">
-  <div class="card-header">
-    {% with c=contacted_applications|length n=not_contacted_applications|length %}
+  {% with c=contacted_applications|length n=not_contacted_applications|length %}
+    <div class="card-header">
 
       Outstanding credential requests  <span class="badge badge-pill badge-info">
         {{c|add:n}}
-    {% endwith %}
-    </span>
-  </div>
-  <div class="card-body">
-    <div class="tab-content">
-      {% if contacted_applications or not_contacted_applications %}
-        <div class="table-responsive">
-          <table class="table table-bordered">
-            <thead>
-              <tr class="header">
-                <th>#</th>
-                <th>User</th>
-                <th>Full Name</th>
-                <th>Email</th>
-                <th>Application Date</th>
-                <th>Reference Contact Date</th>
-                <th>Reference Verification Date</th>
-              </tr>
-            </thead>
-            <tbody>
-            {% if not_contacted_applications %}
-              {% for application in not_contacted_applications %}
-                <tr class="header" id='application_{{ application.user.email }}'>
-                  <td>{{forloop.counter}}</td>
-                  <td><a href="#application_{{ application.id }}">{{ application.user.username }}</td>
-                  <td>{{ application.get_full_name }}</td>
-                  <td>{{ application.user.email }}</td>
-                  <td>{{ application.application_datetime|date }}</td>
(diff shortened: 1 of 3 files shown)


tompollard:
Please could you give a bit more detail about:
- the issue with the ordering
- how this change fixes the issue?

elfeto:
The credentialing applications are not properly sorted. 

This sorts the the applications by date of the column...

Application date and then reference contact date

Screen Shot 2019-05-22 at 10 22 52 AM

tompollard:
Thanks, added a few thoughts below:

- Sorting by application date and reference contact date makes sense. Should reference verification date be included in the sorting?
- Perhaps include a comment at the top explaining how the list is sorted?
- the # on the left now restarts for the `not_contacted_applications` applications

Screen Shot 2019-05-22 at 13 18 59

-  It might be cleaner to use one of the sorting approaches in: https://stackoverflow.com/questions/6540032/sorting-related-items-in-a-django-template (via Meta in the model or via a sorting property), instead of splitting `contacted_applications` and `not_contacted_applications`.

elfeto:
I don't see a reason why would reference verification date should be included in the sorting, if you find one, let me know with the order type you would like. 
There is no need for a explanation of something obvious as to describe from old to new.
Nice catch of the numbers. 
That example is different from the issue of this sorting and it would not work. (I believe)

tompollard:
If all you're trying to do is sort the list, then can't you just add a sort to the `credential_applications` view?:

```
    applications = applications.order_by('reference_contact_datetime',
                                         'application_datetime')
```

elfeto:
So, at least for me in the "production" server it doesn't work. 
The sorting has to be first application date with NO reference contact, then, reference contact.

Trying to find a different way to see if there is something simpler. But nothing yet.

Screen Shot 2019-05-22 at 3 24 55 PM

tompollard:
Caught by another difference between sqlalchemy and postgres then :/

Alistair's suggestion to coalesce dates seems pretty clean. 

Add this method to the `CredentialApplication` class:

```
    def display_order(self):
        """
        Return days between now and reference contact time if available or
        application_datetime if not
        """
        delta = datetime.now(timezone.utc) - (self.reference_contact_datetime
                                              or self.application_datetime)
        return delta.days
```

Then sort in `credential_applications` view with:

```
    applications = CredentialApplication.objects.filter(status=0)
    applications = sorted(applications, reverse=True, key=lambda t: t.display_order())
```

Also nice to display a "time elapsed" column in the template.

elfeto:
I have a different fix. I will continue with Benjamin's suggestion. 
Also, that doesn't work. :)
```
>>> applications = CredentialApplication.objects.filter(status=0)
>>> applications = sorted(applications, reverse=True, key=lambda t: t.display_order())
>>> for a in applications:
...     print(a.application_datetime, '<->', a.reference_contact_datetime, '<->', a.reference_response_datetime)
... 
2019-04-29 11:40:10.163743+00:00 <-> 2019-05-14 20:26:03.124603+00:00 <-> None
2019-05-19 12:42:03.373391+00:00 <-> None <-> None
2019-05-21 20:56:40.854153+00:00 <-> None <-> None
2019-05-16 15:15:29.194675+00:00 <-> 2019-05-22 13:19:06.452379+00:00 <-> 2019-05-23 02:01:32.510484+00:00
2019-05-23 02:31:29.292390+00:00 <-> None <-> None
2019-05-23 13:10:06.472024+00:00 <-> None <-> None```

elfeto:
So, here I replace the hacky way that I was sorting with another hacky way.
Also, I added the time elapsed since last action in days.

tompollard:
thanks!

Issue #519, opened 2019-05-23
Generating SHA256SUMS.txt


bemoody:
SHA256SUMS.txt is currently generated synchronously at "publication time", overwriting any existing file.  This means:

- author never has a chance to check that the hashes are correct
- if author uploaded their own SHA256SUMS.txt file, it will be silently thrown away (*after* the author has approved the supposedly final version of the project)
- publication requires reading every block of every file, which can take a very long time

Of course the editor can notice the presence of the file, verify it themselves and/or warn the author that it will be thrown away, but that relies on the editor noticing and realizing that there could be a problem.

Issue #483, opened 2019-05-10
Check filenames before publication


bemoody:
In the vein of issue #476:

We should do some automated checking of project files and flag such errors prior to publication.

Aside from characters such as '@' that are completely disallowed:

- there should be no two files in the same directory that differ only in case

- no files named "con", "aux", "lpt1", etc.

(We can try to prevent such things via the web UI, but we've also talked about offering other ways of uploading data, such as via SFTP or uploading a tar/zip file.)

PR #523, opened 2019-05-24
Stricter filename validation, and warning messages

bemoody:
These changes address a few things:

- Non-ASCII file names are problematic (we are fortunate, I suppose, to be supporting only English. :/)  Disallow them.

- On certain popular operating systems, you can't create or access files called "con" or "aux.dat" (at least not via the normal filesystem API.)  Disallow these names.

- Limit file names to 50 characters (an arbitrary number, but the rename function already had a limit of 50.)

- If existing file names don't meet these criteria, a warning message is displayed in the edit and preview pages.

(Note: functions for *manipulating* existing files - rename, move, delete - will still probably work, because they don't validate their argument using validate_filename - they validate it by checking that it matches one of the known existing filenames.  This is not the best design, but *probably* is not a security issue.)

- If *two* existing filenames in the same directory differ only in case, show a warning message about that.  On certain popular operating systems, you can't have two filenames that differ only in case.

This is related to issue #483 but not a complete solution.  At some point during the publication process, there should be a *complete* scan and auto-reject, but like computing hashes (see issue #519) this needs to be done asynchronously.  So I'm ignoring that part for now.

### DIFF
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -144,8 +144,8 @@ class CreateFolderForm(ActiveProjectFilesForm):
     """
     Form for creating a new folder in a directory
     """
-    folder_name = forms.CharField(max_length=50, required=False,
-        validators=[validators.validate_filename])
+    folder_name = forms.CharField(max_length=validators.MAX_FILENAME_LENGTH,
+        required=False, validators=[validators.validate_filename])
 
     def clean(self):
         """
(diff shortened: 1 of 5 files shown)


elfeto:
Interesting enough, the function "get_project_file_warning" works on linux, but not on OSX. Since this is a message, and the servers are linux, its ok this is not shown in OSX.

bemoody:
Just to clarify, issue is that if your media directory is in a
case-insensitive filesystem, and you already have a file called
"foo.txt", and you try to upload "FOO.txt", then
project.utility.write_uploaded_file will silently replace the existing
file.

(UploadFilesForm is intended to stop you from doing this, but it
doesn't check whether the target filename *exists*, it checks whether
the target filename *matches the name of an existing file*, which on a
case-insensitive FS is not the same thing.)

See issue #524.

PR #525, opened 2019-05-29
Create the database index page dynamically

tompollard:
Currently the database index page (https://alpha.physionet.org/about/database/) is hardcoded, copied directly from the previous site. This change dynamically generates the page using project topics. Some notes below: 

- So that projects appear only once in the list, they are first grouped into the most popular topic. e.g. a project that is associated with "food" (a rare topic) and "ecg" (a popular topic) will appear in the "ecg" group.
- Once the grouping above is complete, topics that contain only one project are collapsed into "Miscellaneous".
- Independently from this pull request, we have added one or more topics for all projects on the development site.


lbulgarelli:
Wasn't this page kept only to reflect the current PhysioNet Index?

I don't see why we should even have a dedicated page for this otherwise, as it could there could be simply a feature for topic filtering in the "Find" page.

tompollard:
> Wasn't this page kept only to reflect the current PhysioNet Index?

Kind of, to keep current users happy and to facilitate the transition to the new system. We plan to keep the full page summaries for the foreseeable future (6-12 months), so it makes sense to generate them dynamically.

bemoody:
The algorithm for choosing a topic is quite convoluted and I can't
follow it.  I'm fairly certain the 'sorted' bits are not doing what
you want.

Any algorithm like this is also fundamentally flaky: adding a new
database can cause all other databases to be re-categorized in an
unpredictable way.

It's hard to say how this would behave without seeing the complete
list of databases and topics that we have right now.

But I suspect this would lead to one of two things:

- topics are used liberally without regard to the database index, and
the "index" ends up being a nonsensical assortment of topics that vary
from month to month and are useless for navigation; or

- topics are *only* useful for constructing the index; editors avoid
creating new topics in order to avoid screwing up the existing
categorization.

Indeed, with this code as it stands, if I'm an editor preparing to
publish a new project, I don't even have any way of *knowing what will
happen* when I click the "publish" button.


There are many dimensions along which databases might be categorized.
For example:

- databases with ECG signals
- databases from intensive-care patients
- databases with signal quality annotations
- databases used in challenges

If those happened to be four of the most "popular" categories of
databases, would you want to see those as separate sections in the
database index, with each database listed in at most one section?

(In which of those sections would you find the Challenge 2015 training set?)


So: it may be possible to automatically construct a reasonable list of
categories, but this is not the way to do it.

More semantically-structured tags (like Debian package tags or Trove
categories) might be helpful and might be worth considering.

However it works, it needs to be easy for the editors to predict what
will happen when a new database is published (and really, publishing a
single database should never have any significant side effects.)

PR #527, opened 2019-05-29
Check subdir on file form submission

bemoody:
This addresses a critical security issue in project file management.


elfeto:
Seems good :) Nice catch

bemoody:
BTW, I had deployed this patch by hand but didn't update the production and staging branches.  Now I've done so.

Issue #524, opened 2019-05-28
Handle OS errors in file management forms


bemoody:
File management forms should, in general, *handle* OSErrors and raise the appropriate exception when they occur, instead of trying to *anticipate* errors and raise an exception ahead of time.

For example, in UploadFilesForm, instead of checking whether the filename matches an existing file, simply try to create it (invoke write_uploaded_file with overwrite=False) and handle the possible FileExistsError.

Similar issues apply to CreateFolderForm, EditItemsForm (which should be called DeleteItemsForm), RenameItemForm, MoveItemsForm.

bemoody:
I'm fairly certain this is fixed.

PR #532, opened 2019-05-31
Improve resilience of file management forms

bemoody:
Various changes are made here to the file management forms.  In
general:

- Handle exceptions when manipulating files, and display an
  appropriate error message.

- When multiple operations are requested, do as much as possible,
  rather than stopping when one operation fails.

- Don't read the same directory more than once.

This fixes issue #524 along with other issues.


elfeto:
Interesting the use of "format_html", I think up to now we just dumped the html and called it safe in the HTML...

Issue #518, opened 2019-05-23
Disk usage and quota enforcement


bemoody:
Currently disk usage is calculated, and quota is enforced, "by hand".

This requires scanning the entire directory tree every time the file upload form is loaded.  It also makes quota enforcement complicated and buggy, and makes it impractical to offer alternative mechanisms for uploading files.

Instead, filesystem quota should be used to measure and limit the disk usage.

bemoody:
We could use either traditional *group* quota (limit number of files/blocks owned by a particular GID), or XFS *project* quota (limit number of files/blocks within a directory tree; also known as directory quota.)

I was originally leaning towards project quota, then I changed my mind three times, so now I'm leaning towards group quota.  But could easily be swayed either way.

Some possible advantages of group quota:
- more transparent and perhaps easier to manage; GIDs can be viewed and changed using common command-line tools and APIs
- files can be *shared between project versions* without *sharing quota between all versions*
- group permissions might make for easier sandboxing of servers (e.g. sftp)
- could be made to work on ext4 filesystems as well

Some possible advantages of project quota:
- doesn't require reserving a large space of GIDs (still have to manage "project IDs", though)
- filesystem enforces hierarchical consistency so it's not possible to have files within the tree that aren't counted

Either way would require some custom scripts with elevated privileges in order to create project directories and to modify their quota.

To use group quota, I think we'd probably have to do away with IPA auth for local users (to avoid conflicting GIDs).  But that's something I think we ought to do anyway.

Also, last I checked it was not possible to enable both group quota and project quota simultaneously on the same filesystem.

elfeto:
So, I think the easier way to handle this is by project, and keep track of the project id in a table and `/etc/projects`, `/etc/projid`.  You can handle them without the `etc` files, but if you dont use the file, then you can only create them using a numeric id. 
```
Project ID   Used   Soft   Hard Warn/Grace   
---------- --------------------------------- 
...
#1              0   100M   100M  00 [------]
something       0      0     1G  00 [------]
```
I was thinking of having a model to store this information, and on a project create it calls a script in which it sets all of this. (obviously the variables are set and sent from django)

The only thing is that on project publish, there has to be another script to remove the quota, or it would give a error of missing path in `xfs_quota` because the path is moved to another and the quota would try to find the path it was before. 

I incline more in the projects, because if we do a hard link on the files at a different version, then the group of the new project, should be the same than of the published content. 

You can also do all three at the same time, user, group and project.

PR #533, opened 2019-05-31
Quota 1

elfeto:
Adding a fixed disk quota in the projects managed by the system.

In the sudoers file, the user `pn` or `ALL`, depending on what you want, has to be enabled to run the set quota command
`
ALL     ALL=(ALL) NOPASSWD:/usr/local/bin/set-quota.sh
`

There are 4 times the quota should be touched. 
1. On project create
2. On active project publish
3. On project quota increase request
4. On version create
     - On version publish

**The quotas at the moment are cumulative, so all versions would share the same quota.** 
There is a new file `set-quota.sh` that should be placed in `/usr/local/bin/`, and the execution of that would alter the disk quota.

This should be tested a bit more, and the input inn the bash script can be checked before running also.
#518


elfeto:
Note, I forgot to mention that for the projects in the demo, I added the initial quota for one project.

bemoody:
Thanks for working on this.  It's a good start.

The set-quota.sh script has a number of issues - it doesn't validate
its inputs at all.  It's quite difficult to do this sort of thing
safely in a shell script; why not rewrite it in Python (or Perl)?

I think we may not actually need named groups at all (just assign
numeric GIDs from a large reserved range) and if we could avoid that
it would simplify things quite a bit.  xfs_quota command appears to
accept numeric IDs, at least.

Some smaller things:
 - you can just use 'auto_now', instead of 'auto_now_add' and updating
the timestamp by hand
 - use subprocess.check_call (or similar) instead of os.system
 - 'settings.QUOTA' is unclear, maybe call it 'settings.USE_FILESYSTEM_QUOTA' ?

All of this should be better documented as well, with comments
explaining why things are being done this way.

bemoody:
> I think we may not actually need named groups at all (just assign
> numeric GIDs from a large reserved range) and if we could avoid that
> it would simplify things quite a bit.  xfs_quota command appears to
> accept numeric IDs, at least.

We probably should keep track of all GIDs that have been assigned,
even if the project is later deleted; but my thought is that this info
can be kept in the relational database and probably doesn't need to be
pushed out to the system group database as well.  What do you think?

elfeto:
Just in case, the changes done have to be tested now in my VM, but I have to push them here first to then pull in the VM.

elfeto:
I will close this for now, but leave the branch open for future references. It will still need some work after all the PRs.

Issue #480, opened 2019-05-10
Highlight one selected news item prominently on the front page


tompollard:
It would be great if we could:

- select one "highlighted" item in the news menu of the admin console
- display this item prominently on the front page of the site

tompollard:
Completed in https://github.com/MIT-LCP/physionet-build/pull/844

PR #536, opened 2019-06-10
Temporary banner for maintenance

bemoody:
Add a temporary banner message to the top of every page, and disable the login page.

### DIFF
--- a/physionet-django/templates/base.html
+++ b/physionet-django/templates/base.html
@@ -16,7 +16,8 @@
   <body class="flexbody">
     {% include "navbar.html" %}
 
-    <main>{% block content %}{% endblock %}</main>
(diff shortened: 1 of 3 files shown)


tompollard:
Thanks, looks good to me. The banner appears slightly off screen on the front page at the moment. Should it appear above `class="main-header"`?

Screen Shot 2019-06-10 at 13 23 38

bemoody:
Feel free to try to make it prettier if you like :/  Putting it above
main-header doesn't cause main-header to shrink, though...

tompollard:
@lbulgarelli this temporary fix is relevant to: https://github.com/MIT-LCP/physionet-build/issues/480. We may want to consider two levels of highlighting - one for front page and one for all pages.

Issue #504, opened 2019-05-20
HTML file previews


bemoody:
I'd like to have an HTML preview for project files.

For example:
 - CSV files formatted as a table
 - Source files with syntax highlighting
 - Binary data files with a summary of the format
 - Summary of headers from .edf, .mat, .wav

I see this as being something useful both for visitors and for editors, and also as a way of making our data format requirements clearer for authors.

I've done some preliminary work on this, but what I'm not sure about is what the URLs ought to look like.  We'd have two or three different URLs for each file:
 - one to show an HTML preview
 - one to view the file itself as plain text
 - possibly a third which would have the same content as the second, but would also force browsers to download the file (see also issue #310.)

My first thought was that these should have the same path and different query strings (`/content/foo/1.0/bar.c` for the HTML view, `/content/foo/1.0/bar.c?plain` for plain text, `/content/foo/1.0/bar.c?download` to force download).  But that would cause wget to save the file as `bar.c?plain` (and save the HTML as `bar.c`), plus there's no easy and obvious way to recursively wget just the plain files.

My second thought was that we could serve a different default version based on the 'Accept' header, so web browsers would see the HTML version, while wget/curl/wfdb/etc. would see the plain file.  (This is a terrible idea.  I'm just putting it out there so I can point out what a terrible idea it is.)

So what I'm thinking now is to have two separate paths, e.g. `/content/foo/1.0/bar.c` for the HTML view and `/content/foo/raw/1.0/bar.c` for the plain file.  This is how several popular VCS browsers do it.  (It still bugs me a little that wget will save the HTML file as `bar.c` in this case.  Probably this is not a huge concern, but it is slightly at odds with what *PhysioNet* users expect.  One option might be to add either a trailing slash, or a `.html` suffix.)

So here are a few options.  For the HTML version:
 - `/content/foo/view/1.0/bar.c`
 - `/content/foo/1.0/bar.c`
 - `/content/foo/1.0/bar.c?view`
 - `/content/foo/1.0/bar.c/`
 - `/content/foo/1.0/bar.c.html`

For the plain version:
 - `/content/foo/raw/1.0/bar.c`
 - `/content/foo/1.0/bar.c`
 - `/content/foo/1.0/bar.c?raw`

For the force-download version:
 - `/content/foo/raw/1.0/bar.c?download`
 - `/content/foo/download/1.0/bar.c`
 - `/content/foo/1.0/bar.c?download`

[...]

Issue #310, opened 2019-03-25
Add download button next to each file?


alistairewj:
At the moment files can be downloaded one by one, but many browsers open text files on click, rather than opening a dialogue box to download them. This results in a bit of inconsistent UI interaction, e.g. in the [ecgcipa database](https://alpha.physionet.org/content/ecgcipa/) where clicking [addm.csv](https://alpha.physionet.org/static/published-projects/ecgcipa/files/addm.csv) opens the CSV in my browser but clicking [ecgcipa_metadata.docx](https://alpha.physionet.org/static/published-projects/ecgcipa/files/ecgcipa_metadata.docx) opens a dialogue box to download the data.

The worst case is a large text file like [adeg.csv - 25.6 MB](https://alpha.physionet.org/static/published-projects/ecgcipa/files/adeg.csv) being opened in the browser and slowing it down a bit.

I thought maybe adding a little download icon next to each file might help alleviate this and provide the user a consistent place to click to download files.

bemoody:
This is something we need to seriously consider.  AFAIK it's currently inconsistent between public projects and restricted/unpublished projects.

As a general matter (by no means specific to PhysioNet), it can be annoying when:

- You want to *view* a file, and the browser won't let you do so, or makes you go through multiple clicks ("yes, I want to open this file; no, I want to open it in Atril, not GIMP") in order to do so... or worse, you're actually forced to save it to disk somewhere and open it in another program by hand.

- You want to *save* a file, and your browser doesn't give you a way to do so (or you don't know how.)

- You *think* that you want to view a file, without realizing that it's a 25 MB file that will make your browser throw a fit.  (Or else it's a binary data file but your browser still tries to display it as CP1252 text for some reason.)

How can we solve all of these problems, or at least strike a reasonable balance, and avoid making the interface overly complicated?  I have some ideas but I'd like to hear what other people think.

alistairewj:
I like my button idea! Mockup:

![image](https://user-images.githubusercontent.com/1282676/58130354-d5701680-7be9-11e9-89d1-34b250cf98da.png)

tompollard:
That looks good to me. What about having columns for view and download?

bemoody:
"View" is what you'd get by clicking the file name, I think.

At the moment, clicking that link will either display the file as
plain text or try to download it (depending on the file, your browser,
and phase of the moon.)

PR #537, opened 2019-06-12
HTML file previews

bemoody:
This implements the basics of "HTML previews" for project files (addressing issue #504).  When viewing the project files, you can click on a file name to view the file contents as a web page.  Currently:

- images (png / jpeg / gif / bmp) will be displayed in the browser as an image

- CSV files (and *compressed* CSV files, as we use e.g. for MIMIC) will be displayed as a table (truncated to the first N rows and columns, in the case of very large files)

- other text files will be displayed inline

- other files will show "this file cannot be viewed in the browser"

On the page there will be a link to download the file as well as a link to the "plain" file when that makes sense.

----

Note some significant URL changes:

* /content/foo/1.0/bar now shows the HTML preview.

* /files/foo/1.0/bar shows the raw file (equivalent to /static/published-projects/foo/1.0/bar for public projects; there was no equivalent for restricted projects.)

* /files/foo/1.0/bar?download shows the raw file and tries to force the browser to download (what /content/foo/1.0/bar did previously.)

* If 'bar' is a directory, /files/foo/1.0/bar/ will show a directory listing of the raw files.  (This means that 'wget -r -np' should work to download the files as expected.)

* For active projects, /projects/foo/preview/bar is the analog of /content/foo/1.0/bar, and /projects/foo/files/bar (as before) is the analog of /files/foo/1.0/bar.

----

Further work would be useful:

* Implement more views!  Examples:
   - annotated view of WFDB header files
   - program source code with syntax highlighting
   - viewing the metadata of EDF, WAV, MAT files
   - viewing the metadata of image files

* The file_content_type function is not fully consistent with nginx's content types.  It probably should be (i.e., we should copy what we need from /etc/nginx/mime.types into both custom-mime.types and physionet.utility.CONTENT_TYPE and keep these lists in sync.)  (As I described in one of the commit messages... content types are a real mess if you just want files to be *displayed in the browser whenever possible.*)

* Possibly as an alternative to having our own content_type function, we could find a way to override *Python's* notion of content types (which I think is what WSGIServer uses to set content types itself.)

* And, in general, it'd be better if nginx could handle requests for /files/ without involving django at all.  It's not trivial because:
   - we do need to involve django for restricted projects (though we could in theory move those to /files-protected/ or something)
   - we need to set Content-Disposition for '?download' (though this should be possible with regexps in nginx too)

----

So: this is a work in progress, and I'd of course welcome improvements, but I'd like to get *some* form of this deployed, and the new URL structure made stable, before the new site becomes official. :)


elfeto:
SO, I like it, and its a LOOOT of leg work.

This also affects #310

There are only 2 things. 
- The download button I would put it on the file panel and in the file preview. Because if I want just to download 10 files, it would be annoying to click on those 10 files, and then click download again.
- Text files, are not truncated at X rows, but in size. (don't know if this was intentional) just to put it there.

tompollard:
Looks great, thanks Benjamin. I've added a couple of comments below:

#### 1. `<base>` link returns to unexpected page

One minor bug is that I found the `<base>` link sometimes didn't take me to the expected page. For example:

1. create a new project
2. navigate to the slug page http://127.0.0.1:8000/projects/SLUG/files/
3. upload a file
4. clicking on the file takes us to: http://127.0.0.1:8000/projects/SLUG/preview/FILENAME
5. clicking `<base>` in the breadcrumbs takes me to:
http://127.0.0.1:8000/projects/7tfzQfZUSmadHAmdVB9s/preview/#files-panel

I would expect to be taken back to the files page (http://127.0.0.1:8000/projects/SLUG/files/).

#### 2. No prominent "return to previous page" button

Should we have a prominent return to previous page button, in a similar way to the project preview page (shown below)?

Screen Shot 2019-06-19 at 14 19 47

bemoody:
Thanks for your helpful comments.  @elfeto:

- Files panel now shows a download icon for each file (issue #310).

- Text files aren't truncated the way CSV files are; instead it just
skips displaying the file entirely if it's more than 1 MB.  My
reasoning:

CSV will be used for *data* files which can often be very large (MIMIC
being an extreme example.)  I don't expect anyone would want to use
the web interface to extract data even if they could - if you want to
process the data you will want to download the entire CSV file and
parse it in your program, rather than trying to scrape it from an HTML
page.

So the purpose of having the HTML page, in this case, is so that a
visitor (or an editor reviewing a new database) can quickly get a feel
for the database contents before they start writing any code.  And the
reason for truncating the table is to prevent overloading the server
or the browser while doing so.

For arbitary text files, on the other hand, I'd rather not see
anything than to see a truncated version that might give people an
incorrect impression.  And I specifically don't want to encourage
authors to use arbitrary structured text formats for data - CSV is one
exception, we may consider making others, but I don't want to see
e.g. multi-gigabyte XML files.

(There's also the possibility of large *binary* data files that are
incorrectly classified as text, and displaying an excerpt as text
doesn't make sense.)

@tompollard:

- Breadcrumb links should now behave better, I think.

- I don't want to put any more clutter at the top of the page than is
absolutely necessary, and a button is also IMO the wrong metaphor
here.  If the breadcrumb links aren't obvious enough, should we try to
make them more obvious?

I suppose we could make the page *title* be a link back to the
top of the project, but I also don't really like the idea of either
(a) having an H1 that is styled like a link, or (b) having a link that
isn't styled like a link.

tompollard:
Thanks!

PR #542, opened 2019-06-14
Updated the version of Django

elfeto:
The version of django was updated from 2.1.7 to 2.1.9
There was a bug in the links generated by "AdminURLFieldWidget".

We have to think about changing soon to 2.2, but it has some database changes.


tompollard:
Thanks!

PR #545, opened 2019-06-17
Some changes in the "Manage" credentialing

elfeto:
Some minor changes in the Manage credentialing only.

- Made it so the contact can be contacted several times by email client. 
- Added green and black color to differentiate the reference contacted vs not contacted 
- Added sidebar auto colapse
- Added a way to remove credentialing or move a declined application back to observation.
- Changed some help text as requested.


physionet-django/user/templates/user/application_display_table.html
```diff
@@ -125,10 +125,10 @@
       <td>{% if application.status %}{{ application.get_status_display }}{% else %}<i class="fas fa-clock"></i> Pending{% endif %}</td>
     </tr>
     {% if application.responder_comments %}
-    <tr>
+<!--     <tr>
       <td>Admin Comments</td>
       <td>{{ application.responder_comments|linebreaks }}</td>
```
tompollard: This leaves the comment in the raw HTML. To remove it altogether, the `{{ application.responder_comments|linebreaks }}` should be commented too:

```{{# application.responder_comments|linebreaks #}}```

File: physionet-django/user/forms.py [hunk 2]
```diff
@@ -308,7 +308,7 @@ class Meta:
         help_texts = {
             'training_course_name':"The name of the human subjects training course you took. e.g. 'CITI Data or Specimens Only Research Course'",
             'training_completion_date': 'The date on which you finished your human subjects training course. Must match the date in your training completion report.',
-            'training_completion_report':"A pdf of the training completion report from your training program. The CITI completion report lists all modules completed, with dates and scores. Do NOT upload the completion certificate.",
+            'training_completion_report':"Upload the completion report from the CITI \"Data or Specimens Only Research\" training program (PDF or image file). The completion report lists all modules completed, with dates and scores. Do NOT upload the completion certificate. If you would like to submit multiple pages, please combine them into a single pdf file.",
```
tompollard: The text is truncated in the description:

Screen Shot 2019-06-18 at 09 58 12

physionet-django/user/models.py
```diff
@@ -685,3 +685,13 @@ def file_root(self):
 
     def get_full_name(self):
         return ' '.join([self.first_names, self.last_name])
+
+    def get_latest_by_user(self):
+        return CredentialApplication.objects.filter(user=self.user).last()
+
+    def is_last_by_user(self):
```
tompollard: "latest" is clearer than "last", so perhaps also use that here (e.g. `def is_latest_by_user`)

physionet-django/console/views.py
```diff
@@ -779,7 +779,27 @@ def past_credential_applications(request):
...
+            if C_application:
+                C_application = C_application.get()
+                C_application.user.is_credentialed = False
+                C_application.user.credential_datetime = None
+                C_application.user.save()
+                C_application.decision_datetime = None
+                C_application.status = 1
+                C_application.save()
+        elif 'manage_credentialing' in request.POST and request.POST['manage_credentialing'].isdigit():
+            credentiallied_id = request.POST['manage_credentialing']
```
tompollard: - typo in `credentiallied_id` -> `credentialed_id` or maybe just `cid`
- not sure the `C_application` needs capitalisation
- would there be a problem if `C_application.user.save()` was successful but `C_application.save()` failed? If so, the transaction should be atomic (https://docs.djangoproject.com/en/2.2/topics/db/transactions/)

tompollard:
Thanks Felipe, looks good to me. I find the styling of the credentialing page odd, but we've discussed that. I've added a few comments and happy to merge after they've been dealt with.

physionet-django/user/templates/user/application_display_table.html
elfeto: "{{# application.responder_comments|linebreaks #}}" gives a internal server error, and there is no block comment for Django, so I am removing it for now. Because if not then it would be all commented in but seeable in raw HTML or inspect element.

tompollard: my bad, it should just be `{# application.responder_comments|linebreaks #}`

physionet-django/user/forms.py
elfeto: It doesn't work in triple quotes, and in first single then double. So it will have to go double then single.

physionet-django/user/templates/user/application_display_table.html
elfeto: Yea, but you will still see the commented HTML in raw or inspected. So I removed it.

tompollard: Okidokes, we can always add later. With `{# application.responder_comments|linebreaks #}` the user should just see `{# application.responder_comments|linebreaks #}` rather than the comment itself, but removing it all for now is good.

tompollard:
Looks great, thanks

PR #549, opened 2019-06-20
Credentialing 5

elfeto:
A couple of things with credentialing...
- Added a way to view and remove Legacy applications.
- Added a interested project. 
- Added a CSV downloadable with the credentialed project access
- Added a view and count for people that had signed a DUA on credentialed projects.


physionet-django/console/views.py
```diff
@@ -934,3 +945,90 @@ def guidelines_review(request):
...
+@login_required
+@user_passes_test(is_admin, redirect_field_name='project_home')
+def download_credentialed_users(request):
+    """
+    CSV create and download for database access.
+    """
+    # Create the HttpResponse object with the appropriate CSV header.
+    project_access = DUASignature.objects.filter(project__access_policy = 2)
+    added = []
+    List = [['First name', 'Last name', 'E-mail', 'Institution', 'Country',
```
tompollard, line 959: Not a good name for a variable!

elfeto: I agree.... Fixed in the new PR... among other bugs...

Issue #547, opened 2019-06-20
PhysioNet contact reply to email


alistairewj:
Right now emails from PhysioNet contact only have contact@physionet.org in the sender. It would be great to add the users email to the reply to of the message so it's easy to reply to the emails (rather than having to copy paste the email address from the body of the email as we do currently).

PR #550, opened 2019-06-20
changed mail from in contact

elfeto:
Fixes the mail from in the contact form. #547

### DIFF
--- a/physionet-django/notification/utility.py
+++ b/physionet-django/notification/utility.py
@@ -2,6 +2,7 @@
 Module for generating notifications
 """
 from urllib import parse
+from email.utils import formataddr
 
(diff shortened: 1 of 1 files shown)


bemoody:
A little context would be nice.  ("When someone sends a message using
the form at <https://alpha.physionet.org/about/#contact>, the message
should include the person's email address in the From header...")

I would simply use the person's name *and* address as the sender (use
email.utils.formataddr to construct an rfc822-compliant header) and
skip the silly pseudo-header in the body of the message.

bemoody:
(But that's just a suggestion.  Feel free to merge as is if you prefer.)

Issue #530, opened 2019-05-29
Disable auto-filling in credentialing form?


bemoody:
Might be worth setting autocomplete=off for credentialing form, because honestly, this isn't something you're going to be submitting often, and you *should* think carefully and be sure you're entering accurate information.

tompollard:
I had a (quick!) look into this and I couldn't see a well-supported approach for switching off auto-complete for form fields. I think the main issue at the moment is on the original site...it seems that we're bouncing a lot of applications for phone numbers being listed in the suffix.

bemoody:
Oh, there will never be any guarantees!  But there is a standard
('autocomplete' attribute) and it sounds to me like many browsers do
support it partially.  Assuming this is desirable, I don't see any
harm in setting the attribute.

tompollard:
Setting the autocomplete attribute (`autocomplete="off"` or `autocomplete="new-password"`) seems worth trying (https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion). 

Your suggestion of changing the field name in https://github.com/MIT-LCP/physionet-build/issues/529 also seems like it could work.

lbulgarelli:
Fixed on #551

Issue #529, opened 2019-05-29
Credentialing form "suffix" auto filled with phone number?


bemoody:
Apparently some browsers will try to fill in the "suffix" field with a phone number?  (I haven't seen this myself.  But it seems not out of the question since apparently some people refer to the "part of the phone number that doesn't include country or area code" as the "suffix".)

If we can replicate this problem, might be worth trying changing the name of the form field to "name_suffix" or something similarly less ambiguous.

lbulgarelli:
Fixed on #551

PR #551, opened 2019-06-20
tried to remove auto-complete from suffix

elfeto:
Here I address #530 and #529
I added a prefix called "application", that would be prepended to the ID and Name of the input. 
Also, added autocomplete = off.
This SHOULD fix the autocomplete on the suffix, but who knows...


bemoody:
Very thorough :)

(Looks fine to me though I haven't tested it.)

Issue #522, opened 2019-05-23
Linking files upon new version creation


bemoody:
When a new project version is created, we recursively copy the directory structure and hard-link all existing files.

This involves scanning the entire directory tree, so should be done asynchronously (probably fine to make project "active" immediately while linking files in the background.)  Also, we should probably check permission bits at this time and ensure all of the linked files are read-only.

Issue #521, opened 2019-05-23
Setting files read-only upon publication


bemoody:
All files and directories in the project should be made read-only at publication time, to avoid modifying these files inadvertently.

This involves scanning the entire directory tree, so should be done asynchronously (project can be made "public" immediately while this task runs in background.)

PR #554, opened 2019-06-21
Async files

elfeto:
Here I create a background task to set the following:
- Move files on publish to new directory
- Create special files like zip file
- Set file permission as read only

Also changed the resetdb file to re-add permissions to the media, in order for it to delete the files. 

This background task will work as long as the "process_tasks" is running using `python manage.py process_tasks` in a different  terminal

Here I address #522 #521

### **NOTE. Permissions for the folders are kept the same. (Still debating if its a good idea)**

### DIFF
--- a/physionet-django/console/test_views.py
+++ b/physionet-django/console/test_views.py
@@ -3,6 +3,8 @@ import pdb
 
 from django.test import TestCase
 from django.urls import reverse
+from django.test.utils import get_runner
+from background_task.tasks import tasks
 
 from project.models import (ArchivedProject, ActiveProject, PublishedProject,
     Author, AuthorInvitation, License, StorageRequest)
@@ -204,6 +206,8 @@ class TestState(TestMixin, TestCase):
         response = self.client.post(reverse(
             'publish_submission', args=(project.slug,)),
             data={'slug':custom_slug, 'doi':'10.13026/MIT505', 'make_zip':1})
+
+        self.assertTrue(bool(tasks.run_next_task()))
         self.assertTrue(bool(PublishedProject.objects.filter(slug=custom_slug)))
         self.assertFalse(bool(PublishedProject.objects.filter(slug=project_slug)))
         self.assertFalse(bool(ActiveProject.objects.filter(slug=project_slug)))
--- a/physionet-django/notification/utility.py
+++ b/physionet-django/notification/utility.py
@@ -24,14 +24,6 @@ def send_contact_message(contact_form):
     send_mail(contact_form.cleaned_data['subject'], body, mail_from,
         [settings.CONTACT_EMAIL], fail_silently=False)
 
-
-def send_email_to_admin(subject, message):
-    """
-    Send a message to admin
-    """
-    send_mail(subject, message, settings.SERVER_EMAIL, [settings.ADMINS],
-        fail_silently=False)
-
 # ---------- Project App ---------- #
 
 def email_signature():
--- a/physionet-django/notification/utility.py
+++ b/physionet-django/notification/utility.py
@@ -24,6 +24,14 @@ def send_contact_message(contact_form):
     send_mail(contact_form.cleaned_data['subject'], body, mail_from,
         [settings.CONTACT_EMAIL], fail_silently=False)
 
+
+def send_email_to_admin(subject, message):
+    """
+    Send a message to admin
+    """
+    send_mail(subject, message, settings.SERVER_EMAIL, [settings.ADMINS],
+        fail_silently=False)
+
 # ---------- Project App ---------- #
 
 def email_signature():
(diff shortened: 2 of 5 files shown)


bemoody:
Please correct the doc string of move_files_as_readonly
(project/models.py line 34-38).

You're not using Dirs (so you might just as well remove it), but:

- it'd be better for Dirs to be a set rather than a list (so "not in
  Dirs" would be constant rather than linear time)

- please use all lowercase for local variable names

- instead of using get_tree_files() and then trying to pick out
  directories by hand... you could just use os.walk()

If you do want to make the directories read-only (which is probably a
good idea, though not as important as setting the files read-only),
don't forget that you need to set them to 'S_IRUSR | S_IXUSR | S_IRGRP
| S_IXGRP | S_IROTH | S_IXOTH' (aka 0o555, not 0o444).

Finally, I'd say the 'os.rename' step should probably be done
synchronously in publish() rather than asynchronously in
move_files_as_readonly().

bemoody:
> Finally, I'd say the 'os.rename' step should probably be done
> synchronously in publish() rather than asynchronously in
> move_files_as_readonly().

To look at this from another angle, in your first commit you had:

    try:
        os.rename(dir_from, dir_to)
    except:
        print('Files not there')

which is problematic because of the print statement and also because
it catches things that shouldn't be caught, and in your second commit
you changed this to simply:

    os.rename(dir_from, dir_to)

but this is still problematic because if the task previously *failed*
(or was interrupted), then dir_from won't exist and this will fail
with FileNotFoundError.

It's probably best to design background task functions to be
idempotent if possible (calling the function twice with the same
arguments should have exactly the same effect as calling it once) -
which it looks like this one is, apart from the rename and the fact
that make_special_files will modify timestamps.

(Of course, this function *shouldn't* normally fail and if it does
fail there's probably something we need to fix by hand anyway, but
best to make things as robust as we can.)

elfeto:
That print was a test that was committed by mistake. 

Here I add a proper except, and a email to the admins, because this would require some fast attention.

bemoody:
dc8fb72abc9fc144f2c5ead44472e7cb33146299 is not a good idea.

Look at precisely what that change does.

Before you made that change, if for some reason self.file_root() does
not exist (or parent directory of published_project.file_root()
doesn't exist), then:
 - the function will immediately fail
 - the exception will be logged in the usual places
 - the editor who triggered the action will immediately see an error message
 - the ActiveProject is not deleted, which sounds messy but could
actually aid recovery in this abnormal situation

After making that change [1], if for some reason the directory doesn't
exist, then:
 - the function will try to keep going, possibly causing more damage
to a corrupted system
 - the exception will never be logged; instead, a *less* useful
message will be mailed to the admins
 - the editor who triggered the action will not see anything unusual

Just don't do that.  This is an abnormal situation that will almost
certainly never happen [2], and what you've done is to add a new
complicated error handling path that will almost certainly never be
tested, and if it ever does occur, will make the problem harder rather
than easier to diagnose.

Yes, if this particular os.rename call did fail, it would be a serious
issue that we would need to fix immediately.  That can be said of
many, many things!  There's a fine line between making the program
*robust* (handling exceptions, making atomic transactions) and making
it *convoluted* with special cases.


[1] Modulo typos.  Since you misspelled 'send_email_to_admin', and
didn't import it either, what would actually happen is that a
NameError would be raised within the exception handler.  So really,
this commit has no effect other than to make the code more complicated
and harder to read.

[2] Note the key difference between
e0e1a6c9f92f2a9c7f69f5b0e33bfad44b681d48 (which I commented on
previously) and b2cb962ff2a4b56312fc4d2a5d38e00deb55db05: the
os.rename now occurs in publish() rather than
move_files_as_readonly(), so it really should only be called once
(whereas move_files_as_readonly could potentially be called multiple
times.)

bemoody:
Hmm.  Did you mean to use mode 555 for files as well as for directories?

I would say it's wiser to make files non-executable by default, and it
annoys me a little to see bright green everywhere.  On the other hand,
in the rare cases where project files *are* executable scripts
(configure scripts, for example), it's annoying to have to chmod them.

(Of course, downloading files via web browser won't set the executable
bit - but downloading via zip file or other means may do so.)

Perhaps the thing to do would be to set a file executable if and only
if it begins with something that looks like a shebang line.  (On the
grounds that we won't accept binary executables anyway.)

elfeto:
For folders, I think is ideal to use 544, if not I cannot list what its inside for the folders.

bemoody:
Directories should be 555 (= all users can list the contents of the
directory, and can open files within the directory.)  If you set it to
544 then the django server would be able to access the files but
nobody else.

elfeto:
NVM my previous comment... I read wrong what you said. I understood that you wanted the files and folders 444 except for the ones with the shebang... So, NVM :) 

I will post the fixes soon :)

bemoody:
Great!  Just a couple minor things:

- 'folders' variable is not used

- instead of open(...).readline().rstrip(), could just use open(...).read(2)

PR #561, opened 2019-06-21
Fix credentialing error and rename variable

lbulgarelli:
The credentialing page was returning an error because the `research_form` was not being initialized.

Also, rename the variable `List` to `columns`


elfeto:
#560 lol

PR #562, opened 2019-06-24
Cleaned credentialed application show form

elfeto:
Fixed user credentialing application history. (was missing some inputs).
Also, added information on the past application form for the console panel.


physionet-django/user/forms.py
```diff
@@ -332,7 +332,7 @@ class Meta:
             'training_completion_report':"Upload the completion report from the CITI 'Data or Specimens Only Research' training program (PDF or image file). The completion report lists all modules completed, with dates and scores. Do NOT upload the completion certificate. If you would like to submit multiple pages, please combine them into a single pdf file.",
         }
         widgets = {
-            'training_completion_date':forms.SelectDateWidget(years=list(range(1990, timezone.now().year+1))),
+            'training_completion_date':forms.SelectDateWidget(years=list(range(2015, timezone.now().year+1))),
```
tompollard: If we are going to restrict the start date to 2015 then we need to make it clear that we are doing this (otherwise people will complain about the form being broken)

tompollard:
Thanks, this is an improvement. We can work through the other issues that we discussed later (e.g. how to handle approved users for projects).

Issue #566, opened 2019-07-03
nginx.conf includes '/physionet/physionet-build/deploy/uwsgi_params'


bemoody:
Not sure why 'uwsgi_params' is even a thing, but nginx.conf should never be reading configuration files that are stored outside /etc and are writable by users other than root.

PR #567, opened 2019-07-05
Fix handling of old URLs, and other nginx stuff

bemoody:
Trying to redirect old URLs to new structure creates some
interesting problems...

* /physiobank/database/mitdb/ should redirect to /content/mitdb/1.0.0/

* /physiobank/database/mitdb/100.hea should return a header file

* /physiobank/database/fnord/ should return a 404

* /physiotools/wag/ should return the main WAG page (= /physiotools/wag/index.html)

* /physiotools/ecgpuwave/ should redirect to /content/ecgpuwave/1.3.4/

* /physiotools/ecgpuwave/src/ecgpuwave-1.3.4.tar.gz should return a tar file

* /physiotools/fnord/ should return a 404

This should work better and be a little bit cleaner:

* for "top-level" directories (/physiobank/database/foo/, /physiotools/foo/): if an index.html exists, then we serve that; otherwise, we pass the URL to django, which will either redirect or 404.

* other files and directories are served "normally".

Also fixes issue #566.


tompollard:
Thanks, looks good to me! @elfeto please could you double check and merge if you're happy?

elfeto:
So its nice, and I am happy to merge, the only thing is that now it would be ideal to create a 404 page for Nginx to replace the white 404.
For example
- /physiobank/database/mitdb/100.head  (The file really doesn't exists)

Issue #568, opened 2019-07-08
Background task log messages when running manage.py test


bemoody:
After merging pull #554, I now see some log messages when running 'manage.py test':

    Creating test database for alias 'default'...
    System check identified no issues (0 silenced).
    .....2019-07-08 14:50:20,968 background_task.tasks INFO     Running project.models.move_files_as_readonly
    2019-07-08 14:50:21,498 background_task.tasks INFO     Ran task and deleting project.models.move_files_as_readonly
    .......................................
    ----------------------------------------------------------------------
    Ran 44 tests in 15.109s
    
    OK
    Destroying test database for alias 'default'...

These messages should not be printed to the terminal, as they make it hard to see the actual test results.

PR #570, opened 2019-07-08
Background tasks 2

elfeto:
Added a proper name to the action in background tasks.
In the admin page we have now:
   - GCP - \<Project name> 
   - Read Only Files - \<Project name> (This was missing from PR #554)

Adressed #568


bemoody:
Please fix your doc strings.

What does it mean that 'physionet' is added to INSTALLED_APPS?
Why are you doing that here?

I would recommend writing:

 ('foo - %s' % project)

or if you prefer,

 ('foo - {}'.format(project))

rather than

 ('foo - ' + str(project))

elfeto:
Done :)

bemoody:
What's the significance of the changes in project/admin.py?

elfeto:
By default all "INSTALLED_APPS" in the settings/base.py file will be automatically registered to the admin page. What I am doing in the project/admin.py is to unregister those installed applications in particular the "Completed tasks" and "tasks", and I copy pasted the Package page, and removed the "task_name" that was the name of the function (the same for all), and replaced it with the verbose_name that can be edited to what we want. Then after doing the custom admin module, I added it to the admin page.

bemoody:
Sounds good :)

PR #574, opened 2019-07-08
Remove requirement for SubmissionInfo when making updates to active projects using the admin tool

tompollard:
Updating activeprojects using the `/admin/project/activeproject/*/change/` requires the SubmissionInfo fields to be completed (`submission_datetime`, `author_comments`, etc.) when the form is submitted. This change removes that requirement.


bemoody:
I assume this requires a migration.

PR #581, opened 2019-07-09
Equations

bemoody:
These changes allow equations to be included in project pages.

(I was working on this a couple months ago but stopped to focus on more pressing issues.)

Equations are stored as MathML and displayed using MathJax.  So on the django side, this is pretty straightforward. :)

The complicated part is the custom CKEditor plugin, which lets authors edit equations using LaTeX syntax.  There are some weird usability issues (e.g. cursor getting "stuck" and not accepting any keyboard input), but it's mostly working.  Could use a *lot* of testing in different browsers.


tompollard:
I tried creating and publishing projects in MacOS on Chrome (75.0), Safari (13.0), and Opera (53.0). It worked well for me. What is the difference between the pi and sum buttons in the menu?

Screen Shot 2019-07-09 at 17 01 30

tompollard:
> What is the difference between the pi and sum buttons in the menu?

Oh, I see - block vs inline display.

elfeto:
I tested it also and it worked fairly well for me. I will approve and test on staging.

PR #582, opened 2019-07-09
revert the order of accept and reject in credential management

elfeto:
I swapped the approve and reject of the credentialing in the management page.


bemoody:
This code needs some significant cleanup, but OK.

PR #583, opened 2019-07-09
Reorganize 'Metadata' and 'Discovery'

lbulgarelli:
- Change project   as it is more intuitive as to what it actually means.
  - `Metadata` is now named `Content`
  - ` version` field was moved from to the new `Metadata` section


physionet-django/console/templates/console/copyedit_submission.html
```diff
@@ -59,8 +59,8 @@ <h3>Description</h3>
           {% include 'project/item_list.html' with item="reference" item_label=reference_formset.item_label formset=reference_formset form_name=reference_formset.form_name add_item_url=add_item_url remove_item_url=remove_item_url %}
           <h3>Access</h3>
           {% include "project/content_inline_form_snippet.html" with form=access_form %}
-          <h3>Discovery</h3>
-          {% include "project/content_inline_form_snippet.html" with form=discovery_form %}
+          <h3>Metadata</h3>
```
tompollard: I prefer discovery for this section. It helps to make it clear that the information provided here (keywords, short description, etc) is intended to facilitate discovery of the project.

lbulgarelli: Reverted

PR #584, opened 2019-07-09
Altered django to 2.1.10 for CVE

elfeto:
There is a security vulnerability in 2.1.9. Here we change to 2.1.10 temporary while I check the differences in 2.2.3.


tompollard:
thanks felipe

PR #586, opened 2019-07-10
News

elfeto:
Setting news items to be on its on page. 
Now, when you click a news item, it will go to its own page and show the project it belongs to if it has one.
The person that creates the news is recorded but its not mandatory, in case the news is created from the command line.
I am using the UUID for the slug in the news item, because I cant use the news ID because of the yearly news are by number, so if we reach 1000 news, it will be a problem.

Also, check in the news slug if its numeric (current bug right now). Now it assumes everything is numeric and tries to get it, which can fail with a 500 error.


bemoody:
https://alpha.physionet.org/news/PhysioNet%2020170809.1 is bad.
https://alpha.physionet.org/news/e0619142-10c4-4810-9945-b37fc0ab3baa
is terrible.

Some possibly better ideas:

* https://alpha.physionet.org/news/2019/1 (= the first news item
posted in 2019)?
* https://alpha.physionet.org/news/2019/new-database-added-mimic-cxr
* https://alpha.physionet.org/news/article/12345

elfeto:
I used the UUID to see what u guys though, knowing it was going to get rejected :)

In the 3 better ideas, I think I would like more the third, and go:
 - * https://alpha.physionet.org/news/article/<news_id>

The reason not the first, is because we don't keep track of first news item of the year
The reason not the second, is because the title can have special characters that might make it tricky 

I would suggest a slug field, but that would also be a random character which I don't like that much.

elfeto:
@bemoody  Are you ok with the last changes ?

bemoody:
I don't like this at all, actually.  What is the goal here?


What is going on in NewsForm.save?  Looks like it saves once, then
sets created_by, then saves again?  Why not save only once?

News.created_by should be blank=True, and on_delete=SET_NULL (not
CASCADE), I would think.

The URL pattern for news_year should only accept integers
('news/<int:year>/' instead of 'news/<year>/').  The news_year
function shouldn't be trying to deal with malformed inputs, and
/news/fnord/ should be a 404 anyway.

I haven't heard of margin-block-end, and it's apparently not
standardized yet.  Why are you using that, and could you accomplish
the same thing using more widely supported properties?

tompollard:
This looks good to me. I like that this allows us:

- to share direct links to individual news items (with e.g. http://127.0.0.1:8000/news/post/3)
- to create longer news items that can be displayed with a teaser + read more in places like the front page.

elfeto:
So, I removed the year banner that is repeated in every news item because I already know the date of the news. Also, I modified the size of the date and URL and project in the news.

bemoody:
I think the reason that this doesn't make sense to me is that you're
trying to change the *model* of what news is.

That may be a good thing, but we should try to understand what this
new model is and how it relates to the old model, before we go trying
to fit square pegs in round holes.

In the old system, a "news item" consists of:
- a title
- a short description
- a link where you can go to read more

The purpose of a news item is, generally, to point visitors to another
resource (such as a newly posted database) and briefly summarize why
they might find that resource interesting.  The description should
never be longer than a couple of sentences, and the news item isn't
really of any interest in its own right.

As I understand it - correct me if I'm wrong - you want to replace or
supplement this with a more blog-like concept of, let's call them
"news articles", that contain extensive content of their own.

In that case, it makes sense for an article to have a stable URL of
its own, and for the body to have complex HTML formatting, and no
particular size limit.  On the other hand, it *doesn't* make sense for
an entire article to be shown on the front page, or in the RSS feed;
rather, you would want to have a short description for the article
together with a link to read the entire thing.

Do you agree with all of that?

bemoody:
also, do not use <font> ever.  It's not in HTML 5, and it wasn't a
good idea in HTML 4 either.  Use semantically meaningful tags and
classes, and format them using CSS.

tompollard:
@bemoody, in answer to: 

> Do you agree with all of that?

Hmm, mostly yes, except that for now I think it's fine to continue for news items to consist of a title; short description; and link. If we decide to add a long description field later (to allow longer, blog style posts for example) then this might be nice, but it's something that would need more thought.  

I like that the link URL is now explicit rather than wrapped on the title, because it feels wrong to send people to an external site without making it clear that this is what's going to happen.

bemoody:
> If we decide to add a long description field later (to allow longer,
> blog style posts for example) then this might be nice, but it's
> something that would need more thought.

If it's *not* a "longer, blog-style post", then there is no reason
that I, as a reader, would want to click on a front-page news item and
be taken to a page that reiterates exactly the same text.  As a
reader, the only reason I want to click on a link is to read something
I haven't already read.

(The "permalink" issue is a valid one, but the overwhelming majority
of visitors have no reason to care about that.)

> I like that the link URL is now explicit rather than wrapped on the
> title, because it feels wrong to send people to an external site
> without making it clear that this is what's going to happen.

This is the Web; that is what hyperlinks do; if you want to know where
a hyperlink goes before opening it then that's between you and your
browser.

Would it be better if we displayed the host or domain name in
parentheses after the link?

tompollard:
This is good with me, thanks! @bemoody ?

bemoody:
Well, to begin with, I find this very difficult to read since you are
doing a bunch of separate things and your commit messages don't
explain them.

Why did you delete console/templates/console/news_done.html and
console/templates/console/news_list.html?

In console/views.py, it looks like news_edit will change the news
item's created_by.  That seems quite wrong.

notification/templates/notification/news_item.html is nearly a
duplicate of news.html.  The duplicated content should be moved
into another file.  Also, the <title> of news_item.html should be
the title of the news post.  And the indentation of line 40 is
weird.

In notification/views.py, this is wrong in a number of ways:

    except News.DoesNotExist:
        messages.success(request, 'News requested was not found.')
        return redirect('news')

"Does not exist" is not a "success", 'messages' is not defined in this
scope, /news/ doesn't display messages, and a nonexistent item should
not redirect to /news/ anyway (it should just return 404.)

In templates/home.html, your indentation is again wacky.

Please consider rewriting this as a series of logical commits so that
it is possible to follow what exactly is being changed (with respect
to 'dev'.)  I don't want to see changes in one commit that are then
undone in the next commit.  I want to be able to *read* your patches
and commit messages and understand exactly what you're proposing to
change.  I should not have to squash your patches myself, and I should
not have to try to guess your intentions.

And consider dividing these changes into two separate pull requests,
because I don't see any reason that your changes to the model (adding
created_by, and associated changes in console/) have any bearing on
your changes to the public-facing presentation of news items, or vice
versa.

elfeto:
Ok :)

PR #587, opened 2019-07-10
Project types dynamically added from database

lbulgarelli:
Currently, the project types (e.g. Database, Challenge, ...) are hard-coded and need a reasonable amount of work to be changed.

These changes make it so the project types are added from the fixtures written to the database, so any adjustment in the name/description on new types are added without having to change much code.

- Model `ProjectType` is added, containing `id`, `name`, and `description`
- Fixtures for `Database`, `Software`, `Challenge` and `Model` were created
- Views, forms and html pages were updated to use the `id`, which are consistent to the previously used numbering

This is the first change necessary to make all project metadata and content also dynamically added from the database instead of hard coded, but I will make distinct pull requests as it might require a lot of review/tests.


bemoody:
I like the general concept here (obviously need to get it working first. :))

Without having tested it at all, the question that immediately comes to mind is:

        migrations.AlterField(
            model_name='activeproject',
            name='resource_type',
            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE,
related_name='activeprojects', to='project.ProjectType'),
        ),

... does this actually work?!

lbulgarelli:
> I like the general concept here (obviously need to get it working first. :)) Without having tested it at all, the question that immediately comes to mind is: migrations.AlterField( model_name='activeproject', name='resource_type', field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='activeprojects', to='project.ProjectType'), ), ... does this actually work?!

It should, as the type of this foreign key is the same as the current field, as well as the values for each correspondent `resource_type`.

bemoody:
This doesn't work.

* It's not possible to actually perform this migration because you've
  (a) added a new table, and (b) changed other tables to have a
  mandatory dependency on it, in a single step.

  (If you try to run this migration, it will complain that each
  existing project's resource_type refers to a nonexistent
  ProjectType, which of course it does: no ProjectTypes can have been
  created until the table has been created.)

  What you need to do instead is perform one migration to create the
  new table, THEN populate it with data, THEN perform a second
  migration that changes other tables to point to the newly-inserted
  data.

* The second migration can't be done gracefully because it renames the
  column (from 'resource_type' to 'resource_type_id'.)

  (If you perform the migration before restarting the server, then the
  server will continue trying to access the no-longer-defined
  'resource_type' column.  If you restart the server before performing
  the migration, then the server will begin trying to access the
  not-yet-defined 'resource_type_id' column.)

  The simple way to fix this is to set 'db_column'; i.e.

    resource_type = models.ForeignKey('project.ProjectType',
                                      db_column='resource_type',
                                      related_name='%(class)ss',
                                      on_delete=models.CASCADE)

  (tested only with sqlite - I am guessing that the postgres backend
  will work the same way?)

lbulgarelli:
> * It's not possible to actually perform this migration because you've
>   (a) added a new table, and (b) changed other tables to have a
>   mandatory dependency on it, in a single step.

The migration was performed with no problems for me, which can mean that it ignores the current data when adding a foreign key (?). But I agree this can be an issue, and added the loading of the necessary fixtures to the migration.

> * The second migration can't be done gracefully because it renames the
>   column (from 'resource_type' to 'resource_type_id'.)

I also didn't have issues here, but I guess this is related to me having the debug mode on. I've added set the db_column in the code.

Thanks for the tips!

bemoody:
I think we *probably* want to use PROTECT rather than CASCADE on these foreign keys - i.e., if somebody tried to delete a project type, it should croak rather than implicitly deleting all existing projects of that type.

tompollard:
Nice work Lucas, this looks great to me.  Is line 1091 in `physionet-django/project/models.py` missing the trailing `.id` after `self.resource_type` ?

```
        if self.resource_type == 1:
            languages = self.programming_languages.all()
```

I'd be happy to merge, and it would be great to do this before we switch from the green site to the black site. @bemoody, are you good with these changes?

tompollard:
I have finished testing on staging and all seems to work well. It would be good if we could merge soon, if possible. Any objections?

tompollard:
@lbulgarelli please could you rebase on dev and merge the migrations?

tompollard:
Thanks Lucas

PR #591, opened 2019-07-12
Fix the "Return to the admin console" link

bemoody:
Make this link work correctly and point to the submitted-project info page.

### DIFF
--- a/physionet-django/console/templates/console/unsubmitted_projects.html
+++ b/physionet-django/console/templates/console/unsubmitted_projects.html
@@ -20,7 +20,7 @@
         <tbody>
         {% for project in projects %}
           <tr>
-            <td><a href="{% url 'project_preview' project.slug %}">{{ project.title }}</a></td>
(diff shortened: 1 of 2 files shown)


elfeto:
The problem with this, is that if you come from the admin console, or any other place that isnt the project itself, then it will trigger a 403 Forbidden Access.

bemoody:
What are you talking about?

Normal visitors (project authors) will not ever go to
/projects/gfdnklnalfkad/preview/?Admin=true.  If you go there, you are
either an editor or a troublemaker.

If you are an editor for the given project, then you are authorized to
view the console.

If you are a troublemaker, I really could not care less whether you
see a link to a page that you're not authorized to see.  There's
nothing secret about the link itself.

elfeto:
The preview is not only for the editors, you can see a preview of unsubmitted projects. Right now the link appears to be broken, so this might even fix it. I will play with it to see.

elfeto:
So, there was another page in which administrators can see a preview of a project while the project is still active. I fixed the "return" button there.

PR #597, opened 2019-07-15
Install dependencies in a less insecure way

bemoody:
Add package hashes to 'requirements.txt' so that packages will be
verified before being installed.

(I haven't verified that these packages are *legitimate*, but I have
checked that these are the hashes of the files currently hosted at
files.pythonhosted.org.  Trust-on-first-use is not particularly good
but it's better than "trust everyone all the time". :/)

pip3 will enforce these, and refuse to install packages that don't
match the listed hash(es).  When updating to a new version, obviously,
the hashes will need to be changed as well.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,20 +1,108 @@
-bleach==3.1.0
-coverage==4.4.2
-Django==2.2.3
-django-background-tasks==1.2.0
-django-ckeditor==5.7.1
-django-debug-toolbar==1.9.1
-django-js-asset==0.1.1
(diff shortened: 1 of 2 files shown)


elfeto:
The fail on the tests might be because we are running them on a docker container using ubuntu 14.
As for the packages 
- Pillow, is needed for CKeditor file upload. Right now it's not used, and can be removed.
- psycopg2-binary, for my understanding its not used, and not needed. 
- selenium, the selinium tests are broken and not used right now. I don't think they will be fixed any time soon.

PR #598, opened 2019-07-16
Cleaned the deploy files and moved Django to run as www-data

elfeto:
Some minor changes to the deployment files: 
- Set the permissions of the Django process to the www-data user.
- Moved the socket out of the "/physionet" directory, and set the socket permissions.
- Updated the custom mime file.

### DIFF
--- a/deploy/common/etc/systemd/system/django-background-tasks.service
+++ b/deploy/common/etc/systemd/system/django-background-tasks.service
@@ -10,8 +10,8 @@ KillSignal=SIGINT
 Type=simple
 StandardError=syslog
 NotifyAccess=all
-User=pn
-Group=pn
+User=www-data
+Group=www-data
 
 [Install]
 WantedBy=multi-user.target
--- a/deploy/common/etc/systemd/system/emperor.uwsgi.service
+++ b/deploy/common/etc/systemd/system/emperor.uwsgi.service
@@ -3,10 +3,10 @@ Description=uWSGI Emperor
 After=syslog.target
 
 [Service]
-ExecStart=/usr/local/bin/uwsgi --emperor /etc/uwsgi/vassals
+ExecStart=/physionet/python-env/physionet/bin/uwsgi --emperor /etc/uwsgi/vassals 
 # Requires systemd version 211 or newer
-User=pn
-Group=pn
+User=www-data
+Group=www-data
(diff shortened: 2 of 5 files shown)


elfeto:
There might be a issue changing the user to www-data on the collect static. it might be interesting to see if we could use something like 
`runuser -l nginx -c 'python manage.py collectstatic --noinput'`

bemoody:
I changed the default type to text/plain because it seemed to me that browsers were too eager to treat text files as something other than text.

Compare:
- https://archive.physionet.org/pnw/test1/foo.py
- https://archive.physionet.org/pnw/test1/plain/foo.py

(the only difference is that the former has no Content-Type while the latter has a Content-Type of 'text/plain'.)

----

We have this weird situation where a lot of non-static stuff is stored in the "static" directory (mostly, I guess, because django made it simple to do things that way.)

Are there any files or directories *other* than published-projects where the django server needs to have write permission?

elfeto:
I thought that the custom type was backwards, but it was a misunderstanding, I change it back.
In the static question, I don't quite understand your question. Once a project is published, if its completely open without an "Access policy", then it becomes completely static content, in which is why its added to the static content. 

I believe Django only needs to write to:
 - /data/pn-media/*****
 - /data/pn-static/published-projects/*****
 - /tmp/physionet.log

Not sure about the USGI log file
 - /var/log/uwsgi/*****

bemoody:
Regarding the directories, it seems we should be able to

* add 'pn' user to the 'www-data' group
* change group ownership of /data/pn-media/ and
/data/pn-static/published-projects/ to 'www-data' and make them group
writable (mode 775 or perhaps 1775)

bemoody:
Also needs changing the background tasks service (see pull requests #643, #641)

elfeto:
I changed the permissions as follow:
chmod 1775 /data/pn-media/ /data/pn-static/published-projects /var/log/uwsgi/ 
chown -R www-data:www-data /data/pn-media/ /data/pn-static/published-projects/  /tmp/physionet.log /var/log/uwsgi/ /physionet/deploy

Then I changed the user and group for the services for www-data
 - emperor.uwsgi.service
 - django-background-tasks.service 

Also, i changed the user and group for the uwsgi vassal for www-data

I will commit the changes here.

Issue #588, opened 2019-07-10
*_metadata_preview.html: no section IDs


bemoody:
In the project page, the H2 section headers should each have an ID so that they can be linked to directly.  These IDs are included in challenge_metadata.html, database_metadata.html, model_metadata.html, and software_metadata.html; but not in challenge_metadata_preview.html, database_metadata_preview.html, model_metadata_preview.html, and software_metadata_preview.html.  This means authors previewing a project can't check whether their internal links are correct.

Relatedly:

* The IDs are just plain wrong for challenge pages (<https://alpha.physionet.org/content/bhi-2018-challenge/1.0/#usage-notes> links to the section titled "Evaluation").

* There is a huge amount of duplication here.  Probably we should have a single HTML template instead of eight of them (or perhaps one template for "preview" and one for "published"), and the list of sections generated in Python (even if the list is fixed or semi-fixed for each project type.)

* It would be good to have IDs for H3-H6 section headers too, perhaps automatically generated.  Tricky since we need to prevent them from clashing with "built-in" elements.

Issue #543, opened 2019-06-14
Allowing authors of legacy projects to release new versions


tompollard:
Not really an issue, but something that I thought it would be good to record here. We've had a request from an author of a legacy project to create an updated version (amend files). The following process seems safe, from testing on staging:

1. Ensure the user is registered on the new platform.
2. Add the user as the submitting author of the legacy project using the admin tools (update the PUBLISHED AUTHORS field)
3. To submit the update, it may be necessary to complete required fields (e.g. if the legacy project doesn't have a DOI, a DOI should be added).
4. User is then able to create a new version in the usual way, by selecting the project from their project home (https://alpha.physionet.org/projects/).

If there's anything else that needs thinking about when creating an updated version of a legacy project, please comment here...

bemoody:
We need to add authors to legacy projects anyway (see also issue
#386).  But it would be difficult to tie those to new user accounts
automatically - we could try to associate them by email address when
we know their email address, but may be more work than it's worth.

Can any author create a new version of a project?

When an author creates a new version, are the other authors then
automatically invited?

elfeto:
How about, we add a button near the authorship in which creates a request of a logged user, to try and claim authorship, in which we have to decide ? (might be annoying if people want to spam, and it can be a bit informal)

On new versions the authors are automatically re-created without notice.

tompollard:
> But it would be difficult to tie those to new user accounts automatically - we could try to associate them by email address when we know their email address, but may be more work than it's worth.

I agree it doesn't seem worth the effort spending development time on linking authors to legacy projects. It's a task that's fairly easy to deal with manually.

> Can any author create a new version of a project?

Only the submitting author of the original project is able to create a new version.

> When an author creates a new version, are the other authors then automatically invited?

The original authors are automatically included in the new version, but the author list can then be edited by the submitting author (for example, the order can be changed, authors can be removed, new authors can be added).

PR #602, opened 2019-07-17
Project sections dynamically added from database 

lbulgarelli:
This is the second part of the changes required to add more flexibility on the project content.

Project section (e.g. Background, etc) are added from the fixtures written to the database, so the same as the project types PR, any adjustments on sections are added without having to change code.

- New model `ProjectSection`, composed of name, description and a FK to the project type for which that section is available.
- New model `SectionContent`, which contains FKs to both `ProjectSection` and `CoreProject`, storing in the `content` field, the content for each section of each project
- Fixtures for the current sections of all `Database`, `Software`, `Challenge` and `Model` entities were created
- Fixtures for the demo projects were updated to reflect the changes
- Migration handles moving the content for all existing projects in this new approach
- Views, forms and html pages were updated to use the `ProjectSection` and `SectionContent` to show and update the content of projects


bemoody:
This sounds like a good idea in principle (I haven't looked at your changes at all yet), but:

> New model SectionContent, which contains FKs to both ProjectSection and CoreProject, storing in the content field, the content for each section of each project

...this, at least, sounds wrong.  Page content must be tied to a PublishedProject, ActiveProject, or ArchivedProject - i.e., it is tied to a particular version of the project.  CoreProject, IIUC, is associated with *all versions* of a given project.

Also, I'm not so sure about the use of CASCADE in your foreign keys (here and also in pull #587.)

lbulgarelli:
You are correct. 

I've change the FK to use `PublishedProject`, `ActiveProject`, or `ArchivedProject`  to account fot the version number.

Also change `on_delete` to `models.PROTECT` so we don't accidentally delete types or sections being used by any project (same for #587)

bemoody:
Some more issues that jump out at me:

- There is apparently no *order* to the sections, in either the forms
or the rendered pages (e.g., published_project.html just does "{% for
c in project.content.all %}")

- logic like "project.content.all" doesn't belong in the template files anyway

- "content" is not a good name for a field

- now there are *no* section IDs (issue #588) so obviously the table
of contents is broken


And I find this to be poorly documented overall.  You need to
*comment* your code, so that someone reading it can understand the
structure of the database and the logic involved in manipulating it.
You need to write detailed *commit messages*, describing why you are
making changes, and what the impact of the changes will be.  You
should probably reorder and squash many of these commits to make them
more comprehensible.  Overall, you need to write more to explain what
you are doing and why.

bemoody:
- Creating new version doesn't copy over existing text.

- /project/*/content/ refuses to save unless all required fields are filled in.

lbulgarelli:
> - There is apparently no *order* to the sections, in either the forms
> or the rendered pages (e.g., published_project.html just does "{% for
> c in project.content.all %}")

I've added a `default_order` field to `ProjectSection`, which solves this problems.

The idea is to be able to enable authors to add custom sections later and place it between the default sections, but I think the ordering among the default sections should be static, so this is why I placed this field here.

> - logic like "project.content.all" doesn't belong in the template files anyway

Why do you think this can't be here? I've found a lot of examples using it and the django docs actually cite using this as an example to iterate through objects tied to a foreign key. (https://docs.djangoproject.com/en/2.2/ref/templates/language/#accessing-method-calls)

It might be redundant to add a separate argument with all the content in the view, as the foreign keys are already passed with the project. Anyway, I don't have any hard feelings about moving this, just trying to figure out the pros and cons of each approach.

> - "content" is not a good name for a field

I agree and have changed it.

> - now there are *no* section IDs (issue #588) so obviously the table
> of contents is broken

I completely forgot about the table of contents and also had forgotten about adding the `References` section back to the template, but fixed both now.

> - Creating new version doesn't copy over existing text.

True, this should be fixed now.

> - /project/*/content/ refuses to save unless all required fields are filled in.

My mistake, fixed this too.

----

I've also added some comments in the code.

tompollard:
Sorry this has been hanging around so long Lucas. It makes sense for the fields to be in the database and we should try to get this merged. If you have a chance, please could you try resolving the conflicts? Let me know if you don't have time, and I'll give it a go.

physionet-django/project/views.py
```diff
@@ -620,14 +622,54 @@ def project_content(request, project_slug, **kwargs):
...
+    # Creates forms for each section of this project's content type
+    section_forms = []
+    sections = ProjectSection.objects.filter(resource_type=project.resource_type).order_by('default_order')
+    for s in sections:
+        try:
+            # Try to get currently existing content for section
+            section_content = project.project_content.get(project_section=s)
+            section_forms.append(forms.SectionContentForm(instance=section_content))
+        except:
+            # Creates form with empty instace in case content is not found
```
tompollard: Please could you add the exception type? 

> PEP8: "When catching exceptions, mention specific exceptions whenever possible instead of using a bare except". https://www.python.org/dev/peps/pep-0008/#id51.

physionet-django/project/views.py
```diff
@@ -620,14 +622,54 @@ def project_content(request, project_slug, **kwargs):
...
+
+        # Creates forms for each section of this project's content type
+        valid = True
+        section_forms = []
+        for s in sections:
+            try:
+                # Try to get currently existing content for section
+                section_content = project.project_content.get(project_section=s)
+                sf = forms.SectionContentForm(data=request.POST, instance=section_content)
+            except:
```
tompollard: Please could you add the exception type? 

> PEP8: "When catching exceptions, mention specific exceptions whenever possible instead of using a bare except". https://www.python.org/dev/peps/pep-0008/#id51.

tompollard:
Is there a reason not to add `project.abstract` and `project.release_notes`to `project.project_content` too? This would make it easy to deal with the section ordering issue that we discussed.

physionet-django/project/models.py
```diff
@@ -419,16 +456,12 @@ class Metadata(models.Model):
...
     title = models.CharField(max_length=200, validators=[validate_alphaplus])
     abstract = SafeHTMLField(max_length=10000, blank=True)
-    background = SafeHTMLField(blank=True)
-    methods = SafeHTMLField(blank=True)
-    content_description = SafeHTMLField(blank=True)
-    usage_notes = SafeHTMLField(blank=True)
-    installation = SafeHTMLField(blank=True)
-    acknowledgements = SafeHTMLField(blank=True)
-    conflicts_of_interest = SafeHTMLField(blank=True)
+    release_notes = SafeHTMLField(blank=True)
```
tompollard: Perhaps abstract and release_notes should be part of the SectionContent model? This would deal with the ordering issue that we have on the content form at the moment (release notes appearing after the abstract). It's possible that future projects will not have abstracts or release notes.

physionet-django/project/forms.py
```diff
@@ -436,90 +444,47 @@ def save(self):
...
-        data = self.cleaned_data['version']
-        if data in [p.version for p in self.instance.core_project.publishedprojects.all()]:
-            raise forms.ValidationError('Please specify a new unused version.')
-        return data
+    class Meta:
+        model = SectionContent
+        fields = ('section_content',)
+
+    def __init__(self, project=None, project_section=None, *args, **kwargs):
+        # Creates a new instace of SectionContent if none is passed as argument
```
tompollard: nitpick, but there is a typo in "instance"  ("instace") throughout the comments

tompollard (commented):
Currently, it is possible for an editor to accept a project in the copyediting stage (from http://127.0.0.1:8000/console/submitted-projects/SLUG/copyedit/), even if the project doesn't meet requirements for publication. This can leave a project in an unrecoverable state (not publishable, not editable). Should the "Complete copyedit" carry out the same check that is done before a project is submitted?

tompollard:
@bemoody please could you take a look at this pull request and let us know if you would like any more changes before merging? I'd like to check a couple more things, but I'm basically good to merge...

bemoody:
There's a lot here.  2996 lines of changes and no commit messages to
speak of.  What I would *like* is for this to be rewritten as a series
of logical commits that show how to get from A (where we are right
now) to B (where we ought to be.)  What changes are necessary?  Why?


One thing that is apparent is that this requires a two-stage (at
least) migration: need to create and populate the new tables, then
reload the server, then delete the old columns.

You've gone from 19 queries to load a project page, to 37.  I find it
a bit worrying that you're now doing a separate query for every single
section:

SELECT "project_projectsection"."id",
"project_projectsection"."name",
"project_projectsection"."description",
"project_projectsection"."resource_type",
"project_projectsection"."default_order",
"project_projectsection"."required" FROM "project_projectsection"
WHERE "project_projectsection"."id" = 10

SELECT "project_projectsection"."id",
"project_projectsection"."name",
"project_projectsection"."description",
"project_projectsection"."resource_type",
"project_projectsection"."default_order",
"project_projectsection"."required" FROM "project_projectsection"
WHERE "project_projectsection"."id" = 14

SELECT "project_projectsection"."id",
"project_projectsection"."name",
"project_projectsection"."description",
"project_projectsection"."resource_type",
"project_projectsection"."default_order",
"project_projectsection"."required" FROM "project_projectsection"
WHERE "project_projectsection"."id" = 9

...and so on.  (You also seem to be querying the same objects more than once.)

Part of that may be due to the GenericRelation stuff.
https://lukeplant.me.uk/blog/posts/avoid-django-genericforeignkey/
gives a (fairly persuasive, IMHO) explanation of why
GenericForeignKey/GenericRelation shouldn't be used in situations like
this.


The section IDs should stay the same even if section titles change.
For example, here you've changed #description to
#software-description, which will break links.  Instead of dynamically
generating IDs from the section titles, make this an explicit field
("html_id"?) in ProjectSection.

(The field "name" bothers me a little too; it should be "title".)

Messy indentation in ProjectSection and SectionContent:

+    resource_type = models.ForeignKey('project.ProjectType',
+                                    db_column='resource_type',
+                                    related_name='%(class)ss',
+                                    on_delete=models.PROTECT)

+    project_section = models.ForeignKey('project.ProjectSection',
+                                    db_column='project_section',
+                                    related_name='%(class)ss',
+                                    on_delete=models.PROTECT)


There are irrelevant bits of cruft (whitespace changes in
project/views.py, fixing broken HTML in
project/templates/project/published_project.html).  Don't do that.


This is ugly:

+            # Only saves each `section_content` if it's not empty
+            for sf in section_forms:
+                sf.save()
+                text = unescape(strip_tags(sf.instance.section_content))
+                if not text or text.isspace():
+                    sf.instance.delete()

Setting aside the strange
insert-something-nominally-invalid-before-immediately-deleting-it
logic, there is no reason for this to be done in two places.  Just
override save().


This is messy too:

+        # Content
+        sections =
ProjectSection.objects.filter(resource_type=self.resource_type,
required=True)
+        for attr in sections:
+            try:
+                text =
unescape(strip_tags(self.project_content.get(project_section=attr).section_content))
+                if not text or text.isspace():
+                    raise
+            except ObjectDoesNotExist:
+                self.integrity_errors.append('Missing required field:
{0}'.format(attr.name))

raise what, a RuntimeError?

Also, better not to use ObjectDoesNotExist here - better to catch the
particular subtype of ObjectDoesNotExist; in this case, I think it
would be SectionContent.DoesNotExist.

And don't duplicate the unescape/isspace thing - make it a method of
SectionContent, or a utility function if it's something that will be
used in multiple places.

         # Metadata
-        for attr in ActiveProject.REQUIRED_FIELDS[self.resource_type.id]:
+        meta = ['title', 'abstract', 'version', 'license', 'short_description']
+        for attr in meta:
             value = getattr(self, attr)
             text = unescape(strip_tags(str(value)))
             if value is None or not text or text.isspace():
-                l = self.LABELS[self.resource_type.id][attr] if attr
in self.LABELS[self.resource_type.id] else attr.title().replace('_', '
')
+                l = attr.replace('_', ' ').capitalize()
                 self.integrity_errors.append('Missing required field:
{0}'.format(l))

Here you had a constant with a meaningful name (REQUIRED_FIELDS) and
you're replacing it with a hardcoded list buried in the middle of a
function.  Don't do that.

lbulgarelli:
@bemoody I've done some refactoring based on your comments.

Can you please check the changes? I'll squash the commits once we're happy with the content on the PR.

bemoody:
I can see that you've made a number of improvements.  Thanks!

> console/templates/console/copyedit_submission.html
OK

> console/views.py

* Please remove the now-unneeded import statements
(unescape/strip_tags).

> project/fixtures/demo-project.json

* Please remove all of the project.licenses and
project.programminglanguages.  They do not belong here.

* There is a lot of stuff modified here that is not relevant to this
pull request.  Please document the exact commands that you ran in
order to create this file, so that somebody else can repeat the same
process and obtain exactly the same file that you committed.

> project/fixtures/project-sections.json

* Please correct the following:

pk 4 (database, Usage Notes): html_id => "usage-notes"
pk 5 (database, Release Notes): html_id => "release-notes"
pk 7 (database, Conflicts of Interest): html_id => "conflicts-of-interest"
pk 10 (software, Technical Implementation): html_id => "implementation"
pk 12 (software, Usage Notes): html_id => "usage-notes"
pk 13 (software, Release Notes): html_id => "release-notes"
pk 15 (software, Conflicts of Interest): html_id => "conflicts-of-interest"
pk 16 (challenge, Objective): html_id => "objective"
pk 17 (challenge, Participation): html_id => "participation"
pk 19 (challenge, Evaluation): html_id => "evaluation"
pk 20 (challenge, Release Notes): html_id => "release-notes"
pk 22 (challenge, Conflicts of Interest): html_id => "conflicts-of-interest"

and for consistency

pk 24 (model, Technical Implementation): html_id => "implementation"
pk 27 (model, Usage Notes): html_id => "usage-notes"
pk 28 (model, Release Notes): html_id => "release-notes"
pk 30 (model, Conflicts of Interest): html_id => "conflicts-of-interest"

> project/forms.py

* Why is clean_version moved?

* Active project forms are broken!  If you:

- Check out 020c4d1d106a3de77b05819c508e67525876fc9b and load the demo
- Open a browser, log in, and go to /projects/SHuKI1APLrwWCqxSQnSk/content/
- Merge f17b90aeba6392f6678028d19371b4d3a87b006f and run the migration
- Go back to the browser and click "Save Description"

...it erases all of the sections except the abstract.

So we need to be sure that the HTML form field names stay the same, or
explicitly provide a mapping for the old names.

> project/migrations/0026_auto_20190716_1437.py

* Please remove all of the "RemoveField" operations for the
PublishedProject model, and put those into a second migration instead.

* Please rename the new migrations so that they are in numerical order
(i.e. the migration after 0034_auto_20191114_1301.py should be
0035_*.py, not 0026_*.py.)  This doesn't make a huge difference but
should make things less confusing.  While you're at it, you might as
well give these migrations descriptive names.

So you should have:

- '0035_add_project_sections' which depends on '0034_auto_20191114_1301'
   - create new tables
   - copy old column data into the new tables
   - remove columns from ActiveProject and ArchivedProject

- '0036_remove_fixed_content_fields' which depends on
'0035_add_project_sections'
   - remove columns from PublishedProject

(This means that if anyone tries to view/edit an active project during
the transition period, it will throw a 500 error.  That's not good,
but it's better than losing data silently.)

> project/models.py

* ActiveProject.archive (yuck) needs to transfer over the SectionContent
the same way ActiveProject.publish does.

* Instead of:

    class SectionContent(models.Model):
        ...
        project = GenericForeignKey('content_type', 'object_id')
        ...

write something like:

    class SectionContent(models.Model):
        class Meta:
            abstract = True
        ...

    class PublishedSectionContent(SectionContent):
        project = models.ForeignKey('project.PublishedProject', ...)

    class ActiveSectionContent(SectionContent):
        project = models.ForeignKey('project.ActiveProject', ...)

    class ArchivedSectionContent(SectionContent):
        project = models.ForeignKey('project.ArchivedProject', ...)

(Seriously, I want ArchivedProjects to die, but that's a matter for
another day.)

> project/templates/project/challenge_content.html
> project/templates/project/challenge_content_preview.html
> project/templates/project/database_content.html
> project/templates/project/database_content_preview.html
> project/templates/project/model_content.html
> project/templates/project/model_content_preview.html
> project/templates/project/project_content.html
OK

> project/templates/project/project_preview.html
> project/templates/project/published_project.html

* project.release_notes doesn't exist anymore, does it?

* Please submit the bugfix in line 191/197 as a separate pull request.

> project/templates/project/software_content.html
> project/templates/project/software_content_preview.html
OK

> project/views.py

* Please remove trailing whitespace in project_content.

* Please don't add/remove irrelevant blank lines in published_project.

lbulgarelli:
@bemoody Sorry it took some time. I've addressed the points above.

> * Why is clean_version moved?

The version is placed under the discovery tab in the project management. Therefore I moved the function to properly validate the correct form.

> * project.release_notes doesn't exist anymore, does it?

It does, but per Tom's request, it is also a dynamic section now.

bemoody:
Thanks!

A few fairly small things:

* ActiveProject.archive() needs to convert the associated
ActiveSectionContent into ArchivedSectionContent (just as
ActiveProject.publish() converts it into PublishedSectionContent.)

* project_preview.html and published_project.html still refer to
project.release_notes.

* I don't like seeing this sort of stuff in templates:

    {% for c in
project.project_content.all|dictsort:"project_section.default_order"
%}

This is logic that would be much more understandable, more
maintainable, and likely more efficient, if it were written in Python
as a project class method or property.

("default_order" implies that there might be another order.  If you
were going to implement a different order, would it be better to
change the logic in one place, or four?)


A couple of larger things:

* Admin interfaces are broken:

- /admin/project/publishedproject/2/change/ gives a FieldError

- /admin/project/activeproject/2/change/ doesn't... but it also
  doesn't display or give any way to edit the section content.  Same
  for archived projects.

It looks like this should be straightforward to fix; look at how
PublishedAuthor is handled.

* It is not necessary, but it might be convenient, if the legacy
projects' "full_description" were also converted into a section.

We would probably still want a bit of special casing in the HTML
templates, unfortunately, but it's good to simplify as much as we can,
and this should enable us to fix issues #662 and maybe #543 without a
lot of additional effort. :)

lbulgarelli:
> * ActiveProject.archive() needs to convert the associated
> ActiveSectionContent into ArchivedSectionContent (just as
> ActiveProject.publish() converts it into PublishedSectionContent.)

Done.

> * project_preview.html and published_project.html still refer to
> project.release_notes.

Removed.

> * I don't like seeing this sort of stuff in templates:
> 
>     {% for c in
> project.project_content.all|dictsort:"project_section.default_order"
> %}
>
> 
> This is logic that would be much more understandable, more
> maintainable, and likely more efficient, if it were written in Python
> as a project class method or property.
>
> ("default_order" implies that there might be another order.  If you
> were going to implement a different order, would it be better to
> change the logic in one place, or four?)

I've moved the ordering to the view.

> * Admin interfaces are broken:
> 
> - /admin/project/publishedproject/2/change/ gives a FieldError
> 
> - /admin/project/activeproject/2/change/ doesn't... but it also
>   doesn't display or give any way to edit the section content.  Same
>   for archived projects.

Added the content to the admin interface for both active and published projects.

> * It is not necessary, but it might be convenient, if the legacy
> projects' "full_description" were also converted into a section.

Well, I thought of doing this, but one of the things that these changes actually allow us to do is to get rid of legacy projects for good. We can use the html header tags to create custom sections for those projects, and even fit some in our standard sections. This would achieve a lot be a lot better results, but it might be tricky to do.

I can either just move `full_description` on this PR or create another one doing the above, which will take longer. What do you think?

bemoody:
Thanks!

> Well, I thought of doing this, but one of the things that these changes
> actually allow us to do is to get rid of legacy projects for good. We can
> use the html header tags to create custom sections for those projects, and
> even fit some in our standard sections. This would achieve a lot be a lot
> better results, but it might be tricky to do.

Well, I agree that sounds good.  What would that look like (in terms
of models and fields)?

Currently, you have the titles and the order of project sections
defined by ProjectSection objects, which are coupled to resource
types, not to particular projects.

tompollard:
Conflicts again :/ We need to deal with this soon...

elfeto:
from django.core.exceptions import ObjectDoesNotExist
not needed
physionet-django/console/views.py
physionet-django/project/views.py

Wrap the lines 350-359 in an atomic transaction
physionet-django/console/views.py

Wrap the lines 647-650 in an atomic transaction
physionet-django/project/views.py

The "clean_version" in `physionet-django/project/views.py` should be altered

I dont undestant the save on `SectionContentForm`. If the form is not valid then you cannot call the save function
hence line 510 will never occur to my understanging
physionet-django/project/forms.py

I didnt looked at the demo json info

I cant say that the admin page addition is working, but did it ever worked properly?

As an FYI, no section is created when a project is created. they are generated, as when you go to the content page.

Most of the things can be fixed in separate PR (I think)

elfeto:
assuming we rebase, and fix conflicts. Is there any other reason not to merge?

lbulgarelli:
@elfeto #889 should be reviewed and merged into this branch first

I will work on rebasing.

tompollard:
Okay, let's try to get this merged while it's fresh! It's a big change, so everyone needs to be happy. @bemoody and @elfeto please could you:

- note any changes that you think are \*required\* before merging. We should probably pick these up ourselves.
- confirm if you are good to merge to dev

I'll do the same (and @Lucas-Mc if you have thoughts then please add them too). Once we've merged, we can do some heavy testing on staging.

bemoody:
Ugh, that does not make this easier to review. :(

tompollard:
@bemoody You mean merging the legacy PR? Want to chat?

bemoody:
Sorry, I will need to take time to read over this stuff and try
again to understand it.

The main question which I don't think was ever answered is *how
does this actually change the data model*?  Currently a project
has a particular set of fields and some of them are required and
some are optional depending on the project type.  If you're
proposing to break up legacy projects into arbitrary sections,
then that implies a very different data model, and I don't fully
understand how the one maps to the other.  Without understanding
that I can't really assess the quality of implementation or even
whether the new data model is fundamentally the right one for
(how I think) publication ought to work.

elfeto:
I think I can answer you this.
> \*how does this actually change the data model\*?

I would ask you to clarify in what specificly you mean here. 
For example...
At the moment this PR is only to migrate the content we have to from hardcoded to DB based.
This PR is keeping the same requirements of fields as we currently have. 

There is a fixture that sets this.

The ID for the sections are mapped in the following way:
ID Name Required
**Database**
1 Background
2 Methods
3 Data Description
4 Usage Notes
5 Release Notes - False
6 Acknowledgements - False
7 Conflicts of Interest

**Software**
8 Background 
9 Software Description
10 Technical Implementation - False
11 Installation and Requirements
12 Usage Notes
13 Release Notes - False
14 Acknowledgements - False
15 Conflicts of Interest

**Challenge**
16 Objective
17 Participation
18 Data Description
19 Evaluation
20 Release Notes - False
21 Acknowledgements - False
22 Conflicts of Interest

**Model**
23 Background
24 Technical Implementation - False
25 Model Description
26 Installation and Requirements
27 Usage Notes
28 Release Notes - False
29 Acknowledgements - False
30 Conflicts of Interest

I didn't wrote required True because they were too many. 

In small words the mapping of the legacy is done by parsing the content
looking for the header tags and creating a custom published section for that
project. This custom public section is not part of the sections mentioned above.

elfeto:
Note. For the legacy, the only mapping is to the `PublishedSectionContent` from the as a ForeignKey key in the project `section_content`. So if I understand this correctly, all legacy projects will have different sections and its only mapped/accessible from that project itself.

bemoody:
Right.  So then the question is, what does it mean for a particular
project *not* to have all of those sections?  Or to have a section
that is not in the list?

The requirements for now will stay the same - but the idea (I presume)
is to allow those requirements to be changed over time.  Suppose that
in the future, we wanted to replace the "Usage Notes" section with
"Usage Examples", and add a "Research Ethics" section between
"Acknowledgements" and "Conflicts of Interest"?

How would that be represented - assuming that *new* projects are
required to follow the new format, while the format of existing
published projects cannot be changed?

elfeto:
> Right.  So then the question is, what does it mean for a particular project *not* to have all of those sections? 

I would say that it's a legacy project. For a NON-legacy it cannot be published and it should not be possible to be submitted for an editor review. It would be a broken project (this should not happen)

> Or to have a section that is not in the list?

Apart from Legacy, I would say this is not possible. What is done in this PR when a project is created, is that one gets ALL of the section available for the content type. Only those section can exist on the project. If its not in the list it **should not** appear anywhere. 

> Suppose that in the future, we wanted to replace the "Usage Notes" section with "Usage Examples", and add a "Research Ethics" section between "Acknowledgements" and "Conflicts of Interest"?

Since what you mention here is a rename of an existing field in the `ProjectSection` then with the current code the name of the published section will change.

On the add a section in between, for the published projects it will do nothing, but for new projects and ongoing will show the section.

A small note on this, since I just tested it out. If the section is required and it's a new section the red * doesn't show.

>How would that be represented - assuming that *new* projects are required to follow the new format, while the format of existing published projects cannot be changed?

I think I responded this on top, but maybe not. The sections on the published projects are a set of `PublishedSectionContent ` as `ForeignKey` e.project_content. This means that the list of sections is set when the project is published. If you add more sections, it won't make a difference for the project.
At the moment the only difference would be if you rename a `ProjectSection`.

elfeto:
Maybe the best way yo do this, is by using the name `custom_title` and `custom_order` under `PublishedSectionContent` to make this fixed. (like Legacy does)

lbulgarelli:
We do need to implement `custom_order` and `custom_title` in a way that if the default is changed, the projects already published will not be affected.

> A small note on this, since I just tested it out. If the section is required and it's a new section the red * doesn't show.

This is actually because of the way the red star actually works, you need to have a `*` in the end of the description field. I don't think this is ideal, but it is how the current system works and I did not change it in the PR.

lbulgarelli:
One other thing is, the best way to implement this is not allowing `ProjectSection` to ever be removed, but deactivated instead. That way we will always be able to track the changes that were made to the project structure.

We can also make it so that changes in the default order or title be made by creating a new `ProjectSection` and disabling the old one. The actual goal for `custom_title` and `custom_order` when I created it was for both the legacy projects and a possible feature where the users could create extra custom sections to their projects.

bemoody:
Starting from dev (c1af3d66ba16da8755e67085b8cf093b9a403599), I tried
to follow Felipe's instructions for loading legacy projects, which
failed:

    $ manage.py resetdb && manage.py loaddemo && manage.py loaddata
pbank && manage.py loaddata ptools
    $ manage.py shell
    >>> from project.models import LegacyProject, ActiveProject,
PublishedProject
    >>> l = LegacyProject.objects.all()
    >>> for item in l: item.publish()
    ...
      File "<console>", line 1, in <module>
      File "/home/benjamin/physionet-build/physionet-django/project/models.py",
line 1956, in publish
        license=License.objects.get(name='Open Data Commons
Attribution License v1.0')
      File "/home/benjamin/ve-pndjangoplus/lib/python3.7/site-packages/django/db/models/manager.py",
line 82, in manager_method
        return getattr(self.get_queryset(), name)(*args, **kwargs)
      File "/home/benjamin/ve-pndjangoplus/lib/python3.7/site-packages/django/db/models/query.py",
line 420, in create
        obj = self.model(**kwargs)
      File "/home/benjamin/ve-pndjangoplus/lib/python3.7/site-packages/django/db/models/base.py",
line 483, in __init__
        _setattr(self, field.name, rel_obj)
      File "/home/benjamin/ve-pndjangoplus/lib/python3.7/site-packages/django/db/models/fields/related_descriptors.py",
line 211, in __set__
        self.field.remote_field.model._meta.object_name,
    ValueError: Cannot assign "0": "PublishedProject.resource_type"
must be a "ProjectType" instance.

Having this stuff stored as "LegacyProject" objects is not really helpful. :/

elfeto:
I fixed the piece of code by hand and forgot to mention it there. I just did a PR to fix that.

Also, before merging #889, I did a commit to fix that.
I just did a PR to fix that in dev #947

tompollard:
I'm just trying to work out where we are with this. There's a conflict on the fixture but that's easy to resolve. @elfeto are the issues mentioned in your comment above dealt with? Does that address the error mentioned by @bemoody?

elfeto:
@tompollard 
Yea the error on the publish is resolved.
You can try publishing the legacy using this:
```
./manage.py resetdb && ./manage.py loaddemo && ./manage.py loaddata pbank

./manage.py shell

for item in LegacyProject.objects.all(): item.publish()
```
I will take a look at the fixture and test it again to create a checklist of things to do.
From memory I don't think or cant remember there being something that popped out to me that had to be dealt with before merging if all are content.

tompollard:
Conflicts again :/ @bemoody, are you in general happy with this PR? If so, we should make a push to get it merged. If not, please could you highlight the major issues that you think need addressing?

lbulgarelli:
Alright, should I rebase this? 😨

tompollard:
@lbulgarelli  to avoid your wasting time, it's probably best to save the rebase until we are completely ready to merge. I'll take a look and will post again here in a few days. 

I think we are at a point where we are seriously thinking about adding new content types (e.g. tutorial), so we have a strong motivation to get this done. Holiday season might be a good time for this...fewer users to disrupt if things don't go to plan!

lbulgarelli:
🎂

tompollard:
erm @lbulgarelli how would you feel about rebasing this?

physionet-django/console/views.py
```diff
@@ -340,57 +349,55 @@ def copyedit_submission(request, project_slug, *args, **kwargs):
...
+    # according to its content type
+    valid = True
+    section_forms = []
+    sections = ProjectSection.objects.filter(resource_type=project.resource_type).order_by('default_order')
+    for s in sections:
+        form = project_forms.SectionContentForm(project=project, 
+            project_section=s, data=data_edit)
+        section_forms.append(form)
+        # Validation of all `section_content` forms
+        valid = valid and form.is_valid()
```
fiedosiukr: Probably the forms shouldn't be validated unless a POST request is sent. We should only initialize them here.

physionet-django/console/views.py
```diff
@@ -340,57 +349,55 @@ def copyedit_submission(request, project_slug, *args, **kwargs):
...
+                    and valid):
+
                 description_form.save()
                 access_form.save()
                 discovery_form.save()
                 reference_formset.save()
                 publication_formset.save()
                 topic_formset.save()
+                for sf in section_forms:
+                    sf.save()
```
fiedosiukr: I would suggest to use more meaningful variable names.

physionet-django/project/migrations/0050_parse_legacy.py
```diff
@@ -0,0 +1,24 @@
...
+
+def parse_legacy(apps, schema_editor):
+    projects = PublishedProject.objects.filter(is_legacy=True)
+    for p in projects:
+        p.parse_legacy_content()
+
+def unparse_legacy(apps, schema_editor):
+    projects = PublishedProject.objects.filter(is_legacy=True)
+    for p in projects:
+        p.project_content.all().delete()
```
fiedosiukr: This for loop leads to the famous N+1 queries problem. When you access a related model, a separate query is executed. In order to improve it, you should add `prefetch_related('project_content')` to the initial queryset.

You can also achieve that in one line:
 ```suggestion
def unparse_legacy(apps, schema_editor):
    PublishedSectionContent.objects.filter(project__is_legacy=True).delete()
```

File: physionet-django/project/modelcomponents/activeproject.py [hunk 8]
```diff
@@ -249,6 +226,14 @@ def archive(self, archive_reason):
...
                 archived_project.programming_languages.add(*list(languages))
 
+        # Copy content
+        content = self.project_content.all()
+        for c in content:
+            ArchivedSectionContent.objects.create(
+                project=archived_project,
+                section_content=c.section_content,
+                project_section=c.project_section)
+
```
fiedosiukr: Once again this leads to N+1 queries. You should initialize a list of `ArchivedSectionContent` instances and then use `bulk_create()` to push them into the database.

physionet-django/project/views.py
```diff
@@ -642,43 +646,60 @@ def project_content(request, project_slug, **kwargs):
...
-
+    valid = True
+    section_forms = []
+    sections = ProjectSection.objects.filter(resource_type=project.resource_type).order_by('default_order')
+    for s in sections:
+        form = forms.SectionContentForm(project=project, 
+            project_section=s, data=data, editable=editable)
+        section_forms.append(form)
+        # Validation of all `section_content` forms
+        valid = valid and form.is_valid()
```
fiedosiukr: Same as above.

physionet-django/project/forms.py
```diff
@@ -388,6 +391,13 @@ def save(self):
 
         project.save()
 
+        # Copy content for each project section
+        for c in self.latest_project.project_content.all():
+            ActiveSectionContent.objects.create(
+                project=project,
+                project_section=c.project_section,
+                section_content=c.section_content)
+
```
fiedosiukr: Once again N+1 queries problem.

physionet-django/project/modelcomponents/activeproject.py
```diff
@@ -295,12 +280,24 @@ def check_integrity(self):
...
+        sections = ProjectSection.objects.filter(
+            resource_type=self.resource_type, required=True)
+        for attr in sections:
+            try:
+                section = self.project_content.get(project_section=attr)
+                if not section.is_valid():
+                    raise ActiveSectionContent.DoesNotExist
+            except ActiveSectionContent.DoesNotExist:
+                self.integrity_errors.append(
+                    'Missing required field: {0}'.format(attr.title))
```
fiedosiukr: The related names are kind of complicated. When you access a field called `project_content`, I guess you should expect a single object instead of multiple. I would consider changing all of the related names.

fiedosiukr: The amount of database queries could also be reduced.

physionet-django/project/modelcomponents/activeproject.py
```diff
@@ -486,6 +482,15 @@ def publish(self, slug=None, make_zip=True, title=None):
...
                 published_project.save()
+ 
+                # Copy content
+                content = self.project_content.all()
+                for c in content:
+                    PublishedSectionContent.objects.create(
+                        project=published_project,
+                        section_content=c.section_content,
+                        project_section=c.project_section
+                    )
```
fiedosiukr: `bulk_create()` should be used.

fiedosiukr (changes requested):
The test cases should also be updated to verify whether everything works as intended or not.

physionet-django/console/views.py
lbulgarelli: This only verifies if all forms was valid or not and stores the result, actions are only taken when a POST request is sent. This avoids having to loop twice over the sections and does not affect functionality on requests other than POST.

physionet-django/project/migrations/0050_parse_legacy.py
lbulgarelli: Done.

physionet-django/project/modelcomponents/activeproject.py
lbulgarelli: Done.

physionet-django/project/modelcomponents/activeproject.py
lbulgarelli: Done.

physionet-django/project/modelcomponents/activeproject.py
lbulgarelli: Done.

physionet-django/project/forms.py
lbulgarelli: Done

tompollard:
We should revisit this at some point, but I'm closing for now.

Issue #590, opened 2019-07-12
References in unspecified order


bemoody:
"References" in project pages are displayed as a numbered list, but the order in which they are shown is unspecified, AFAICT.

(In project_preview and published_project, they are retrieved by 'project.references.all()'.)

Even if you assume that the order that references are created is preserved, there's no reasonable way for authors to change that order.

References *should* be listed in alphabetical order anyway (though in some cases it may be hard to do this correctly and automatically.)

bemoody:
Relatedly, each reference should also have an id attribute by which it can be referred to in the text.

It could just be "ref-{%forloop.counter%}", but that relies on there actually being a stable sorting key!

But as an author, I'd actually prefer that the id be something like "ref-goldberger-2003".

tompollard:
I came across this (or a related) issue when moving a URL from a project description to the reference section during the copyediting stage.

Brief summary:

1. visited the copyedit page: https://physionet.org/console/submitted-projects/SLUG/copyedit/
2. clicked the "+" button next to the references to add a reference, as shown in the screenshot below:

![Screen Shot 2020-04-25 at 14 38 17](https://user-images.githubusercontent.com/822601/80287975-a79d2700-8702-11ea-8061-a7058e5b1b21.png)

3. added a new reference at the end of the sequence (ref 9 in this case).

![Screen Shot 2020-04-25 at 14 42 19](https://user-images.githubusercontent.com/822601/80288019-fea2fc00-8702-11ea-99ee-ce8539c4df8a.png)


4. in the project preview (https://physionet.org/projects/SLUG/preview/?Admin=True), the new reference appeared as reference 1.

![Screen Shot 2020-04-25 at 14 42 52](https://user-images.githubusercontent.com/822601/80288028-0f537200-8703-11ea-8f5f-d2a6e638c247.png)

tompollard:
> References should be listed in alphabetical order anyway (though in some cases it may be hard to do this correctly and automatically.)

I'm not sure I agree that alphabetical is best, but we definitely need to come up with a plan. Personally I like numerical citations (e.g. [Vancouver Style](https://guides.lib.monash.edu/citing-referencing/vancouver)):

>> Vancouver is a numbered referencing style commonly used in medicine and science, and consists of:

>> - Citations to someone else's work in the text, indicated by the use of a number.
>> - A sequentially numbered reference list at the end of the document providing full details of the corresponding in-text reference.

tompollard:
[...]

PR #603, opened 2019-07-18
Randomize order of query results by default

bemoody:
In many places, the code requires loading many objects from the database, and displaying the results in some particular order.

When testing using sqlite, the order that objects are returned from a query is generally very predictable, so it's easy not to notice if you have neglected to define an explicit ordering criterion.  This can lead to bugs like #419, #590, new issues in #587 and #602, and probably many others.

So here is one approach: in the development configuration, force sqlite to always *randomize* the order of last resort, so that hopefully such mistakes will become more obvious during initial testing.

----

I'm not entirely sure this is a good idea.  I looked to see if there were any other implementations of this idea, or built-in Django options to do anything similar, and didn't find any, but I may very well have missed something.

(One possible approach would be to set 'ordering = ["?"]' in each model's "Meta" class.  This seems like a poor idea as it requires additional effort per class, when the entire point is to *avoid* such work; and it would have undesirable effects on production, and in cases where the order really doesn't matter.)

### DIFF
--- a/physionet-django/physionet/db/chaotic/compiler.py
+++ b/physionet-django/physionet/db/chaotic/compiler.py
@@ -1,0 +1,40 @@
+from django.db.models.sql import compiler
+from django.db.models.expressions import OrderBy, Random
+
+
+class SQLCompiler(compiler.SQLCompiler):
+    """
+    SQL compiler class that shuffles results by default.
+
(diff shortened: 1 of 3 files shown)


tompollard:
really like this. very useful for testing!

Issue #546, opened 2019-06-18
"Editor Response" form should include additional options (currently just yes/no)


tompollard:
When reviewing a project, the editor is required to select Yes/No for a list of questions before selecting "Accept"; "Reject" or "Revisions accepted". e.g.:

- Has the data been produced in a sound manner? Yes/No
- Is the data adequately described? Yes/No
- Are the data files machine-readable? Yes/No
- Is the project free of protected health information? Yes/No
- Is the content suitable for PhysioNet? Yes/No

Sometimes the editor may want to select something like "More detail needed" or "To be confirmed" or similar, instead of yes or no. Reasons for this include:

-  To determine whether "the data been produced in a sound manner?" may require additional information.
- The editor may choose not to review data files until the authors have provided a detailed descriptive information, so the answer to "Are the data files machine-readable?" etc may be unclear.

elfeto:
I think this was addressed right ?

tompollard:
Yes, thanks - this was addressed, so the issue can be closed.

PR #616, opened 2019-07-24
Allow N/A response to QA form. ref #546

tompollard:
When reviewing a project, the editor is required to select Yes/No for a list of questions before selecting "Accept"; "Reject" or "Revisions accepted". Sometimes these responses are not appropriate (e.g. we might not want to check whether files are machine readable if there are much larger issues that need to be resolved first). This change allows the editor to select N/A as an alternative to yes/no.

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -29,6 +29,13 @@ YES_NO = (
     (0, 'No')
 )
 
+YES_NO_NA = (
+    ('', '-----------'),
+    (1, 'Yes'),
+    (0, 'No'),
+    ('', 'N/A')
+)
+
 
 class AssignEditorForm(forms.Form):
     """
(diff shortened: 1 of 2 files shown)


bemoody:
This is a great idea.

I'm not sold on labelling the third option as "N/A"; that abbreviation suggests to me "this criterion has not been checked because it's irrelevant to this project", rather than "this criterion may or may not be relevant, but hasn't yet been checked because there are other issues that need to be addressed first."

I would instead label the third option as "Unconfirmed", "Unverified", "To be determined", or something along those lines.

As a style issue, instead of this:

```
NO_YES = ('No', 'Yes')
qa_str = '{} {}'.format(labels[f],
    NO_YES[getattr(self, f)] if isinstance(getattr(self, f), bool) else 'N/A')
```

I would write:

```
RESPONSE_LABEL = {True: 'Yes', False: 'No', None: 'N/A'}
qa_str = '{} {}'.format(labels[f], RESPONSE_LABEL[getattr(self, f)])
```

(or whatever that third value actually is.)

I also somewhat dislike having the same text defined in two places (here, we have "YES_NO_NA" defined in forms.py and "NO_YES" defined in models.py).  If it's not convenient to use a single variable for both purposes, then I would add explicit comments cross-referencing the two, e.g. in forms.py:

```
# See also RESPONSE_LABEL in project/models.py
YES_NO_NA = (
    ('', '-----------'),
    (1, 'Yes'),
    (0, 'No'),
    ('', 'N/A')
)
```

and in models.py:

```
# See also YES_NO_NA in console/forms.py
RESPONSE_LABEL = {
   True: 'Yes',
   False: 'No',
   None: 'N/A'
}
```
(well, the variable names are not great, but you get the point.)

tompollard:
Nice suggestions, thanks. I couldn't think of a good way of removing the similar variables, so I took your suggestion and added comments.

PR #617, opened 2019-07-24
Allow the corresponding author to approve publications for others

elfeto:
Here we allow the **submitting** not the **corresponding** author to approve the publications for other authors.


tompollard:
This is great, thanks! `project_submission()` in `project.views` is getting pretty difficult to follow, but that's not the fault of these changes.

PR #620, opened 2019-07-25
Fixes user username change bug

elfeto:
There is a bug that if the user has never added a picture, and tries to
change the username, it will assume that there is a user folder under
the media directory, and try to move it. Here we just check if that
path exists, and if does, then we rename to the new username.


tompollard:
Thanks!

PR #622, opened 2019-07-25
Fixing second user bug

elfeto:
On registration the username was being forced to lower case.
I removed that, and changed the defined user get queries to
be case insensite.

Now admin is the same as Admin.

The thing to note here is that for future code, we need to
use "User.objects.get(username__iexact="
instead of "User.objects.get(username="


tompollard:
Looks great, thanks

bemoody:
Suddenly it becomes messy and inefficient to look up users.  Why allow
non-lowercase usernames?

PR #631, opened 2019-07-29
Adding known reference

elfeto:
Here I clean the code, and use the SAME sorting function we previously used.
I removed the lazy two variables contacted and not contacted applications, 
and use only one "applications"

I also add the known reference if that specific email has been contacted before.

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -14,7 +14,7 @@ from django.shortcuts import redirect, render
 from django.urls import reverse
 from django.utils import timezone
 from django.db import DatabaseError, transaction
-from django.db.models import Q, CharField, Value, IntegerField
+from django.db.models import Q, CharField, Value, IntegerField, BooleanField
 from background_task import background
 
 from notification.models import News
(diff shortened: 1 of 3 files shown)


physionet-django/console/views.py
```diff
@@ -726,27 +726,42 @@ def complete_credential_applications(request):
...
+    for item in applications:
+        if item.reference_contact_datetime:
+            if item.reference_response_datetime:
+                item.time_elapsed = (timezone.now() - item.reference_response_datetime).days
+            else:
+                item.time_elapsed = (timezone.now() - item.reference_contact_datetime).days
+            temp.append([1, item.reference_contact_datetime, item])
+        else:
+            item.time_elapsed = (timezone.now() - item.application_datetime).days
+            temp.append([0, item.application_datetime, item])
```
lbulgarelli: You propably can replace this logic with `.annotate` and then sort by it.

lbulgarelli (changes requested):

physionet-django/console/views.py
elfeto: Resolved in latest commit

lbulgarelli:
Looks good.

PR #635, opened 2019-07-30
News Part 2

elfeto:
Taking into consideration the previous pull request #586, I add here the news by item.

I removed a unnecessary landing page and replaced it with a simple message.
Then I took the same content from the previous pull request that was agreed on and set the proper templates.

Also, if the news is not found it raises a 404 instead of a redirect to the news page.


tompollard:
Works well for me. I like this.

bemoody:
Looks good.  Thanks.

PR #637, opened 2019-07-31
Made page requested by KP

elfeto:
This page will help KP with the credentialing. Its a page for him and him only.
This page is exactly the same as the one he uses now.

Here I just remove ALL of the HTML and leave him with the same as he has in the green site.


tompollard:
It would be good to work out a way of clearly separating these changes from the standard site, so that people reusing the platform don't see them.

PR #638, opened 2019-07-31
Added user pagination in the console

elfeto:
Added user pagination for the console page. It was taking TOO LONG to load.

I used exactly the SAME as Lucas used in the search. word for word, I just replaced project for user and set the amount of users to 100 per page.

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -16,6 +16,7 @@ from django.utils import timezone
 from django.db import DatabaseError, transaction
 from django.db.models import Q, CharField, Value, IntegerField
 from background_task import background
+from django.core.paginator import Paginator
 
 from notification.models import News
 import notification.utility as notification
@@ -617,8 +618,22 @@ def users(request):
     """
     List of users
     """
-    users = User.objects.all().order_by('username')
-    return render(request, 'console/users.html', {'users':users})
+    all_users = User.objects.all().order_by('username')
+
+    # PAGINATION
+    page = request.GET.get('page', 1)
+    paginator = Paginator(all_users, 100)
+    try:
+        users = paginator.page(page)
+    except:
+        users = paginator.page(1)
+
+    querystring = re.sub(r'\&*page=\d*', '', request.GET.urlencode())
+    if querystring != '':
+        querystring += '&'
+
+    return render(request, 'console/users.html', {'users': users,
+        'querystring': querystring})
 
 @login_required
 @user_passes_test(is_admin, redirect_field_name='project_home')
@@ -651,10 +666,24 @@ def users_inactive(request):
     """
     List of users
     """
-    inactive_users = User.objects.filter(Q(is_active=False) | Q(last_login__lt=timezone.now() 
-                + timezone.timedelta(days=-90))).order_by('username')
+    all_inactive_users = User.objects.filter(Q(is_active=False) | Q(
+        last_login__lt=timezone.now() + timezone.timedelta(days=-90))
+        ).order_by('username')
 
-    return render(request, 'console/users.html', {'users':inactive_users})
+    # PAGINATION
+    page = request.GET.get('page', 1)
+    paginator = Paginator(all_inactive_users, 100)
+    try:
+        inactive_users = paginator.page(page)
+    except:
+        inactive_users = paginator.page(1)
+
+    querystring = re.sub(r'\&*page=\d*', '', request.GET.urlencode())
+    if querystring != '':
+        querystring += '&'
+
+    return render(request, 'console/users.html', {'users': inactive_users,
+        'querystring': querystring})
 
 
 @login_required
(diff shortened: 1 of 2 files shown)


physionet-django/console/views.py
```diff
@@ -617,8 +618,22 @@ def users(request):
...
-    users = User.objects.all().order_by('username')
-    return render(request, 'console/users.html', {'users':users})
+    all_users = User.objects.all().order_by('username')
+
+    # PAGINATION
+    page = request.GET.get('page', 1)
+    paginator = Paginator(all_users, 100)
+    try:
+        users = paginator.page(page)
+    except:
```
tompollard: perhaps `except EmptyPage:`?

physionet-django/console/views.py
```diff
@@ -651,10 +666,24 @@ def users_inactive(request):
...
+    all_inactive_users = User.objects.filter(Q(is_active=False) | Q(
+        last_login__lt=timezone.now() + timezone.timedelta(days=-90))
+        ).order_by('username')
+
+    # PAGINATION
+    page = request.GET.get('page', 1)
+    paginator = Paginator(all_inactive_users, 100)
+    try:
+        inactive_users = paginator.page(page)
+    except:
```
tompollard: as above, `except EmptyPage:`?

bemoody:
Thanks.

As you say, this is the same logic as search.views.content_index.  I
don't like it:

    page = request.GET.get('page', 1)
    paginator = Paginator(all_inactive_users, 100)
    try:
        inactive_users = paginator.page(page)
    except:
        inactive_users = paginator.page(1)

    querystring = re.sub(r'\&*page=\d*', '', request.GET.urlencode())
    if querystring != '':
        querystring += '&'

This would be better written as:

    params = request.GET.copy()
    page = params.pop('page', None)
    paginator = Paginator(all_inactive_users, 100)
    inactive_users = paginator.get_page(page)

    querystring = params.urlencode()
    if querystring != '':
        querystring += '&'

although personally I'd like to see something like:

    params = request.GET.copy()
    page = params.pop('page', None)
    limit = params.get('limit', 100)
    paginator = Paginator(all_inactive_users, limit)
    inactive_users = paginator.get_page(page)

    querystring = params.urlencode()
    if querystring != '':
        querystring += '&'

:)

bemoody:
>     page = params.pop('page', None)

or params.pop('page', 1), whatever.

tompollard:
The rewrite looks cleaner, though I don't really see the problem with the original version. One issue with the `pop` approach is that it causes the pagination bar at the end to misreport the page.

```
    all_users = User.objects.all().order_by('username')

    # PAGINATION
    params = request.GET.copy()
    page = params.pop('page', None)
    paginator = Paginator(all_users, 100)
    users = paginator.get_page(page)

    querystring = params.urlencode()
    if querystring != '':
        querystring += '&'
```

elfeto:
Bt doing a simple pop, there is a small problem.

If the variable 'page' does NOT exists,  then its fine.
If it doesn't, then it will return an array.
The function “paginator.page” takes in a number and not an array.

There are so many ways to do this then, I would do:
page = params.pop('page', 1)
if 'page' in request.GET:
    page = page[0]

page = params.get('page', 1)
if 'page' in params:
    params.pop('page')

or just ignore the whole params variable and keep it how it was in the beginning.

bemoody:
1. "except:" without a list of exception types is always wrong (I
think it's pretty safe to say.)  Anyway, Paginator provides a function
that does more or less exactly what you want, so use it.

2. If your goal is "take the current query parameters, excluding the
'page' parameter, and encode it as a new query string", then it's
better to write the logic that way, rather than "take the current
query parameters, encode as a query string, then delete the 'page'
parameter from that encoded string."

That's because removing a parameter from an encoded query string
requires writing a complicated regular expression (note that the
regexp used there is buggy in many cases), while deleting parameters
from a QueryDict is simple.

bemoody:
Sorry, I didn't realize that QueryDict.pop returns a list.  Guess
that's the problem with ignoring type errors :/

elfeto:
Apparently the query string wasn't needed by the pagination in this case.
So, I removed it, and got the page variable the same way that I did in the first commit.
Also, now there are excepts for not integer, and empty pages.

tompollard:
Looks good to me. I'm happy to merge.

bemoody:
Do you dislike the behavior of get_page?  ISTM that the only
difference between get_page and what you have is that with get_page,
?page=99999 will show the *last* page, whereas with yours it'll show
the *first* page.

The code looks okay - @tompollard, please go ahead and merge if you're
satisfied that it works.  But I'll probably make another pull request
later to fix search.views.content_index and unify the logic.

tompollard:
@bemoody thanks, I'll merge for now and I'll look out for your clean up shortly.

PR #639, opened 2019-08-01
Moved pagination to another file and cleaned past credentialed applications

elfeto:
Here remove the try except and use the built ‘get_page’ in as suggested.
I moved it to another file, and just call the function instead of having the same thing in multiple places.

Also, I cleaned the past credential applications page. Now we don’t use 2 variables to differentiate legacy and new applications.

I was going to add pagination here, but it’s a bit more complicated because of the tab. To add pagination, the pages has to be separated.

### DIFF
--- a/physionet-django/console/templates/console/pagination.html
+++ b/physionet-django/console/templates/console/pagination.html
@@ -3,7 +3,7 @@
     <ul class="pagination">
       <li><a href="?{{ querystring }}page=1">&laquo;</a></li>
       {% if pagination.has_previous %}
-        <li><a href="?{{ querystring }}page={{ users.previous_page_number }}">&lsaquo;</a></li>
+        <li><a href="?{{ querystring }}page={{ pagination.previous_page_number }}">&lsaquo;</a></li>
       {% else %}
         <li class="disabled"><span>&lsaquo;</span></li>
       {% endif %}
--- a/physionet-django/console/templates/console/pagination.html
+++ b/physionet-django/console/templates/console/pagination.html
@@ -1,9 +1,9 @@
 {% if pagination.has_other_pages %}
   <div style="text-align: center;">
     <ul class="pagination">
-      <li><a href="?{{ querystring }}page=1">&laquo;</a></li>
+      <li><a href="?{{ querystring }}&page=1">&laquo;</a></li>
       {% if pagination.has_previous %}
-        <li><a href="?{{ querystring }}page={{ pagination.previous_page_number }}">&lsaquo;</a></li>
+        <li><a href="?{{ querystring }}&page={{ pagination.previous_page_number }}">&lsaquo;</a></li>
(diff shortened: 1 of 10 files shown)


bemoody:
I like this.  Some small things:

- I'd put it in physionet.utility, not console.utility

- I'd move the pertinent CSS rules into physionet.css (or else create
  a new 'pagination.css')

- Typo: 'maximun'

- Your doc string doesn't explain what 'paginate' does, what the
  arguments mean or what the function returns.

I still don't like the "querystring" logic in
search.views.content_index and think this should be rewritten.  And
maybe that should be part of 'paginate' as well, although it's hard to
see how to do so elegantly.

I don't follow your changes in past_credential_applications.

elfeto:
The problem with past_credential_applications is that it takes three different variables.
 - s_applications (successful applications)
 - u_applications (unsuccessful applications)
 - l_applications (legacy applications)

A legacy application automatically credentials a person, and since there is no real application in this new server, then its its own structure.

A simple way to remove the excess of variables in the same template is to get all the credentialed users and sort them by credentialed date (which is the real credentialing date).
Here I just rename the rejected or unsuccessful applications and set the proper logic in the template. 

The easy way to know if the person has been credentialed by legacy is to check the amount of credential applications, the rest is just reorganizing the same html.

bemoody:
From what I understand, your assumption is that users fall into one of
these categories:

* never applied, and thus have no CredentialApplications nor
  LegacyCredentials, and is_credentialed is False

* submitted an application to the old system that was accepted, and
  thus have a LegacyCredential, and is_credentialed is True

* submitted one or more applications to the new system that were
  rejected, and thus have one or more CredentialApplications with
  status=1, and is_credentialed is False

* submitted one application to the new system that was accepted, and
  thus have one CredentialApplication with status=2 (which is also
  their most recent), and is_credentialed is True.


I'm not convinced that this assumption holds - there are a lot of
potential corner cases here.  What if someone submits an application
that is rejected, and then associates their account with an email
address that was approved in the old system?  What if someone submits
an application that is accepted, and their access is later revoked?

I appreciate your effort to simplify things, but I think in this case
it's more important to present the information as transparently as
possible.  If I were managing this system, I don't think I'd actually
want to see a complete list of all applications, but I might want to
see a list of all *people who have applied*, along with each person's
current status, and a link to view their past/current applications.

If the page is titled "Successful Applications", though, then I don't
think it should be a list of *users who are currently authorized*.


Also: you changed 'u_applications' to 'rejected_applications' in
Python but you didn't change the template.


(Thanks for your improvements to the pagination issue!)

elfeto:
I understand what you say on the logic. For the time being, I removed the past applications from this PR, and just left the pagination.

elfeto:
@bemoody  we good ?

bemoody:
LGTM, thanks!

Issue #604, opened 2019-07-22
Running of background tasks


bemoody:
This needs to be cleaned up and documented.

It looks like currently on physionet-production, there is a script called '/physionet/process_tasks.sh'.  This script is not in git anywhere AFAICT, and I don't know what invokes this script or when.

Whatever needs to be done to run background tasks, it needs to be fully automated and any necessary configuration files added to 'deploy/common'.

PR #641, opened 2019-08-01
Created service for background tasks

elfeto:
I created a service for the background tasks to run automatically. The service SHOULD start after the uwsgi service. It will log the output to the system log.
Can be managed by:
- systemctl COMMAND django-background-tasks.service

This address #604


bemoody:
Looks good.  One thing is that I would suggest KillSignal=SIGINT
rather than SIGQUIT (because python will catch SIGINT by default, so
the program has a chance to clean up before halting.)

Seems process_tasks also handles SIGTSTP for a *graceful* stop - i.e.,
wait until the currently-running task finishes before exiting - but I
don't think that's what we want here, because we are using
background_tasks specifically to handle things that may take a long
time to finish.

Issue #595, opened 2019-07-15
Set a canonical hostname


bemoody:
In various places, we need to specify the hostname of the web server.

The most obvious place is in email message templates:
 * user/templates/user/email/verify_email_email.html
 * user/templates/user/email/reset_password_email.html
 * user/templates/user/email/register_email.html
 * notification/templates/notification/email/accept_submission_notify.html
 * notification/templates/notification/email/contact_reference.html
 * notification/templates/notification/email/process_credential_complete.html
 * notification/templates/notification/email/copyedit_complete_notify.html
 * notification/templates/notification/email/invite_author.html
 * notification/templates/notification/email/publish_notify.html
 * notification/templates/notification/email/contact_supervisor.html
in which the hostname is specified as '{{ domain }}'.

Another place is in project/templates/project/published_versions.html, where the hostname is specified as '{{ current_site }}'.

We probably want all of the above to use a single canonical hostname, which should be 'alpha.physionet.org' initially but can be switched over to 'physionet.org' once the new server has been promoted and we can be confident it isn't going to explode. :)

In all of the above cases (I think), the hostname is obtained via django.contrib.sites.shortcuts.get_current_site.  Since we are not currently using the sites system, this simply returns the hostname specified in the current page URL.

So probably the "right way" to define a canonical hostname would be to enable the django sites system (https://docs.djangoproject.com/en/2.2/ref/contrib/sites/).  (This could be done just for the staging/production configurations and leave the development configuration alone.)

The "easy way" would be to write our own kludgy replacement for get_current_site.

bemoody:
This is resolved by pull #642 and #804.  If we find any other references to 'alpha.physionet.org' or 'www.physionet.org', please raise those as separate issues.

PR #642, opened 2019-08-01
Enable django site mechanism

bemoody:
As discussed in issue #595, there are various places where the
server's hostname is displayed to the user.

We want to use a single canonical hostname; e.g. we don't want
notification emails to be different depending on whether the person
who triggered them was logged into https://alpha.physionet.org/,
https://physionet.org/, or https://www.physionet.org/.

The canonical hostname for production is set to alpha.physionet.org.
We will of course want to change that later, which is easy.

This should be straightforward, but test it on staging to be sure it
won't break anything.

### DIFF
--- a/physionet-django/physionet/fixtures/sites.json
+++ b/physionet-django/physionet/fixtures/sites.json
@@ -1,0 +1,18 @@
+[
+    {
+        "model": "sites.Site",
+        "pk": 1,
+        "fields": {
+            "name": "PhysioNet",
+            "domain": "alpha.physionet.org"
+        }
+    },
+    {
+        "model": "sites.Site",
+        "pk": 2,
+        "fields": {
+            "name": "PhysioNet-staging",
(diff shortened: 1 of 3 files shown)


elfeto:
I think there should be one also for dev.
Then we could change the function we use `get_current_site` for the example below.
```
from django.contrib.sites.models import Site
Site.objects.get_current()
<Site: dev.physionet.org>
```
The thing about Site is that is cached after the first call, so it doesn't do that many database calls.

bemoody:
We could.  I thought it would be somewhat inconvenient for testing, to
have localhost:8000 generate URLs pointing to dev.physionet.org.

I think it would be unwise to change the current code (which works
*without* sites) to make sites mandatory.  If you follow the logic, it
doesn't look like there's a real difference performance-wise.

elfeto:
Works for me.

PR #626, opened 2019-07-25
Cloud integration

elfeto:
The administrator under published project manage, can add several ways to download or use the data.
 - There is AWS for the open data platform, that just gives the URL for open data.
 - There is AWS bucket, in which if given a bucket, it will generate a preview command to download the data. 
     - (This is custom software our AWS open data) because function that adds the amazon ID, runs with a custom header to AWS.
- There is a GCP email that manages Big Query.
     - This has to be done with the google groups organizational account. NOT regular google groups.

I added to the users a model that would store their google email address (assuming they have added it to the email list) and their AWS ID.  

I also fixed a bug with the "bucket_policy_only_enabled", that was not being set.
I unified the place where the google credentials are stored.

PR #646, opened 2019-08-05
Missing hash for installation on python 3.5

bemoody:
Add a hash for the cffi binary package for python 3.5 on amd64 linux (Debian oldstable.)

(I haven't looked at the changes in pull #626 at all yet.  And yes, I should update my development system, but I'm doing one thing at a time. :))


elfeto:
For apple I get different hashes for the following:
 - cryptography
 - cffi
 - protobuf 
 - Pillow

You think should be useful to add to the file ?

bemoody:
> For apple I get different hashes for the following:
>  - cryptography
>  - cffi
>  - protobuf
>  - Pillow
>
> You think should be useful to add to the file ?

Yeah, by all means.

PR #647, opened 2019-08-05
Add the newly published 2019 challenge to the hardcoded challenge index

tompollard:
The 2019 PhysioNet Challenge is now listed at: https://alpha.physionet.org/content/2019-challenge/1.0.0/. This is a minor change to add the challenge to the hardcoded page at: https://alpha.physionet.org/about/challenge/

### DIFF
--- a/physionet-django/templates/about/challenge_content.html
+++ b/physionet-django/templates/about/challenge_content.html
@@ -27,7 +27,7 @@
     of challenges, inviting participants to tackle clinically interesting
     problems that are either unsolved or not well-solved.</p>
   <ul>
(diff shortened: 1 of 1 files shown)


elfeto:
Just checked and in between Benjamin's comment in #648 and the old website backwards compatibility this is a bad idea... is better to go in the backend and set the proper slug. 

Right now on you can access the challenges by this URL:http://physionet.org/challenge/2018 and if you pre-pend the word alpha to the hostname, you would get the same... http://alpha.physionet.org/challenge/2018 and both url's would resolve, but if we do it for the 2019 it fails.

tompollard:
I'll set the slug from the shell and will update this pull request.

PR #648, opened 2019-08-05
Increase max length of slug from 20 characters to 30 characters

tompollard:
This change increases the length of the slug to 30 characters. 20 is too short (e.g. "2019-physionet-challenge") would be a good name, but this is >20.


elfeto:
You are missing the slug for the PublishedProject model, its still max = 20

bemoody:
I do discourage the use of long database names, as they take more effort to type.  (And long record names for the same reason.)

There are also bugs in various programs (e.g. older versions ecgpuwave) but that's not much of an argument.

As for *this* project, of course, it should be challenge-2019 (yes, I realize there's also another bug at play there...)

tompollard:
Well spotted Felipe, now fixed! @bemoody I agree that we should try to keep slugs short, but I think we need a little more flexibility than 20 characters.

elfeto:
If you are going to pursue this, please redo the migrations in a single file. There is no need for two files.

tompollard:
Squashed changes - now single commit, with single migration.

bemoody:
Agreed.

We don't *need* to change the slug field of active/archived projects
here but it also shouldn't hurt anything.

However, I think we do need to update project.validators.validate_slug
function.  I would probably change it by adding a
'MAX_PROJECT_SLUG_LENGTH' (like I did for MAX_FILENAME_LENGTH) and
referring to that constant both in validate_slug and in the model
field definitions.

tompollard:
Thanks for the feedback. I've defined MAX_PROJECT_SLUG_LENGTH' in project.validators and referenced this in:

- project.models, to set the slug field length
- project.views, to avoid hardcoding the max length in the HTML template (also removed duplicate pattern check)
- console.forms, to validate the field

bemoody:
Looks good, thanks.

PR #650, opened 2019-08-06
Assorted email style fixes

bemoody:
- Word-wrap email messages

- Use 'https' URLs instead of 'http'

- Assorted punctuation and grammar fixes

I would probably have used "they" instead of "he/she" in the "reference contact" message templates.  I'm guessing, though, that "he/she" was used deliberately to emphasize that credentialing is for a single person, not a group.


tompollard:
Looks great, thanks.

PR #652, opened 2019-08-06
Fixed file link in wget and gsutil

elfeto:
Here I fix the location of where I get the files in the wget command and also i prepend the gs:// to the gsutil command.


bemoody:
Do *not* tell people to use 'wget -nd'.  Possibly --cut-dirs would be
appropriate.

*Do* tell people to use '-N' and '-c'.

'-P DESTINATION' is really, really ugly too.

Looks like this will vary by subdirectory in an inconsistent way.
That needs rethinking.

PR #654, opened 2019-08-06
Added hashes for OSX

elfeto:
When trying to install the requirement in OSX it complaints for the hashes because the OS. Here I add the pertinent OSX hashes.


tompollard:
Thanks, works for me!

PR #657, opened 2019-08-06
change all the alpha.physionet.org for physionet.org

elfeto:
This PR is to replace all the places where alpha.physionet.org appears in django.
In the mailto notification email, I just replace alpha for the output of "get_current_site(request)"
which should receive either alpha or physionet.org


tompollard:
Thanks Felipe

PR #660, opened 2019-08-06
Fix section header IDs, and other small things

bemoody:
This fixes the section header IDs in challenge pages (we also need to fix an internal link to #background in the challenge-2019 page.)

Also, add IDs to project previews (issue #588), show Usage Notes as required (it is required, it just wasn't displayed as such in database previews), and don't do weird word breaking in References.


tompollard:
Thanks Benjamin. I'll update the links in the 2019 challenge news items once this is pushed to production.

PR #665, opened 2019-08-07
Added a way to not store private information in tracebacks

elfeto:
Django has a neat way to handle protected information.

It can be done with either of the following:
 - sensitive_variables
 - sensitive_post_parameters

In this case I use post because the variables are in a post
request and I dont touch them.

The traceback will replace the specified fields for ***
E.g.
email =3D 'felipe.torres.cs@gmail.com'
username =3D 'ftorres'
first_names =3D 'Felipe'
last_name =3D 'F=C3=A1bregas'
password1 =3D '********************'
password2 =3D '********************'

### DIFF
--- a/physionet-django/user/views.py
+++ b/physionet-django/user/views.py
@@ -19,6 +19,7 @@ from django.urls import reverse
 from django.utils import timezone
 from django.utils.encoding import force_bytes, force_text
 from django.utils.http import urlsafe_base64_encode, urlsafe_base64_decode
+from django.views.decorators.debug import sensitive_post_parameters
 
 from user import forms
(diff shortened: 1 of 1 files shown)


elfeto:
Note!
For this to work, the DEBUG in the setting has to be False. 
(which is the case for production and staging)

bemoody:
Please get rid of ALL of the unrelated changes here.  There is no
reason for you to be touching user/forms.py here, or user/urls.py, or
adding and removing blank lines in unrelated functions.

There are only two lines here that have anything to do with your pull
request.  Please just change those two lines, and squash into a single
commit.

elfeto:
Done.

PR #670, opened 2019-08-07
Links and codes

lbulgarelli:
Links and code snippets should be set to `word-break: break-word` site-wide, as both can contain long paths that would result in undesirable scrolling in small window sizes. 

Breaking words in these cases to avoid this shouldn't be a problem.

### DIFF
--- a/physionet-django/project/static/project/css/project-content.css
+++ b/physionet-django/project/static/project/css/project-content.css
@@ -14,10 +14,6 @@ blockquote {
   background: #F0F0F0;
 }
 
-a{
-  word-break: break-word;
-}
-
 @media (min-width: 768px){
   .btn-right{
     float: right;
--- a/physionet-django/static/custom/css/physionet.css
+++ b/physionet-django/static/custom/css/physionet.css
@@ -525,3 +525,7 @@ table.files-panel {
   text-align: center;
   width: 1.25em;
 }
+
+a, code{
+  word-break: break-word;
+}


bemoody:
Not convinced this is wise.  Forcibly breaking words is *always* a
problem, even if sometimes it is the least bad thing you can do.

Better in general is to make sure things can be broken in appropriate
places (e.g. with soft hyphens or <wbr>)... and avoid stuffing things
into overly-narrow boxes to begin with (see: the current front page.)

There also seem to be a lot of different CSS properties that do
similar things, with varying degrees of formal standardization and
varying degrees of actual browser support.

Please point to some specific examples of the elements that you are
trying to fix here.

lbulgarelli:
- This only breaks words in `<a>` and `<code>` tags
- `break-word` only breaks a sequence of characters when it exceeds the size of the container, which is either very rare or nonexistent for words in the dictionary that would be used on PhysioNet content.

The combination of the two properties above will ensure that this never or almost never occurs beyond the scope we are trying to solve here, like:

- urls (e.g. `http://circ.ahajournals.org/content/101/23/e215.full`) with no alias
- code snippets with path/url (e.g. `wget -r -N -c -np http://alpha.physionet.org/files/gmse/1.0.0/`)

The boxes in the front page are wide enough for its content, and are also switched to full width at narrow windows.

bemoody:
At the moment, if you look at the front page in a window that is
roughly 1000 pixels wide, you'll see:

Computed
Tomography
Images for
Intracranial
Hemorrhage
Detection
and
Segmentatio
n

which is in this case due to an *inline style* rather than your pull
request here, but it's the same problem.

As for URLs, it'd be better to add word breaks at the slashes if
possible.  Where does
'http://circ.ahajournals.org/content/101/23/e215.full' appear on the
site?

PR #675, opened 2019-08-08
Fix featured ordering on postgresql

lbulgarelli:
PostgreSQL doesn't support casting boolean to integer and Django simply ignores this is a thing, which causes PR #668 to not work on staging/production.

This fixes it by saving the featured content in a variable, removing the field and then recreating it as integer and inserting the data.

### DIFF
--- a/physionet-django/project/migrations/0030_auto_20190807_1220.py
+++ b/physionet-django/project/migrations/0030_auto_20190807_1220.py
@@ -3,18 +3,19 @@
 from django.db import migrations, models
 from project.models import PublishedProject
 
-def featured(apps, schema_editor):
-    projects = PublishedProject.objects.filter()
+
+featured = {}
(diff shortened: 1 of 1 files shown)


physionet-django/project/migrations/0030_auto_20190807_1220.py
```diff
@@ -3,18 +3,19 @@
 from django.db import migrations, models
 from project.models import PublishedProject
 
-def featured(apps, schema_editor):
-    projects = PublishedProject.objects.filter()
+
+featured = {}
```
elfeto, line 7: What is the use of this "featured" ?

lbulgarelli: This is a dict that will save the projects currently flagged as featured and assign an integer which will be used to populate the field again after it is recreated.

PR #676, opened 2019-08-08
Fixed the filter of the migration for postgres

elfeto:
Setting a proper import in the migration.


bemoody:
WHOA WHOA WHOA

bemoody:
Please make an effort to explain what you are doing, and why, in your
comments and in your commit messages.

In this case, the migration logic is decidedly non-trivial.  Explain
what the existing boolean field means, why you are replacing it with
an integer field, and how you are assigning values to the new integer
field.

Modifying an existing migration script is a little dicey too.  Explain
in your commit messages why you are changing it.  What would happen if
you ran the previous version of the migration script?  What happens if
you now run this version?  Is the end result the same?  Explain all of
that in your commit message.

As we discovered in trying to test this on the staging server, it
would be wise to ensure that the migration is reversible (which I
think should be pretty easy in this case.)

I believe you also want to use 'apps.get_model' in *both* internal
functions here (not just the first one) and not to import
project.models.  Otherwise stuff will break in the future.

Overall I think you've made this more complicated than it's worth: I
would just have added a new field, populated it by hand, and dropped
the old field afterwards.

But this has been a good learning experience. :)

bemoody:
To be clear: that was directed at @lbulgarelli.

And also to be clear: this *is* a good change overall, so thank you
for making it.

lbulgarelli:
> Please make an effort to explain what you are doing, and why, in your
> comments and in your commit messages.

> In this case, the migration logic is decidedly non-trivial.  Explain
> what the existing boolean field means, why you are replacing it with
> an integer field, and how you are assigning values to the new integer
> field.

I added comments to the code to address this.

> Modifying an existing migration script is a little dicey too.  Explain
> in your commit messages why you are changing it.  What would happen if
> you ran the previous version of the migration script?  What happens if
> you now run this version?  Is the end result the same?  Explain all of
> that in your commit message.

Done.

> As we discovered in trying to test this on the staging server, it
> would be wise to ensure that the migration is reversible (which I
> think should be pretty easy in this case.)

Done.

> I believe you also want to use 'apps.get_model' in *both* internal
> functions here (not just the first one) and not to import
> project.models.  Otherwise stuff will break in the future.

Done.

> Overall I think you've made this more complicated than it's worth: I
> would just have added a new field, populated it by hand, and dropped
> the old field afterwards.

I agree. I was blindsided by the fact that Django does not deal with the whole boolean/int mess in Postgres. I then tried to keep the same outcome for the migration in order to avoid any undesirable effects, which resulted in something more complicated.

> But this has been a good learning experience. :)

Agree

elfeto:
Well... We need this migration to not have dev broken. I will merge to dev, and staging because both are broken... If there are no issues and/or concerns, then we'll merge this and the other changes to production....

tompollard:
Felipe, I assume that this fixes the problem in: https://github.com/MIT-LCP/physionet-build/pull/681. Shall I rebase the branch on dev?

elfeto:
I don't know if it fixes the problem in #681, I know it fixes the DB problems for Postgres vs sqlite for that specific migration. (was tested in staging, and its running on staging)

PR #678, opened 2019-08-09
renamed reverse url

elfeto:
The name for the URL 'project_access' was used already. It should have given errors before, but for some reason it worked fine. 

Here I rename the URL reverse.

### DIFF
--- a/physionet-django/console/templates/console/console_navbar.html
+++ b/physionet-django/console/templates/console/console_navbar.html
@@ -94,11 +94,11 @@
         </ul>
       </li>
 
-      {% url 'project_access' as project_access %}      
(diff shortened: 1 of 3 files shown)


bemoody:
What are the two URLs / views that are clashing?  Which links should
be pointing to which URLs?  Please explain that in your commit
message.

You changed an 'id' here - is there something that refers to that element by id?

elfeto:
The ID was for consistency. 
The reverse URL for:
 - /console/project-access/
 - /projects/<project_slug>/access/
have the same, they probably didn't clash because of the project_slug, but just to be safe I wanted to rename one of them.

PR #684, opened 2019-08-12
Added new changes by KP and fixing known references

elfeto:
More HTML changes by KP, and altering the logic of the known references in the HTML


physionet-django/console/templates/console/complete_application_display.html
```diff
@@ -1,82 +1,87 @@
...
-		{% endif %}
-	    {% if application.known_ref %} [Reference is known] {% endif %}
+	{% csrf_token %}
+	<button name="contact_reference" value="{{ application.id }}" type="submit" style="margin-right: 1em;">
+	  Contact Reference
+	</button> 
+	{{ application.reference_email }} 
+	{% if application.reference_contact_datetime %}
+		[The reference was contacted on {{ application.reference_contact_datetime }}]
+	{% else %}
```
tompollard: `{% elif application.known_ref %}` ?

physionet-django/console/templates/console/complete_application_display.html
```diff
@@ -1,82 +1,87 @@
...
-			<div class="alert alert-danger">
-		  	  <strong>{{ error|escape }}</strong>
-			</div> <!-- alert alert-danger -->
-		  {% endfor %}
+	{% csrf_token %}
+        <div class="form-group row">
+	  <label for="inputPassword">Reason for approval:</label>
+	  <div class="col-sm-6">
+	    {{ process_credential_form.responder_comments }}
+	  </div> <!-- col-sm-6 -->
```
tompollard, line 31: What is the purpose of `<!-- col-sm-6 -->`  comment? Can this be removed?

physionet-django/console/templates/console/complete_credential_applications.html
```diff
@@ -48,25 +61,24 @@
...
-                {% endif %}
-                <a href="#application_{{ application.id }}">{{ application.user.username }}</a>
-                </li>
-              {% endfor %}
-	          </ol>
+	         <ol>
             {% for application in applications %}
-              {% if application.reference_contact_datetime %}
-                {% include "console/complete_application_display.html" with color='green' %}
+              {% if not application.reference_contact_datetime %}
```
tompollard: Don't need to worry for this pull request, but in general I think it is good practice (more readable) to have the positive case first. e.g. 

{% if application.reference_contact_datetime %}
DO X
{% else %}
DO X
{% endif %}

tompollard:
I'm okay with this being merged, but I think it's got to the point where this page is going to cause problems for people reusing the platform (the hacks are affecting things they shouldn't, like the menu, header etc). Perhaps we could remove the item from the menu to avoid this issue?

physionet-django/console/templates/console/complete_application_display.html
elfeto: Its good to know when the divs end, at least for this code...

tompollard: yep, makes sense.

tompollard:
thanks!

PR #690, opened 2019-08-15
Show project admin task status and prevent conflicts

bemoody:
If I click the "Make Checksum File" button, I shouldn't be able
to immediately click "Make Zip" - depending on how things are
scheduled, or if the process gets interrupted, I might end up
with a zip file containing the old SHA256SUMS file, or half of
the new one, or no SHA256SUMS file at all.

Moreover, I'd like to see some visual indication of tasks that
have been scheduled.

This does all of that.

### DIFF
--- a/physionet-django/console/tasks.py
+++ b/physionet-django/console/tasks.py
@@ -1,0 +1,148 @@
+import inspect
+import json
+
+from background_task.models import Task
+
+
+_model_tasks = {}
+
+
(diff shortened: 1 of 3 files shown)


elfeto:
Looks nice

elfeto:
@bemoody  This lock is not enforced in any way in the DB right? This is just a "lock" that one sees in the browser which doesn't allows the user to click?

bemoody:
That's right.

PR #693, opened 2019-08-16
removed autoescape and linewrap

kpierceHST:
removed autoescape and linewrap


bemoody:
This doesn't work (it will raise a TemplateSyntaxError.)

You probably don't want to remove the "{% autoescape off %}".
("autoescape off" disables autoescape, so removing that would *enable*
autoescaping, not disable it.)  "Autoescaping" means that Django will
replace the characters < > & " ' with the corresponding HTML entities;
for example, if the applicant's research summary includes apostrophes,
those would be shown as "&#39;" in the email message, which is
probably not what you want.

If you want to disable the word wrapping, then in addition to removing
the "{% filter wordwrap:70 %}" you also need to remove the
corresponding "{% endfilter %}".

I would guess you probably want to make the same change in
mailto_contact_reference.html as well.

PR #694, opened 2019-08-16
Avoid recursively scanning project directories in console

bemoody:
There are still a few places that perform a complete recursive
directory scan as part of a page load.  We want to avoid such things,
especially if they enable external authors to break the admin console.

(Just for fun: I tried invoking 'get_tree_size' on one-tenth of
the MIMIC-III Waveform Database.  It took about 90 seconds.)

This fixes two of those issues:

- 'manage_published_project' wants to display the complete size of a
  (published) project.  Instead of calculating the size, just use the
  stored 'main_storage_size' (which is calculated by set_storage_info
  at the time the project is published.)

- 'submission_info' (and 'edit_submission' and 'copyedit_submission'
  and 'awaiting_authors' and 'publish_submission') wants to display
  the complete size of an active project.  For the time being, I've
  changed this to say "unknown".

(This is a kludge and I promise it will eventually be fixed!  But for
the time being, I don't think it's crucial to have disk usage
displayed there, and it's more important to allow us to review huge
projects without using even uglier kludges on the server.)

The issues that (I know) are still unfixed are:

* 'project_files' will still do a complete scan every time the page is
  loaded.

* UploadFilesForm will still do a complete scan every time one or more
  files are uploaded.

* NewProjectVersionForm will still copy over the directory structure
  synchronously.

I am ignoring those issues for now because... well, the system still
doesn't have any good mechanism that external contributors can use to
easily upload a project containing millions of files.  If it only
inconveniences me and Alistair, we can deal with that. :)

### DIFF
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -647,13 +647,20 @@ class UnpublishedProject(models.Model):
         """
         return os.path.join(self.__class__.FILE_ROOT, self.slug)
 
-    def get_storage_info(self):
+    def get_storage_info(self, force_calculate=True):
         """
         Return an object containing information about the project's
         storage usage.
+
+        If force_calculate is true, calculate the size by recursively
+        scanning the directory tree.  This is deprecated.
         """
+        if force_calculate:
+            used = self.storage_used()
+        else:
+            used = None
         return StorageInfo(allowance=self.core_project.storage_allowance,
-            used=self.storage_used(), include_remaining=True)
+                           used=used, include_remaining=True)
 
     def get_previous_slug(self):
         """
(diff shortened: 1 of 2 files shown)


bemoody:
Okay... a curious side note which I didn't notice at first.

copyedit_submission uses project.info_card to obtain storage_info, so
that will show "unknown" in the files form as well as in the "info
card".  (The function still uses UploadFilesForm, so it still enforces
quota on uploads - the editor just can't see how much space is used.)

(The fact that it's used in two places may also be the reason why
'storage_info' is listed twice in the same dictionary, at lines 336
and 346.)

tompollard:
@lbulgarelli if you have time, please could you take a look at this and pass on feedback to Benjamin?

tompollard:
Thanks Benjamin, the description with each commit made this very easy to review!

PR #696, opened 2019-08-20
Fix syntax of mailto template

bemoody:
Syntax of templates was broken in pull #692 and partially fixed in #695.  This fixes the syntax of the other template.


bemoody:
@kpierceHST Ken, this change ought to fix the error you were seeing.  If it looks OK to you, I'll merge it.

Issue #345, opened 2019-04-08
Allowing anonymous reviewers to see project previews


cx1111:
Some journals require the data to be viewed by their reviewers before they will accept a paper. Authors are sometimes willing to publish the data first on PN, but occasionally they do not want to before their paper is published.

Should we build in a feature allowing reviewers to see the project preview of submitted projects? If so, we can work out the details. Remember: the reviewers must be anonymous to the data authors.

Possible methods:
1. A long random URL is generated for authors to send to the journal to directly view the preview. Simple, but nothing is tracked.
2. A long random URL is generated for authors to send to the journal, which can be used by the reviewers to make their PN accounts reviewers of the project.
3. We can assign PN accounts to be reviewers to submitted projects using the admin console. More work for us, requiring us to interface more with the journal/author, etc. May be most 'official' method if we decide to partner with journals.

Thoughts? Other solutions?

bemoody:
"Long random URL" sounds easier from the POV of the guest reviewers.
Also it'd mean we don't have to worry about leaking the reviewer's
identity if we don't know who they are.

On the other hand, there is no way of controlling or even knowing who
has accessed that URL once it has been shared with the journal.  This
makes me slightly nervous.  It also definitely wouldn't work for
restricted-access projects.

It could also get weird when considering multiple revisions of the
project (does this long random URL point to the "current revision", or
to a snapshot of the project at one point in time?  can the URL only
be generated after the project has been "frozen"?)

Either way, users shouldn't have the ability to create a publicly
visible URL without the approval of PN editors.

Would the reviewers have strictly read-only access, or would we (for
example) want to allow them to make comments as well?

(I also touched on this briefly in issue #350)

cx1111:
Good point about the safety and access (for protected projects). It may be good for editors to control this.

Just as a reminder, we have the ActiveProject model, one of which is created for each version. We create a PublishedProject from it when it becomes published, and delete the original ActiveProject object. So access would be given to each version.

One possible solution is that when a project is under submission, an admin can assign PN accounts 'reviewer' status, allowing them to view the project preview. When the submitting author submits the project for review, we should have a box asking whether the authors are expecting a journal to contact us for preview rights. Then they can send us emails or use some form to request access.

tompollard:
For the reasons mentioned by @bemoody, I don't think we should provide authors with the ability to share unreviewed projects via an anonymous link (method 1).

[...]

PR #697, opened 2019-08-20
Allow anonymous access to projects

lbulgarelli:
- A new AnonymousAccess model was created
  - `passphrase` is a hash from a random string of length 20
and is set by the function `set_passphrase`, which returns the raw key.
Once the raw key has been set it cannot be recovered, as it is not stored.
- Functions to generate and check a passphrase was added to Metadata,
and the function `has_access` has been updated and tries to validate
a anonymous access if the argument passed is a string.
- A `reviewer` path was added to published and active projects
  - $URL/content/slug/version/reviewer for published projects
  - $URL/projects/slug/reviewer
- `project_preview` and `published_project` were altered to allow access using the anonymous interface. The access granted by a session with `max_age` of one hour
- New `Manage` page added in the author's project list


bemoody:
Conceptual problems:

This isn't something that should be available as an option until the
project has been accepted by the editor.  I'm not at all convinced
that the option should be available to *authors* at all.

It's rather sketchy to give a third party a URL to "review a project
for publication" when the content of that URL is subject to change.
This is less of an issue if the project has already been "accepted
for publication", and thus we can assume that the editors will not,
and the author cannot, make any major changes.  Even so, I rather feel
like if you are handing a URL to somebody to review, the content
should never change - otherwise, what was it that they reviewed?  If
it's not practical to make the URL point to a permanent snapshot, then
at least we should ensure that once the content is changed in any way,
the URL becomes non-functional.

The URL is also supposed to be *secret*.


Implementation problems:

"AnonymousAccess" is not a good name for this class.  The name of a
class should indicate what an object "is", and I don't know what an
"anonymous access" is.  The doc string doesn't tell me what the object
is, either.

A comment like "Model fields" is pretty useless.  I can see that these
are fields.  Tell me what they mean.

The function names are not good, either.  A function called
"is_valid_reviewer" should... I don't know, check whether a User has
appropriate affiliations or something.  Nothing about the name
"is_valid_reviewer" suggests that the function has anything to do with
passwords.  Or passphrases.  What's a passphrase?  How is it different
from a password?

I see "project.anonymous.first()" in a lot of places.  This seems like
a bad sign: either there is a possibility that there might be more
than one object (in which case, you've got to handle more than just
the first), or there isn't (in which case, you should be using an API
that returns "zero or one object", not "a list of objects".)

Messing with the API of 'has_access' in that way is rather sketchy.

Putting the *passphrase itself* into a cookie is dangerous, and
pointless if you're using a cryptographic signature anyway.  Or to put
it another way, having a token that expires is pointless if anyone who
has a copy of the expired token also has everything they need to
obtain a fresh one.

In 'published_manage.html', you have a bunch of inline styles (don't
do that) and you are using a *permanently-disabled text input field*
as a *warning message*.

I realize that the API of 'project_auth' is bad, but please don't make
it worse.  What is the difference between auth_mode=2 and auth_mode=3?
How did you decide which of the functions that were auth_mode=2 should
now be auth_mode=3, and which should remain auth_mode=2?

Your URL 'content/<project_slug>/<version>/reviewer/' (aside from not
being secret) clashes with the project file namespace.

lbulgarelli:
> This isn't something that should be available as an option until the
> project has been accepted by the editor.  I'm not at all convinced
> that the option should be available to *authors* at all.
> 
> It's rather sketchy to give a third party a URL to "review a project
> for publication" when the content of that URL is subject to change.
> This is less of an issue if the project has already been "accepted
> for publication", and thus we can assume that the editors will not,
> and the author cannot, make any major changes.  Even so, I rather feel
> like if you are handing a URL to somebody to review, the content
> should never change - otherwise, what was it that they reviewed?  If
> it's not practical to make the URL point to a permanent snapshot, then
> at least we should ensure that once the content is changed in any way,
> the URL becomes non-functional.
> 
> The URL is also supposed to be *secret*.

This is a different concept: Imagine I submitted an article to a journal, and I'm publishing the data I used on PhysioNet.

The editor/reviewers for the journal might want to access the data to reviewer your article, and as reviewers in peer-reviewed journals are anonymous, we should allow anonymous access to projects, if the author desires.

In this case, we should not be worrying about changes that might be made during a review that happens totally independent from PhysioNet.

As `reviewer` is a role we want to implement internally in the future, I've removed any reference to this role from this PR, as this is not the goal here and might be confusing.

> "AnonymousAccess" is not a good name for this class.  The name of a
> class should indicate what an object "is", and I don't know what an
> "anonymous access" is.  The doc string doesn't tell me what the object
> is, either.

As explained above, this is exactly what it is. It grants anonymous access (without user authentication). Regardless, I've updated the doc string to better describe this.

> A comment like "Model fields" is pretty useless.  I can see that these
> are fields.  Tell me what they mean.

> The function names are not good, either.  A function called
> "is_valid_reviewer" should... I don't know, check whether a User has
> appropriate affiliations or something.  Nothing about the name
> "is_valid_reviewer" suggests that the function has anything to do with
> passwords.  Or passphrases.  What's a passphrase?  How is it different
> from a password?

Sorry, I'm updating those.

> I see "project.anonymous.first()" in a lot of places.  This seems like
> a bad sign: either there is a possibility that there might be more
> than one object (in which case, you've got to handle more than just
> the first), or there isn't (in which case, you should be using an API
> that returns "zero or one object", not "a list of objects".)

`anonymous` is a `GenericRelation` to `AnonymousAccess`, and each project has only one 'AnonymousAccess`.

Accessing a `GenericRelation` always returns a `QuerySet` which is an iterable.

`first()` is the API that returns "zero or one object".

> Messing with the API of 'has_access' in that way is rather sketchy.

I've changed that.

> Putting the *passphrase itself* into a cookie is dangerous, and
> pointless if you're using a cryptographic signature anyway.  Or to put
> it another way, having a token that expires is pointless if anyone who
> has a copy of the expired token also has everything they need to
> obtain a fresh one.

I agree, I meant to change this but forgot, I've updated the way we handle the cookie. 

> In 'published_manage.html', you have a bunch of inline styles (don't
> do that) and you are using a *permanently-disabled text input field*
> as a *warning message*.

I got rid of most of them.

> I realize that the API of 'project_auth' is bad, but please don't make
> it worse.  What is the difference between auth_mode=2 and auth_mode=3?
> How did you decide which of the functions that were auth_mode=2 should
> now be auth_mode=3, and which should remain auth_mode=2?

`auth_mode=2` is for pages that can be accessed by authors and admins
`auth_mode=3` is for pages that can be accessed by authors, admins and those who has been authenticated by using a passphrase

I actually made a mistake here, as I left the admins out of `auth_mode=3`, but already fixed it.

I changed the functions from `auth_mode=2` to `auth_mode=3` based on what a reviewer from a peer-reviewed journal needs, which is the files and project description.

> Your URL 'content/<project_slug>/<version>/reviewer/' (aside from not
> being secret) clashes with the project file namespace.

I don't think the URL needs to necessarily be hidden, although it might be a nice to implement it in the future. I've updated the url to not clash with the file namespace.

bemoody:
> This is a different concept: Imagine I submitted an article to
> a journal, and I'm publishing the data I used on PhysioNet.

You are *hoping* and *intending* to publish the data you used on
PhysioNet.  Before you have gone through the process, you have not
actually done so.  And if you don't complete the process, or your
project is not of sufficient quality, you won't.  Keep that in mind.

> The editor/reviewers for the journal might want to access the
> data to reviewer your article, and as reviewers in
> peer-reviewed journals are anonymous, we should allow anonymous
> access to projects, if the author desires.

*After* the project has passed at least a preliminary legal and
technical review (isn't full of PHI, isn't full of copyright
violations, isn't full of malware, etc.), possibly.

Before, no.  That's not something I'm willing to support.

> In this case, we should not be worrying about changes that might be
> made during a review that happens totally independent from
> PhysioNet.

I think it's entirely reasonable for us to think about the reviewer's
perspective and interests here, and not just "do whatever the author
thinks they want, and their interactions with reviewers are somebody
else's problem."

To a naive outside observer, simply seeing a page, hosted on
physionet.org and looking vaguely legitimate, gives the impression
that the page is in some way endorsed or supported by PhysioNet.
Someone who has a passing familiarity with us, and is presented with a
page that is described as "this is my project that I'm going to
publish on PhysioNet", could be forgiven for thinking that the page in
question is actually in the process of being published and *will* soon
be published in something very close to its current form.

I mean, if you just want to throw up some files at a fly-by-night URL
somewhere and share them with a reviewer, do that.  If you're hosting
them on PhysioNet, you're (intentionally or not) giving the impression
that they are, or will eventually become, part of a large, respected
public data repository.

If I were asked to review a database (or a paper based on a database;
either way), and while I was doing so, it was silently modified in the
background, I would be *furious*.  And whether or not "but we haven't
bothered to build a version control system" is a valid reason, it's
not an excuse.  And it makes *us* look bad, not the author (who didn't
read the fine print and didn't realize that it was a potential issue
in the first place.)

Snapshots would of course be preferable, but a simple "this page is no
longer available" would be far, far superior to just changing the
content out from under them.  This (in addition to security, and to
allow for the future possibility of keeping snapshots) is another
reason why I think the "reviewer login" URL should include a unique
random token.

Setting aside invisible pre-submission modifications, there is also
the fact that the project will likely undergo some editing before it's
finally published.  As a reviewer, if I reviewed a paper based on the
promise that the accompanying data would eventually be published, and
then a couple months later the database was published and it looked
nothing at all like what I'd reviewed... well, I wouldn't be quite so
ticked off as in the previous case, but I'd still be a bit annoyed.

(Again, this doesn't mean we shouldn't do it, but it's another
argument in favor of "only with the editor's approval.")

And I really do not want to hear from somebody submitting a database,
"but I can't change my proprietary file formats / cryptic labels /
craptastic directory structure because *that's what my paper says* and
I can't change my paper now."


> I agree, I meant to change this but forgot, I've updated the way we handle
> the cookie.

Unfortunately, your new version is broken too - it's not tied to a
project at all.


> `auth_mode=2` is for pages that can be accessed by authors and admins
> `auth_mode=3` is for pages that can be accessed by authors, admins and those
> who has been authenticated by using a passphrase
>
> I actually made a mistake here, as I left the admins out of `auth_mode=3`,
> but already fixed it.
>
> I changed the functions from `auth_mode=2` to `auth_mode=3` based on what a
> reviewer from a peer-reviewed journal needs, which is the files and project
> description.

Agh, I didn't notice that project_files_panel was screwed up to begin
with.  That ought to have the same permissions as project_files,
except that (I'm guessing, not going to bother to check now) the same
function/URL is *also* used from the console page.

Anyway, the other two functions (lightwave_project_home and
lightwave_project_server) should be accessible to reviewers as well.

So out of eight functions that previously used auth_mode=2, that makes
seven that should be using the new behavior and one that's broken in
its own special way.

I'd probably just change the behavior of auth_mode=2 and fix the one
broken case separately.  But it's okay either way, so long as you
explain your reasoning.

lbulgarelli:
I have changed the feature to only allow anonymous access when a project was already accepted by the editor.

As to changes that might still happen before, and if, the project is published, I think you are right. However, it is also inconvenient to reviewers to try to access a link that they have been provided only to realize it is suddenly not valid anymore, as well as the author having to regenerate it every time he makes a change. That said, the fact that the link could be possibly regenerated means invalidating the url would only be effective to the point that the author would be force to communicate with the reviewers.

Here's an idea: I added an alert box at the top of the page with the following content:

> This is a preview of a resource that is undergoing our editorial process and is subject to change.
> 
> It is not guaranteed that this resource will be published, or that the content will be the same at time of publication.
> 
> Last modified: `modified_datetime`

In addition, we also inform the author of the consequences of utilizing the anonymous access feature:

> Note: When providing anonymous access to third parties, keep in mind that any changes made during the PhysioNet review process will impact anyone viewing your project through that access.
> 
> Warning: generating a new passphrase will revoke the existing access

Of course, snapshots and versioning are still better, but those are separate and laborious tasks.

Sorry about the cookie stuff, mistakes were made, rs.

I've opted to keep `auth_mode=3`, the main reason is that we might want to implement something that we might not want to be accessible to reviewers. But also, we will (or should) probably refactor the access control once we implement roles into the platform.

alistairewj:
> I have changed the feature to only allow anonymous access when a project was already accepted by the editor.
> 
> As to changes that might still happen before, and if, the project is published, I think you are right. However, it is also inconvenient to reviewers to try to access a link that they have been provided only to realize it is suddenly not valid anymore, as well as the author having to regenerate it every time he makes a change. That said, the fact that the link could be possibly regenerated means invalidating the url would only be effective to the point that the author would be force to communicate with the reviewers.

As the main use case is reviewing data for a submitted paper, I can easily imagine the anonymous reviewer suggesting changes to the draft paper which require changes on the PhysioNet page. How will we allow this behavior? Correct me if I'm wrong, but there are two options:

1. The PhysioNet project is published (and now publicly accessible), then the author makes a new version.
2. The editor makes the changes on behalf of the author during copy-edit (eek!).

Obviously the first is preferred of the two but it still seems like an awkward workflow.

tompollard:
Just catching up on this now. Will take a look today. An outline of the functionality we'd aimed to build is at: https://github.com/MIT-LCP/physionet-build/issues/345#issuecomment-517813271. We definitely don't want to allow the author to generate or share these links, and it shouldn't be necessary for the project to have been accepted.

lbulgarelli:
I've addressed your concerns in the latest commit.

tompollard:
@lbulgarelli thanks, this looks excellent :) I'm in the middle of doing some testing - is there any chance that you could take a look at the points below in the meantime?

- set the links to automatically expire after a fixed period (perhaps 60 days)? i think this is especially important for the post-publication links, because it would be easy to forget about them.
- add functionality to allow the editor to manually expire the link?

Let us know if you don't have time in the next day or two and we'll work something else out!

tompollard:
Great work @lbulgarelli. @bemoody is there anything in the implementation that you would addressed, specifically relating to security? If not, I think we should go ahead and merge.

There are a few minor things that we might want to clean up later, but these can be dealt with in a separate issue. e.g.:

- Javascript popups appearing on login page (see screenshot below).

Screen Shot 2019-08-29 at 11 06 29

- Removing hardcoded "PhysioNet" where it appears ("access to a PhysioNet resource...").
- "Generate passphrase" is clear, but "Generate access" not so much. Instead of separate "Generate access" and "Remove passphrase" buttons, perhaps just one button that flips between "Grant access" and "Revoke access" might be good?

Screen Shot 2019-08-29 at 11 08 42

tompollard:
Just merged https://github.com/MIT-LCP/physionet-build/pull/713, so you'll need to remove the following chunk in the conflict:

```
=======
    if not request.user.is_authenticated:
        pass
        # return redirect_to_login(request.get_full_path())
    raise PermissionDenied()
>>>>>>> reviewer login
```

tompollard:
Thanks again Lucas, nice work.

tompollard:
@bemoody I'm going to go ahead and merge this. I hope you're okay with that. If there is anything that you're uncomfortable with then let me know and we'll address in a new pull request.

PR #681, opened 2019-08-09
MK2: Switch Shippable to run tests on Postgres (ref #677) #680

tompollard:
This is a first attempt at switching Shippable to run tests on Postgres instead of SQLite (ref #677). I'm expecting this to fail on first attempt, but having the pull request open will help me to troubleshoot. Please don't merge for now!

PR #699, opened 2019-08-20
Allow multiple background tasks to run in parallel

bemoody:
This is desirable since some tasks can take a very long time to run.  Note that currently, background tasks are only ever triggered by administrator actions, but that will probably change.

By default, the daemon will limit itself to running as many tasks in parallel as there are CPUs.

This should not be merged until after pull #690.

### DIFF
--- a/physionet-django/console/test_views.py
+++ b/physionet-django/console/test_views.py
@@ -5,7 +5,6 @@ import pdb
 
 
 import requests_mock
-from background_task.tasks import tasks
 from django.contrib.sites.models import Site
 from django.test import TestCase
 from django.test.utils import get_runner
@@ -433,7 +432,7 @@ class TestState(TestMixin):
             data={'slug': custom_slug, 'doi': False, 'make_zip': 1, 'georestricted': False})
 
         # Run background tasks
-        self.assertTrue(bool(tasks.run_next_task()))
+        self.assertBackgroundTasks(1)
 
         self.assertTrue(bool(PublishedProject.objects.filter(slug=custom_slug)))
         self.assertFalse(bool(PublishedProject.objects.filter(slug=project_slug)))
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -238,7 +238,9 @@ Q_CLUSTER = {
     'label': 'Django Q2',
 }
 
-# Django background tasks max attempts
+# Background tasks
+
+BACKGROUND_TASK_RUN_ASYNC = True
 MAX_ATTEMPTS = 5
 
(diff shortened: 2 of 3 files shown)


bemoody:
Clearly this breaks 'manage.py test' in some way.  Not sure why but I also don't know what exactly 'manage.py test' does.  Preferably we *should* try to run tests in a parallel configuration (if there are any cases where two tasks might be spawned at once)... and also we should run tests against a postgres server instead of sqlite.

I'm leaving this open but it shouldn't be merged until it's fixed :)

tompollard:
This is the same error that I'm getting in https://github.com/MIT-LCP/physionet-build/pull/681. Was keeping my fingers crossed that you'd come up with the fix! 

```
self.assertEqual(response.status_code, 200)
AssertionError: 404 != 200
```

elfeto:
The problem is a lock on the DB, but I cant seem to find the problem just after trying to execute the background task.

lbulgarelli:
This is happening because django runs each test in a transaction and setting the background tasks to be asynchronous is causing `test_publish` to trigger the task which will then be executed outside that transaction.

I'm guessing this will happen despite the change in databases for tests.

We can correct this easily if we change the test to execute that task synchronously, by changing:

```python
self.assertTrue(bool(tasks.run_next_task()))
```

to

```python
task = tasks._tasks['project.models.move_files_as_readonly']
self.assertTrue(bool(task.now))
```

bemoody:
Hey, maybe we should consider fixing this issue before doing lots of uploads to Amazon (pull #2086).

bemoody:
Updated, and I hope this should work now.

Very belated thanks to @lbulgarelli for debugging.

> This is happening because django runs each test in a transaction and setting the background tasks to be asynchronous is causing test_publish to trigger the task which will then be executed outside that transaction.

I still don't fully understand how Django handles the database in the test framework, but I'm sure you're right.

> ```
> task = tasks._tasks['project.models.move_files_as_readonly']
> self.assertTrue(bool(task.now))
> ```

I'm not sure what exactly you meant here (maybe you meant to write `bool(task.now())`, but that alone doesn't work.)  It should indeed be possible to invoke the task function "now", but to do that we would need to know what function to invoke with what arguments.

What I want for test_publish and other test cases is to simulate the behavior of background_tasks (taking Task objects from the queue and executing them the same way process_tasks would).

So here I've opted for a simpler kludge (use run_next_task, using override_settings to keep the original single-thread behavior.)

Meanwhile, I also want test cases to fail (and report a useful traceback) if the task function raises an exception.  So the new method assertBackgroundTasks does both those things.

tompollard (approved):
thanks, looks good to me!

PR #700, opened 2019-08-20
Bottom spacing in the content list

lbulgarelli:
The content list in both home and find pages is without any spacing between the short description and the project tags, as can be easily noticed either on staging or production.

This is not clear when you are on dev or in some projects because they already contain `<p>` tags in the short description.

Here I wrap it in a `<div>` with bottom margin to fix the issue.

### DIFF
--- a/physionet-django/search/templates/search/content_list.html
+++ b/physionet-django/search/templates/search/content_list.html
@@ -13,11 +13,13 @@
           {% endfor %}
       </p>
     {% endif %}
-    {% if not published_project.short_description %}
-      {{ published_project.abstract|safe|truncatechars_html:250 }}
-    {% else %}
-      {{ published_project.short_description }}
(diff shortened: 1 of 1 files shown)


elfeto:
The problem here is not the spacing, but an inconsistency. When the project has the short description it's just dumped there wit no <p> tags. But when it doesn't that its truncated, its wrapped under the <p> tags. 
If we just wrap the short description in <p> tags, then this is fixed.

lbulgarelli:
The truncated text is also not wrapped with a tag, it happens that the abstract that is truncated sometimes contains tags.

In any case, wrapping the content with a tag is what I did, spacing is just the visual effect of this problem. I used a `<div>` because the short description might  contain `<p>` tags, which cannot be nested.

elfeto:
Makes sense :)

PR #701, opened 2019-08-20
Better styling for static pages

lbulgarelli:
This is an attempt to improve the template of the `Share` and `About` pages.

Instead of using a bunch of `<br>` and `<hr>` tags, I wrap subsections in an indented box with a left border.


tompollard:
Thanks Lucas, this is an improvement.

PR #704, opened 2019-08-22
turn off autoescape in copy of dua used in e-mail to applicant

kpierceHST:
turn off autoescape in copy of dua used in e-mail to applicant

### DIFF
--- a/physionet-django/notification/templates/notification/email/mailto_contact_applicant.html
+++ b/physionet-django/notification/templates/notification/email/mailto_contact_applicant.html
@@ -1,4 +1,4 @@
-----------------
+{% autoescape off %}----------------
 {{ dua }}
 Application Date: {{ application.application_datetime|date }}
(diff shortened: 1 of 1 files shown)


lbulgarelli:
Since the only field that uses html is the DUA, it is more appropriate and safer to replace `{{ dua }}` with `{{ dua|safe }}` instead of turning autoescape off in the entire message.

bemoody:
Lucas, you're right that if you're substituting HTML strings into an
HTML template, it's better to use 'safe' instead of turning off
autoescape.

In this case, though, neither the message nor the DUA is actually
HTML: the message is plain text which is eventually converted to a
mailto link, and the DUA is plain text generated by html2text (see
dua_text_content).

It's strange that the email template files have a .html suffix; as far
as I know, all of them are plain text.  I thought about fixing that,
but it's a minor issue, all things considered.

PR #706, opened 2019-08-26
Reference is not a required field in the credentialing form

tompollard:
KP asked for the credentialing form to be updated so that reference is only a required field for postdocs and students (for others, reference is not required). This pull request implements that change.


tompollard:
Updated the pull request:

- fixed issue with the default setting for models.PositiveSmallIntegerField
- updated the template to display a message if the reference is not provided.

PR #708, opened 2019-08-27
Changes to the credentialing form requested by KP

tompollard:
We met today to discuss the credentialing form and agreed on a number of changes. The changes are summarized below and implemented in this pull request:

## Introductory text

1. Update the wording to clarify that a reference is not always required.

## Training course

2. Change the title to "CITI Completion report"
3. Remove the date and name fields. These are not required because the CITI course is currently the only course that we accept.

## Course Survey

4. Remove the section. It is not useful for us, and is not usually completed.

## Research area

5. Change the text to "Research topic", to encourage more concise responses.
6. Reduce the size of the text box.
7. Remove "Project of interest", which was causing some confusion. People are typically interested in multiple databases.

## Notes

- As a result of these changes, there will be missing data in the submitted forms that are reviewed by Ken (no error is raised). I will remove these fields from the templates in a separate pull request.
- In the personal information, KP would like "state" to be required if the country is USA or Canada. This is not implemented here.


bemoody:
If a reference is provided, should the applicant be required to fill
in all four fields (reference category / name / email / job title or
position)?

tompollard:
> If a reference is provided, should the applicant be required to fill in all four fields (reference category / name / email / job title or position)?

Yes, I think so, though this isn't addressed here. This comment seems more related to: https://github.com/MIT-LCP/physionet-build/pull/706/

tompollard:
Rebased on dev and added the merged migration

bemoody:
LGTM.  See also issue #709 about the optional fields.

tompollard:
Thanks! I'll try to deal with the points in https://github.com/MIT-LCP/physionet-build/issues/709 shortly.

PR #711, opened 2019-08-27
Update templates to reflect changes in #708

tompollard:
In https://github.com/MIT-LCP/physionet-build/pull/708 we updated the form for applying for credentialed access to databases. Several fields/sections were removed, including:

- Training report name and completion date
- Course survey (whether the applicant was applying for access as part of a course)
- Project of interest in the "research" section

This pull request updates the templates for completed submissions, so that these fields are not displayed.

### DIFF
--- a/physionet-django/console/templates/console/application_display_table.html
+++ b/physionet-django/console/templates/console/application_display_table.html
@@ -1,18 +1,18 @@
 {# Status section #}
 
 <p><strong>Status:</strong><br>
-Application Date: {{ application.application_datetime|date }}<br>
+Application date: {{ application.application_datetime|date }}<br>
 
 {% if application.reference_contact_datetime %}
-  Reference Contact Date: {{ application.reference_contact_datetime|date }}<br>
+  Reference contact date: {{ application.reference_contact_datetime|date }}<br>
   {% if application.reference_response_datetime %}
-    Reference Response Date: {{ application.reference_response_datetime|date }}<br>
-    Reference Verified: {{ application.get_reference_response_display }}<br>
+    Reference response date: {{ application.reference_response_datetime|date }}<br>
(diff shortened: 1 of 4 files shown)


tompollard:
Oh, and there were some inconsistencies in the case of section headers (Sentence case vs title case). I made a weak (sorry) attempt at trying to make the text more consistent. We should agree on a style and then fix this throughout the site.

bemoody:
I think that, to the extent that it's practical, we should try to
avoid having fields labelled differently, or shown in a different
order, in the management and history pages, compared to the actual
application form.

I do, however, think it's beneficial that the management pages are
consistent.  I wouldn't want to see older applications displayed with
different labels just because that's how the form was labelled at the
time it was submitted.

So it's a little weird.  I think, on balance, that if you remove
fields from the form, then those fields should still be shown for
*historical* applications where the field was not blank.  But maybe
they should be moved to the bottom in that case.

Looking at past applications:

- course_info was largely used incorrectly: people would put "THESIS",
or "Data mining", or "Data or Specimens Only Research".  But a few
people do appear to have used it correctly, listing the name/number of
a course they were taking.

- project_of_interest was supplied for every application submitted
since June 22.  Was that field previously required?  (It's not easy to
tell.)

tompollard:
> I think that, to the extent that it's practical, we should try to
avoid having fields labelled differently, or shown in a different
order, in the management and history pages, compared to the actual
application form.

Yep, I agree. If it's good with you, I'll address this once the text on the credentialing form itself is settled.

> I think, on balance, that if you remove
fields from the form, then those fields should still be shown for
*historical* applications where the field was not blank. 

I agree this would be nice, but I also would prefer to avoid logic in the history forms that apply only to a relatively small number of applications.

tompollard:
Thanks Benjamin

PR #713, opened 2019-08-28
Better error message for unauthorized file preview

bemoody:
If you visit a published file preview (e.g. /content/demoeicu/2.0.0/SHA256SUMS.txt), currently:

- if you are not logged in, it bounces you to a login page

- after logging in, if you are not authorized to view the project, it displays a generic error message

Instead of this, display an error message similar to what is shown in the files panel of the main project page, explaining that you need to be credentialed and sign the DUA.


tompollard:
Thanks, looks good to me!

Issue #688, opened 2019-08-13
Provide suggested citation(s) in BibTeX format


bemoody:
The recommended citation for a published project is currently shown as, e.g.:

> Pollard, T., Johnson, A., Raffa, J., Celi, L. A., Badawi, O., Mark, R. (2019). eICU Collaborative Research Database. PhysioNet. doi:10.13026/C2WM1R

This citation is missing an essential piece of information: the version number.

(This citation looks similar, but not identical, to the APA bibliography style for a journal article.  In APA style, the title should be lowercase except for the abbreviation - "eICU collaborative research database" - and the title of the journal should be italicized.)

(Note also that the three citations on that page are written in three completely different styles.  The first is "Authors (Year). Title. Publication.", the second is "Title. Authors. Publication (Year).", and the third is "Authors. Title (Year). Publication.")

Relatedly, I think the citation information ought to be presented in BibTeX format (and perhaps other formats) - see issue #406.  But since the PhysioNet citation doesn't fit any of the standard BibTeX entry types, this raises the question of *how* it should be presented.

Although people can edit the information if they want to (or even write their own BibTeX style), the hope would be that they can copy and paste it as-is into their .bib file, and the result should look reasonable, regardless of what style they're using.

The citation shown on the web page also doesn't *need* to match any particular BibTeX style, but it might be nice if it did.

I tried a few different ways of formatting the entry; see https://archive.physionet.org/users/shared/benjamin/bib/ for the results.

tompollard:
The core of this issue (adding version number to suggested citation) was fixed in https://github.com/MIT-LCP/physionet-build/pull/861. 

The secondary issue of allowing users to download the suggested citation isn't yet tracked (though https://github.com/MIT-LCP/physionet-build/issues/406 is related), so I'm leaving this open as a reminder to open a new issue on this topic.

PR #717, opened 2019-09-03
add published projects to public profile

tompollard:
The public profile page for physionet users (e.g. http://localhost:8000/users/rgmark/) is fairly sparse at the moment. This is a minor change to display published projects on the page. Example screenshot below:

Screen Shot 2019-09-03 at 13 50 54

I would like to improve the way that citation details are displayed across the website, and will do this in a separate pull request.

### DIFF
--- a/physionet-django/user/templates/user/public_profile.html
+++ b/physionet-django/user/templates/user/public_profile.html
@@ -84,7 +84,7 @@ Profile for {{ public_user.username }}
         <div class="row mb-1">
           <ul>
             {% for project in projects %}
-            <li><a href="{% url 'published_project' project.slug project.version %}">{{ project.citation_text }}</a>
+            <li><a href="{% url 'published_project' project.slug project.version %}">{{ project.citation_text }}</a> (v{{ project.version }})
             {% endfor %}
           </ul>
         </div>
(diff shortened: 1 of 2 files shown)


bemoody:
Looks good.  In public_profile(), could you do:

    PublishedProject.objects.filter(authors__user=public_user, ...)

instead of

    PublishedProject.objects.filter(authors__user__username__iexact=username,
...)

or are there corner cases I'm missing?

Also, it will become more complicated as we move into a future where
projects have multiple versions, and the authors for version 1.0 might
not be the same as version 2.0.  If there are a huge number of
versions, then that becomes a problem as well, but for now it might be
better to err on the side of displaying all published versions.

tompollard:
Thanks for the helpful comments Benjamin. 

> PublishedProject.objects.filter(authors__user=public_user, ...)

Yep, not sure why I didn't!

> for now it might be better to err on the side of displaying all published versions.

Good point, I'm now displaying all versions. The version is displayed with `(v{{ project.version }})` at the end of the citation text, as shown below. 

Screen Shot 2019-09-04 at 11 07 10

bemoody:
Great!  I really think that ultimately, the version number needs to be
part of the citation text (see issue #688) but this is good for now.

PR #718, opened 2019-09-03
changed reference_email comparison to case-insensitive

kpierceHST:
following Tom's suggestions, I added an “iexact” keyword to comparisons that check to see whether reference's e-mail address is on a list of known reference e-mail addresses.

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -777,11 +777,11 @@ def complete_credential_applications(request):
     for application in applications:
         application.mailto = notification.mailto_process_credential_complete(
             request, application, comments=False)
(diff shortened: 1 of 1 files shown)


bemoody:
Oh boy.  Comparing email addresses is an interesting business.  Is
benjamin.moody@gmail.com the same as benjaminMOODY@googlemail.com?

But yeah, this is probably a good idea.

tompollard:
Thanks @bemoody, I'll take this as you being happy to merge!

bemoody:
I was just about to do that, but thanks :)

tompollard:
Sorry to jump the gun!

PR #719, opened 2019-09-03
Update wording in credentialing form based on feedback from KP

tompollard:
These changes fix some of the wording in the credentialing form at:  http://localhost:8000/credential-application/ to respond to feedback from KP. Some notes on the changes are below:

### Text in the help popups (changed from "new" to "preferred") 

- **new**: "Your first and middle names." ->  **preferred**: "First (given) name"

- **new**: "Your last name" -> **preferred**: "Last (family) name"

- **new**: "Additional information about your name, for example 'Jr' or 'Sr'. Not a number" -> **preferred**: "Please leave the suffix blank if your name does not include a suffix like "Jr." or "III". Do not list degrees. Do not put a prefix like "Mr" or "Ms". Do not put "not applicable". Especially, do not leave a string of digits from browser autofill."

- **new**: "The name of your organization. Put 'None' if you are an independent researcher." -> **preferred**: "Your employer or primary affiliation. Put 'None' if you are an independent researcher."

- **new**: "The title of your job/role." -> **preferred**: "You job title or position (e.g., student) within your institution or organization"

- **new**: "The title of your reference. e.g. Professor, Dr." -> **preferred**: "Your reference's professional title or position" (I'd like to elicit a bit of information about what the reference does, rather than merely that s/he is Dr. or Ph.D.)

### Formatting/style

- please surround the two sentences below with ```<em>...</em>```:

```
"Submissions that are clearly incomplete, incorrect, or frivolous may be discarded without notice."
"Do not list yourself as reference." (this one appears in two places)
```

- please format the following sentences as indicated:

```
"<em>If you agree to all of these terms and conditions, access to restricted information within PhysioNet clinical databases may be granted to you <strong>as an individual.</strong> Your colleagues may obtain access to these data as individuals via the same procedure you are following.<em>"
```

- "Reference Topic" should be "Research Topic"


bemoody:
Several of these help strings use single quotes; my American instincts
are telling me they should be double quotes instead. :)

I don't know for sure whether the "suffix" auto-fill issues are
completely fixed (because I never figured out how to reproduce the
problem to begin with).  But strings of digits are rejected by the
form, so I don't think it's necessary to specifically tell people not
to enter a string of digits.

In "research_summary", "Brief description on your research" should be
"Brief description of your research".  Or maybe "your intended
research" or "your proposed research"?

The help text for "training_completion_report" is unnecessary; it
contains exactly the same text that is shown above the field:

> Instructions for taking the course are provided here. Upload the
> completion report from the CITI 'Data or Specimens Only Research'
> training program (PDF or image file). The completion report lists
> all modules completed, with dates and scores. Do NOT upload the
> completion certificate. If you would like to submit multiple pages,
> please combine them into a single pdf file.

I would change this:

> Instructions for taking the course are provided here. Upload the
> completion report (PDF file) from the CITI "Data or Specimens Only
> Research" training program. The completion report lists all modules
> completed, with dates and scores. Do <em>not</em> upload the
> completion certificate.

(Side note: it is severely irritating that CITI's website uses exactly
the same filename - "citiCompletionReportXXXXXXX.pdf" - for both the
"completion report" and "certificate" files.  Any chance we could get
them to fix this?)

In descriptive_inline_form_snippet.html, the label is forced to
titlecase, which is why some of the labels are displayed particularly
weirdly ("My First (Given) Name(S)").  This should be fixed.

I also dislike the "My" in "My First (Given) Name(S)".

I also think the help string for "first_names" should probably retain
"first and middle".

tompollard:
Thanks Benjamin. I've implemented the changes that you requested above. Some notes below:

> Several of these help strings use single quotes; my American instincts
are telling me they should be double quotes instead. :)

Fixed, with the help of @lbulgarelli !

> strings of digits are rejected by the form, so I don't think it's necessary to specifically tell people not to enter a string of digits.

Good point. Updated, to remove the text about digits.

> In "research_summary", "Brief description on your research" should be "Brief description of your research".  Or maybe "your intended research" or "your proposed research"?

Well spotted. Changed to "your proposed research" as suggested.

> The help text for "training_completion_report" is unnecessary; it contains exactly the same text that is shown above the field

I have kept the help text, but made it a little more succinct, so that it complements the text displayed above the field.

> I would change this: Instructions for taking the course are provided here...

I agree that your version is clearer. Now changed.

> (Side note: it is severely irritating that CITI's website uses exactly the same filename - "citiCompletionReportXXXXXXX.pdf" - for both the "completion report" and "certificate" files.  Any chance we could get them to fix this?)

I'll try dropping someone an email to see if we can get this changed.

> In descriptive_inline_form_snippet.html, the label is forced to titlecase, which is why some of the labels are displayed particularly weirdly ("My First (Given) Name(S)").  This should be fixed.

Well spotted, I was wondering why this was happening. Now fixed! I like the look of sentence case, and my preference would be to use this across the site for titles.

> I also dislike the "My" in "My First (Given) Name(S)".

"My" has been removed.

> I also think the help string for "first_names" should probably retain "first and middle".

The help text now says "first and middle" names.

bemoody:
Thanks, looks great.

One small thing about 02e1b5997e1b0a7317c0fed119eac26e24b32567:
{{foo|escape|safe}} is the same as just {{foo|escape}}, which (if
auto-escape is enabled) is the same as {{foo}}.  The help-text
shouldn't have been marked as "safe" to begin with.

tompollard:
Thanks @bemoody, `{{foo|escape|safe}} ` is now `{{foo}}`

PR #721, opened 2019-09-03
Display help text for credentialing form on mouseover, instead of on click

tompollard:
Currently, viewing the help text on the credentialing form at http://localhost:8000/credential-application/ requires the user to click on the "?" icon. As requested by KP, this update instead displays the text on mouseover. I think this is an improvement and should perhaps be implemented site wide.

Screen Shot 2019-09-03 at 16 33 47


bemoody:
Looks good to me, but this now hits a merge conflict with pull #719.

tompollard:
Thanks, I have rebased on dev and resolved the conflict.

PR #722, opened 2019-09-03
Fix error in the js control for the credentialing form

tompollard:
Currently, the credentialing form at http://localhost:8000/credential-application/ raises the following js error:

```
credential-form-control.js:6 Uncaught TypeError: Cannot read property 'value' of null
    at controlCourses (credential-form-control.js:6)
    at credential-form-control.js:21
```

The error is raised because the course survey section was removed in https://github.com/MIT-LCP/physionet-build/pull/708. This pull request fixes the error by removing the form control function for the section.

### DIFF
--- a/physionet-django/user/static/user/js/credential-form-control.js
+++ b/physionet-django/user/static/user/js/credential-form-control.js
@@ -1,27 +1,5 @@
 course_category_input = document.getElementById("id_application-course_category");
 
-function controlCourses() {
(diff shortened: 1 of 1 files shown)


bemoody:
Looks good to me.

Shippable seems to be having a meltdown at the moment, so I'll wait
and hope it is fixed tomorrow...

tompollard:
Force pushed the same change to (hopefully) trigger a rebuild

PR #723, opened 2019-09-04
Fixes the test error in the console

elfeto:
There was a 40 != 25 error in a test of the console. The error I received was while trying to delete the file, the except returned a “PermissionError” instead of a “IsADirectoryError”.

Here I now check if it’s a link because of versions, and then unlink.
OR, if it’s a folder, and delete the folder with its content.
Last if it’s a file, and then delete it.

### DIFF
--- a/physionet-django/project/utility.py
+++ b/physionet-django/project/utility.py
@@ -155,12 +155,11 @@ def remove_items(items, ignore_missing=True):
     """
     for item in items:
         try:
-            if  os.path.islink(item):
-                os.unlink(item)
-            elif os.path.isdir(item):
-                shutil.rmtree(item)
-            else:
-                os.remove(item)
+            os.unlink(item)
+        except OSError as e:
+            if e.errno not in (errno.EISDIR, errno.EPERM):
+                raise
+            shutil.rmtree(item)
         except FileNotFoundError:
             if not ignore_missing:
                 raise
--- a/physionet-django/project/utility.py
+++ b/physionet-django/project/utility.py
@@ -156,13 +156,13 @@ def remove_items(items, ignore_missing=True):
     for item in items:
         try:
             os.unlink(item)
+        except FileNotFoundError:
+            if not ignore_missing:
+                raise
         except OSError as e:
             if e.errno not in (errno.EISDIR, errno.EPERM):
                 raise
             shutil.rmtree(item)
-        except FileNotFoundError:
-            if not ignore_missing:
-                raise
 
 def clear_directory(directory):
     """
--- a/physionet-django/project/utility.py
+++ b/physionet-django/project/utility.py
@@ -152,9 +152,14 @@ def list_items(directory, return_separate=True):
 def remove_items(items, ignore_missing=True):
     """
     Delete the list of (full file path) files/directories.
+
+    Args:
+      items: Is a list of files or directories to delete.
+      ignore_missing: Flag to ignore missing files or directories.
     """
-    for item in items:
+    while items:
         try:
+            item = items.pop()
             os.unlink(item)
         except FileNotFoundError:
             if not ignore_missing:
@@ -163,6 +168,8 @@ def remove_items(items, ignore_missing=True):
             if e.errno not in (errno.EISDIR, errno.EPERM):
                 raise
             shutil.rmtree(item)
+        except (TypeError, AttributeError):
+            raise
 
 def clear_directory(directory):
     """
--- a/physionet-django/project/utility.py
+++ b/physionet-django/project/utility.py
@@ -155,9 +155,12 @@ def remove_items(items, ignore_missing=True):
     """
     for item in items:
         try:
-            os.remove(item)
-        except IsADirectoryError:
-            shutil.rmtree(item)
+            if  os.path.islink(item):
+                os.unlink(item)
+            elif os.path.isdir(item):
+                shutil.rmtree(item)
+            else:
+                os.remove(item)
         except FileNotFoundError:
             if not ignore_missing:
                 raise
--- a/physionet-django/project/utility.py
+++ b/physionet-django/project/utility.py
@@ -157,19 +157,21 @@ def remove_items(items, ignore_missing=True):
       items: Is a list of files or directories to delete.
       ignore_missing: Flag to ignore missing files or directories.
     """
-    while items:
-        try:
-            item = items.pop()
-            os.unlink(item)
-        except FileNotFoundError:
-            if not ignore_missing:
-                raise
-        except OSError as e:
-            if e.errno not in (errno.EISDIR, errno.EPERM):
-                raise
-            shutil.rmtree(item)
-        except (TypeError, AttributeError):
-            raise
+    if isinstance(items,(tuple,list)):
+        for item in items:
+            try:
+                os.unlink(item)
+            except FileNotFoundError:
+                if not ignore_missing:
+                    raise
+            except OSError as e:
+                if e.errno not in (errno.EISDIR, errno.EPERM):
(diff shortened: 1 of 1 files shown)


tompollard:
Excellent, thanks for getting to the bottom of this! Looks good to me. Also explains why I ran into issues deleting files on a project recently :) Um, please could you remove the extra space in `if  os.path.islink` before we merge (sorry!).

bemoody:
What?  Sorry, this makes no sense.

os.unlink is the same as os.remove (both of them invoke unlink(), not
remove().)  They are documented as being identical.

remove_items is never supposed to follow symlinks.  If a symlink is
one of the arguments, it should be deleted (but the referent, if it
exists, is untouched.)  If one of the arguments is a directory and a
symlink is found within that directory, then likewise, shutil.rmtree
deletes the link without affecting the referent.

What exactly is going on in the situation where you are encountering
an error and why would this change fix it?

tompollard:
One issue that this fixes is that folders cannot be deleted. For example, try loading the demo data and navigating to: http://localhost:8000/projects/T108xFtYkRAxiRiuOLEJ/files/

Currently, it is not possible to delete the "notes" folder in the screenshot below:

Screen Shot 2019-09-04 at 15 23 29

...so at the least, isn't this chunk needed?:

```python
elif os.path.isdir(item):
    shutil.rmtree(item)
else:
    os.remove(item)
```

bemoody:
On 9/4/19, Tom Pollard <notifications@github.com> wrote:
> One issue that this fixes is that folders cannot be deleted. For example,
> try loading the demo data and navigating to:
> http://localhost:8000/projects/T108xFtYkRAxiRiuOLEJ/files/
>
> Currently, it is not possible to delete the "notes" folder in the screenshot
> below:

Why not?  It works for me.  What happens for you?

After resetdb && loaddemo:

$ ls -la ../media/active-projects/T108xFtYkRAxiRiuOLEJ/
total 2700
drwxr-xr-x 3 benjamin benjamin    4096 Feb 21  2019 .
drwxr-xr-x 5 benjamin benjamin    4096 Sep  4 15:33 ..
-rw-r--r-- 1 benjamin benjamin  187922 Feb 21  2019 D_ITEMS.csv.gz
-rw-r--r-- 1 benjamin benjamin 1990193 Feb 21  2019 ICUSTAYS.csv.gz
-rw-r--r-- 1 benjamin benjamin  571615 Feb 21  2019 PATIENTS.csv.gz
drwxr-xr-x 2 benjamin benjamin    4096 Feb 21  2019 notes

... no symlinks in sight.

bemoody:
OK, so Tom demonstrated that unlink returning EISDIR is a Linux-ism
and OSX does EPERM instead.

Other options...

# messy
try:
    os.unlink(item)
except OSError as e:
    if e.errno not in (errno.EISDIR, errno.EPERM):
        raise
    shutil.rmtree(item)

# harder to read and slightly less efficient
s = os.lstat(item)
if stat.S_ISDIR(s.st_mode):
    shutil.rmtree(item)
else:
    os.unlink(item)

# also less efficient, maybe more readable
if os.path.isdir(item) and not os.path.islink(item):
    shutil.rmtree(item)
else:
    os.unlink(item)

I don't really like any of these.  Whatever you do, please explain
your reasoning with comments (especially if the behavior differs
between systems).

elfeto:
Why not do it backwards ?
try:
    shutil.rmtree(item)
except NotADirectoryError:
    os.remove(item)

In both OSX and Linux-ism it shows "NotADirectoryError" when trying to delete a file using "shutil.rmtree", and since "os.remove" and "os.unlink" are identical, we should keep what we had.

tompollard:
Makes sense to me. Does `shutil.rmtree` also raise `NotADirectoryError` for symlinks?

bemoody:
Unfortunately rmtree can raise all sorts of errors when passed a symbolic link:

```
#!/usr/bin/python3

import os
import shutil
import tempfile

def test(expr):
    with tempfile.TemporaryDirectory() as tmpdir:
        os.chdir(tmpdir)
        os.mkdir('dir')
        os.system('touch file')
        os.symlink('dir', 'dir-link')
        os.symlink('file', 'file-link')
        os.symlink('aaaaaa', 'broken-link')
        try:
            eval(expr)
            print('%-28s OK' % (expr,))
        except Exception as e:
            print('%-28s %r' % (expr, e))

test('os.unlink("missing")')
test('os.unlink("file")')
test('os.unlink("dir")')
test('os.unlink("file-link")')
test('os.unlink("dir-link")')
test('os.unlink("broken-link")')

test('shutil.rmtree("missing")')
test('shutil.rmtree("file")')
test('shutil.rmtree("dir")')
test('shutil.rmtree("file-link")')
test('shutil.rmtree("dir-link")')
test('shutil.rmtree("broken-link")')
```

->

```
os.unlink("missing")         FileNotFoundError(2, 'No such file or directory')
os.unlink("file")            OK
os.unlink("dir")             IsADirectoryError(21, 'Is a directory')
os.unlink("file-link")       OK
os.unlink("dir-link")        OK
os.unlink("broken-link")     OK
shutil.rmtree("missing")     FileNotFoundError(2, 'No such file or directory')
shutil.rmtree("file")        NotADirectoryError(20, 'Not a directory')
shutil.rmtree("dir")         OK
shutil.rmtree("file-link")   OSError('Cannot call rmtree on a symbolic link',)
shutil.rmtree("dir-link")    OSError('Cannot call rmtree on a symbolic link',)
shutil.rmtree("broken-link") FileNotFoundError(2, 'No such file or directory')
```

elfeto:
I like the option “#messy” because it differentiates between permission denied and operation not permitted.

Since the errors can be between is a directory or Operation not permitted depending on the OS.
Here we catch them both. If the error is not one of those, then it raises the error.

Note, this is different from permission denied.
I say this because we could use "except PermissionError" but that would capture permission denied.

bemoody:
Fair enough.

One problem, though, is that now FileNotFoundError isn't being caught
- FileNotFoundError is a subclass of OSError.  If ignore_missing is
true then FileNotFoundErrors are supposed to be ignored.

Also I think you need to import errno.

elfeto:
errno was already imported

I moved the FileNotFoundError to the top, and if it's not that error, then it will continue with the OS exceptions.

bemoody:
OK... this isn't your fault, but I just tried to test it by running

    # DO NOT DO THIS
    # project.utility.remove_items('/tmp/foo')
    # DO NOT DO THIS

and since this function interprets its first argument as a *sequence*
of things to delete, it promptly started trying to recursively delete
/.

I hope this didn't actually delete anything important. :|

But maybe this function should raise a TypeError if 'items' is a str
or bytes, just to stop people doing stupid things.

elfeto:
This is not as pretty, BUT I think it works.
Since a list contains the function pop, I check if the argument passed contains it.

It wont return a TypeError, but it will a AttributeError that would imply that the argument is not a list.

bemoody:
I guess that works, but I'd rather be more explicit and say "if items
is not a list then raise a TypeError."

Do all existing callers use a list of filenames, or are there any that
use a tuple?

bemoody:
(Sorry to complicate things when this started out as such a simple issue! :P)

elfeto:
Its fine, one learns things with every commit.

To be frank, I have no idea. I **THINK**, that they are all lists of filenames.

tompollard:
Checking for tuple and list should be safe `isinstance(item,(tuple,list))`

elfeto:
So, just turned it to check specifically for a tuple or a list. If its not that, then I will log it, and raise a TypeError.

PR #725, opened 2019-09-04
Fix encoding of mailto URLs within 403.html and 404.html

bemoody:
If I visit:

https://physionet.org/foo/bar/?asdf=ghjk&cc=nobody@example.com

it displays a mailto link which, when clicked, generates a message:

```
To: contact@physionet.org
Cc: nobody@example.com
Subject: Missing URL in physionet.org

Dear Physionet Team,

The following URL is missing from the website.

/foo/bar/?asdf=ghjk
```

(which is a silly example, but if there are actually links pointing to old resources with query strings - e.g., ATM - this could be important.)

Fix this by encoding the message properly.

### DIFF
--- a/physionet-django/templates/403.html
+++ b/physionet-django/templates/403.html
@@ -11,7 +11,7 @@
   <h1>Forbidden Access<br />(403)</h1>
   <br /><br />
   <p>You dont have access to see this page.</p>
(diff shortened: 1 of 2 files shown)


tompollard:
Thanks Benjamin

PR #728, opened 2019-09-05
Fix anonymous authentication and add DUA

lbulgarelli:
This fixes anonymous authentication for published project not working because of missing parenthesis on a function call.

Also, it adds the DUA in the anonymous login page, as long as it is provided by the license chosen for the project requested.

### DIFF
--- a/physionet-django/project/templates/project/anonymous_login.html
+++ b/physionet-django/project/templates/project/anonymous_login.html
@@ -9,7 +9,7 @@
 {% endblock %}
 
 {% block content %}
-<div class="container">
+<div class="container text-center">
   {% include "message_snippet.html" %}
   <form action="{% url 'anonymous_login' anonymous_url %}" method="post">
     {% csrf_token %}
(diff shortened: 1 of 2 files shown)


bemoody:
> This fixes anonymous authentication for published project not working
> because of missing parenthesis on a function call.

Don't you think it would be better to put this logic in one function,
rather than 5 and counting?  Replace or extend has_access?

lbulgarelli:
I do, but I have a separate branch where I'm working on auth/access stuff that includes this.

This PR is just a quick fix so the feature can be used.

bemoody:
Okay.  Thanks.

The button style is not good.  Setting "width: 25% !important" makes
the button either too small or too large (depending on window size),
and inconsistent with site style.  In general, please do not use
classes like "w-25".  Use classes with meaningful names, like
"btn-rsp" (which is probably what you should use here.)

Otherwise, this looks good to me.

PR #729, opened 2019-09-05
Bullet lists in static pages

lbulgarelli:
I don't quite like bullets lists, so I tried to give them a better appearance and some meaning here.

I created a new CSS class and added them in the bulleted lists in both `About` and `Share` pages.


tompollard:
When I view http://localhost:8000/about/publish/, I now see a mixture of different symbols to mark the bulleted items (✓, ∨ and •). Is this what you were aiming for?

Screen Shot 2019-09-06 at 15 35 09

Issue #720, opened 2019-09-03
/settings/cloud/ is uninformative


bemoody:
The page /settings/cloud/ contains a form field labelled "GCP Email" and a field labelled "AWS ID".

The page explains absolutely nothing about what these fields mean, what they do, why somebody might want to complete them (or might not want to complete them), or what exactly is supposed to go in them.

tompollard:
Funny you say that, this was bothering me earlier too. Definitely needs improvement!

elfeto:
Happier with the description added? or still a bit uninformative ?

PR #739, opened 2019-09-09
Added basic instructions for user cloud information

elfeto:
Here I add some basic information of the usage of the page.

#720

### DIFF
--- a/physionet-django/user/templates/user/edit_cloud.html
+++ b/physionet-django/user/templates/user/edit_cloud.html
@@ -1,17 +1,17 @@
 {% extends "user/settings.html" %}
 
-{% block title %} User Cloud information {% endblock %}
+{% block title %} Edit Cloud Details {% endblock %}
 
 {% block main_content %}
-<h1 class="form-signin-heading">User Cloud information</h1>
+<h1 class="form-signin-heading">Edit Cloud Details</h1>
 
-<p>Some of our databases are on Google Cloud Platform or Amazon Web Services.</p>
-<p>To gain access to them:</p>
-<ol>
-<li>Enter your google email address and/or your AWS numerical ID.</li>
-<li>Go to a published database and above the files click "Request access".</li>
-</ol>
-<p>Note: If the database does not have a "Request access", then that database is not available in the cloud at the moment.</p>
+<p>PhysioNet is integrated with several cloud services. To access data via these services, you will need to:</p>
+<ul>
+<li>Provide the appropriate account identifier using the form below.</li>
+<li>Navigate to the "Files" section of the relevant project (for example, https://physionet.org/content/eicu-crd-demo/#files).</li>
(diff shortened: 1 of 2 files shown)


physionet-django/user/templates/user/edit_cloud.html
```diff
@@ -1,10 +1,17 @@
 {% extends "user/settings.html" %}
 
-{% block title %} Edit Cloud information {% endblock %}
+{% block title %} User Cloud information {% endblock %}
```
tompollard: "Edit Cloud Details" might be better, for consistency with the other sections (all are in title case, and most begin with "Edit")

physionet-django/user/templates/user/edit_cloud.html
```diff
@@ -1,10 +1,17 @@
...
 
 {% block main_content %}
-<h1 class="form-signin-heading">Manage cloud settings</h1>
+<h1 class="form-signin-heading">User Cloud information</h1>
```
tompollard: "Edit Cloud Details"

physionet-django/user/templates/user/edit_cloud.html
```diff
@@ -1,10 +1,17 @@
...
 
+<p>Some of our databases are on Google Cloud Platform or Amazon Web Services.</p>
```
tompollard: I had a play around with the wording slightly to avoid hardcoding service names (so we can add integration without having to change it) and to provide a link to an example project.

Please could you use the version below if you're happy with it? Ideally `physionet` would be generated from a variable, rather than hardcoded.
```
PhysioNet is integrated with several cloud services. To access data via these services, you will need to:

- Provide the appropriate account identifier using the form below.
- Navigate to the "Files" section of the relevant project (for example, https://physionet.org/content/eicu-crd-demo/#files).
- Follow the instructions to request access. If instructions for cloud access are not shown, the project is not currently available on the cloud.
```

physionet-django/user/templates/user/edit_cloud.html
```diff
@@ -1,10 +1,17 @@
...
+<p>To gain access to them:</p>
+<ol>
+<li>Enter your google email address and/or your AWS numerical ID.</li>
+<li>Go to a published database and above the files click "Request access".</li>
+</ol>
+<p>Note: If the database does not have a "Request access", then that database is not available in the cloud at the moment.</p>
 <form action="{% url 'edit_cloud' %}" method="post" class="form-signin no-pd">
```
tompollard, line 15: Instead of acronyms on the form fields (`GCP Email` and `AWS ID`), please could we have the full name (`Google Cloud Platform (Email)` and `Amazon Web Services (ID)`)? If those are too long, then they could be shortened to `Google (Email)` and `Amazon (ID)`). We might want to add a `?` with instructions later, but I think this is fine for now.

elfeto:
I just addressed to all the comments.

tompollard:
Looks good, thanks Felipe.

PR #742, opened 2019-09-11
Add short description to project info card

bemoody:
The project's short description is not visible anywhere in the published project page itself, or in the project preview.  Add it to the "submission info card" shown in the console so that editors can easily review it.


elfeto:
Should this also be added in the: "manage_published_project.html" or should we just include the "submission_info_card.html" into the manage file ?

bemoody:
Good call.  I've added it there too.

It does seem like these two templates (submission_info_card.html and
manage_published_project.html) have a lot of the same content, so it
would be good to unify them if possible.

Some aspects of these templates are different because published/active
projects need to be handled differently, but other things are
different for no real reason I can see.

PR #746, opened 2019-09-16
Allow QueryBuilder to use Basic Authentication

elfeto:
Here we add a custom user agent string to be used for the Query Builder authentication.

### DIFF
--- a/physionet-django/project/utility.py
+++ b/physionet-django/project/utility.py
@@ -338,6 +338,7 @@ def grant_gcp_group_access(user, project, data_access, request):
 HTTP_AUTH_USER_AGENT = re.compile('|'.join((
     'Wget/',
     'libwfdb/',
+    'QueryBuilder',
 )))
 
 


bemoody:
How and why would Query Builder access files over HTTP?

elfeto:
The current method of authentication is verifying if the person has access to "https://archive.physionet.org/works/MIMICIIIClinicalDatabase/files/".

That page will open up a dialog for user credentials, and depending on the return code, I know if the person has access to MIMIC, and I check his credentials.

Since this function gives relatively the same, I can check the status code for the page and will tell me the same things without downloading any files.

I can try to bypass the BasicAuth, but that can get a bit messy in Query Builder.

bemoody:
As someone who has used similar kludges in the past, it's a bad idea.

I'd go so far as to say that people should not be asked to enter their
PhysioNet password to any web page other than
https://physionet.org/login/, and they shouldn't be taught to think
such a thing is normal.

Aside from convenience and other "layer 8" concerns: having multiple
login forms, multiple pieces of code that handle passwords, define
login and session policies, etc., makes for a security mess and should
be avoided if at all possible.

elfeto:
You are suggesting that we should create another set of credentials (username/email and password) that the user will have to use and memorize and we would have to keep track of that and maintain ??

bemoody:
No, I'm suggesting that the user should log in only once, and having
done so, should have access to both sites.

elfeto:
We could also implemente a oauth type authentication. But I would say this is messy.

bemoody:
@elfeto

Can you point me to where authentication is handled in QueryBuilder?
I can take a look and see if I can think of any easier ways to
integrate the login process.

PR #747, opened 2019-09-16
Minor update to the wording for credentialing decision time

tompollard:
When a user submits the form at http://127.0.0.1:8000/settings/credentialing/ to request credentialed access, they see the following message: 

> Your credentialing application was submitted on Sept. 16, 2019, 1:58 p.m.. We will respond shortly.

Delays are almost always a result of an unresponsive reference, so this changes the wording to:

> Your credentialing application was submitted on Sept. 16, 2019, 1:58 p.m..
>
> We aim to reach a decision within two weeks. If you have not received a decision within this time, it is likely that we are awaiting a response from your reference.


tompollard:
force push to trigger rebuild (builds were failing earlier).

PR #749, opened 2019-09-16
Allow applications for credentialed access to be withdrawn

tompollard:
This pull request allows a user to withdraw their application for credentialed access. It is useful in cases where the user discovers that they have made a mistake in their application.

- Adds a new "Withdrawn" status to the `user.CredentialApplication` model.
- Moves the process of accepting/rejecting/withdrawing applications to an `apply_decision` method in the CredentialApplication class. 
- Allows open applications to be withdrawn by the submitting user via http://127.0.0.1:8000/settings/credentialing/
- Allows open applications to be withdrawn by admin users via the "Process Credential Application" page
- Updates the notification email to account for the new status (text below).

## Not approved:

> Dear Tom Pollard,

> Thank you for applying for credentialed access to PhysioNet. Your
> application was not approved. This may be for one of the following
> reasons:

> - It was incomplete, or included obviously incorrect information
> (perhaps as a result of browser auto-fill).
> - You did not submit the required CITI completion report, listing the
> training modules you completed, with dates and scores.
> - Your CITI training is out of date.
> - You are a student, postdoc, intern, or trainee, but did not list
> your supervisor (a faculty member or someone with a senior research
> appointment at your institution) as reference.
> - Your research summary did not include sufficient information, or was
> in some other way inadequate.

> The following comments were added to your application:

> A CUSTOM MESSAGE HERE

> If you are able to address these issues, please open a new
> credentialing application at
> http://127.0.0.1:8000/credential-application/.

> Regards,

> The PhysioNet Team,
> MIT Laboratory for Computational Physiology,
> Institute for Medical Engineering and Science,
> MIT, E25-505 77 Massachusetts Ave. Cambridge, MA 02139

## Withdrawn:

> Dear Tom Pollard,

> Your application for credentialed access to PhysioNet was withdrawn.

> The following comments were added to your application:

> A CUSTOM MESSAGE HERE

> Regards,

> The PhysioNet Team,
> MIT Laboratory for Computational Physiology,
> Institute for Medical Engineering and Science,
> MIT, E25-505 77 Massachusetts Ave. Cambridge, MA 02139

## Approved

> Dear Tom Pollard,

> Thank you for your interest in the PhysioNet Clinical Databases. We
> are pleased to say that your application for credentialed access has
> been approved. You are now able to access protected databases upon
> agreeing to the terms of usage. For example, you can access MIMIC-III
> by following the steps below:

> - Go to the project page at http://127.0.0.1:8000/content/mimiciii/
> - Find the “Files” section in the project description
> - Click “Sign the data use agreement” to agree to the terms of usage
> for this dataset

> Regards,

> The PhysioNet Team,
> MIT Laboratory for Computational Physiology,
> Institute for Medical Engineering and Science,
> MIT, E25-505 77 Massachusetts Ave. Cambridge, MA 02139


bemoody:
It seems wrong to say "withdrawn" if the person doing the
"withdrawing" is not the person who submitted the application in the
first place.

What does it mean, and what is the use case, for an administrator to
"withdraw" rather than "rejecting" an application?

This also creates a race condition:

1. Applicant submits an application.

2. Administrator opens either
   (a) /console/complete-credential-applications/ or
   (b) /console/credential-applications/1tabTZn5VCpj2fHq1Anv/process/
   in their browser, to view the application.

3. Applicant withdraws their application.

4. Administrator clicks "accept" or "reject".

What happens then?  In case (a), the application is marked as
accepted/rejected, and applicant receives a corresponding email.  In
case (b), the administrator sees a 500 error (CredentialApplication
matching query does not exist).

I would say that in both cases, the administrator should see an error
message ("This application was withdrawn by the applicant") and the
application should remain "withdrawn".

(Of course, you can also have problems if multiple administrators try
to handle the same application simultaneously.  We should guard
against that, too, if we're not already doing so.)

tompollard:
Thanks for the quick and helpful feedback Benjamin. I have made a couple of updates which I think address the points raised:

- Applications can now only be withdrawn by the applicant.
- If the applicant clicks "Withdraw application" when the administrator has already accepted/rejected the application, the error is caught and "The application has already been processed" is displayed.
- If the administrator attempts to accept/reject an application that has been withdrawn by the applicant, the error is caught and "The application has already been processed. It may have been withdrawn by the applicant or handled by another administrator." is displayed.

physionet-django/console/views.py
```diff
@@ -815,8 +816,15 @@ def process_credential_application(request, application_slug):
     Process a credential application. View details, contact reference,
     and make final decision.
     """
-    application = CredentialApplication.objects.get(slug=application_slug,
-        status=0)
+    try:
+        application = CredentialApplication.objects.get(slug=application_slug,
+            status=0)
+    except ObjectDoesNotExist:
```
elfeto: The model already has a way to handle NotExists with "CredentialApplication.DoesNotExist"

physionet-django/user/models.py
```diff
@@ -701,6 +702,26 @@ def is_latest_by_user(self):
...
 
+    def apply_decision(self, decision, responder):
+        """
+        Reject (1), accept (2), or withdraw (3) a credentialing application.
+
+        Args:
+            decision (int): 1 is reject, 2 is accept, 3 is withdraw.
+            responder (str): User object
+        """
+        # if accepted, credential the user
```
elfeto: Shouldn't this be atomic? In the case that he has been approved, we need both to happen at the same time.

physionet-django/user/views.py
```diff
@@ -361,9 +361,21 @@ def edit_credentialing(request):
     applications = CredentialApplication.objects.filter(user=request.user)
     current_application = applications.filter(status=0).first()
 
+    if request.method == 'POST' and 'withdraw_credentialing' in request.POST:
+        try:
+            current_application.apply_decision(decision=3, responder=request.user)
+            return redirect('withdraw_credentialing_success')
```
elfeto: Why don't you render the template here?

lbulgarelli:
I just don't like the idea of propagating this kind of thing:

```python
MY_OPTIONS = (
        ('', '-----------'),
        (1, 'OPTION 1'),
        (2, 'OPTION 1'')
        (..., '...'),
        (N, 'OPTION N')
    )
```

We should try and start to properly represent states in our code.

We could use Django's proxy models and custom managers to do that.

Also

```python
def apply_decision(self, decision, responder):
        """
        Reject (1), accept (2), or withdraw (3) a credentialing application.
        Args:
            decision (int): 1 is reject, 2 is accept, 3 is withdraw.
            responder (str): User object
        """
        # if accepted, credential the user
        if decision == 2:
            user = self.user
            user.is_credentialed = True
            user.credential_datetime = timezone.now()
            user.save()

        self.responder = responder
        self.status = decision
        self.decision_datetime = timezone.now()
        self.save()
```

Why not separate functions for each decision? This style should be avoided. No reason why you have to spread this and make the code really hard to read.

We don't want `current_application.apply_decision(decision=3, responder=request.user)` all over the place and having to figure out for every function we have what a code used as a argument means.

tompollard:
> Why not separate functions for each decision? This style should be avoided. No reason why you have to spread this and make the code really hard to read.

Yeah, I did this originally, but you end up with a bit of duplicate code, so switched. Agree separate is clearer though, so will switch back.

physionet-django/user/models.py
tompollard: yep, good call!

physionet-django/console/views.py
tompollard: Good point. Switched!

physionet-django/user/views.py
tompollard: Done!

tompollard:
Thanks for the great feedback everyone. I think everything is addressed, except:

> I just don't like the idea of propagating this kind of thing:

> ```python
> MY_OPTIONS = (
>         ('', '-----------'),
>         (1, 'OPTION 1'),
>         (2, 'OPTION 1'')
>         (..., '...'),
>         (N, 'OPTION N')
>     )
> ```

I agree, but would like to deal with this another time.

bemoody:
In user.views.edit_credentialing, you're using AttributeError to
distinguish between an existing and non-existing object.  That's a
little confusing and seems flaky.  Probably it would be better to
explicitly check whether current_application is None (or else avoid
using first(), use get() instead and handle the DoesNotExist case.)

console.views.complete_credential_applications is still using the same
logic as before, so the "complete applications" interface still has
the issue of allowing the admin to approve/reject an application that
has already been processed or withdrawn.

One way to deal with this would be to do the check at the level of
user.models.CredentialApplication._apply_decision.  Another way would
be to do the check in console.forms.ProcessCredentialForm.

Either way, since we have two separate console interfaces doing the
same task, I'd prefer to minimize duplicated logic. :/

tompollard:
@bemoody thanks for the comments.

> In user.views.edit_credentialing, you're using AttributeError to distinguish between an existing and non-existing object.  That's a little confusing and seems flaky.

Now checking whether the result is none with `if current_application:`

> console.views.complete_credential_applications is still using the same logic as before, so the "complete applications" interface still has the issue...since we have two separate console interfaces doing the
same task, I'd prefer to minimize duplicated logic. :/

The issue is fixed, with a small chunk of repeated logic in the view (`console/views.py`). I tried alternatives, but this seemed cleanest. 

The model (`user.models.CredentialApplication._apply_decision`) didn't seem like the right place, and I couldn't get the same functionality with the form validator (`console.forms.ProcessCredentialForm`) without passing more arguments, which made this more complicated that it needs to be. 

I see the two separate console thing as a temporary hack, and I've been letting pretty much anything go into the `complete_xxx` versions up to now, so the duplication doesn't worry me too much. If you have alternative suggestions, let me know!

elfeto:
In the past applications page, if the application is withdrawn, you should list there that it was withdrawn, in the "reason".

tompollard:
Please could you point to the specific place that you would like it to appear?

The application history (e.g. http://localhost:8000/console/view-credential-applications/ATYEEruTTHCdiq6HuA6A/) lists "withdrawn":

Screen Shot 2019-10-08 at 09 32 25

I assume you are thinking about http://localhost:8000/console/past-credential-applications/unsuccessful which is a good idea, but there isn't a "reason" field. You mean the "Response" field or somewhere else?

Screen Shot 2019-10-08 at 09 35 10

elfeto:
Yea, you got it, I was speaking of the "Response". Since it's already in the history, no need to duplicate it there. 

The word "Reason" came from the credentialing management page that KP uses.

tompollard:
I had a play around and ended up adding a "decision" column, which seemed a bit clearer than mixing the "withdrawn" status with the response comments.

bemoody:
Thanks.  This addresses my major concern, which was the potential for
sending confusing messages depending on who clicks the button first.
I'm happy to merge this as is and try to clean up the implementation
another day. :)

In commit 16f3d0862eaf2b3b6de0d34c0bc249775ff77486 you reverted
a6420f82e76c548016b1f1835bfa804d029c198f ("We will respond shortly" ->
"We aim to reach a decision within two weeks...")  Did you intend to
change that?

PR #755, opened 2019-09-20
Added functionality to automaticly create and assign organizational e…

elfeto:
Here I check if the project is protected, if so, then
I check if there exists a group email address to manage
access for it. If so, then I do nothing.

If there isn't, then I create it, and assign it to the new
bucket.


tompollard:
I've been putting this off because I'm never sure how best to test things that involve integration with the cloud services. Will try to get over that and look in the next few days!

physionet-django/console/utility.py
```diff
@@ -71,9 +75,57 @@ def remove_bucket_permissions(bucket):
...
-        logger.info("Removed all read permissions from bucket {}".format(bucket.name))
+        LOGGER.info("Removed all read permissions from bucket {}".format(bucket.name))
 
-def add_email_bucket_access(project, email):
+def create_grant_access_group(bucket, project_slug, version, title):
+    """
+    Function to automatically create organizational group emails
+    that manage access to private projects.
+    """
+    email = '{0}-{1}@physionet.org'.format(project_slug, version)
```
tompollard: Can these hardcoded `@physionet.org`s be avoided?

physionet-django/console/utility.py
```diff
@@ -71,9 +75,57 @@ def remove_bucket_permissions(bucket):
...
+                project {0} version {1}'.format(title, version)}).execute()
+            if creation['kind'] == 'admin#directory#group':
+                service = create_directory_service(settings.GCP_DELEGATION_EMAIL, group=True)
+                update = service.groups().update(groupUniqueId=email, body={
+                    'allowExternalMembers':'true',
+                    'whoCanPostMessage':'ALL_OWNERS_CAN_POST',
+                    'whoCanModerateMembers':'OWNERS_AND_MANAGERS',
+                    'whoCanJoin': 'INVITED_CAN_JOIN'}).execute()
+                if update['kind'] == 'groupsSettings#groups':
+                    LOGGER.info("The access group {0} was succesfully created".format(
```
tompollard: succesfully -> successfully

physionet-django/console/utility.py
```diff
@@ -82,15 +134,19 @@ def add_email_bucket_access(project, email):
     storage_client = storage.Client()
     bucket = storage_client.get_bucket(project.gcp.bucket_name)
     policy = bucket.get_iam_policy()
-    for role in Public_Roles:
-        policy[role].add('user:'+email)
+    if group:
+        for role in Public_Roles:
```
tompollard: Not directly part of this pull request, but `Public_Roles` is an odd (/unPEP8) variable name. If this is a constant, then perhaps all caps is better? https://www.python.org/dev/peps/pep-0008/#constants

physionet-django/console/utility.py
```diff
@@ -71,9 +75,57 @@ def remove_bucket_permissions(bucket):
         policy[item[0]].discard(item[1])
     if to_remove:
         bucket.set_iam_policy(policy)
-        logger.info("Removed all read permissions from bucket {}".format(bucket.name))
+        LOGGER.info("Removed all read permissions from bucket {}".format(bucket.name))
 
-def add_email_bucket_access(project, email):
+def create_grant_access_group(bucket, project_slug, version, title):
+    """
```
tompollard: As discussed, I'd find it helpful to have more detailed docstrings and comments throughout. No need to go overboard, but giving the basics of how things work would be helpful. 

An alternative is to try to document the Google Cloud integration elsewhere, but I think within the code is kind of nice, because documentation tends to get lost/stale, while keeping things in the code should help us to keep it up to date.

physionet-django/console/utility.py
```diff
@@ -71,9 +75,57 @@ def remove_bucket_permissions(bucket):
...
+    Function to automatically create organizational group emails
+    that manage access to private projects.
+    """
+    email = '{0}-{1}@physionet.org'.format(project_slug, version)
+    if 'production' not in settings.SETTINGS_MODULE:
```
tompollard: I think we recently merged a new approach to checking the server status?

physionet-django/console/utility.py
```diff
@@ -71,9 +75,57 @@ def remove_bucket_permissions(bucket):
...
+    service = create_directory_service(settings.GCP_DELEGATION_EMAIL)
+    # Get all the members of the Google group
+    try:
+        found = False
+        outcome = service.groups().list(domain='physionet.org').execute()
+        for group in outcome['groups']:
+            if group['email'] == email:
+                found = True
+        if not found:
+            creation = service.groups().insert(body={'email': email,
```
tompollard: This (and other functions like `create_bucket` are pretty big and might be easier to follow if broken into one or more sub-functions. e.g. if `if not found` calls a function called `create_group(x,y)` (or whatever) then it becomes easier to follow what's happening, and maybe reduces the need to keep track of the state of some of the variables.

physionet-django/console/views.py
```diff
@@ -614,11 +614,22 @@ def manage_published_project(request, project_slug, version):
...
                     if not bucket_name:
-                        bucket_name = utility.create_bucket(project=project.slug,
-                            protected=is_private, version=project.version)
+                        bucket_name, group = utility.create_bucket(
+                            project=project.slug, version=project.version,
+                            title=project.title, protected=is_private)
                     GCP.objects.create(project=project, bucket_name=bucket_name,
                         managed_by=user, is_private=is_private)
-                    messages.success(request, "The GCP bucket for project {0} was \
+                    if group:
```
tompollard: Doesn't need to be dealt with in this pull request, but same point about the size of `manage_published_project`. At some point it would be good if we could break some of the functionality into sub-functions.

physionet-django/console/views.py
```diff
@@ -614,11 +614,22 @@ def manage_published_project(request, project_slug, version):
                 except GCP.DoesNotExist:
                     bucket_name = utility.check_bucket(project.slug, project.version)
                     if not bucket_name:
-                        bucket_name = utility.create_bucket(project=project.slug,
-                            protected=is_private, version=project.version)
+                        bucket_name, group = utility.create_bucket(
```
tompollard: Again, this isn't directly related to the pull request, but will mention just because otherwise I'll forget. On line 606 we assign `is_private = False` and then update to `True` if the `project.access_policy > 0`. It seems safer to do this the other way around (i.e. assume `is_private = True` unless explicitly defined otherwise?).

tompollard (commented):
Thanks @elfeto, I'm still getting my head around the cloud integration, but looks like nice functionality to add. I've made a few comments throughout.

My main thought is that it would be helpful to (1) document the core aspects of how the integration is set up somewhere, preferably in docstrings/comments (2) try to break some of the larger functions into smaller, easier to read chunks/functions. These tasks can be dealt with later as separate issues.

physionet-django/console/utility.py
elfeto: I did a workaround for this.

physionet-django/console/utility.py
elfeto: Added more docstrings and separated some functions

elfeto:
I did a rebase and broke the code into smaller chunks. I also added a way to handle the hardcoded physionet.org.

physionet-django/console/views.py
elfeto: Most of the databases are not private, and I added other code to be executed if the project is private which would be an inconvenient if we change this now.

physionet-django/console/utility.py
```diff
@@ -3,169 +3,268 @@
...
+        make_bucket_public(bucket)
+        LOGGER.info("Made bucket {0} public".format(bucket_name.lower()))
+
+
+def get_bucket_name(project, version):
+    """
+    Generate the bucket name.
+    """
+    current_site = Site.objects.get_current()
+    production_site = Site.objects.get(id=3)
```
tompollard: @bemoody do you have any suggestions for improving the way we handle this step (checking the domain)?

physionet-django/user/test_views.py
```diff
@@ -244,7 +244,7 @@ def test_edit_emails(self):
 
         # Test 5: Verify the newly added email
         # Get the activation info from the sent email
-        uidb64, token = re.findall('http://testserver/verify/(?P<uidb64>[0-9A-Za-z_\-]+)/(?P<token>[0-9A-Za-z]{1,13}-[0-9A-Za-z]{1,20})/',
+        uidb64, token = re.findall('http://localhost:8000/verify/(?P<uidb64>[0-9A-Za-z_\-]+)/(?P<token>[0-9A-Za-z]{1,13}-[0-9A-Za-z]{1,20})/',
```
tompollard: Please could you briefly explain this change Felipe?

physionet-django/console/views.py
```diff
@@ -716,6 +699,48 @@ def manage_published_project(request, project_slug, version):
...
+        group = utility.get_bucket_email(project.slug, project.version)
+
+    bucket_name = utility.get_bucket_name(project.slug, project.version)
+
+    if GCP.objects.filter(bucket_name=bucket_name):
+        messages.success(request, "The bucket already exists. Resending the \
+            files for the project {0}.".format(project))
+    else:
+        if utility.check_bucket_exists(project.slug, project.version):
+            LOGGER.info("The bucket {0} already exists, skipping bucket and \
```
tompollard, line 741: `GCP.objects.create` and the `if group` clause both appear outside the `else`, so the logger message doesn't appear to be true.

physionet-django/console/views.py
```diff
@@ -716,6 +699,48 @@ def manage_published_project(request, project_slug, version):
...
+
+    if project.access_policy == 0:
+        is_private = False
+        group = None
+    else:
+        group = utility.get_bucket_email(project.slug, project.version)
+
+    bucket_name = utility.get_bucket_name(project.slug, project.version)
+
+    if GCP.objects.filter(bucket_name=bucket_name):
```
tompollard: Is it possible to simplify this logic, perhaps by removing the nesting and/or moving some of the content to another function? As it is now, I find the steps difficult to follow.

bemoody:
> +def get_bucket_name(project, version):
> +    """
> +    Generate the bucket name.
> +    """
> +    current_site = Site.objects.get_current()
> +    production_site = Site.objects.get(id=3)
>
> @bemoody do you have any suggestions for improving the way we handle this
> step (checking the domain)?

It would be better to define a variable in the settings module; for
example, you could define in settings/base.py:

    GCP_BUCKET_PREFIX = 'testing-delete.'

and then in settings/production.py:

    GCP_BUCKET_PREFIX = ''

TBH, I'm not sure if this is the *best* way to do custom settings
(should we have a prefix to avoid clashing with Django or other
packages?) ...but the settings module is where we are keeping most
host-specific configuration.

The sites module should really only be used to answer the question
"what is the canonical name of this host?"

physionet-django/console/views.py
elfeto: If the object already exists, there is no reason to re-enter to the group creation section. I don't understand what you want to say in this comment.

physionet-django/console/views.py
elfeto: I would not make sense to move to another function since this was already moved to another function, plus the function is not that long nor complicates since half the function are messages and loggers. I can "simplify the logic" by converting an if/else for a try/except.

physionet-django/user/test_views.py
elfeto: It simply sets the proper URL for the tests as added in the Sites fixtures.

physionet-django/console/views.py
tompollard: Looking at this in more detail, I now understand. I'd missed that `create_bucket` was also creating the access group, and had thought that "skipping group creation" referred to the `DataAccess.objects.create` step (that is not being skipped.

physionet-django/console/views.py
tompollard: I'll take another look and will give it a bit more thought. It feels like a minor rearrangement could make the function easier to follow.

elfeto:
I addressed all the comments.

The only thing that stands out would be this:
>The sites module should really only be used to answer the question
"what is the canonical name of this host?"

In some way I have to use the words physionet.org for the bucket and email creation.
The only way I can think of without hardcoding it is using the sites and get the domain of the production by id.
Technically, I am not using sites to ask a question, but more as a statement :p 
If you have an alternative let me know.

bemoody:
> In some way I have to use the words physionet.org for the bucket and email creation.

I don't know what you mean by "email creation", and I've forgotten most of what was going on with these changes. :/ But I'll try to address what I think you're asking.

You've created an account with Google that lets us publish databases with the suffix "physionet.org".

I understand that creating that account requires you to somehow prove that you are an administrator for that domain.

However, having done that, *using* that account to publish databases is independent of the fact that the web server is running on the host "physionet.org".

If example.org and example.com are operated by the same organization, there's no reason that a web server at example.com should not be able to upload data to Google Cloud within the "example.org" namespace.  There's no reason that a web server at "publishing.physionet.org" shouldn't be able to upload to the "physionet.org" namespace, or that a web server at "physionet.org" shouldn't be able to upload to the "databases.physionet.org" namespace.  If that's what we wanted.

That is what I mean when I say that these two configuration elements - the hostname of the server, and the suffix of the cloud bucket - are independent.  If the two are related in some way, that is a matter of Policy, not Mechanism, and that relationship should be defined by "configuration" rather than "application logic".

> The only way I can think of without hardcoding it is using the sites and get the domain of the production by id.

If the configuration is defined in "physionet/settings/production.py", I generally wouldn't consider that to be "hard-coding."  In contrast, if it is defined in "console/utility.py", I would consider that to be "hard-coding" and something to avoid.

elfeto:
The problem is that all: staging, dev and production need to use the same name.

You are correct in your example, staging.physionet.org, dev.physionet.org and physionet.org would be able to use the name physionet.org.

We use organizational google groups email addresses to handle access in GCP. These emails are hidden to the world and are only seen by organizational members. These emails can be "anything@physionet.org". Having the google group exists in google does not mean that the email actually works, just means that google will look for the members of that to do X action.

For example BigQuery/storage access for MIMIC is handled by the email "mimic-iii-users@physionet.org". All members of the email group will ave access to GCP.

As for the bucket it's the same mechanism.

I can define a domain variable in base in which will contain physionet.org if that simplifies things?

bemoody:
Actually, the staging server should not be capable of giving anyone
access to any real restricted-access data.  It should not have the
necessary passwords to create or modify buckets, or grant or deny
access.

From what you describe, it sounds like the "GCP fake email domain"
should be configurable separately, vs. "GCP public bucket domain".

And TBH, this probably should not be something that could ever be
mistaken for a real routable email address.  It should be
"mimic-iii-users@google-cloud.physionet.org" or something like that.

elfeto:
>Actually, the staging server should not be capable of giving anyone
access to any real restricted-access data.  It should not have the
necessary passwords to create or modify buckets, or grant or deny
access.

Only reason it should work on staging is to test with the same configuration as production. Also there is no real data in staging.

>From what you describe, it sounds like the "GCP fake email domain"
should be configurable separately, vs. "GCP public bucket domain".

I don't understand what you mean here.
The access for both buckets and databases (bigquery) are handled by google email groups. 
In order to grant access to a "GCP public bucket domain" we use a "GCP fake email domain".

>And TBH, this probably should not be something that could ever be
mistaken for a real routable email address.  It should be
"mimic-iii-users@google-cloud.physionet.org" or something like that.

I agree, but the admin console doesnt allow me to do that. For me to do that, then we would have to create the alias "google-cloud.physionet.org" and then proof to google that we own it, and do do the same things we are doing in this PR but to a different host

bemoody:
> Only reason it should work on staging is to test with the same configuration
> as production. Also there is no real data in staging.

There is real data in *Google*.  If you can use the staging server to
give somebody access to Google, then you can use the staging server to
give somebody access to restricted data.  Please do your best to
ensure that isn't possible.

tompollard:
Thanks, looks good to me. Before we merge, I have a few questions/comments (all or most minor I think):

- are the points above resolved? if not, can they be raised as new issues or do they need to be dealt with before we merge?
- in `create_bucket()` in `/console/utility.py`, the following chunk assigns two different objects to `bucket`:

```py
    bucket = storage_client.create_bucket(bucket_name)
    bucket = storage_client.bucket(bucket_name)
```

Why not call `storage_client.create_bucket()` without the assignment?

- the behavior of the `email` argument of `bucket_info()` in `/console/utility.py` is slightly unusual (if `email = True` then return `email` else return `name`). Why not drop the email argument and always return name and email? 

- I don't completely understand the two `Site.objects.get(id=3)` that appear in `console/utility.py`. Do these need to be hardcoded? Don't we want to check the domain (localhost/staging/production) and then use the result of this check to decide on the appropriate behaviour?

```py
def bucket_info(project, version, email=False):
    ...
    production_site = Site.objects.get(id=3)


def create_access_group(bucket, project, version, title):
...
    production_domain = Site.objects.get(id=3)
```

- the docstring for `grant_gcp_group_access` in `project/utility.py` suggests that it can handle different values for the `data_access` argument, but it defaults to Big Query access unless `data_access` is equal to 3. It would be nice to handle the argument explicitly (e.g. if data_access == 3 then x elif data_access == 4 then x else x).

- the "Returns" comment in the docstring for `create_access_group()` in `console/utility.py` doesn't reflect behaviour:

```py
def create_access_group(bucket, project, version, title):
    """
    Create an access group with appropriate permissions, if the group does not
    already exist.

    Returns:
        bool: False if there was an error or change to the API. True otherwise.
    """

    ... 

    return email
```

- the `project_request_access()` function in `project/views.py` would be much more readable with  some line breaks (e.g. before and after `if` and `try` clauses, before and after `return`). 

- lastly (not directly related to this pull request, so let's save for another issue), it would be good for us to spend a bit more time on the add/remove cloud options section in the project management form:

Screen Shot 2020-03-11 at 18 38 38

I'd assumed that clicking "remove" would remove access to the bucket and possibly delete the bucket too, but from our chat it sounds like it only removes the cloud option from the "files" section of the project page.

elfeto:
>are the points above resolved? if not, can they be raised as new issues or do they need to be dealt with before we merge?

Up to my understanding yes.

>Why not call storage_client.create_bucket() without the assignment?

One creates the bucket, but doesn't let you edit, and the other get the already created bucket and allows you to make updates

>the behavior of the email argument of bucket_info() in /console/utility.py is slightly unusual (if email = True then return email else return name). Why not drop the email argument and always return name and email?

Because not all buckets will have an email. Only Credentialed and protected projects do. I can return a tuple with name/email and name/False

>I don't completely understand the two Site.objects.get(id=3) that appear in console/utility.py. Do these need to be hardcoded? Don't we want to check the domain (localhost/staging/production) and then use the result of this check to decide on the appropriate behaviour?

No checks are done. No matter where I am, I need a way to get "physionet.org" in all localhost/staging/production. The behaviour is decided using the prefix that is set in the settings file. All three email and buckets will end in either "@physionet.org" or ".physionet.org" One will go to GCP and the other will go to the fake PN GCP organization.

>the docstring for grant_gcp_group_access in project/utility.py suggests that it can handle different values for the data_access argument, but it defaults to Big Query access unless data_access is equal to 3. It would be nice to handle the argument explicitly (e.g. if data_access == 3 then x elif data_access == 4 then x else x).

Sure

>the "Returns" comment in the docstring for create_access_group() in console/utility.py doesn't reflect behaviour:

You are right, I will fix that

>the project_request_access() function in project/views.py would be much more readable with some line breaks (e.g. before and after if and try clauses, before and after return).

ok

>lastly (not directly related to this pull request, so let's save for another issue), it would be good for us to spend a bit more time on the add/remove cloud options section in the project management form:

Yea I know is not pretty and there is much improvement to be done there.

>I'd assumed that clicking "remove" would remove access to the bucket and possibly delete the bucket too, but from our chat it sounds like it only removes the cloud option from the "files" section of the project page.

No, if you click remove it will only delete the option to request access. The bucket is still active, and the access that has been granted before is still active as well. If this is something you don't like, we can create an issue and I can see how this can be fixed depending on the desired action here.

I will try to make the changes early tomorrow :)

tompollard:
Thanks, sounds good! Couple of quick comments on the points above:

> I don't completely understand the two Site.objects.get(id=3) that appear in console/utility.py. Do these need to be hardcoded? Don't we want to check the domain (localhost/staging/production) and then use the result of this check to decide on the appropriate behaviour?

>> No checks are done. No matter where I am, I need a way to get "physionet.org" in all localhost/staging/production. The behaviour is decided using the prefix that is set in the settings file. All three email and buckets will end in either "@physionet.org" or ".physionet.org" One will go to GCP and the other will go to the fake PN GCP organization.

Thanks for explaining, makes sense. In this case, could we request the domain/host in the configuration file? Feels like this would be more generalisable. e.g.:

```
GCP_BUCKET_DOMAIN = "physionet.org"
``` 

> the behavior of the email argument of bucket_info() in /console/utility.py is slightly unusual (if email = True then return email else return name). Why not drop the email argument and always return name and email?

>> Because not all buckets will have an email. Only Credentialed and protected projects do. I can return a tuple with name/email and name/False

Or just two variables and email is empty if the bucket field is empty?

elfeto:
I believe I addressed all your comments above.

tompollard:
looks good to me, thanks!

PR #756, opened 2019-09-23
Changed all usernames to lowercase

elfeto:
There is a problem with the username login when the person added the username as a capitalized letter. This is because Django by default on login turns the input to lower.

Here we set the default for usernames to lower case.

### DIFF
--- a/physionet-django/user/forms.py
+++ b/physionet-django/user/forms.py
@@ -137,7 +137,7 @@ class UsernameChangeForm(forms.ModelForm):
         self.old_file_root = self.instance.file_root()
         if User.objects.filter(username__iexact=self.cleaned_data['username']):
             raise forms.ValidationError("A user with that username already exists.")
-        return self.cleaned_data['username']
+        return self.cleaned_data['username'].lower()
 
     def save(self):
         """
@@ -221,7 +221,7 @@ class RegistrationForm(forms.ModelForm):
         "Record the original username in case it is needed"
         if User.objects.filter(username__iexact=self.cleaned_data['username']):
             raise forms.ValidationError("A user with that username already exists.")
-        return self.cleaned_data['username']
+        return self.cleaned_data['username'].lower()
 
 
     def clean_password2(self):


bemoody:
Can you please describe the problem in detail?

Give an example of a series of steps to reproduce it.

elfeto:
Reproduce:
 - ./manage.py resetdb && ./manage.py loaddemo
 - Login as "rgmark"
 - Change the username to one containing the capitalized letter "RGmark1"
 - Logout
 - Login using the username "RGmark1"
 
Like I said in the description.... 
The login function used by default in Django converts username/email into lowercase.

If there is a username with a capitalized letter, it will fail because it will check upper vs lower case...
RGmark1 != rgmark1 so it fails

Here I change the change username and registration forms to lowercase usernames.

bemoody:
Okay, I can see that it's a problem, but this doesn't fix it (after
applying this change, "RGmark1" still can't log in.)

I don't think it's Django; I think this is implemented by
user.models.DualAuthModelBackend.

Now, I don't have any problem with forcing usernames to be lowercase,
but I guess somebody didn't like that, since it was changed. :)

If we *do* want to allow usernames to be variable case, then I don't
think putting "iexact" everywhere is a good way to do that - better, I
think, would be to store the canonical username as lowercase, and
separately store the variable-case username which is only used for
display.  That should make lookups more efficient, and allow enforcing
uniqueness at the database level.

So, do we want to do that (and keep the variable-case usernames that
people have already registered?)  Or do we want to squash all existing
users down to lowercase?

tompollard:
Thanks for looking at this so quickly Felipe. It would be great to get this fixed because several people have reported problems logging in.

> Or do we want to squash all existing users down to lowercase?

My vote is to squash everything to lower case. If there is demand for cased usernames later, we can think about @bemoody's fix (create a new "display_username" field).

tompollard:
> Okay, I can see that it's a problem, but this doesn't fix it (after applying this change, "RGmark1" still can't log in.)

This fix seems good to me. We would need to manually change all existing usernames in the database to lower case, but that seems okay. 

@bemoody  I'd like to merge if possible. Do you have any objections (assuming that we're all fine with forcing usernames to lowercase)?

elfeto:
I am with Tom, I feel that creating another field / method would be too much of a hassle, for something this silly. It's easier just to force all usernames lowercase. 

I don't know of anywhere that has case sensitive usernames.

tompollard:
It looks there were a few new registrations with upper case usernames:

```
from models.users import User
User.objects.filter(username__regex="[A-Z]+")
```

I converted these to lower case.

PR #757, opened 2019-09-23
Move log when user connect to the signal

elfeto:
There is an unknown bug that is filling the log file with repetitions of
"User logged in email"

This is wrong because there are thousands of repetitions of the same 6
people.

Here I move the logging to the signal that happens after the connection is done.


tompollard:
Thanks Felipe, just had a chat to Benjamin about this, who pointed out that the logs are people downloading lots of files with wget (new log for each file!).

Issue #752, opened 2019-09-18
Hyperlinks from HTML previews are broken if project is unpublished


alistairewj:
If you navigate to a file and then click it to open the HTML preview, there is a hyperlink bar at the top:

```[PREVIEW]: Project Name
File: <base> / test1 / filename.txt (10 bytes)
```

The folders are links, but they are broken when the project is unpublished. In my BERT-deid project the `test1` link points to https://physionet.org/projects/eVT6feluFB2RPWmd6JbF/files/test1#files-panel which then an raises nginx 404. The links work on published projects though (I tested a few). Also seems to not work regardless of the type of file previewed.

Oddly enough the folder navigation bar in the file panel works fine (`test1` links to https://physionet.org/projects/eVT6feluFB2RPWmd6JbF/#files-panel ), it's only when you load a preview that it breaks.

bemoody:
Buggy generation of the breadcrumbs, but also buggy redirection...

https://physionet.org/projects/eVT6feluFB2RPWmd6JbF/files/test1

in principle, ought to redirect to

https://physionet.org/projects/eVT6feluFB2RPWmd6JbF/files/test1/

not to

https://physionet.org/protected/active-projects/eVT6feluFB2RPWmd6JbF/test1/

alistairewj:
I just tested this on production with one of my projects and the fix seems to work.

PR #758, opened 2019-09-23
Fix breadcrumb links in project file preview

bemoody:
File previews that are linked from the file management page will show breadcrumb links pointing back to the file management page.  For example (if logged in as rgmark):

    http://localhost:8000/projects/SHuKI1APLrwWCqxSQnSk/preview/subject-100/100.hea?return=files

shows a link `subject-100` which points to:

    /projects/SHuKI1APLrwWCqxSQnSk/files/subject-100#files-panel

but this should instead be:

    /projects/SHuKI1APLrwWCqxSQnSk/files/subject-100/#files-panel

(with a trailing slash); the former URL works on a development server but does not currently work on staging/production.

Fixes issue #752.

### DIFF
--- a/physionet-django/project/fileviews/base.py
+++ b/physionet-django/project/fileviews/base.py
@@ -53,7 +53,7 @@ class FileView:
                 and not hasattr(self.project, 'publish_datetime')):
             base_url = reverse('project_files', args=(self.project.slug,))
             for b in breadcrumbs:
-                b.rel_path = os.path.join(base_url, b.full_subdir)
+                b.rel_path = os.path.join(base_url, b.full_subdir, '')
 
         response = render(request, template, {
             'breadcrumbs': breadcrumbs,


elfeto:
For me using Chrome in OSX, `https://physionet.org/content/mitdb/1.0.0/mitdbdir#files-panel` automatically redirects to `https://physionet.org/content/mitdb/1.0.0/mitdbdir/#files-panel`

You are trying to do this by default ? or ?

bemoody:
That redirection works for published projects, and for *previews*, but
not for the file management page.

To demonstrate, I've just created a directory "hello" in your project
"vfgE4V34NH6rnZuolsSK", so if you're logged in:
https://physionet.org/projects/vfgE4V34NH6rnZuolsSK/preview/hello
will redirect you to
https://physionet.org/projects/vfgE4V34NH6rnZuolsSK/preview/hello/

but

https://physionet.org/projects/vfgE4V34NH6rnZuolsSK/files/hello
will redirect you to
https://physionet.org/protected/active-projects/vfgE4V34NH6rnZuolsSK/hello/

It might be easy to fix this if I knew the right incantation, but that
would require a bit of digging into nginx's logic.

elfeto:
OH, sorry for the delay. What you would want is something like this in the nginx config ?
 - rewrite ^([^.]*[^/])$ $1/ permanent;
it would add a trailing slash to anything after the last slash that doesn't contain a dot. (Currently running on staging as a test)

for example
`https://staging.physionet.org/projects/43LPirNlsYC6BERSU5d0/files/test`
redirects to 
`https://staging.physionet.org/projects/43LPirNlsYC6BERSU5d0/files/test/`

I just added that line under: 

       location / {
           uwsgi_pass  django;
           include     uwsgi_params;
           rewrite ^([^.]*[^/])$ $1/ permanent;
       }

I mean, I am happy to merge this too if you prefer while we sort out the config.

bemoody:
No... I don't fully understand what that's doing, but if I have a
regular file called "RECORDS", then:
https://staging.physionet.org/projects/taPfYhFm7G4nnkrBwm79/preview/RECORDS?return=files
would then give an HTTP redirection to:
https://staging.physionet.org/projects/taPfYhFm7G4nnkrBwm79/preview/RECORDS/?return=files

You can't assume that directories don't have dots in their names, or
that regular files do.

(Also, I guess you meant (.*/[^/.]+), not ([^.]*[^/]), anyway.)

elfeto:
Interesting. I forgot about files with no dots. It might be a bit trickier then.

bemoody:
Hmm... it would probably be okay for serve_active_project_file to
check the file's type before returning an X-Accel-Redirect.  I
wouldn't do that for, say, serve_published_project_file, but for
active projects the slight overhead doesn't really matter.

I feel weird about ignoring a TOCTOU issue - especially when dealing
with files that are directly controlled by an untrusted user - but the
failure mode here is that you might possibly get redirected to an
inaccessible URL, so maybe it's okay?

PR #760, opened 2019-09-24
set real number of item counts in pagination

elfeto:
The current total if items is the numbers of items per page. Here it sets the amount to the total of items in pagination.

### DIFF
--- a/physionet-django/console/templates/console/past_credential_applications.html
+++ b/physionet-django/console/templates/console/past_credential_applications.html
@@ -28,7 +28,7 @@
         {% else %}
           <a class="nav-link" id="u-tab" href="{% url 'past_credential_applications' 'unsuccessful' %}" role="tab">
         {% endif %}
-            Unsuccessful Applications <span class="badge badge-pill badge-info">{{ u_application.paginator.counts }}</span></a>
+            Unsuccessful Applications <span class="badge badge-pill badge-info">{{ u_applications.paginator.count }}</span></a>
       </li>
     </ul>
   </div>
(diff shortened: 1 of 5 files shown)


physionet-django/console/templates/console/past_credential_applications.html
```diff
@@ -20,15 +20,15 @@
...
+            Successful Applications <span class="badge badge-pill badge-info">{{ applications.paginator.count }}</span></a>
       </li>
       <li class="nav-item">
         {% if "unsuccessful" in request.path %}
           <a class="nav-link active" id="u-tab" href="{% url 'past_credential_applications' 'unsuccessful' %}" role="tab">
         {% else %}
           <a class="nav-link" id="u-tab" href="{% url 'past_credential_applications' 'unsuccessful' %}" role="tab">
         {% endif %}
-            Unsuccessful Applications <span class="badge badge-pill badge-info">{{ u_applications|length }}</span></a>
+            Unsuccessful Applications <span class="badge badge-pill badge-info">{{ u_application.paginator.counts }}</span></a>
```
tompollard: looks like the `s` got misplaced? should be: `u_applications.paginator.count`

physionet-django/console/templates/console/published_projects.html
```diff
@@ -11,7 +11,7 @@
 {% block content %}
 <div class="card mb-3">
   <div class="card-header">
-    Published Projects <span class="badge badge-pill badge-info">{{ projects|length}}</span>
+    Published Projects <span class="badge badge-pill badge-info">{{ projects.paginator.count}}</span>
```
tompollard: While you're updating the thing above, please could you add a space between the variables and the `}}` template tags for consistency?

Issue #750, opened 2019-09-17
Registration activation link should tell the user if the account is already activated


tompollard:
If an account is already activated, the activation link in the registration email should say something along the lines of "Your account is now activated. Click here to login."

> I got the activation link. When I try the to click on it , it says “Invalid Activation Link. There was an error with your link.

>> Once the account is activated it will display that message. 
>> I verified and the account was successfully activated.

PR #764, opened 2019-09-26
Display a message when activating an active account

elfeto:
There are a lot of emails complaining that the account activation failed.

At the moment when the person tries to activate the account the email link is valid for only one time. After the account is active, and you use either the same link or refresh the page, it will show a link is invalid message. 

This here is replaced with a the account is already active message.

### DIFF
--- a/physionet-django/user/views.py
+++ b/physionet-django/user/views.py
@@ -53,7 +53,7 @@ def activate_user(request, uidb64, token):
         email.save()
         logger.info('User activated - {0}'.format(user.email))
         context = {'title':'Activation Successful', 'isvalid':True}
-    elif user is not None and user.is_active and default_token_generator.check_token(user, token):
+    elif user is not None and user.is_active:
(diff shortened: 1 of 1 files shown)


bemoody:
(This is issue #750.)

I don't think it's necessary to display a *different* message if the
link has already been used; why not just say "Activation Successful"
rather than "The accouint [sic] is active"?

You're also duplicating logic.  Why not just move lines 46-53 into a
conditional "if not user.is_active"?

Please write commit messages to explain what you're changing and why.

tompollard:
Thanks, fixes the issue for someone who has not logged in before. Once the user has logged in, `default_token_generator.check_token(user, token)` is `False`, so the user will see the `Invalid Activation Link` message again. 

Would you like to suggest a fix for the logged in user too?  If not, this is a step forward, so I'm okay to merge as is, but it would be nice to fix both cases!

tompollard:
Nice, thanks. I'll merge when the test finishes.

PR #765, opened 2019-09-30
Fix crash when trying to access a regular file as a directory

bemoody:
These changes fix a crash that occurs when trying to preview a non-existent file, such as:

https://physionet.org/content/ptbdb/1.0.0/patient001/s0010_re.xyz/ws

(where 's0010_re.xyz' is an existing file, not a directory, so this raises NotADirectoryError rather than FileNotFoundError).

In addition, restructure display_project_file to be more careful with exceptions, and add some tests.

Thanks to @alistairewj for reporting (?) this issue. :P


tompollard:
nice, thanks!

Issue #727, opened 2019-09-05
'project_request_access' is not an active project view


bemoody:
'project_request_access' is a *published* project view, not an active project view.  To be consistent with other views, this:

- should be renamed to "published_project_request_access" or similar

- should be located somewhere under /content/<project_slug>/, not /projects/<project_slug>/

PR #767, opened 2019-10-01
Adds option for editors/authors to download all files

### LINKED ISSUE #499: Mass file download for editorss
There should be a way to allow editors download the files, either create a zip and allow download, or a scp/rsync

elfeto:
The functionality to people download all the files for protected projects
exists, the only issue is that it works for published projects.

Here I add the functionality for active projects also. If the person is
either the editor or one of the authors (which they should not have the
link, but still is possible for them because is their project), then
they can download all the files from the project, ether to have them or
to do the "editing".

Here I use a lot of the groundwork made by Benjamin, and even it might
be possible to eventually merge with "serve_published_project_file"
since it uses basically the same things.

Here it resolves #499


bemoody:
Definitely a useful feature to have.  Thanks for working on this!

One problem, though, is that this mixes up the "active" and "published"
project namespaces.  *Active* project URLs should go under /projects/,
*published* projects should not.

(See issue #727 for a similar problem in the opposite direction.)

Apart from keeping things semantically clean, notice that, for
example, this wouldn't work if an active project has a top-level
subdirectory whose name is entirely dots and/or digits.

elfeto:
So, you mean that the url should change to something similar to this: "https://physionet.org/projects/download/vfgE4V34NH6rnZuolsSK/files/" ?

Could you give me an example of a subdirectory whose name is entirely dots and/or digits ?

bemoody:
> Could you give me an example of a subdirectory whose name is entirely dots
> and/or digits ?

https://physionet.org/projects/j0dgl3L9LKUaY0Y4haEn/preview/30/
for one ;)

"https://physionet.org/files/j0dgl3L9LKUaY0Y4haEn/30/" would match
that first pattern, i.e. version 30 of a published project named
"j0dgl3L9LKUaY0Y4haEn".

elfeto:
OH!!! Now I understand what you meant. In theory then, this should be resolved if we add a proper URL, for example "https://physionet.org/projects/download/vfgE4V34NH6rnZuolsSK/files/" ?

bemoody:
Yep.  I'd swap it and put the project slug first (e.g.
/projects/vfgE4V34NH6rnZuolsSK/download/) for consistency, but that's
mostly just nit-picking.

bemoody:
Thanks!

Do you intend to make this available only to the assigned editor?
You're checking whether "user in project.author_list()", but that
doesn't work (project.author_list() is a QuerySet of Authors, not
Users).

You're also checking and verifying the "anonymousaccess" cookie, but
not doing anything with that information.  (I would be inclined not to
support that mechanism here, at least initially.)

Please add some test cases, along the lines of what I did in commit
13101c8af4db1b06a7f71399d7214c630bc56b57 - check that the authorized
user(s) can download an existing file, and that unauthorized or
anonymous users can't.

bemoody:
Looks good to me.  Thanks!

PR #768, opened 2019-10-02
Removes old way of granting access to GCP buckets

elfeto:
Before the organizational emails were possible, the only way to
grant access to google Buckets were to add the email individually
to each bucket. 

Now this is handled by the organizational email.

I didn't removed the actual functions because they might be useful.


tompollard:
I don't think there is any need to keep `add_email_bucket_access` if it isn't used (we have it in the history if we ever need to go back). Please could you remove it before we merge?

elfeto:
Its being used in PR #755

PR #770, opened 2019-10-02
Deprecate static indexes

lbulgarelli:
The landing page buttons redirect to the search page with this change.

### DIFF
--- a/physionet-django/templates/home.html
+++ b/physionet-django/templates/home.html
@@ -26,7 +26,7 @@ PhysioNet
       </p>
       <br>
       <div>
-        <a href="{% url 'database_overview' %}">Data</a>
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks Lucas, I'll chat to the others to see if we can push this through.

tompollard:
Lucas, when you have 5 minutes, please could you find me for a quick chat about this? I'm keen to drop the static lists, but I think we need a slightly softer introduction to the change. Perhaps:

- begin by implementing this switch only for datasets.
- somewhere on the search page, say: to see a complete list of datasets, click here (link to page that lists all datasets and descriptions).

tompollard:
thanks Lucas, let's call this an experiment...

PR #773, opened 2019-10-08
Minor changes to credentialing application form, requested by KP

tompollard:
This is a collection of minor changes to the credentialing application form that were requested by KP.

> If Canada or USA is listed as the applicant's country, state should be required.

Added to the form validation.

> Students and postdocs must provide their supervisor name and contact details as a reference.

Added to the form validation.

> "tooltip for first name still says "First and middle names", which isn't right.  Should be just ""First name(s)"

Fixed

> "it would be nice to adjust the width of the tooltip box so that, for example, the suffix tooltip isn't so tall.  Or at least set as ragged-right, rather than centered."

Style updated.

> "second instance of "Do not list yourself as reference." should be wrapped with `<em>`...`</em>` please."

Done

> "tooltip for Research topic says "Brief description on your research...".  Should be "...of your research..."."

Fixed.

>  "you agree to the terms and conditions described on this page." doesn't seem quite right.  The terms and conditions aren't described; they're enumerated.  How about ... "you agree to the terms and conditions above"?

Updated

### DIFF
--- a/physionet-django/static/custom/css/physionet.css
+++ b/physionet-django/static/custom/css/physionet.css
@@ -544,3 +544,8 @@ a, code{
   border-image: linear-gradient(to bottom, silver 0%, silver 60%, transparent 100%) 1% 100%;
           
 }
+
+.tooltip-inner {
+  text-align: left;
+  max-width: 350px;
+}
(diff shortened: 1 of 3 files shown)


elfeto:
Do you care that it works if the person just enters a space in all three inputs?

tompollard:
> Do you care that it works if the person just enters a space in all three inputs?

That did cross my mind, but hopefully it won't be too common! If you think it's an issue, I can strip the white space when we check.

PR #774, opened 2019-10-09
Removes approved_users duplicates DUASignature

### LINKED ISSUE #712: "approved_users" duplicates "DUASignature"
There are two database tables that appear to be fully redundant:

* project_duasignature (= model 'DUASignature')

* project_publishedproject_approved_users (= field 'approved_users')

They are redundant in that there is no way to become an approved user other than by signing a DUA and no way to sign a DUA without becoming an approved user.

However:

- some functions refer to approved_users while others refer to DUASignatures

- functions that add or remove a user need to do so in both places

- project.views.sign_dua currently doesn't do so atomically

- there's no /admin/ interface to approved_users AFAICS

So I think it would be better to remove 'approved_users' and refer exclusively to DUASignatures.

elfeto:
There is a duplication on how the projects with Licenses  and Credentialing
handles access. The DUASignature has a bit more information than
approved_users, it has the date of when it was signed.

Here I removed the duplication by removing approved_users from the project
since it's duplicated in the DUASignature.

This should fix #712


tompollard:
Thanks, looks good to me. @bemoody this change deletes data, so shout now if you have any questions before I merge!

PR #775, opened 2019-10-09
Altered public contributions message

elfeto:
I replaced the message for the public contributions to the title and version. 

Looks cleaner. opinions ?


tompollard:
Much cleaner, thanks!

PR #780, opened 2019-10-11
Added site-wide cache

elfeto:
After some time researching cache in Django, here I add a site wide cache removing the views to serve the files from protected projects.

In the developing I left two options, the cache that will be used for the moment in the servers, and the cache designated for developing commented out.

At the moment, this is a test to see if the site speed improves.

There are multiple types of cache, and multiple types of doing this. I prefer memory cache, but I want to see how this goes. Or at least start a discussion about this.

### DIFF
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -52,6 +52,7 @@ INSTALLED_APPS = [
 ]
 
 MIDDLEWARE = [
+    'django.middleware.cache.UpdateCacheMiddleware',
     'django.middleware.security.SecurityMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
     'django.middleware.common.CommonMiddleware',
@@ -59,6 +60,7 @@ MIDDLEWARE = [
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
+    'django.middleware.cache.FetchFromCacheMiddleware',
 ]
 
 CRON_CLASSES = [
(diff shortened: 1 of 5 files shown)


bemoody:
Hmm.  I have not tested this but here are my general, possibly
uninformed, thoughts about caching. :)

I don't know anything about django.middleware.cache.  My instinct is
to say that the Python layer is probably not the best place to be
doing HTTP-like caching, because it's heavyweight - every request
still has to be handled by one of the small number of worker
processes.  Real HTTP caching at the level of nginx would likely give
better bang for your buck.  It'd also scale better and provide
client-side benefits too.

From the "application" side, the requirements are more or less the
same: you need to determine which responses make sense to cache, for
how long they should be cached (max-age) and which request parameters
affect the response (Vary).

That said, HTTP caching headers are definitely complicated and
confusing and I'd be interested to know if Django has ways to make it
easier or more intuitive for the programmer.  And in particular, if
doing something like this brings us closer to being able to use HTTP
caching in the future, that's a good thing.


The other piece of the picture is that there's stuff that *could* in
principle be cached, but is impossible for "dumb" mechanisms, like
either HTTP or Django middleware, to take advantage of.  Most public
HTML pages look exactly the same regardless of whether you're logged
in, except for one tiny block of HTML in the nav bar.  That tiny block
of HTML makes it impossible to serve the same cached copy of the page
to two different clients, if one's logged in and one isn't.  I'd love
to hear of any good solutions to this.

elfeto:
So there are 3 types of catching for Django, one would be this one. The dumb catching, this "works" for the entire website catching every page. Another would be a per-view catching, which the the variables like "max-age" would be added as part as the decorator. The last one would be a template catching, I am not very familiarized with this yet. 

I thought as a test, we could run this and see if there was a difference in database calls and wait time, and it would make it faster.

PR #783, opened 2019-10-14
Changed location for nginx and uwsgi logs

elfeto:
The logs for both Nginx and uWSGI keep getting bigger and bigger.
Since there is limited space on the "/var/" partition I would feel safer of they are in a place with lots of space.

### DIFF
--- a/deploy/common/etc/logrotate.d/nginx
+++ b/deploy/common/etc/logrotate.d/nginx
@@ -1,0 +1,18 @@
+/data/log/nginx/*.log {
+	daily
+	missingok
+	rotate 14
+	compress
+	delaycompress
+	notifempty
+	create 0640 www-data adm
+	sharedscripts
+	prerotate
+		if [ -d /etc/logrotate.d/httpd-prerotate ]; then \
+			run-parts /etc/logrotate.d/httpd-prerotate; \
+		fi \
+	endscript
+	postrotate
+		invoke-rc.d nginx rotate >/dev/null 2>&1
+	endscript
+}
--- a/deploy/common/etc/logrotate.d/uwsgi
+++ b/deploy/common/etc/logrotate.d/uwsgi
@@ -1,4 +1,4 @@
-/var/log/uwsgi/*.log 
+/data/log/uwsgi/*.log
 {
         rotate 7
         daily
@@ -6,9 +6,9 @@
         notifempty
         delaycompress
         compress
-        create 0640 pn pn
+        create 0640 www-data www-data
         sharedscripts
         postrotate
-                systemctl restart uwsgi
+                systemctl reload emperor.uwsgi
         endscript
 }
--- a/deploy/common/etc/logrotate.d/uwsgi
+++ b/deploy/common/etc/logrotate.d/uwsgi
@@ -4,7 +4,6 @@
         daily
         missingok
         notifempty
-        copytruncate
         delaycompress
         compress
         create 0640 www-data www-data
--- a/deploy/common/etc/logrotate.d/uwsgi
+++ b/deploy/common/etc/logrotate.d/uwsgi
@@ -4,6 +4,7 @@
         daily
         missingok
         notifempty
+        copytruncate
         delaycompress
         compress
         create 0640 www-data www-data
--- a/deploy/common/etc/logrotate.d/uwsgi
+++ b/deploy/common/etc/logrotate.d/uwsgi
@@ -6,6 +6,7 @@
         notifempty
         delaycompress
         compress
+        su www-data www-data
         create 0640 www-data www-data
         sharedscripts
         postrotate
--- a/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
+++ b/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
@@ -64,8 +64,8 @@ server {
 
     #### End of public data ####
 
-    error_log /var/log/nginx/physionet_error.log warn;
-    access_log /var/log/nginx/physionet_access.log;
+    error_log /data/log/nginx/physionet_error.log warn;
+    access_log /data/log/nginx/physionet_access.log;
     log_not_found off;
 
     # Finally, send all non-media requests to the Django server.
@@ -165,8 +165,8 @@ server {
 
     #### End of public data ####
 
-    error_log /var/log/nginx/physionet_error.log warn;
-    access_log /var/log/nginx/physionet_access.log;
+    error_log /data/log/nginx/physionet_error.log warn;
+    access_log /data/log/nginx/physionet_access.log;
     log_not_found off;
 
     location / {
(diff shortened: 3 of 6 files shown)


tompollard:
@bemoody this is more your area. happy with the new location?

bemoody:
We also need to rotate the logs.  This is done by
/etc/logrotate.d/nginx for nginx.

/etc/logrotate.d/uwsgi is seemingly not working:

* As the error message says,

  error: skipping "/var/log/uwsgi/physionet_uwsgi.log" because parent
directory has insecure permissions (It's world writable or writable by
group which is not "root") Set "su" directive in config file to tell
logrotate which user/group should be used for rotation.

* I don't think "systemctl restart uwsgi" will do anything.  I think
it needs to be "systemctl restart emperor.uwsgi" to do a hard
restart... but we don't want to do a hard restart!  I'm not sure
whether "systemctl reload emperor.uwsgi" would work (but I would hope
so.)

https://code.luasoftware.com/tutorials/uwsgi/uwsgi-log-rotation-without-restart/
suggests that the only way to do this is "copytruncate" mode... but
that page also seems to be talking about a non-emperor configuration.

uwsgi also has its own built-in log rotation thing.  I'd prefer to use
logrotate if possible but that might be an alternative.

bemoody:
Thanks!  To be clear, this means the log *directories* should all be
owned by root (so that logrotate works safely) but the uwsgi log
*files* should be owned by www-data (so that they can be written to.)

I don't think it should be necessary to do *both* the postrotate
command and copytruncate.  Did you test whether it works with just the
postrotate command?

(I'm not sure if we're more likely to break something by automatically
reloading uwsgi every morning, or if we're more likely to break
something by neglecting to reload it when we should.  All other things
being equal, probably better to do it than not :))

elfeto:
Yea I tested it out in staging.

```
croot@physionet-staging:~#logrotate /etc/logrotate.d/uwsgi --verbose -f
reading config file /etc/logrotate.d/uwsgi
Reading state from file: /var/lib/logrotate/status
Allocating hash table for state file, size 64 entries
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state
Creating new state

Handling 1 logs

rotating pattern: /data/log/uwsgi/*.log 
 forced from command line (7 rotations)
empty log files are not rotated, old logs are removed
considering log /data/log/uwsgi/physionet_uwsgi.log
  Now: 2019-10-23 16:37
  Last rotated at 2019-10-23 14:24
  log needs rotating
considering log /data/log/uwsgi/physionet_uwsgi-req.log
  Now: 2019-10-23 16:37
  Last rotated at 2019-10-23 14:24
  log does not need rotating (log is empty)
rotating log /data/log/uwsgi/physionet_uwsgi.log, log->rotateCount is 7
dateext suffix '-20191023'
glob pattern '-[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]'
compressing log with: /bin/gzip
renaming /data/log/uwsgi/physionet_uwsgi.log.7.gz to /data/log/uwsgi/physionet_uwsgi.log.8.gz (rotatecount 7, logstart 1, i 7), 
old log /data/log/uwsgi/physionet_uwsgi.log.7.gz does not exist
renaming /data/log/uwsgi/physionet_uwsgi.log.6.gz to /data/log/uwsgi/physionet_uwsgi.log.7.gz (rotatecount 7, logstart 1, i 6), 
old log /data/log/uwsgi/physionet_uwsgi.log.6.gz does not exist
renaming /data/log/uwsgi/physionet_uwsgi.log.5.gz to /data/log/uwsgi/physionet_uwsgi.log.6.gz (rotatecount 7, logstart 1, i 5), 
old log /data/log/uwsgi/physionet_uwsgi.log.5.gz does not exist
renaming /data/log/uwsgi/physionet_uwsgi.log.4.gz to /data/log/uwsgi/physionet_uwsgi.log.5.gz (rotatecount 7, logstart 1, i 4), 
old log /data/log/uwsgi/physionet_uwsgi.log.4.gz does not exist
renaming /data/log/uwsgi/physionet_uwsgi.log.3.gz to /data/log/uwsgi/physionet_uwsgi.log.4.gz (rotatecount 7, logstart 1, i 3), 
old log /data/log/uwsgi/physionet_uwsgi.log.3.gz does not exist
renaming /data/log/uwsgi/physionet_uwsgi.log.2.gz to /data/log/uwsgi/physionet_uwsgi.log.3.gz (rotatecount 7, logstart 1, i 2), 
renaming /data/log/uwsgi/physionet_uwsgi.log.1.gz to /data/log/uwsgi/physionet_uwsgi.log.2.gz (rotatecount 7, logstart 1, i 1), 
renaming /data/log/uwsgi/physionet_uwsgi.log.0.gz to /data/log/uwsgi/physionet_uwsgi.log.1.gz (rotatecount 7, logstart 1, i 0), 
old log /data/log/uwsgi/physionet_uwsgi.log.0.gz does not exist
log /data/log/uwsgi/physionet_uwsgi.log.8.gz doesn't exist -- won't try to dispose of it
copying /data/log/uwsgi/physionet_uwsgi.log to /data/log/uwsgi/physionet_uwsgi.log.1
truncating /data/log/uwsgi/physionet_uwsgi.log
running postrotate script
root@physionet-staging:~# 
```

bemoody:
> Yea I tested it out in staging.

Thanks.  But did you test it *without* the copytruncate option?
I've tested it just now in a VM and it seems to work correctly,
but perhaps I'm missing something.

Regarding permissions, logrotate's behavior seems bizarre to me -
I can see that it's unsafe to manipulate files as the superuser
when the directory is world-writable, but why is it considered
acceptable to do so when the directory is *owned* by a
non-superuser?  I would strongly suggest using the "su" option
when the directory isn't owned by root.

Additionally, I notice that /data itself is currently owned by
pn:pn and mode 775.  This seems risky, I think it should be owned
by and only writable by root.  Do you see a reason not to change
that?

elfeto:
So, I just tested it without, and it worked fine (on staging).

 The data in staging I changed it to root:root, and 755 it works fine.

bemoody:
Thanks!

PR #785, opened 2019-10-17
Added active users to the console 

### LINKED ISSUE #776: Add "Active Users" item to the Console?
Usually when searching for a user, we are interested in active users. The "All Users" list is getting pretty noisy, so for now (until we do something about this), it might be helpful to add a new "Active Users" item to the list, which displays all of the users that are not in "Inactive Users". The "All users" item could be moved to the bottom of the list.

Screen Shot 2019-10-09 at 14 40 02

elfeto:
Here I add active users to the console moving "all users" down. I also try to fix the bugginess of the search by aborting the previos ajax call before executing the next one. 

I also add an icon next to the person to easily identify if the account is active.  

This closes #776

### DIFF
--- a/physionet-django/console/templates/console/console_navbar.html
+++ b/physionet-django/console/templates/console/console_navbar.html
@@ -106,10 +106,11 @@
       </li>
       <!-- users -->
       {% url 'user_list' as user_list %}
-      {% url 'user_list_inactive' as user_list_inactive %}
       {% url 'user_list_admin' as user_list_admin %}
+      {% url 'user_status_list' 'active' as user_list_active %}
+      {% url 'user_status_list' 'inactive' as user_list_inactive %}
       <li class="nav-item" data-toggle="tooltip" data-placement="right">
-      {% if request.path == user_list or request.path == user_list_inactive or request.path == user_list_admin  %}
+      {% if request.path == user_list or request.path == user_list_inactive or request.path == user_list_admin or request.path == user_list_active %}
         <a id="nav_users_dropdown" class="nav-link nav-link-collapse drop" data-toggle="collapse" href="#userComponents" data-parent="#sideAccordion" aria-expanded="true">
       {% else %}
         <a id="nav_users_dropdown" class="nav-link nav-link-collapse drop collapsed" data-toggle="collapse" href="#userComponents" data-parent="#sideAccordion" aria-expanded="false">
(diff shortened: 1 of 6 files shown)


tompollard:
Thanks, this is useful. For me, the icons are the same for everyone:

Screen Shot 2019-10-23 at 18 38 46

How are they supposed to differ for active vs inactive users? Not sure they are necessary really...

elfeto:
The icons is for activated accounts.

tompollard:
> The icons is for activated accounts.

Okay, I see. This kind of makes sense, but the dual use of "active" also becomes quite confusing. In the inactive users menu, I now have a list of "inactive users", some of who have activated accounts and some who don't. We need to think about this a little more...it probably needs a slight reorganization to clarify this point.

elfeto:
I believe all 4 user pages use the same html template.
This said, they all would have the icon.

The icon is useful for the pages of "All users" and "Inactive users", since it will show if the account was never verified by email.

All the users in the "Active users" by definition have their accounts activated, but since it uses the same template then it's shown.

I don't see how confusing it can be that an icon will show if the account was verified by email and thus activated once. If you could explain, I can re-arrange things for you.

tompollard:
I feel like this would have been easier as three separate pull requests because all of the changes are useful but they are all independent, so it would be easier to review and discuss them separately (1. fix ajax; 2. add new menu item in the console for active users 3. introduce new icon for highlighting users who have not activated their accounts).

On the point that we discuss in the earlier posts, I like the idea of highlighting accounts that are activated, but it isn't clear that this is what the picture of the little guy means (even to a user like me who is pretty familiar with the system). This makes me think that the current implementation isn't quite right.

tompollard:
My current thought is:

- "Active" should mean that someone has created an account and "activated" the account by clicking the link in the email (list at: https://physionet.org/console/users/active/).
- "Inactive" should mean that someone has created an account but not clicked the link in the email (list at https://physionet.org/console/users/active/).

As mentioned in https://github.com/MIT-LCP/physionet-build/issues/795, we should purge the inactive accounts on a regular basis.

What do you think?

elfeto:
The current definition of "inactive" is an account who is not activated OR hasn't been logged in the last 3 months.

tompollard:
> The current definition of "inactive" is an account who is not activated OR hasn't been logged in the last 3 months.

Yep, so we have two different definitions of inactive (one for the list, one for the icon). My suggestion is to define inactive as "has not activated account". I don't think the new icon is necessary.

elfeto:
ok, you want to change the current definition that we are using of "inactive users" then.

So now you only want:
 - In inactive the accounts that have not been activated by email.
 - In active, ALL the accounts of the users that have activated the account once.
 - Administrators 
 - All users both inactive and active? -> why would you want this then ?

The users that have not logged in the last 6 months or 2 years will be count as active by your definition.

tompollard:
> So now you only want:
> 
> "inactive" the accounts that have not been activated by email.
> "active", ALL the accounts of the users that have activated the account once.
> "administrators"
> "all users" both inactive and active?

Yep, I think this makes sense.

> why would you want "all users" then ?

having "active", "inactive" and "admin" might be enough, but i don't see any harm in keeping "all users" for now. the menus are likely to change again as our user types develop.

Issue #772, opened 2019-10-08
Add "Remind author" email for projects that are awaiting revisions


tompollard:
Projects are getting stuck in the "Awaiting Author Revisions" stage. To help move these projects along, a good first step would be to add a "reminder" email.

We could add a "Remind author" and "Last reminded" column to the table in the "Awaiting Author Revisions" tab at: https://physionet.org/console/submitted-projects/, along with a button to send a reminder email. The "Last reminded" column should be populated with the most recent date. 

This is a first step towards slightly more sophisticated functionality for handling projects that are awaiting revisions. At some point we probably want to be able to archive a project that has been sitting for too long.

elfeto:
Would the same be helpful backwards ? If the author is waiting for X time for the "Awaiting Decision", then send an email as a reminder? 

Also, how long until emails can be sent?  1 day? 2 days ?

tompollard:
Thanks for the pull request - sorry, missed these questions.

> If the author is waiting for X time for the "Awaiting Decision", then send an email as a reminder?

Think I'd prefer not to do this initially, but if you've included it in your pull request then we can give it a try and see how it goes.

> Also, how long until emails can be sent? 1 day? 2 days ?

A week maybe?

elfeto:
So, I only included editor reminding the authors.

Then the wait time between emails right now is set for 1 day. But the editor is the person that can click the button. I can set it for a week if you want to...

tompollard:
> Then the wait time between emails right now is set for 1 day. But the editor is the person that can click the button. I can set it for a week if you want to...

As it's the editor, 1 day sounds good to me.

PR #786, opened 2019-10-18
Added reminder emails when waiting on author input

elfeto:
Here I add a reminder button that will send an email to the authors
if there is something needed from them.

This button can only be clicked once every 24 hours.
The content of the emails are the same as the emails sent initially.
#772


physionet-django/notification/utility.py
```diff
@@ -292,21 +292,21 @@ def copyedit_complete_notify(request, project, copyedit_log):
...
-                'domain': get_current_site(request),
-                'url_prefix': get_url_prefix(request),
-                'signature': email_signature(),
-                'project_info': email_project_info(project),
-                'footer': email_footer()
-            })
-
-        send_mail(subject, body, settings.DEFAULT_FROM_EMAIL,
-                  [email], fail_silently=False)
+    for person in project.author_list():
```
tompollard, line 295: please could you briefly explain the need for changing this chunk? if someone has approved, and then we reopen and edit, shouldn't they be required to approve again?

physionet-django/console/views.py
```diff
@@ -126,16 +126,38 @@ def submitted_projects(request):
 
     assign_editor_form = forms.AssignEditorForm()
 
+    # Time to check if the reminder email can be sent
+    yesterday = timezone.now() + timezone.timedelta(days=-1)
```
tompollard, line 130: doesn't really matter, but why not `timezone.now() - timezone.timedelta(days=1)`? seems clearer to me.

tompollard (commented):
Thanks Felipe, this looks great. I've listed a few thoughts below:

### Emails should be "reminder" emails

I'm happy to merge and then deal with this later, but ideally the reminder emails should not just be copies of the previous message. It would be better if they included "Reminder" in the header, and make it clear that we are reminding them in the main text of the email.

### Display "Last reminded" date

In the tables at http://localhost:8000/console/editor-home/, it would be good if we could display "Last reminded" columns, with either a date or number of days since the last reminder. Again, happy to merge and then deal with this later. 

### Rename columns in http://localhost:8000/console/editor-home/

Not directly linked to this pull request, but it might make more space in the table if you decide to add a "last reminded" column. The following column headers don't need to include "date" because they are clearly dates: 

- "Submission Date" -> "Submitted"
- "Revision Request Date" ->"Revision requested" 
- "Editor Accept Date" -> "Accepted"
- "Copyedit Completion Date " -> "Editing completed"

physionet-django/notification/utility.py
elfeto: In the previos code, the message would be sent to ALL of the authors. I converted this to work as a reminder and its initial function to notify all authors for the first time.

In the reminder part, I don't want to SPAM the author that has already approved, hence I check first if he approved, and if he did not then I continue with the email.

tompollard: Got it, thanks for explaining. The reminder email should be different from the original notification email (the reminder should make it clear that this is a reminder), but we can deal with this as a separate issue.

elfeto:
Perfect, then since "happy to merge", then I will do 2 new PR. One adding the word reminder and the other with the "last reminded date"

I think simply adding reminder on the title should suffice.

tompollard:
Thanks Felipe.

> I think simply adding reminder on the title should suffice.

I agree, definitely good enough to begin.

tompollard:
getting a migration conflict, so will need to make a new PR with the fix. i checked this before merging, but i guess my local dev branch wasn't up to date :/

elfeto:
huh ?

PR #787, opened 2019-10-18
Added validation when some reference is added.

elfeto:
Right now if a person being credentialed adds a reference category, but no information of the reference, then its passed.

Here I force that if anything in the reference category is set, then he has to fill all of the information.


elfeto:
Here are several  commits, This was due to a long conversation with TP.

We found a way to simplify the logic of the validations

tompollard:
Looks great, thanks :)

PR #790, opened 2019-10-21
More efficient authentication for wget

bemoody:
This wound up being more complicated than I expected.

These changes are an attempt to fix the performance problems we've been seeing due to bulk downloads from wget: instead of verifying the password every time (which is slow), use a cookie that can be verified quickly.

This relies on existing Django mechanisms to guarantee integrity of the session data; in fact, this works almost the same way as the standard Django login mechanism.

(The user ID is stored in a session variable, and a hash of the hash of the password is stored in a second session variable.  Thus, if the person changes their password, existing wget jobs should be interrupted.  On the other hand, if they change their username or their primary email address, existing wget jobs should now continue running.)

I considered just using the standard login mechanism (i.e., once you've supplied a valid password, just call auth.login()), but I think it is still useful to restrict HTTP-auth clients to a subset of views, so that for functions such as creating projects or requesting access, you are required to go through the standard login page.

One problem with all this is that the client might not support cookies - this is a problem because the session data is stored server-side, so we may not want to create hundreds of thousands of sessions for a single user.

To deal with that issue, I recommend that we first deploy these changes to production, then wait a day, then deploy a second change that will set the session-ID only for clients that are known to support cookies.  The reason for the delay is so that any clients *currently in the process* of downloading mimic-cxr will be assigned session-IDs.

### DIFF
--- a/physionet-django/project/test_views.py
+++ b/physionet-django/project/test_views.py
@@ -395,6 +395,7 @@ class TestAccessPublished(TestMixin, TestCase):
             secure=True,
             HTTP_USER_AGENT='Wget/1.18')
         self.assertEqual(response.status_code, 401)
+        self.client.logout()
         response = self.client.get(
             reverse('serve_published_project_file',
                     args=(project.slug, project.version, 'SHA256SUMS.txt')),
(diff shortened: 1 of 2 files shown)


elfeto:
Looks nice, have any thoughts on restricting the session lifetime? I think the default is one week.

bemoody:
It's a good question.

I'm not sure what exactly will happen if the session expires while
wget is running (if it takes a very long time to download, or if you
suspend it and try to resume days later, etc.)  Preferably, it should
fall back to username/password authentication for the first request,
and then generate a new session ID for you.  But I should test it.

I think a one-week lifetime is probably okay, but we could set it
shorter to be safe.  I don't know if there's an easy way to do that on
a per-session basis.

elfeto:
I like this PR, and for now I will merge... If there is something else form sessions can be added to another one.

PR #793, opened 2019-10-23
Python requirementes updates

elfeto:
There is a "low" security issue on Pillow for images.

"An issue was discovered in Pillow before 6.2.0. When reading specially crafted invalid image files, the library can either allocate very large amounts of memory or take an extremely long period of time to process the image."

Here I do an update to that and Django.
Also, for some reason my psycopg2-binary hash was incorrect, so I add one for OSX.


bemoody:
Looks good.  However, this raises a number of concerns about the
handling of profile images, which is something I haven't really paid
any attention to.  I'll raise a separate issue for that.

PR #799, opened 2019-10-30
Move javascript search function from console templates to file

tompollard:
Some of our console pages include a javascript search bar:
- https://physionet.org/console/users/all/
- https://physionet.org/console/news/

The templates for these pages include variations of the following function:

```javascript
<script type="text/javascript">
 function search(value){
  var csrftoken = getCookie('csrftoken');
  $.ajax({
    type: "POST",
    url: "{% url 'users_list_search' %}",
    data: {'csrfmiddlewaretoken':csrftoken,
            'search_field':value,
            {% if request.path == user_list_inactive %}
              'inactive':true,
            {% endif %}
    },
    success: function reloadSection(result){
        $("#userlist").html(result);
    },
  });
 }
</script>
```

We'd like to include the search on several other pages, so this is a first step to avoid duplication of the function in every template. This pull request:

- moves the function to a file at: https://github.com/MIT-LCP/physionet-build/blob/mv-console-search/physionet-django/static/custom/js/search-console.js

Notes:

- I ended up replacing the `badge-danger` style for inactive users count with `badge-info` in the process. The style didn't seem necessary. Hopefully not controversial.
- The user search required an argument to allow filtering on inactive users. I wasn't sure of the best way of handling this outside the javascript function (which is how it was handled before), and ended up including it in the URL. If there's a cleaner way, let me know!


elfeto:
Why put the JS on static custom, instead of console static ? You think this would be use in another place apart from the console? 

Why are you sending the arguments as an array instead of as 2 separate arguments with names that states what they are?

In your doc string you say:
 - group (str): group of users to filter search (currently only 'inactive')

But you send both "all" and "inactive"

Instead of 'group' variable, I would prefer 'item' or 'criteria', since this can be used for credentialing eventually.

tompollard:
Thanks for looking at this. 

> Why put the JS on static custom, instead of console static ? You think this would be use in another place apart from the console?

No good reason, other than I thought we might end up using it elsewhere. Now moved to console static.

> Why are you sending the arguments as an array instead of as 2 separate arguments with names that states what they are?

Erm, I don't think there is a good answer to that question. Now changed!

> In your doc string you say: group (str): group of users to filter search (currently only 'inactive'), But you send both "all" and "inactive"

Good point, I've updated the docstring to mention "all"

> Instead of 'group' variable, I would prefer 'item' or 'criteria', since this can be used for credentialing eventually.

I looked at this for a bit and decided that "group" seems clearer than "item" or "criteria" in the `users_search` view, because it makes clear that we are searching within a group of users.

When implementing the search for credentialing (e.g. adding search at http://127.0.0.1:8000/console/past-credential-applications/successful), "group" would filter for the group of people who were successfully credentialed. Can you give a bit more detail about why you prefer "item"?

lbulgarelli:
I think we can simplify the ajax by using:

`$( "#searchitems" ).load(url, data);`

Also, there's a race condition happening in all the pages using this:

The first letters typed - let's say `a` -  returns a large set of users, and therefore take longer to return results, which causes it to eventually be rendered after the other calls generated by new characters inputed by the user.

physionet-django/console/static/console/js/search-console.js
```diff
@@ -0,0 +1,16 @@
...
+         type: "POST",
+         url: url,
+         data: {'csrfmiddlewaretoken':csrftoken,
+                'search': value
+         },
+         success: function reloadSection(result){
+             $("#searchitems").html(result);
+         },
+ });
+}
```
lbulgarelli, line 8: data = {'csrfmiddlewaretoken':csrftoken,  'search': value}
$( "#searchitems" ).load(url, data);

tompollard: Thanks, Lucas - added  in https://github.com/MIT-LCP/physionet-build/pull/799/commits/feb610040a6faa3bedec5b2c05ccb78d911b4be2

tompollard:
> Also, there's a race condition happening in all the pages 

This has been bugging me for a while! I'll look into possible solutions...let me know if you find anything in the meantime.

elfeto:
Thanks tom, looks good. BTW, the possible solution is a submit button.

PR #804, opened 2019-11-07
Set the site config as physionet.org

elfeto:
Here I change the site config to set "physionet.org" the default site instead of "alpha.physionet.org"


tompollard:
thanks felipe

PR #805, opened 2019-11-07
Added a checkmark on activated inactive accounts

elfeto:
There should be an easy way to know if the account has been activated by the user.

Here I add a checkmark on the users of the inactive accounts page.
The checkmark says if the account has been activated by the link sent via email.

No checkmark means that the account hasn't been activated by the user.

### DIFF
--- a/physionet-django/console/templates/console/console_navbar.html
+++ b/physionet-django/console/templates/console/console_navbar.html
@@ -105,11 +105,12 @@
         </a>
       </li>
       <!-- users -->
-      {% url 'user_list' as user_list %}
-      {% url 'user_list_inactive' as user_list_inactive %}
-      {% url 'user_list_admin' as user_list_admin %}
+      {% url 'users' 'active' as active_user_list %}
+      {% url 'users' 'inactive' as inactive_user_list %}
+      {% url 'users' 'all' as all_user_list %}
+      {% url 'users' 'admin' as user_list_admin %}
       <li class="nav-item" data-toggle="tooltip" data-placement="right">
-      {% if request.path == user_list or request.path == user_list_inactive or request.path == user_list_admin  %}
+      {% if request.path == all_user_list or request.path == active_user_list or request.path == inactive_user_list or request.path == user_list_admin  %}
         <a id="nav_users_dropdown" class="nav-link nav-link-collapse drop" data-toggle="collapse" href="#userComponents" data-parent="#sideAccordion" aria-expanded="true">
       {% else %}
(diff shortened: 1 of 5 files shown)


bemoody:
I'm not sure why we have an "All Users" and an "Inactive Users" list
to begin with; it seems like it would make more sense to have "Active
Users" and "All Users".

But, so long as we have a list of "All Users", it makes sense to
indicate which accounts are activated.  So I'm in favor of this
change.

As a style matter, I don't think it's good for the behavior of the
template to be controlled by request.path.  Better would be for the
behavior to be controlled by a template parameter that is specified by
the view function.

elfeto:
Here I add the "Inactive users" to the console panel, and by popular vote I modify the
current definition to strictly user accounts that have not been activated via the 
link sent to the email.

I also unite all the user functions into a single function.

I am keeping the checkmark that shows if the account has been activated ONLY in the "all users" page.

bemoody:
Sorry for the delay.  This looks pretty good; a few minor things:

> console/templates/console/console_navbar.html

This isn't pretty, although I understand why it's done this way.  I'd
like to replace this with a more generic approach (enumerating the
available pages in Python) but that's a matter for another pull
request.

> console/templates/console/users.html

Here, on the other hand, I don't like the use of request.path.  Why
not make this generic, by making "group" (or "category", if you
prefer) a template parameter?

> console/templates/console/users_list.html

Please don't use <i class> as a substitute for an inline image or
glyph.  It's bad practice and needs to be rooted out throughout the
site.

Instead of '<i class="fa fa-check" aria-hidden="true"></i>',
how about '&#10003;' (✓) or '&#10004;' (✔) or even
'<span class="fa fa-check"><span class="visually-hidden">&#10004;</span></span>'
?

> physionet-django/console/urls.py

Why do you use "group" as the parameter name for views.users_search
and "category" for views.users?

> console/views.py

Please put a space after the colon in line 753.

I think you need to add the 'show_inactive' template parameter to
users_search as well.

elfeto:
I addressed all but the:
> This isn't pretty, although I understand why it's done this way.  I'd
like to replace this with a more generic approach (enumerating the
available pages in Python) but that's a matter for another pull
request.

But I will do another PR for that.

bemoody:
Great, thanks!

It looks like users_search ought to set 'group' as a template
parameter (line 784).  Otherwise, if you are looking at the "all
users" page and do a search, the checkmark column would disappear.

elfeto:
Good catch, I just fixed it

bemoody:
Looks good!

PR #807, opened 2019-11-08
Display PDF files

bemoody:
This allows displaying PDF files from active/published projects in an inline frame (assuming the browser supports it, which current firefox/chromium do.)


tompollard:
nice, thanks

PR #810, opened 2019-11-12
Citation text in project preview

bemoody:
In the project preview, rather than displaying:

    When using this resource, please cite:
    [Your citation will appear here]

which might be confusing to authors, display:

    When using this resource, please cite:
    Mark, R. G. (2019). MIT-BIH Arrhythmia Database. PhysioNet. doi:10.13026/*****

Does this seem clear enough?

### DIFF
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -210,6 +210,15 @@ class Author(BaseAuthor):
         """
         return self.user.profile.get_full_name()
 
+    def initialed_name(self):
+        """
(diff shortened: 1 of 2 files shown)


tompollard:
This seems clear to me. One minor thing that I've been meaning to fix in the suggested citation field is that DOIs should be displayed in their full URL form: 
https://support.datacite.org/docs/datacite-doi-display-guidelines

Screen Shot 2019-11-12 at 16 29 22

PR #813, opened 2019-11-13
Adding search on past credentialed applications

elfeto:
Here I add the search function on the past credential applications.
It searches on:
 - Username
 - Email
 - First Name

I just moved the population of the two tables to separate files and load them in the normal page, and in the search.

### DIFF
--- a/physionet-django/console/templates/console/past_credential_applications.html
+++ b/physionet-django/console/templates/console/past_credential_applications.html
@@ -10,6 +10,12 @@
 
 {% load console_templatetags %}
 
+{% block local_js_top %}
+<script src="{% static 'custom/js/cookie.js' %}"></script>
+<script src="{% static 'project/js/dynamic-formset.js' %}"></script>
+<script src="{% static 'console/js/search-console.js' %}"></script>
+{% endblock %}
+
 {% block content %}
 <div class="card">
   <div class="card-header">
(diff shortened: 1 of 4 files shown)


elfeto:
No big changes on that last commit, I just move some of the functionalities out of the main function to add some readability.

tompollard:
Thanks, this looks good to me. I made some minor changes as discussed:

- Updated the docstring for search_credential_applications and made some minor PEP8 style changes
- Removed the unused `status` argument from the `search_credential_applications` view
- Renamed some of the variables for clarity (e.g. `u_applications` to `unsuccessful_apps`, `l_applications` to `legacy_apps`, and `s_applications` to `successful_apps`)
- Moved a chunk from `search_credential_applications` back to `past_credential_applications`.

Does this sound okay? If so, and you are happy, please could you merge?

elfeto:
I like it and im fine... there is only one thing that was an error form the beginning.
I don't know if you care about it, once you do a search, the number of applications is keep the same. 

For example, you have 2 successful applications (admin and rgmark), and you search for rgmark, the red icon on the tab, will still list 2 applications, but there will only be one application shown.

tompollard:
> I don't know if you care about it, once you do a search, the number of applications is keep the same.

Good point. I think this is okay - if someone decides they need to see the number based on the search, we can update later.

PR #815, opened 2019-11-14
Trying to fix server 502 error

elfeto:
There was an error in which uWSGI stopped working with no errors in the log and
just stopped responding. To make it work again it was restarted, but it happened
several times after that.

This is an attempt to fix that error, adding a bit more logging, and enabling
"thunder-lock" to see if that was the cause of the problem or part of it.
Enabling thunder-lock should put a stop to the thundering herd problem if we have it.

Also, here I introduce the concept of worker recycling.
Worker recycling can prevent issues that become apparent over time such as memory
leaks or unintentional states. In some circumstances, however, it can improve
performance because newer processes have “fresh memory space”.

uWSGI provides multiple methods for recycling workers. Assuming the app is relatively
quick to reload, all three of the methods below should be effectively harmless and
provide protection against different failure scenarios.

### DIFF
--- a/deploy/production/etc/uwsgi/vassals/physionet_uwsgi.ini
+++ b/deploy/production/etc/uwsgi/vassals/physionet_uwsgi.ini
@@ -3,20 +3,25 @@
 # Django-related settings
 # the base directory
 chdir           = /physionet/physionet-build/physionet-django
+
 # Django's wsgi file
 module          = physionet.wsgi
+
 # the virtualenv
 home            = /physionet/python-env/physionet
 
 # process-related settings
 # master
 master          = true
+
 # maximum number of worker processes
 processes       = 10
+
 # the socket
 socket          = /physionet/deploy/physionet.sock
 # ... with appropriate permissions - may be needed
 chmod-socket    = 664
+
 # clear environment on exit
 vacuum          = true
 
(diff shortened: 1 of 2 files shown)


bemoody:
Can you elaborate on a few of these?

> # This will give memory usage of each process in log file at location
> memory-report = true

Where is this report written and how often?

> reload-on-rss = 2048                 ; Restart workers after this much resident memory

I assume this means the worker is restarted *after it has finished
handling a request* if it has used a lot of memory.  What is 2048?
Kilobytes?

If this limit is exceeded regularly it would be a problem - does this
trigger a log message we could look for?

> py-call-osafterfork = true			 ; Allow them to receive signals to attempt to gracefully time out requests

What does this mean?

Restarting workers automatically after N requests or after N minutes
is a fine idea.  We do need to be cautious, though, about migrations.
When we add a new column to a hot table, we need to be sure that the
migration is run before any workers are restarted with the new code.
This could be accomodated by a more careful workflow for deploying
migrations - but we should figure out what that workflow will be.

elfeto:
In the uWSGI documentation of best practices, they recommend turning this on.
https://uwsgi-docs.readthedocs.io/en/latest/ThingsToKnow.html
> memory-report = true

This will just grant more information on the log, or at least it should.
The output on the log should look something like this:
`{address space usage: 771027046400 bytes/735308MB} {rss usage: 5484544 bytes/5MB} [pid: 884|app: 0|req: 3/8] 127.0.0.1 () {42 vars in 1090 bytes} [Tue Apr  9 11:30:30 2019] GET / => generated 11 bytes in 0 msecs (HTTP/1.1 200) 2 headers in 79 bytes (1 switches on core 5)`


> reload-on-rss = 2048                 ; Restart workers after this much resident memory

is in megabytes, although 2gb might be a bit excessive, we might want to start with a smaller number and work our way up ?
Also, this will be restarted after the request and all is finished. There is also the EVIL option, which is called `evil-reload-on-rss` which will kill the worker even if is still running. I thought of adding this, but until we don't know a number on the graceful reload, I didn't want to introduce this.

>  py-call-osafterfork = true			 ; Allow them to receive signals to attempt to gracefully time out requests

From what I have read, the uWSGI workers don't receive signals properly or at least they don't receive OS signals, if we are going to kill the workers after X factors, we have to make sure they can receive the kill signal properly. Also apparently uWSGI by default doesn't execute cleanup code on threads after forking into different workers. Enabling `py-call-osafterfork` should attend this.

bemoody:
Thanks!  I'll merge this now; please be sure to test on staging before
pushing to production.

PR #816, opened 2019-11-14
Changing the project title validation

elfeto:
Here I add the "validate_title" function under the project validators.
That function forces the title to start with an alphabetical word and then
followed by a Unicode character and a specific set of special characters


tompollard:
Looks great, thanks. We talked about including the model/object name in the function name (e.g. `validate_project_title`) - it can always be added later if needed.

elfeto:
Yea, I know, I did it in the beginning, but since its in the project validators file, I thought the extra redundancy was not needed. but I can add it for the next PR of most of the validators if you want.

tompollard:
Yeah, it might be fine as it is. Let's see how it goes...depends on how many we end up needing!

Issue #791, opened 2019-10-21
Clean up old sessions


bemoody:
We ought to remove old sessions from the database (django_session) once they've expired.

From the documentation, it sounds like all we need is to run "manage.py clearsessions" in a daily cron job.  I haven't tested it.

PR #817, opened 2019-11-14
Clean up expired sessions

bemoody:
Add a cron job to remove expired sessions from the database.  Fixes issue #791.

Testing this is fun.

It's a bit annoying that, yet again, the configuration needs to be duplicated for staging and production... we should have a better way of handling that.


elfeto:
> Testing this is fun.

I agree

PR #822, opened 2019-11-19
Remove right border from indented-box class

lbulgarelli:
For some reason a border appeared in the right of the boxes in the `About` and `Share` pages, making it rather ugly. This should simply remove it.


tompollard:
The left and right bars appear in Safari, but not Chrome. In Safari, the bars look like this:

Screen Shot 2019-11-19 at 09 13 55

tompollard:
I still don't see the border (left or right) in chrome, but I get what you're doing here and it seems to fix the issue.

Issue #818, opened 2019-11-15
The search in the console is bugged


elfeto:
There is a problem with the search in the console and pagination. 
This affects, user, news and credentialing.
- The number of items of the output of the search is not altered.
- The pagination numbers in the button is kept from the initial default, and never altered.
- If one clicks a page of the pagination, the query of the search is lost.

elfeto:
The number of items of the output of the search is not altered.

But I don't see a harm on closing this.

PR #824, opened 2019-11-19
Remove paging when searching

lbulgarelli:
This is an approach to solve #818.

I remove the paging if there's a search string.

Maybe not the best solution but solves the problem.


elfeto:
I was going to add the search as a get request, or try... but its nice... If I am searching for someone, I would like it to be in the same page and not paginated to just do a "Ctrl-F" and get it fast.

If @tompollard doesn't mind, then I will merge

tompollard:
Thanks for checking, and this seems good to me from a quick look. @elfeto please go ahead and merge if you are happy!

Issue #820, opened 2019-11-18
Improve visibility of "latest version" on old versions of projects


alistairewj:
At the moment if you access an older version of a project, the top of it looks something like:

--

![image](https://user-images.githubusercontent.com/1282676/69072427-9f291d80-09f9-11ea-8ffd-b09cb2b98216.png)

--

Personally, I think this is a bit too subtle. I vote we go for a much more prominent indication that the version is outdated. I think a new div with a red background highlighting that this is an older version would work well - something like what scikit-learn have:

![image](https://user-images.githubusercontent.com/1282676/69072728-2ffff900-09fa-11ea-9578-21b5a7720f6e.png)

tompollard:
Adding Zenodo as another example:

![Screen Shot 2019-11-18 at 12 03 53](https://user-images.githubusercontent.com/822601/69073489-8e79a700-09fb-11ea-89ec-945f76c696ed.png)

tompollard:
Fixed in https://github.com/MIT-LCP/physionet-build/pull/828 by @elfeto (thanks!). cc @matthewreyna

PR #828, opened 2019-11-19
Add banner to old project versions

elfeto:
This solves #820

This will just display a simple banner saying that this is NOT the last version, and a link to the last version.


tompollard:
Looks great, thanks!

PR #830, opened 2019-11-21
Make news without a required project

### LINKED ISSUE #829: "Project" should not be a required field for new news items
It looks like "Project" is now required when adding news items at:
https://www.physionet.org/console/news/add/

The project field should be optional, not required.

elfeto:
Fixes #829


tompollard:
that was quick!

Issue #831, opened 2019-11-22
Validation for the "Research topic" field in the credentialing form is too strict


tompollard:
Related to  #811.  The validation for the "Research topic" field in the credentialing form is too strict. As a start, we should also allow:

- question mark ?
- comma ,
- colon :
- at sign @
- parentheses ()
- hash # 
- percent %
- dollar $ 
- etc

<img width="914" alt="Screen Shot 2019-11-22 at 16 51 25" src="https://user-images.githubusercontent.com/822601/69463189-0243e880-0d49-11ea-9c07-38f6281486e8.png">

tompollard:
Fixed in #832

Issue #714, opened 2019-08-29
Validation on "Reference topic" field of credentialing form is too strict


tompollard:
The validation on the "Reference topic" field of the credentialing form is too strict. Currently the field allows "Letters, numbers, spaces, underscores, hyphens, apostrophes, periods, and commas" only.

We should allow all common punctuation to avoid making things difficult for people (e.g. parentheses, exclamation, hash, percent, ampersand, double quote, colon, semi colon, plus, equals, m-rule, asterisk...).

<img width="1143" alt="Screen Shot 2019-08-29 at 23 25 09" src="https://user-images.githubusercontent.com/822601/63978618-0a2c1d80-cab7-11e9-8814-c72632a199bd.png">

Issue #515, opened 2019-05-21
Affiliation text forbids most punctuation characters


bemoody:
User's affiliation is permitted to contain only "Letters, numbers, spaces, underscores, hyphens, apostrophes, periods, and commas only. Must begin with a letter or number."

This excludes many punctation characters; in particular, I can easily imagine parentheses, slashes, and ampersands being needed here.  But I don't see any particular reason to restrict this field.

Note, however, that project.templatetags.project_templatetags.author_popover uses this field as raw HTML, so before permitting other characters in this field, that function would need to be fixed as well.

Issue #544, opened 2019-06-14
Unable to assign slugs ending in -[0-9]


alistairewj:
It looks like there is a bug in the code to prevent ending on a hyphen when assigning a slug for a project. You can't input `test-123` even though it should be valid.

slug | valid
--- | ---
`test` | yes
`test123` | yes
`test-123` | no (bug)
`test-` | no (intentional)

bemoody:
That's deliberate; we don't want 'test-123' as a project name because
it could be interpreted as version '123' of 'test'.

alistairewj:
This eliminates project slugs like `embc-challenge-2018`, which was a bit unfortunate. Doesn't slug versioning append `v1`, `v2.0`, etc?

bemoody:
Hmm.  We could require that all version numbers include a dot, and
avoid possible ambiguity that way.

(We've traditionally had databases labelled as "version 2", "version
3", etc. - but those were somewhat informal.  (There were at least two
different releases of "version 2" of the MIMIC II Waveform Database
and at least three releases of "version 3"; each release added new
records but didn't change the database structure.))

"wfdb-10.6.2" is standard practice; "wfdb-v10.6.2" looks hideous. :P

alistairewj:
That's fair! I like the idea of requiring dots; we should push semantic versioning anyway. Seems like a good compromise.

tompollard:
Currently the rule is "Version may only contain numbers and dots, and must begin and end with a number.", so we do kind of require semantic versioning. It is possible to list a single integer though (e.g. version "1"). So is the suggestion here to require at least one dot (e.g. "1.0") or at least two dots ("1.0.0")?

alistairewj:
At least one dot seems the way to go since we've already adopted that for MIMIC, eICU, etc.

PR #832, opened 2019-11-25
Removed plusplusplus validations

elfeto:
I am replacing all the "plusplus*******" validations into input specific validations.

Now slugs ending with dash numbers are accepted, this is useful for challenges.

We are adding more special characters in titles and credential research summary.

Most of the validations were kept the same as the "plusplus" for future changes.

This affects #831, #714 and #515.

When you mention fix "project.templatetags.project_templatetags.author_popover" what you mean?

This function will assume the input is clean and then display it, what would be needed to be changed?


bemoody:
> Now slugs ending with dash numbers are accepted, this is useful for
> challenges.

This is issue #544.

As I mentioned in that issue, we should also require dots in version
numbers, to avoid ambiguity (the star in validate_version should
become a plus.)

> We are adding more special characters in titles and credential research
> summary.

Can you please make a list of exactly what characters will be allowed,
in each field, that are not allowed currently?

> When you mention fix
> &quot;project.templatetags.project_templatetags.author_popover&quot; what
> you mean?
>
> This function will assume the input is clean and then display it, what would
> be needed to be changed?

author_popover (and probably many of the other functions in that file)
does not escape the inputs properly.

If your affiliation is "Texas A&M University", currently it would be
rendered as

    data-content="<b>Affiliations</b><p>Texas A&M University</p>"

but to be valid HTML, it would have to be:

    data-content="&lt;b&gt;Affiliations&lt;/b&gt;&lt;p&gt;Texas
A&amp;amp;M University&lt;/p&gt;"

Yeah, that's two levels of escaping - you need to escape the author's
affiliation string when pasting it into the HTML snippet, and then you
need to escape the HTML snippet when pasting it into the attribute.

elfeto:
> As I mentioned in that issue, we should also require dots in version
numbers, to avoid ambiguity (the star in validate_version should
become a plus.)

Done

>Can you please make a list of exactly what characters will be allowed,
in each field, that are not allowed currently?

I fixed the docstring and error messages. If you want a detailed list of all the changes let me know.

For the HTML escape, maybe this should be another PR ?

bemoody:
> I fixed the docstring and error messages. If you want a detailed list of all
> the changes let me know.

Thank you for doing that.  Documentation is important and too often
neglected. :)

However, that doesn't answer the question of what exactly has *changed*.

What is the practical impact of applying the changes you've proposed?

This should be spelled out in the git log.  If I'm reading the git log
a year from now, I should be able to see "prior to this change, titles
beginning with digits were forbidden, but after this change, titles
beginning with digits were allowed."

In this case, you're changing the validation of several fields at
once.  It's not obvious, at a glance, exactly what is permitted now
that wouldn't be permitted before, or vice versa.

So please write this out explicitly so we can see what changes you're
making, and discuss whether they might cause any problems.  I doubt
they will, but please spell it out for us.


> For the HTML escape, maybe this should be another PR ?

It's not really an unrelated issue.  It *is* a separate bug, but it's
a bug that is currently not a serious problem because of the strict
validation.

Loosening the validation is a good thing, but it also makes the other
bug more serious.  So I would like to see that fixed first.  If you'd
like, go ahead and make it a separate pull request, and we can merge
that one first. :)

bemoody:
Some specific issues about project titles:

I'm somewhat inclined *not* to allow titles beginning with digits,
because it feels unfair to put those at the top of the list (and
there's no other obvious place to put them.)

On the other hand, we *should* allow accented letters in titles,
including at the start of a title.  This leads to some further
questions:
- how do we alphabetize them?
- how do we represent them in the database (composed or decomposed)?
- should we allow non-Latin letters as well?

I've also said that I think we should avoid ASCII ', ", and - in
titles, replacing those characters with ‘, ’, “, ”, ‐, –, — (if you
must), and −, as appropriate.  (I realize " isn't currently allowed,
but I think it should be.)  Ideally, the form should fix up all of
those characters automatically so authors never know the
difference. :)  That doesn't need to be done right now, but I think
it's worth mentioning.

elfeto:
Here I am adding a verbose commit log. 

I will now think of a way to do the 
>  For the HTML escape, maybe this should be another PR ?

so that it can be merged before this.

elfeto:
Added the double escape for the affiliations. I cant find another place that would need it in the project template tags.

bemoody:
Please do not:

   import user.validators as validations

In this case, to be consistent with project/models.py I would import
the individual validators by name.  But it'd also be okay to import
the module as "validators".  Just don't import the module by a name
that is subtly different from its actual name.


Thank you for the detailed breakdown!  Looking through these in order:

> validate_name is used for the user `first_names` and `last_name`
>    r'[^\W\d_]([^\W_]|[\'\ -])*' (OLD)
>    r'[a-zA-Z]([\w\' -])+' (NEW)

> `validate_reference_name` was previously `validate_name`
>    r'[a-zA-Z][\w\',. -]+' (NEW)

People's names can absolutely begin with non-ASCII letters such as É
or Þ.

(Remember, in a Unicode pattern, "\w" means "any Unicode letter, digit,
or underscore."  That is a much broader category than [A-Za-z0-9_].)

I think that if a project author's name is written in a non-Latin
script, we should ask them to provide a transliteration (alongside the
native spelling of their name, if they would like.)  The same probably
goes for people applying for credentialing.  But, these are issues of
policy rather than mechanism - for now, let's just have the form
accept any letter.

(It is unfortunate that Python's regular expressions support only a
very limited subset of Unicode character properties - as compared to
Perl or even POSIX.  But, "any string of Unicode letters" should be
good enough for now - he says with his ugly American hat on - and we
can improve this later.)

@tompollard, does DataCite, or the DOI system, have a policy regarding
the spelling of authors' names?

> validate_alphaplus was used to validate the topic and some published
> affiliations
>    r'[\w][\w\ -]*' (OLD)

> `validate_topic` was previously `validate_alphaplus`
>    r'[a-zA-Z\d][\w -]*' (NEW)

Probably OK; it doesn't make much difference.

> `validate_affiliation` (for published affiliations only) was
> previously `validate_alphaplus`
>    r'[a-zA-Z][\w\',()/&. -]+' or '..' (NEW)

Same-ish concerns as for authors' names, though it's probably a less
sensitive issue.

Thanks for pointing out the discrepancy between active/published
affiliations - that could have caused a major headache!

> validate_alphaplusplus was used in ALL other fields including affiliations
>    r'[\w][\'\,\.\w\ -]*' (OLD)

> Changed fields from `validate_alphaplusplus`:
>  - validate_research_summary    r'[a-zA-Z][\w\'!@#$%&*()+={};:"<>?,./` -]+'

So, you don't like tildes or brackets?  :P

Actually, this field is a textarea, so it really *should* accept \n
(and \r, do we get \r from textareas?)... is there any reason not to?

I would say "no control characters except \n, \r, and \t, and must
contain at least one non-whitespace character".  Is there a reason to
make it more complicated than that?

>  - validate_course              r'[a-zA-Z0-9][\w\',. -]+'
>  - validate_reference_title     r'[a-zA-Z][\w\',. -]+'
>  - validate_reference_response  r'[a-zA-Z][\w\',. -]+'
>  - validate_training_course     r'[a-zA-Z][\w\',. -]+'
>  - validate_location            r'[a-zA-Z][\w\',. -]+'
>  - validate_organization        r'[a-zA-Z][\w\',. -]+'
>  - validate_job_title           r'[a-zA-Z][\w\',. -]+'
>  - validate_city                r'[a-zA-Z][\w\',. -]+'
>  - validate_state               r'[a-zA-Z][\w\',. -]+'

All of these should probably allow non-ASCII letters at the start.

>  - validate_zipcode             r'[a-zA-Z][\w\',. -]+'

ZIP codes start with a digit!  So, clearly, this isn't being tested
and needs a test case.

(Apparently loaddata doesn't do any validation.  That'd be a quick way
to detect a lot of potential problems!  Maybe I'll try to add that as
a feature to loaddemo.)

https://en.wikipedia.org/wiki/Postal_code#Presentation tells me that
postal codes worldwide include only ASCII letters, digits, spaces, and
hyphens, so [A-Za-z0-9-][A-Za-z0-9- ]* ought to do it.

> suffix was a mix of `validate_alphaplusplus` and `validate_nan` now
> replace by `validate_suffix`
> First validation is explained above and the second `validate_nan`is
> the following:
> r'[0-9\-+()]*'
> Marked as invalid if it was formed only of numbers, hyphens, plus,
> parenthesis.
>
> That was merged into `validate_suffix` adding the following:
> r'[a-zA-Z0-9][\w\',. -]+'
> Now it forces to start with an alphabetical character or digits,
> followed by word characters, digits, underscores, apostrophes,
> comas, dots, spaces and hyphens.

Looks good.

> `validate_title` had minor escape changes but kept the same.
>    r'[a-zA-Z][\w \,\-\'\*\?\:\(\)]+'
>    r'[a-zA-Z][\w\'*,?:( )-]+'

OK; we can improve this later.

elfeto:
I believe the only thing I am missing from the comments would be the tests cases in loaddemo.

bemoody:
Thanks for your work on this!

bemoody:
The following should be done to fix up version numbers:

for p in project.models.ActiveProject.objects.all():
    if p.version.isdigit():
         print('%s: %s -> %s' % (p.slug, p.version, p.version + '.0'))
         p.version += '.0'
         p.save(update_fields=('version',))

tompollard:
I ran this on staging:

```python
from project.models import ActiveProject

for p in ActiveProject.objects.all():
    if p.version.isdigit():
         print('%s: %s -> %s' % (p.slug, p.version, p.version + '.0'))
         p.version += '.0'
         p.save(update_fields=('version',))
```

..which outputs:

```
zOBgn601PvGu6vBir3qD: 102118 -> 102118.0
VYZVJIvxVH6K7oSz0Ljh: 1 -> 1.0
IUnpNrDEkdirow1DDcT3: 1 -> 1.0
rj3hXPzJU3LiZX8lgIf8: 1 -> 1.0
hwKE3E4idfeTGTAy2KUu: 1 -> 1.0
qcS7gOOFkAvjTbAXBPYC: 1 -> 1.0
3mJhsf8lyjJqPKBEl1Qw: 1 -> 1.0
```

I'll do some testing and will move to production if things seem okay.

tompollard:
It looks like the new validation rules do not allow a project title to begin with a number. Is this intentional? Added as new issue at https://github.com/MIT-LCP/physionet-build/issues/850

Issue #611, opened 2019-07-23
Histories in inconsistent order


alistairewj:
Reverse chronological order:

* The submission section for an activate project

Chronological order:

* The submission history section for a rejected database

Reverse chronological order seems best.

elfeto:
@tompollard  ?

tompollard:
There is still an issue with the ordering. Try looking at the history for a page and hitting refresh multiple times. Thanks to https://github.com/MIT-LCP/physionet-build/pull/603 the sequence of events jumps around.

alistairewj:
There is an issue around the ordering of the history. I'm reviewing a project that has been through 2 review cycles, and the order on the "edit submission" page is as follows:

Date | Event
--- | ---
Nov 23, 2019 | Editor requested resubmission (third)
Nov 8, 2019 | Project resubmitted for review (second)
Nov 1, 2019 | Editor requested resubmission (second)
Sep 9, 2019 | Project resubmitted for review (first)
Currently | Waiting for editor decision
Nov 25, 2019 | Project resubmitted for review (third)
Aug 24, 2019 | Editor requested resubmission (first)
Aug 19, 2019 | Project assigned
Aug 19, 2019 | Project submitted

As you can see currently and the third/final resubmit for review late in the order.

PR #834, opened 2019-11-27
Display submission history in correct order

bemoody:
This should fix the order that submission histories are displayed, in:

- `/projects/*/submission/`
- `/console/submitted-projects/*/edit/`
- `/projects/rejected/*/submission-history/`
- `/projects/published/*/*/submission-history/`

This should fix issue #611.

This is quite a mess - look at how many special cases there are in active_submission_timeline.html, not to mention static_submission_timeline.html.  So, no guarantees that the order won't get screwed up again. :/

### DIFF
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -551,10 +551,10 @@ class Metadata(models.Model):
         """
         authors, author_emails = self.get_author_info(include_emails=include_emails)
         storage_info = self.get_storage_info(force_calculate=force_calculate)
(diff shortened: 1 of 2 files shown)


tompollard:
Thanks @bemoody, this looks good to me, and it fixes the issue as far as I've been able to test it.

> This is quite a mess - look at how many special cases there are in active_submission_timeline.html, not to mention static_submission_timeline.html.

Yep, the templates are super confusing right now.

Issue #814, opened 2019-11-14
Performance tweaks


bemoody:
Some small things that might improve performance:

* Blackhole /favicon.ico

Spammers keep hitting this URL (which doesn't exist) and getting a 400
(probably because they specify an unknown Host.)  These also clutter
the logs with "writing to a closed pipe/socket/fd" so I'd like to
filter them out even if it doesn't do anything for performance.

* Increase proxy_buffers

If I understand correctly, "an upstream response is buffered to a
temporary file" means that nginx tries to store the whole HTTP
response in memory (so that uwsgi can go off and do something else)
and it has run out of memory in which to do so, so it stores the
response to a file instead; and increasing the number and/or size of
"proxy_buffers" might help.

PR #837, opened 2019-12-09
Fixing the favicon.ico Blackhole

elfeto:
The best way I could find performance-wise to fix this was to alter the
Nginx config and handle it there.

The other option is via Django, but since its a static file, it makes more sense this way.

This is part of #814


bemoody:
That's not a "black hole", but this works too. :)

Something in the back of my mind is telling me that very old web
browsers would unconditionally request "/favicon.ico" rather than
looking for a proper <link>.  So, probably not a bad idea all around.

PR #838, opened 2019-12-09
Increased the proxy_buffers

elfeto:
Increasing the number of proxy_buffers can allow you to buffer more information.
I am swapping 8 buffers of 4k to 24 buffers of 4k.

Since the proxy_buffer_size holds the header, I am making it smaller.

Second part of #814

### DIFF
--- a/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
+++ b/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
@@ -119,6 +119,10 @@ server {
     # Security headers
     add_header Strict-Transport-Security "max-age=2592000; includeSubDomains; preload";
     add_header Accept-Ranges bytes;
+
+    # Proxy buffers
+    proxy_buffers 24 4k;
+    proxy_buffer_size 1k;
 }
 
 server {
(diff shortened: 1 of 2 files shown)


bemoody:
Hmm.  This change to proxy_buffer_size doesn't appear to do anything.
The limit for response headers from uwsgi is still 4k.

It looks like for uwsgi, you need to use 'uwsgi_buffer_size' rather
than 'proxy_buffer_size', and presumably likewise for the other
options.

But, I'm not sure it does any good to set the header size limit that
low.  The uwsgi server is far more of a bottleneck than a paltry 3k of
RAM, and if that limit is exceeded then nginx gives a 502 error.
Response headers currently seem to be smaller than 1k - max from
recent logs is 597 - but it's a little close for comfort.


To generate a big response header:

--- a/physionet-django/physionet/views.py
+++ b/physionet-django/physionet/views.py
@@ -23,10 +23,13 @@ def home(request):
     latest = PublishedProject.objects.all().order_by('-publish_datetime')[:6]
     news_pieces = News.objects.all().order_by('-publish_datetime')[:5]

-    return render(request, 'home.html', {'featured': featured,
+    resp = render(request, 'home.html', {'featured': featured,
                                          'latest': latest,
                                          'news_pieces': news_pieces})

+    for i in range(36):
+        resp['X-Blah%d' % i] = 'x' * 100
+    return resp

 def about_publish(request):
     """

bemoody:
Closing as I don't think this is needed at present.

Issue #761, opened 2019-09-25
Need to handle existing email address in /register/


bemoody:
The simple case: someone tries to register, then tries to register again with the same email address (either because the message hasn't been delivered yet, or they deleted it or their spam filter deleted it, or whatever.)

In this case, they should just see the same message as before: "An email has been sent to {{ email }}. Please click on the confirmation link to activate your account."

The email probably *shouldn't* be sent again if it is still "in flight" or if too little time has elapsed.  (Obviously I'm handwaving over some tricky issues here...)  If the message *is* sent again, the verification token should be the same.

The complicated case is where somebody else tried to use your email address.  Perhaps the easiest way to address this would be:
- *remove* the password fields from the initial registration form
- *add* a form to the 'activate_user' page, allowing them to view and edit their username and profile, and requiring them to supply a password, before their account can be activated

Perhaps there should also be an option to say "I didn't request an account and have no interest in creating one; please don't contact me again."

Similar problems can apply to 'add_email'.

elfeto:
If somebody else tried to use an email that is not theirs, then the account will never be activated. Since the account is not activated they cant reset the password, and the account is "locked" until they email us. I think this would rarely happen. 
With add emails, the email will never be activated, and "locked".

Why not do a weekly cleanup, for the accounts/emails that haven't been activated in a week just delete them?

As for the existing email, we could in the registration form do a check if the account was created in the previos seconds or minutes. I think this happens because it was clicked fairly fast. Since the registration form will show if the email/username are taken.

bemoody:
I have an email address that is my real name on a popular free mail
service.  I frequently receive emails due to other people thinking my
address is theirs, and entering it into web registration forms.  It's
not at all a rare occurrence.

> If somebody else tried to use an email that is not theirs, then the
> account will never be activated.

That's not true: it is entirely possible that the person who *did*
receive the confirmation message will click on the link (perhaps
hoping that it will lead to a form where they can say "this isn't my
account, please stop sending me email").

Of course, if that person *does* click on the link, they will discover
that they've just "activated" somebody else's account (whatever that
means), and are given no further options.  If they have a lot of free
time, they might dig around to try to find a way to report the
problem.  Most likely they won't.  (Remember, they don't know or care
what PhysioNet is.  They probably don't know the person who created
the account.)

I think the general principle is that it shouldn't be possible to
activate an account, or add an email address, without verifying who
you are.  If you are activating a new account, you should have to
[...]

Issue #833, opened 2019-11-27
ActiveProject has no attribute 'integrity_errors'


bemoody:
* Create a project and submit it for publication

* Ctrl-R

* Browser says "this was a POST, are you sure?"; say yes

* AttributeError: 'ActiveProject' object has no attribute 'integrity_errors'

tompollard:
```
Traceback:

File "/Users/tompollard/projects/physionet-build/env/lib/python3.8/site-packages/django/core/handlers/exception.py" in inner
  34.             response = get_response(request)

File "/Users/tompollard/projects/physionet-build/env/lib/python3.8/site-packages/django/core/handlers/base.py" in _get_response
  115.                 response = self.process_exception_by_middleware(e, request)

File "/Users/tompollard/projects/physionet-build/env/lib/python3.8/site-packages/django/core/handlers/base.py" in _get_response
  113.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/Users/tompollard/projects/physionet-build/physionet-django/project/views.py" in view_wrapper
  120.                 return base_view(request, *args, **kwargs)

File "/Users/tompollard/projects/physionet-build/physionet-django/project/views.py" in project_submission
  1206.             elif project.integrity_errors:

Exception Type: AttributeError at /projects/bogtNdgj8VnkQLwR6ZwN/submission/
Exception Value: 'ActiveProject' object has no attribute 'integrity_errors'
Request information:
USER: admin
```

PR #840, opened 2019-12-12
Adding integrity error for account registration

elfeto:
Fixing integrity errors on account registrations.

This doesn't fix the issues #761 and #833, but its a start.

I created a template that I believe should be used for ALL integrity errors under the user model, and handle them this way.

I am doing this initial PR to get your opinions.

### DIFF
--- a/physionet-django/notification/utility.py
+++ b/physionet-django/notification/utility.py
@@ -10,6 +10,7 @@ from django.core.mail import EmailMessage, send_mail, mail_admins
 from django.template import loader
 from django.utils import timezone
 
+
 from project.models import DataAccessRequest, License
 
 RESPONSE_ACTIONS = {0:'rejected', 1:'accepted'}
@@ -706,3 +707,22 @@ def task_rescheduled_notify(name, attempts, last_error, date_time, task_name, ta
         })
     subject = name + " has been rescheduled"
     mail_admins(subject, body, settings.DEFAULT_FROM_EMAIL)
+
+
+def notify_account_registration(request, user, uidb64, token):
+    """
+    Send the registration email.
+    """
+    # Send an email with the activation link
+    subject = "PhysioNet Account Activation"
+    context = {
+        'name': user.get_full_name(),
+        'domain': get_current_site(request),
+        'url_prefix': get_url_prefix(request),
+        'uidb64': uidb64,
+        'token': token
+    }
+    body = loader.render_to_string('user/email/register_email.html', context)
+    # Not resend the email if there was an integrity error
+    send_mail(subject, body, settings.DEFAULT_FROM_EMAIL,
+              [user.email], fail_silently=False)
--- a/physionet-django/user/forms.py
+++ b/physionet-django/user/forms.py
@@ -218,9 +218,11 @@ class RegistrationForm(forms.ModelForm):
         return self.cleaned_data['username'].lower()
 
     def save(self):
-        # Save the provided password in hashed format
-
-        if self.errors: return
+        """
+        Process the registration form
+        """
+        if self.errors:
+            return
 
         user = super(RegistrationForm, self).save(commit=False)
         user.email = user.email.lower()
@@ -228,7 +230,7 @@ class RegistrationForm(forms.ModelForm):
         with transaction.atomic():
             user.save()
             # Save additional fields in Profile model
-            profile = Profile.objects.create(user=user,
+            Profile.objects.create(user=user,
                 first_names=self.cleaned_data['first_names'],
                 last_name=self.cleaned_data['last_name'])
             return user
--- a/physionet-django/user/views.py
+++ b/physionet-django/user/views.py
@@ -31,7 +31,7 @@ from physionet import utility
 from project.models import Author, License, PublishedProject
 from notification.utility import (process_credential_complete,
                                   credential_application_request,
-                                  get_url_prefix)
+                                  get_url_prefix, notify_account_registration)
 
 
 logger = logging.getLogger(__name__)
@@ -311,43 +311,22 @@ def register(request):
 
     if request.method == 'POST':
         form = forms.RegistrationForm(request.POST)
-        saved = False
-        integrity_catch = False
         if form.is_valid():
             # Create the new user
             try:
                 user = form.save()
-                saved = True
+                uidb64 = force_text(urlsafe_base64_encode(force_bytes(
+                    user.pk)))
+                token = default_token_generator.make_token(user)
+                notify_account_registration(request, user, uidb64, token)
             except IntegrityError:
                 form.full_clean()
                 if form.is_valid():
                     raise
-                integrity_catch = True
-            if saved or integrity_catch:
-                if user.is_anonymous and integrity_catch:
-                    user = User.objects.get(username=form.data['username'])
+                user = User.objects.get(username=form.data['username'])
 
-                if not integrity_catch:
-                    # Send an email with the activation link
-                    uidb64 = force_text(urlsafe_base64_encode(force_bytes(
-                        user.pk)))
-                    token = default_token_generator.make_token(user)
-                    subject = "PhysioNet Account Activation"
-                    context = {
-                        'name': user.get_full_name(),
-                        'domain': get_current_site(request),
-                        'url_prefix': get_url_prefix(request),
-                        'uidb64': uidb64,
-                        'token': token
-                    }
-                    body = loader.render_to_string(
-                        'user/email/register_email.html', context)
-                    # Not resend the email if there was an integrity error
-                    send_mail(subject, body, settings.DEFAULT_FROM_EMAIL,
-                        [user.email], fail_silently=False)
             return render(request, 'user/register_done.html', {
                 'email': user.email})
-
     else:
         form = forms.RegistrationForm()
 
--- a/physionet-django/user/views.py
+++ b/physionet-django/user/views.py
@@ -12,6 +12,7 @@ from django.contrib.auth.tokens import default_token_generator
 from django.contrib.sites.shortcuts import get_current_site
 from django.core.mail import send_mail
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
+from django.db import IntegrityError
 from django.forms import inlineformset_factory, HiddenInput, CheckboxInput
 from django.http import HttpResponse, Http404, HttpResponseRedirect
 from django.shortcuts import redirect, render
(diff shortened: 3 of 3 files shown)


bemoody:
I would say that "integrity errors" should not happen.

Or, more specifically: catch *specific* errors that can be triggered
by remote actions (such as trying to create the same username twice.)

We do not want to catch arbitrary errors from the database.  If, let's
say, Django mistakenly tries to assign a primary key that is already
in use, that should raise an error and be logged.

That's just a general statement of principle.  As to how to actually
*implement it*, well...

- we could use get_or_create (appears to have TOCTOU problems, yuck)
  https://docs.djangoproject.com/en/dev/ref/models/querysets/#get-or-create

- we could dig into the exception details (also yuck)

elfeto:
The problem with the "get-or-create" is that since this is a model form with several things being saved at the same time a get or create would be a bit complicated.

bemoody:
What exactly is the problem you're trying to solve here?

The ModelForm *does* check uniqueness, or tries to.  If that is broken
then we should fix it, not paper over it.

elfeto:
There is an integrity error that happens when someone tries to register for an account and double clicks the "Register" button.

Both requests are handled and at the time they are handled there is a check for uniqueness and they are both unique at that point in time. Since one request is executed first, AFTER the uniqueness check, the DB is changed, and then the second request comes and it fails with an IntegrityError.

I cannot think of any other way of fixing this, rather than with a "try: & except IntegrityError:"

bemoody:
The first thing we need is a way to reproduce the issue.  Can you
think of a way to do that in a test case?

ISTM one way to fix it would be to wrap "form.is_valid" + "form.save"
in transaction.atomic.  However, I don't fully understand what that
would do.

* It would presumably mean that if Alice is trying to register an
account, and Bob is also trying to register an account, Bob must wait
for the transaction to finish.

* Would it also mean that if Alice is trying to register an account,
and Bob is trying to change his password, Bob must wait for the
transaction to finish?

* Would it also mean that if Alice is trying to register an account,
and Bob is trying to *log in*, Bob must wait for the transaction to
finish?

bemoody:
Okay, here's another, possibly not so great, idea.

What if we did something like:

    saved = False
    if form.is_valid():
        try:
            form.save()
            saved = True
        except IntegrityError:
            if form.is_valid():
                raise
    if saved:
        # do more stuff
    else:
        # do other stuff

Not so efficient on a per-process level since you might have to run
some checks twice, and lack of locking means it's not completely
foolproof (the person whose name you clashed with might have changed
their name in between the "save" and the second "is_valid") - but it
may be worth it to avoid any locking on these tables.  What do you
think?

If we did something like that, I would suggest it should be wrapped up
into an abstract form class that could be used in multiple places (at
least in user there are many *potential* places for these sorts of
problems to occur.)

elfeto:
To reproduce is quite easy... Just register for an account and double click (fast) the registration button.
For the first one, if we lock the table with the atomic, then for all user related it will have to wait for each registration. All three statements are correct. 

I dislike the second one, it adds a layer of complexity, but might work tho...

What about in the user registration form, the save section do check if an account exists
```

        users_exists = User.objects.filter(Q(username=user.username) |
            Q(email=user.email))

        if not users_exists:
            with transaction.atomic():
                user.save()
                # Save additional fields in Profile model
                profile = Profile.objects.create(user=user,
                    first_names=self.cleaned_data['first_names'],
                    last_name=self.cleaned_data['last_name'])
        else:
            return users_exists.get()
```
That will just send the activation email twice.

bemoody:
Well, I see a couple of practical problems with that:
- since the check doesn't happen within the "atomic", it's still racy
- you're not checking AssociatedEmails

and it seems to me very messy and fragile to have *two* different
places that are supposed to be validating uniqueness, and *two*
different error code paths.

But suppose we did that.  Who would receive the second activation
email, and what would happen to the first one?

- Alice tries to register username 'alice', email address
  'alice@example.com', password 'Tester11!', at the same time Bob
  tries to register username 'bob', email address 'alice@example.com',
  password 'Tester22!'.

- Alice tries to register username 'alice', email address
  'alice@example.com', password 'Tester11!', at the same time Bob
  tries to register username 'alice', email address 'bob@example.com',
  password 'Tester22!'.

In each of these cases, what emails are sent, and what happens if
either or both of the respective activation links are clicked?

bemoody:
test script
----
#!/usr/bin/perl -w

use strict;
use LWP::UserAgent;

my $HOST = 'http://localhost:8000';

my $ua1 = LWP::UserAgent->new;
# $ua1->add_handler("request_send",  sub { shift->dump; return });
# $ua1->add_handler("response_done", sub { shift->dump; return });
$ua1->cookie_jar({});
$_ = $ua1->get("$HOST/register/")->decoded_content;
my ($token1) = /name="csrfmiddlewaretoken" value="([^"]*)"/;

my $ua2 = LWP::UserAgent->new;
$ua2->cookie_jar({});
$_ = $ua2->get("$HOST/register/")->decoded_content;
my ($token2) = /name="csrfmiddlewaretoken" value="([^"]*)"/;

if (fork) {
  $_ = $ua1->post("$HOST/register/",
                  { csrfmiddlewaretoken => $token1,
                    username => 'aaaa',
                    email => 'aaaa@example.com',
                    first_names => 'Aaaa',
                    last_name => 'Aaaa',
                    password1 => 'Tester11!',
                    password2 => 'Tester11!' });
  print $_->status_line, "\n";
}
else {
  $_ = $ua2->post("$HOST/register/",
                  { csrfmiddlewaretoken => $token2,
                    username => 'aaaa',
                    email => 'aaaa@example.com',
                    first_names => 'Aaaa',
                    last_name => 'Aaaa',
                    password1 => 'Tester11!',
                    password2 => 'Tester11!' });
  wait;
  print $_->status_line, "\n";
}
----

elfeto:
Then what about this now. 

I moved the search to inside the atomic transaction.
Now i search for 2 things. If the exact parameters of the registration are being re-submited and if the email already exists....

If the account doesn't exists nor the AssociatedEmail, then create it. 

If the account with the exact same parameters exists, then return it

If the account doesn't exists but the email has been taken, then raise the error?

bemoody:
If I'm reading this right:

(a) If you submit two requests with the same
    username+email+firstname+lastname:

  - very quickly: the second request will send you to a blank
    form with no error message (and the second password will be
    ignored.)

  - not so quickly: the second request will show "User with this
    email already exists" + "A user with that username already
    exists".

(b) If you submit two requests with the same username and different
    email addresses:

  - very quickly: the second request will raise an IntegrityError.

  - not so quickly: the second request will show "A user with
    that username already exists".

(c) If you submit two requests with different usernames and the
    same email address:

  - very quickly: the second request will raise a
    RuntimeError.  (You can't "raise" if you don't have an
    exception to raise.)

  - not so quickly: the second request will show "User with this
    email already exists".


It seems to me that you are trying to do two things at once:
- you're trying to fix a race condition
- you're trying to add special handling for duplicate submissions

Try to figure out how to fix the race condition *without*
changing the existing behavior (i.e. if you submit the same thing
twice in rapid succession, you should see an error just as if you
waited a few seconds.)

*Or* find a way to handle duplicate submissions in a more
friendly way (i.e. if you submit the same thing twice after a few
seconds' delay, show "Thank you for registering" instead of an
error message.)

Don't try to do both of these things at once.


>     saved = False
>     if form.is_valid():
>         try:
>             form.save()
>             saved = True
>         except IntegrityError:
>             if form.is_valid():
>                 raise
>     if saved:
>         # do more stuff
>     else:
>         # do other stuff

In this example, I missed that you also need to call
'form.full_clean()' before calling 'is_valid' the second time.

elfeto:
Here I am going with your approach of the try catch integrity error.

elfeto:
I simplified the commits.
Let me know if this works better or?

elfeto:
Ping to @bemoody  :)

bemoody:
It looks like it should work, but I'm a little confused.  The way this
is written seems more complicated than it needs to be.

Why do you write "if saved or integrity_catch"?  It looks like those
are the only two possibilities - either saved is true and
integrity_catch is false, or saved is false and integrity_catch is
true.  What is the point of this if statement?  What is the point of
having two separate variables in the first place?

Why do you write "if user.is_anonymous and integrity_catch"?  If
integrity_catch is true then user was not assigned in the preceding
block (so you want to instead fetch the conflicting user object.)  The
existing "user" variable must in fact be anonymous, but that's
unrelated.

elfeto:
Now is simpler...
If it's saved, then send the email. If there is an integrity error it means that the user exist, so grab the user to display the proper page.

I removed the variables and unneeded if statements.

bemoody:
Thanks.  I like that this makes the register function shorter and
easier to read.

But now you're catching IntegrityError outside of form.save, which I
think is not a good idea.  If you want to avoid using a variable you
could do "try: / X / except IntegrityError: / Y / else: / Z" although
I'm not a huge fan of that syntax either.

"account_activation_email" is not the best name for that function.
It's a function that Does Something (sends an email); there should be
an imperative verb somewhere in there. :)  The other functions in that
module are... not entirely consistent, but most are either
SOMETHING_notify or notify_SOMETHING.

elfeto:
> "account_activation_email" is not the best name for that function.

Changed.

> But now you're catching IntegrityError outside of form.save, which I
think is not a good idea.  If you want to avoid using a variable you
could do "try: / X / except IntegrityError: / Y / else: / Z" although
I'm not a huge fan of that syntax either.

If I set the try/except inside the save function of the form, then 2 emails will be sent. (if you dont mind, then I can do that. 
```
        try:
            with transaction.atomic():
                user.save()
                # Save additional fields in Profile model
                Profile.objects.create(user=user,
                    first_names=self.cleaned_data['first_names'],
                    last_name=self.cleaned_data['last_name'])
                return user
        except IntegrityError:
            return User.objects.get(username=user.username)
        else:
            raise Exception("Something")
```
If we don't want to send the email 2 times, then the only thing I can think of is the way I did it to set the try/except before the save function is called.
```
            try:
                user = form.save()
                uidb64 = force_text(urlsafe_base64_encode(force_bytes(
                    user.pk)))
                token = default_token_generator.make_token(user)
                notify_account_registration(request, user, uidb64, token)
            except IntegrityError:
                form.full_clean()
                if form.is_valid():
                    raise
                user = User.objects.get(username=form.data['username'])
```
PS. I haven't been able to get the error again. Not sure if this PR is worth it.

bemoody:
> If I set the try/except inside the save function of the form, then 2 emails
> will be sent. (if you dont mind, then I can do that.

I understand.  I think it's better not to send two emails in the
common case where somebody just clicked twice and the form takes a
long time for some reason.

I think you could in fact send the email within RegistrationForm.save
(after the transaction is finished), but I don't like that so much
because then you would be catching IntegrityErrors (if there were any)
that were raised while sending the email, not just IntegrityErrors
raised by the saving of the form data.

Though there's definitely an argument to be made for taking the logic
out of the view functions and into the forms!

> If we don't want to send the email 2 times, then the only thing I can think
> of is the way I did it to set the try/except before the save function is
> called.

You could do it with a temporary variable as you did it in
909763b0f9d68738a0d2bd7ca94fa1822e99f4f1:

try:
    user = form.save()
    saved = True
except IntegrityError:
    form.full_clean()
    if form.is_valid():
        raise
    user = User.objects.get(username=form.data['username'])
    saved = False
if saved:
    uidb64 = force_text(urlsafe_base64_encode(force_bytes(
        user.pk)))
    token = default_token_generator.make_token(user)
    notify_account_registration(request, user, uidb64, token)

Or this would be equivalent, but I think it's less clear:

try:
    user = form.save()
except IntegrityError:
    form.full_clean()
    if form.is_valid():
        raise
    user = User.objects.get(username=form.data['username'])
else:
    uidb64 = force_text(urlsafe_base64_encode(force_bytes(
        user.pk)))
    token = default_token_generator.make_token(user)
    notify_account_registration(request, user, uidb64, token)

elfeto:
What would be the diference in
```
try:
    user = form.save()
except IntegrityError:
    form.full_clean()
    if form.is_valid():
        raise
    user = User.objects.get(username=form.data['username'])
else:
    uidb64 = force_text(urlsafe_base64_encode(force_bytes(
        user.pk)))
    token = default_token_generator.make_token(user)
    notify_account_registration(request, user, uidb64, token)
```
vs 
```
try:
    user = form.save()
    uidb64 = force_text(urlsafe_base64_encode(force_bytes(
        user.pk)))
    token = default_token_generator.make_token(user)
    notify_account_registration(request, user, uidb64, token)
except IntegrityError:
    form.full_clean()
    if form.is_valid():
        raise
    user = User.objects.get(username=form.data['username'])
```
Once the save is done, if there is an integrity error, it till occur before/after the email is sent. (I think)

bemoody:
The difference is that if force_bytes or urlsafe_base64_encode or
force_text or make_token or notify_account_registration raises an
IntegrityError, the first will not catch it and the second will.

(Presumably, if form.save() finishes without an exception, then
subsequently form.is_valid() will be false, because now the user
exists.)

I don't like the fact that Django doesn't make it easy to catch
particular constraint violations.  I don't know of a good way to solve
that problem, but we have to deal with it.  This means we should both
double-check the inputs (as you're doing with form.full_clean() /
form.is_valid()), and also keep the 'try' body as small as possible.

elfeto:
The following functions will not raise an `IntegrityError` because they don't touch the database.
 - force_bytes
 - urlsafe_base64_encode
 - force_text
 - make_token
 - notify_account_registration

Either way, I used your except/else method (I kinda like it)

bemoody:
Thanks for your work on this and humoring my nitpicking!

PR #844, opened 2019-12-16
Pinned news

elfeto:
Simple way to have a single news banner in the home page.

#480


tompollard:
I have tried to simplify this a little. @bemoody, please could you take a look and merge if you are happy? I've briefly outlined my updates below:

- Replaced the extra table with a banner column at: http://localhost:8000/console/news/
- Changed the nav bar icon back to fa-newspaper (the menu item had been copied from `nav_editor_home` so was showing a book)
- Removed the "global banner" code because this hasn't been implemented as far as I can see.
- Removed the code that restricted the banner to one item of news (this was preventing the user from updating a news item that was displayed as a banner, and it may be helpful to display multiple items anyway).

A couple of notes:

- the style of the banner could be improved, but it seems fine for now.
- the "banner" field on the form for editing news is awkward (for me, a small checkbox is displayed in the center of the page). This can be fixed later.

Hope you don't mind me adding to the PR Felipe. I'm traveling at the moment and it seemed quickest to make a few changes directly, as I know Roger is keen to have this done.

elfeto:
It would be nice if the commits had a proper meaning to each one and not just one with several different changes. 

I believe we had this conversation before, but I strongly believe you should only have 1 "front page banner" and in the need one global site banner for updates or maintenance.
Hence the "global banner" was there to do less database changes, and the restrictions were also there for a reason. 

Now I find the front page banner a little awkward. The way you left the banners are dumped there in the same alert without any separation, which looks like an incomplete banner item.

Also, the banners are now not dismissible, which I think was useful, not sure was an intended change since was not listed in the changes made.

bemoody:
This looks good, but I have some small nits to pick in
templates/home.html:

* news.title should not be marked "safe"

* indentation is a little weird: you have the opening "<li>" on a line
  by itself but the "</li>" tucked at the end of the following line

* html comment isn't needed

* extraneous blank line added after <div class="main-header">

* don't use inline style attributes; instead, add a class such as
  "front-page-banner" to the div and add rules to physionet.css (or
  home.css)

tompollard:
good points, thanks benjamin. i think they should be addressed, but let me know if i missed anything.

bemoody:
Looks good.  Thanks!

PR #846, opened 2020-01-06
Display only latest version on Latest Resources

elfeto:
Right now if there are two versions of the same project published at a similar time, then they might end up in the "Latest Resources" section of the index page. 

Here I only allow the latest versions for each publications.


tompollard:
Looks good, thanks Felipe

PR #852, opened 2020-01-17
More robust server upgrades

bemoody:
These are some changes to hopefully improve the workflow for
deploying upgrades to the server.

(I wrote most of this a few weeks ago but haven't touched it since.
Further testing is needed, and there may be some corner cases that
need to be addressed, which I don't have time to work on right now -
but I thought it would be useful to put this up as-is, as a preview of
possible things to come!)

Key changes here:

- Migrations will be applied automatically.

- "Late" migrations (those that should be applied after upgrading,
  rather than before) will need to be indicated using a custom
  attribute of the Migration class.

- Servers will be restarted/reloaded automatically.

- Much less verbose output from 'git push'; details will be logged in
  /physionet/update.log.

### DIFF
--- a/deploy/common/etc/sudoers.d/physionet
+++ b/deploy/common/etc/sudoers.d/physionet
@@ -1,0 +1,18 @@
+#### PhysioNet privileged operations ####
+
+# This file defines specific commands that may be invoked by automated
+# scripts using 'sudo'.  Each command listed in this file must have a
+# single, clearly-defined function.  The program must be carefully
+# written to check all of its inputs to avoid privilege escalation.
+# Do not add commands to this list without carefully considering the
+# consequences.
+
+Defaults:pn env_reset, !env_keep, !env_check
+
+# Reload emperor.uwsgi.service (if it is running)
+pn ALL=(root:root) NOPASSWD: \
+    /bin/systemctl reload emperor.uwsgi.service
+
+# Restart django-background-tasks.service (if it is running)
+pn ALL=(root:root) NOPASSWD: \
+    /bin/systemctl try-restart django-background-tasks.service
--- a/deploy/post-receive
+++ b/deploy/post-receive
@@ -2,36 +2,113 @@
 
 set -e
 
-destination=/physionet/physionet-build/
+install_dir=/physionet/physionet-build
+working_dir=/physionet/physionet-build.new
+venv_dir=/physionet/python-env/physionet
+log_file=/physionet/update.log
 branch=$(cat /physionet/deploy-branch) # production or staging
 
-echo "Deploying $branch into $destination"
-GIT_WORK_TREE=$destination git checkout --force $branch
+while read -r old new refname; do
+    if [ "$refname" = "refs/heads/$branch" ]; then
+        oldrev=$old
+        newrev=$new
+    fi
+done
+if [ -z "$oldrev" ] || [ -z "$newrev" ]; then
+    exit
+fi
 
-echo "Updating content"
-export DJANGO_SETTINGS_MODULE=physionet.settings.$branch
-cd $destination
-source /physionet/python-env/physionet/bin/activate
-pip install -r requirements.txt
-cd physionet-django
-python manage.py collectstatic --noinput
+################################################################
 
-echo "Checking system configuration files"
+# If the following commands fail, it is too late to abort the push.
+# Try to avoid doing things that might fail!
+
+echo "$(date '+%F %T %z'): $branch: post-receive started" >> $log_file
 exec 3>&1
+exec &> >(tee -a $log_file)
+
+export DJANGO_SETTINGS_MODULE=physionet.settings.$branch
+
+. $venv_dir/bin/activate
+
+# Unpack files into $working_dir
+rm -rf $working_dir
+mkdir -p $working_dir
+git archive "$newrev" | tar -x -C $working_dir
+ln -s $install_dir/.env $working_dir/.env
+
+# Check what migrations should be applied
+(
+    cd $working_dir/physionet-django
+    ./manage.py getmigrationtargets --current > OLD-TARGETS
+    ./manage.py getmigrationtargets --early -v2 > EARLY-TARGETS 2>> $log_file
+    ./manage.py getmigrationtargets > LATE-TARGETS
+)
+
+# Run early migrations
+(
+    cd $working_dir/physionet-django
+    if cmp -s OLD-TARGETS EARLY-TARGETS; then
+        echo "- No early migrations to apply."
+    else
+        echo "* Applying early migrations..."
+        ./manage.py migratetargets EARLY-TARGETS --no-input |
+            tee -a $log_file | grep 'Applying' >&3
+        echo >> $log_file
+    fi
+)
+
+echo "* Installing new static files..."
+(
+    cd $working_dir/physionet-django
+    ./manage.py collectstatic --no-input >> $log_file
+)
+
+echo "* Installing new server code..."
+GIT_WORK_TREE=$install_dir git checkout --force "$branch" 2>> $log_file
+
+echo "* Reloading uwsgi server..."
+sudo systemctl reload emperor.uwsgi.service
+(
+    cd $working_dir/physionet-django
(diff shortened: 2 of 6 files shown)


elfeto:
I have been going through this for some time, but there is no better test than staging.

The only comment I can think of is to set the `log_file` to 'log_file=/data/log/update' or something under /data/log since we moved the logs there.

bemoody:
> The only comment I can think of is to set the `log_file` to
> 'log_file=/data/log/update' or something under /data/log since we moved the
> logs there.

Good call.  I'll fix that.

Thanks for reviewing.  What I'd still like to do is test how it
behaves when things are broken, e.g. you try to perform a migration
that makes an existing non-unique field unique or something like that.

And I should take the time to update the docs :/

bemoody:
I'd like it if we could deploy these changes in the near future.  If it's okay with everyone, we could try installing this on staging and testing by deploying pull #897.  Hopefully everything goes smoothly, but if not, we go back to the old method.


Somewhat independent of the deployment scripts themselves, I've added a script "test-upgrade.sh" to test whether an upgrade is "safe", in a highly pedantic sense - it checks that the test suite works both before and after each stage of the upgrade.

(This script is *not* run by the Shippable CI tests, though perhaps it could be in the future.)

Satisfying this condition requires a bit more diligence than the way we've handled migrations up to now, and personally I think it'd be a good idea... but I realize it may be asking a lot.  So I'd like to know what you all think.

Here is a simple example of a patch that would not be considered safe:

```
--- /dev/null
+++ b/physionet-django/user/migrations/0024_profile_favorite_color.py
@@ -0,0 +1,10 @@
+from django.db import migrations, models
+class Migration(migrations.Migration):
+    dependencies = [('user', '0023_auto_20191125_1500')]
+    operations = [
+        migrations.AddField(
+            model_name='profile',
+            name='favorite_color',
+            field=models.CharField(default='blue', max_length=50),
+        ),
+    ]
--- a/physionet-django/user/models.py
+++ b/physionet-django/user/models.py
@@ -530,6 +530,8 @@ class Profile(models.Model):
         validators=[FileExtensionValidator(['png', 'jpg', 'jpeg'],
         'Allowed filetypes are png and jpg only.')])
 
+    favorite_color = models.CharField(max_length=50, default='blue')
+
     MAX_PHOTO_SIZE = 2 * 1024 ** 2
 
     # Where all the users' files are kept
```

To make it safe, it can be rewritten as follows:

```
--- /dev/null
+++ b/physionet-django/user/migrations/0024_profile_favorite_color.py
@@ -0,0 +1,10 @@
+from django.db import migrations, models
+class Migration(migrations.Migration):
+    dependencies = [('user', '0023_auto_20191125_1500')]
+    operations = [
+        migrations.AddField(
+            model_name='profile',
+            name='favorite_color',
+            field=models.CharField(default='blue', max_length=50, null=True),
+        ),
+    ]
--- /dev/null
+++ b/physionet-django/user/migrations/0025_profile_favorite_color_2.py
@@ -0,0 +1,11 @@
+from django.db import migrations, models
+class Migration(migrations.Migration):
+    dependencies = [('user', '0024_profile_favorite_color')]
+    operations = [
+        migrations.AlterField(
+            model_name='profile',
+            name='favorite_color',
+            field=models.CharField(default='blue', max_length=50),
+        ),
+    ]
+    MIGRATE_AFTER_INSTALL = True
--- a/physionet-django/user/models.py
+++ b/physionet-django/user/models.py
@@ -530,6 +530,8 @@ class Profile(models.Model):
         validators=[FileExtensionValidator(['png', 'jpg', 'jpeg'],
         'Allowed filetypes are png and jpg only.')])
 
+    favorite_color = models.CharField(max_length=50, default='blue')
+
     MAX_PHOTO_SIZE = 2 * 1024 ** 2
 
     # Where all the users' files are kept
```

bemoody:
ping @elfeto - what are your thoughts about testing this as mentioned
in my previous comment?

elfeto:
Sorry just saw this because of the ping. For some reason I never got the email of the comment and commits.

I saw the comments and it all looks good to try to test it and begging with merge you mention.

I went into staging and I altered the post-received and the update scripts, alongside the sudoers.

Then locally, I merged your branch onto my latest staging then pushed to staging.
This was the output:
```
Enumerating objects: 1042, done.
Counting objects: 100% (814/814), done.
Delta compression using up to 8 threads
Compressing objects: 100% (415/415), done.
Writing objects: 100% (712/712), 112.49 KiB | 7.03 MiB/s, done.
Total 712 (delta 541), reused 337 (delta 254)
remote: Resolving deltas: 100% (541/541), completed with 59 local objects.
hint: The 'hooks/update' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
remote: CommandError: Some migrations that have already been applied (django_cron.0001_initial, django_cron.0002_remove_max_length_from_CronJobLog_message) are not present in the source tree.
To 192.168.11.101:/physionet/physionet-build.git
   e2ea38d4..b8127ab7  staging -> staging
```

Then I tried checking out your script. Can you provide with an example of the script? 

For example, I blindly tried several arguments but all were failed with the same error:
```
./test-upgrade.sh -v 
./test-upgrade.sh staging 
./test-upgrade.sh better-push 
./test-upgrade.sh physionet.settings.staging 
./test-upgrade.sh staging/staging 
./test-upgrade.sh origin/staging 

Unsupported DJANGO_SETTINGS_MODULE=physionet.settings.staging
```
Please let me know what I am doing wrong in the testing of this.

Also, if I wanted to run this manually like one can do right now the ./post-receive (since there is no update) What would be the order? first run post-receive then update ?

bemoody:
Ah, good questions!  Sorry to complicate things by adding the
test-upgrade script into the mix.

The test-upgrade script should *not* be run on the live database
on production or staging (it will wipe out anything currently in
the database!)  That's why I put that check in. :)

If you want to try out the test-upgrade script, locally, then one
way to do so is:

- check out the branch you want to test (e.g. 'git checkout
  origin/dev')

- merge this branch without committing ('git merge --no-commit
  origin/better-push')

- run './test-upgrade.sh origin/production' (and yes, at present
  some tests will fail for the production-to-dev case!)

- 'git merge --abort' to undo the temporary merge


As for the git hooks:

I haven't thought about running the git hooks by hand.  You
probably could, but you'd need to call them correctly, the way
git does.

But of course this comes back to the question of how to clean up
when something breaks.  Most of these installation steps should
be idempotent... but if something breaks in installation then
presumably it needs to be fixed, and pushing the same thing again
won't help.

django_cron thing, duly noted.  Presumably we just want to get
rid of any django_cron tables/migrations that exist at present.
But it's true, I didn't think at all about the possibility of
removing an app that was formerly enabled.

bemoody:
Oh, and:

> hint: The 'hooks/update' hook was ignored because it's not set as executable

...yes, it needs to be executable. ;)

elfeto:
To fix the issue there, I re-added the app in the settings file, then I undid the migration by hand.

Uncommented
```
    # 'django_cron',
```

Ran:
```
(physionet) pn@physionet-staging:~/physionet-django$ ./manage.py migrate --fake django_cron zero
Operations to perform:
  Unapply all migrations: django_cron
Running migrations:
  Rendering model states... DONE
  Unapplying django_cron.0002_remove_max_length_from_CronJobLog_message... FAKED
  Unapplying django_cron.0001_initial... FAKED
```

Then the error went away.
Locally I merged `better-push` to `staging` with the files added with the correct permissions (executable :P ) I pushed, and it worked.

Then I deleted the branches locally, got a clean `better-push` and `staging`, merged `better-push` to my latest dev, rebased dev on staging, and pushed to see all the changes from #897.

That PR had a migration and it worked fine.

elfeto:
BTW, I cannot run the script because my dev has differences than the system itself, and it gave me some errors on paths, so I ignored it.

elfeto:
As an FYI for the future... Django apps added by pip install have the migration in the python package... not in the django app.

For example one of the migrations of the error above is here:
> /physionet/python-env/physionet/lib/python3.7/site-packages/django_cron/migrations/0002_remove_max_length_from_CronJobLog_message.py

PR #853, opened 2020-01-21
Database index updates

bemoody:
Add various recently-published databases to /about/database/, and
a little bit of reorganization and style cleanup.

### DIFF
--- a/physionet-django/templates/about/database_content.html
+++ b/physionet-django/templates/about/database_content.html
@@ -154,6 +154,12 @@ alphabetically by the name of the database.
           ongoing studies.
       </li>
     </ul></li>
+
+  <li>[Class 3] *<a href="{% url 'published_project_latest' 'picdb' %}">Paediatric Intensive Care Database</a>.
+    The Paediatric Intensive Care (PIC) database is a large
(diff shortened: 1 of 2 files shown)


tompollard:
Thanks, looks good from a quick review. Before I merge, what is the `*` that appears before some database titles? e.g.:

```py
<li>[Class 3] *<a href="{% url 'published_project_latest' 'picdb' %}">
Paediatric Intensive Care Database</a>.
```

bemoody:
The stars were meant to indicate credentialing required, but I agree
that isn't clear.  I think there was a footnote that got lost
somewhere along the way.  Should we get rid of them or should we add
some text to clarify?

tompollard:
Oh okay, that makes sense. Perhaps we remove for now, and then come up with a clearer way of indicating permissions/licenses later? e.g. we could display each entry in a similar style to https://physionet.org/content/ (but more compact).

bemoody:
I thought about trying to replace these, and the "[Class 1]" - "[Class
3]" labels, with little color-coded icons, but I wasn't really happy
with how it looked.  I'm just going to remove the stars for now.

PR #855, opened 2020-01-23
DOI automatic assignment

elfeto:
Here I add the automatic assignment of DOI on publish.

There is still some work to be done with this, but it works.
The things to add are:
 - Handling tests (Right now tests are omitting DOI assignment.)
 - If the publish page is refreshed, one will create another DOI, and it will end up in a duplicate DRAFT.
 - Delete draft DOI if not used

We can add more information onto the DOI later if needed. 

The contact email in DataCite should be changed to something under PhysioNet, and maybe the account password also.

For documentation on how to use the API, I used: https://support.datacite.org/docs/api


elfeto:
I cant seem to find it, but where in shippable one sets variables and passwords ?

tompollard:
Nice, thanks for doing this. Looks like tests are passing, so does this mean the question about variables/passwords in Shippable is resolved?

bemoody:
Well, this "secret" information is obviously no longer secret.  What
is it used for?

Are there test cases here that are accessing DataCite services?

elfeto:
@tompollard  No, not resolved, I skipped it.
@bemoody That is how they say to do it -> http://docs.shippable.com/platform/tutorial/security/encrypt-vars/
But I don't like it, and it didn't worked.

There are NO tests, they are skipped at the moment like I wrote earlier, I am omitting tests in this PR.  I can add them to the PR tho. (have to create them, and alter a lot of publishing)

tompollard:
> But I don't like it, and it didn't worked.

Even if we could get it to work, I assume that people aren't comfortable with trusting Shippable to keep the login details safe? It's okay not to have a live test of the system I think, and this can be annoying when the system is down anyway. Simulating the API for tests seems fine.

bemoody:
Who has the username/password for 'https://api.datacite.org/dois'?
How do we generate that password and how do we change it?

Who has the username/password for
'https://api.test.datacite.org/dois'?  How do we generate that
password and how do we change it?

If you supplied any passwords to shippable or committed them into git,
please be sure that those passwords are revoked.

In principle it'd be good to have tests that run against
api.test.datacite.org, if it's possible to do that safely and without
causing a lot of load to those folks.  Barring that, it would be good
to test by mocking the HTTP requests (which you could do using
https://github.com/bemoody/islandhack :P, but maybe Python/Django has
a better way.)

elfeto:
@tompollard  and I have the username and password, for both. 

To change the password it seems simple through the web browser.

All NON production execution (dev, test and staging) are using the test API.

bemoody:
Sorry to belabor the point, but:

- Are you saying that *the same* username and password must be used
  for both services, or is it possible to have a separate account that
  is *only* permitted to access the "test" service?

- Information about how to change the password should be documented
  with a comment in ".env" (or wherever we ultimately decide to store
  this information.)

In particular, include information about what exactly the password is
used for (what specific services/sites it can be used to access), how
to change it (where is the web site you mentioned?), when it was last
updated, and any side effects of changing it (does it render existing
data inaccessible? does it need to be updated in other places?)

(This goes for everything that is stored in ".env".)

elfeto:
Before, BOTH accounts had the same user/pass. 
Now each one has its own password.

If you try to use the credentials for one in the other it won't work, the same as if you use the wrong prefix.

I did 2 commits, because I saw ur comments after I did the first one.

bemoody:
Thanks, that's helpful!


Don't change behavior based on Sites.  You can use Sites to determine
a canonical URL for a given page.  Don't use them for anything else.

You should have *one* variable for the DataCite API URL, and *one*
variable for the DOI prefix.  These variables will have different
values in the production and staging configurations.  But the hostname
of the server has nothing to do with which API URL you will use.

Instead of storing "DATACITE_KEY" as a single base64-encoded string,
better to have one variable for the username and a second variable for
the password.  (Not any more or less secure - just easier to
understand and to update when necessary.)  Then pass these to
requests.put using the "auth" argument.


IIUC, the point of "draft" DOI is that we can provide limited metadata
and reserve a DOI for a database that is not yet ready to be
published.  This information can then be updated prior to publication,
and finally once the project is actually published, the DOI status and
URL are changed.

So, when we create a draft DOI, it should be saved as an attribute of
the ActiveProject.  We might want to do this at any stage of
publication after the project is accepted - not necessarily when the
editor is ready to actually publish the project.

Reloading the page (e.g. /console/submitted-projects/foo/publish/)
should *not* cause a new DOI to be created; in fact, a DOI should only
be created in response to a form submission (a POST, not a GET
request), and it should be saved to the database immediately so that
it doesn't get lost.

When the project actually *is* published, and the DOI is changed from
draft to published status, the available metadata (title, authors, and
so on) also needs to be updated.  We might also want the ability to
update this information multiple times prior to publication.

elfeto:
I get your point (that is why I wrote the things to do, I just wanted to provide some foot work to get input and continue), and the draft can be deleted without any problem whatsoever. 

Since I am unaware of the actual point of a "draft" I used the DataCite documentation, https://support.datacite.org/docs/doi-states. Since it says it can be made with no information, I assumed it could be used for this.

> We might also want the ability to
update this information multiple times prior to publication.

Since at the point of "publishing" (after copyedit)  ALL the information should be correct, then the only thing to edit would be the URL and to publish the draft.

bemoody:
> Since I am unaware of the actual point of a "draft" I used the DataCite
> documentation, https://support.datacite.org/docs/doi-states. Since it says
> it can be made with no information, I assumed it could be used for this.

Makes sense to me, and it sounds like maybe you can't create a DOI in
"findable" state to begin with - it has to first be created as a
draft, and then published as a separate action?

> Since at the point of "publishing" (after copyedit)  ALL the information
> should be correct, then the only thing to edit would be the URL and to
> publish the draft.

If you assume that "publish_doa_draft" is always called immediately
after "create_doa_draft", with exactly the same project and DOI,
that's true.

However, we have cases where the author would like to have a DOI
assigned before the project is published (so they can refer to it in a
paper that is published at the same time.)  It's important that the
"draft" DOI we give to the author is in fact exactly the same DOI that
will be assigned at publication time.  So it would be good for this to
be tracked by the PhysioNet site, and not something that the editor
has to remember to do by hand.

tompollard:
My understanding is that the main purpose of the "draft" state is to reserve a DOI in the system, so that it can't be taken by someone else (which is unlikely to happen at the moment, but might become more likely if the frequency of publication increases).

bemoody:
Can I (a random unauthorized individual) view the metadata of a draft DOI?

tompollard:
> Can I (a random unauthorized individual) view the metadata of a draft DOI?

It's a good question. My guess is no, but I can't say for sure. "Draft DOIs" and "Registered DOIs" are not discoverable in the DataCite Metadata Store, but I don't know for certain that there isn't an alternative mechanism for discovery.

elfeto:
Up tp my knowledge while playing with this, the answer is no.

elfeto:
I added to the model of unpublished projects the DOI object, to once I get the DOI for the project, store it there, and if the page is refreshed, then no other DOI will be requested. This might also help for the tests, assuming I discover how to use credentials for shippable.

I also separated the key into user/pass and I am using the auth in the request instead of sending it in the header.

physionet-django/console/utility.py
```diff
@@ -152,7 +158,7 @@ def create_directory_service(user_email):
     Returns:
       Admin SDK directory service object.
     """
-    logging.getLogger('googleapiclient.discovery_cache').setLevel(logging.ERROR)
+    logging.getLOGGER('googleapiclient.discovery_cache').setLevel(logging.ERROR)
```
tompollard: Looks like search and replace hit this by accident?

physionet-django/console/utility.py
```diff
@@ -171,3 +177,120 @@ def paginate(request, to_paginate, maximun):
     paginator = Paginator(to_paginate, maximun)
     paginated = paginator.get_page(page)
     return paginated
+
+
+def create_doa_draft(project):
```
tompollard: should this be `create_doi_draft` ?

physionet-django/console/views.py
```diff
@@ -476,25 +477,32 @@ def publish_submission(request, project_slug, *args, **kwargs):
...
-            published_project = project.publish(
-                doi=publish_form.cleaned_data['doi'],
-                slug=slug,
+            published_project = project.publish(slug=slug,
                 make_zip=int(publish_form.cleaned_data['make_zip']))
+
+            current_site = Site.objects.get_current()
+            url = 'https://{0}/content/{1}/{2}'.format(current_site, slug, project.version)
+            if utility.publish_doa_draft(url, project.doi):
+                messages.success(request, 'Succesfully created DOI.')
```
tompollard: Succesfully -> Successfully

physionet-django/console/utility.py
```diff
@@ -171,3 +177,120 @@ def paginate(request, to_paginate, maximun):
...
+    current_site = Site.objects.get_current()
+    if settings.PRODUCTION:
+        prefix = settings.DATACITE_PREFIX
+        url = settings.DATACITE_URL
+        username = settings.DATACITE_USER
+        password = settings.DATACITE_PASS
+
+    headers = {'Content-Type': 'application/vnd.api+json'}
+
+    Dataset = ['Database', 'Challenge', 'Model']
```
tompollard: Is there a way to avoid hardcoding the project type here? The project `Metadata` class has a `schema_org_resource_type()` method to return the schema.org resource type. Could we do something similar here? e.g. modify the `schema_org_resource_type()` method to `resource_type()` and require `schema` as an argument?

physionet-django/console/utility.py
```diff
@@ -171,3 +177,120 @@ def paginate(request, to_paginate, maximun):
...
+            "descriptionType": "Abstract"
+          }],
+        }
+      }
+    }
+
+    response = post(url, data=json.dumps(payload), headers=headers,
+        auth=HTTPBasicAuth(username, password))
+    if response.status_code < 200 or response.status_code >= 300:
+        raise Exception("There was an unkown error submitting the DOI, here is \
```
tompollard: unkown -> unknown

physionet-django/console/utility.py
```diff
@@ -171,3 +177,120 @@ def paginate(request, to_paginate, maximun):
...
+            the response text: {}".format(response.text))
+
+    content = json.loads(response.text)
+    LOGGER.info("DOI draft for project {0} was created with DOI: {1}.".format(
+        project.title, content['data']['id']))
+    return content['data']['id']
+
+def publish_doa_draft(project_url, doi):
```
tompollard: `publish_doa_draft` -> `publish_doi_draft`

physionet-django/console/utility.py
```diff
@@ -171,3 +177,120 @@ def paginate(request, to_paginate, maximun):
...
+    """
+    Upate a DOI from draft to publish.
+
+    A PUT request is done in order to update the information od the DOI.
```
tompollard: "od the DOI" -> "for the DOI"?

physionet-django/console/utility.py
```diff
@@ -171,3 +177,120 @@ def paginate(request, to_paginate, maximun):
...
+          "state": "publish",
+          "event": "publish",
+          "url": project_url
+        }
+      }
+    }
+    response = put(url, data=json.dumps(payload), headers=headers,
+        auth=HTTPBasicAuth(username, password))
+    if response.status_code < 200 or response.status_code >= 300:
+        raise Exception("There was an unkown error updating the DOI, here is \
```
tompollard: "unkown" - "unknown"

physionet-django/console/utility.py
tompollard: Are the caps needed in `Dataset` and `Software`? For consistency, perhaps all lower case.

physionet-django/console/views.py
```diff
@@ -476,25 +477,32 @@ def publish_submission(request, project_slug, *args, **kwargs):
...
             notification.publish_notify(request, published_project)
             return render(request, 'console/publish_complete.html',
                 {'published_project':published_project})
 
     publishable = project.is_publishable()
     url_prefix = notification.get_url_prefix(request)
+    if project.doi == '':
+        project.doi = utility.create_doa_draft(project)
```
tompollard: create_doa_draft -> create_doi_draft

physionet-django/console/utility.py
```diff
@@ -171,3 +177,120 @@ def paginate(request, to_paginate, maximun):
...
+
+def create_doa_draft(project):
+    """
+    Create a draft DOI with some basic information about the project.
+
+    A POST request is done to set the base information.
+    The assigned DOI is returned to be used in the template.
+    """
+    url = settings.DATACITE_TEST_URL
+    if url == False:
```
tompollard: Instead of `if url == False:` -> `if not url:`

physionet-django/console/utility.py
```diff
@@ -171,3 +177,120 @@ def paginate(request, to_paginate, maximun):
...
+        auth=HTTPBasicAuth(username, password))
+    if response.status_code < 200 or response.status_code >= 300:
+        raise Exception("There was an unkown error submitting the DOI, here is \
+            the response text: {}".format(response.text))
+
+    content = json.loads(response.text)
+    LOGGER.info("DOI draft for project {0} was created with DOI: {1}.".format(
+        project.title, content['data']['id']))
+    return content['data']['id']
+
```
tompollard, line 348: nitpick, but PEP8 suggests two breaks here

physionet-django/console/utility.py
```diff
@@ -171,3 +177,120 @@ def paginate(request, to_paginate, maximun):
...
+def publish_doa_draft(project_url, doi):
+    """
+    Upate a DOI from draft to publish.
+
+    A PUT request is done in order to update the information od the DOI.
+    The URL is made using the prefix and doi assigned to the project.
+    """
+    url = settings.DATACITE_TEST_URL
+    if url == False:
```
tompollard: Instead of `if url == False:` -> `if not url:`

physionet-django/console/utility.py
```diff
@@ -171,3 +177,120 @@ def paginate(request, to_paginate, maximun):
...
+          "publisher": current_site.name,
+          "publicationYear": timezone.now().year,
+
+          "types": {
+            "resourceTypeGeneral": resource_type
+          },
+          "url": 'https://{}'.format(current_site.domain),
+          "descriptions": [
+          {
+            "description": description,
```
tompollard: As we are assigning each version a new DOI, we should populate the version property (ID 15) in the DataCite metadata. See page 26 of https://schema.datacite.org/meta/kernel-4.3/doc/DataCite-MetadataKernel_v4.3.pdf

> Suggested practice: track
major_version.minor_version.
Register a new identifier for a
major version change.
Individual stewards need to
determine which are major
vs. minor versions27
.
Software engineering practice
follows this approach of
tracking changes and giving
new version numbers.
May be used in conjunction
with properties 11 and 12

tompollard (commented):
Thanks Felipe, this looks good. I've added some comments inline. Mostly minor changes, so it seems more than it is. Just for reference, an example of the metadata on the test DataCite system is below:

```json
{
  "id": "https://doi.org/10.7966/j7fx-wj46",
  "doi": "10.7966/J7FX-WJ46",
  "url": "https://localhost:8000/content/1234/1.0.0",
  "types": {
    "resourceTypeGeneral": "Dataset"
  },
  "creators": [
    {
      "name": "Roger Greenwood Mark"
    }
  ],
  "titles": [
    {
      "title": "MIT-BIH Arrhythmia Database"
    }
  ],
  "publisher": "PhysioNet-dev",
  "publicationYear": 2020,
  "descriptions": [
    {
      "description": "The MIT-BIH Arrhythmia Database contains 48 half-hour excerpts of two-channel ambulatory ECG recordings, obtained from 47 subjects studied by the BIH Arrhythmia Laboratory between 1975 and 1979. Twenty-three recordings were chosen at random from a set of 4000 24-hour ambulatory ECG recordings collected from a mixed population of inpatients (about 60%) and outpatients (about 40%) at Boston&#39;s Beth Israel Hospital; the remaining 25 recordings were selected from the same set to include less common but clinically significant arrhythmias that would not be well-represented in a small random sample.",
      "descriptionType": "Abstract"
    }
  ],
  "providerId": "physiont",
  "clientId": "physiont.physiont",
  "agency": "DataCite",
  "state": "findable"
}
```

In the past I have been assigning the same DOI across versions (even though this has been populating the version specific DOI field in the metadata).

As this update introduces version-specific DOIs, I think it's fairly important that we introduce the latest-version DOI functionality soon after this is merged. Otherwise we can expect some fuss!

physionet-django/console/utility.py
elfeto: There are two DataCite resource types we sue when registering a DOI, Dataset and Software. 

If you are ok saying that EVERYTHING that DOES NOT have software as its resource type is a Dataset, then yea. I can even remove the variables.

The example you mention, either I don't understand it or doesn't make sense.

tompollard: > The example you mention, either I don't understand it or doesn't make sense.

The general idea is that we shouldn't be hardcoding mappings between metadata profiles. I'll look into this in a little more detail when I have time.

elfeto: This is just following the schema from DataCite. (https://support.datacite.org/docs/schema-40)

bemoody:
Can you please explain why you'd write something like this:

    .env:
        DATACITE_URL=https://api.datacite.org/dois
        DATACITE_TEST_URL=https://api.test.datacite.org/dois

    physionet/settings/base.py:
        DATACITE_URL = config('DATACITE_URL', default=False)
        DATACITE_TEST_URL = config('DATACITE_TEST_URL', default=False)
        PRODUCTION = False

    physionet/settings/production.py:
        PRODUCTION = True

    console/utility.py:
        url = settings.DATACITE_TEST_URL
        if settings.PRODUCTION:
            url = settings.DATACITE_URL

instead of writing:

    physionet/settings/base.py:
        DATACITE_API_URL = 'https://api.test.datacite.org/dois'

    physionet/settings/production.py:
        DATACITE_API_URL = 'https://api.datacite.org/dois'

    console/utility.py:
        url = settings.DATACITE_API_URL

Please explain this to me.  What is the advantage?  I honestly do not
understand.

physionet-django/console/utility.py
tompollard: This is a variable called `Dataset`. Seems unusual to use the capital letter.

elfeto:
@bemoody There is none. Just adding the words tests to the variable so that its known that we are working with the test information.

The same can be done with the user/password and prefix.

physionet-django/console/utility.py
elfeto: I understand your feelings in hardcoding things, At some point you have to write something to differentiate between data and software.

If you say that ALL non software as resource type, then that can be simplified and less hardcodes mappings.

elfeto: mmm, you meant those, I would say then consistency.

elfeto:
@bemoody Also the `PRODUCTION = True` is a way to differentiate between anything and production which can also be used for GCP.

physionet-django/console/utility.py
elfeto: I can set the version, but for some ODD reason it doesn't show in the test site once I add it.

Also, there are 3 variables that could be used for version:
 - version
 - metadataVersion
 - schemaVersion

I added all 3, and only "schemaVersion" is shown.
I can keep them all if you want it.

elfeto: As an example of the DataCite JSON output of a DOI...
```
{
  "id": "https://doi.org/XX.XXXX/XXX-XXX",
  "doi": "XX.XXXX/XXX-XXX",
  "url": "https://localhost:8000",
  "types": {
    "resourceTypeGeneral": "Dataset"
  },
  "creators": [
    {
      "name": "PERSON LASTNAME"
    }
  ],
  "titles": [
    {
      "title": "project.title"
    }
  ],
  "publisher": "PhysioNet-dev",
  "publicationYear": 2020,
  "descriptions": [
    {
      "description": "description",
      "descriptionType": "Abstract"
    }
  ],
  "schemaVersion": "2.3",
  "providerId": "physiont",
  "clientId": "physiont.physiont",
  "agency": "DataCite",
  "state": "draft"
}
```

elfeto: Since I am using the API reference only the things here can be sent. 
https://support.datacite.org/reference/dois-2#post_dois

ALSO, there is NO version in the online form to register a DOI.

elfeto: Just in case, only set the following:
```
          "version": project.version,
          "metadataVersion": project.version,
          "schemaVersion": project.version,
```

tompollard: `schemaVersion` and `metadataVersion` refer to the version of the metadata schema, so they are not relevant here and can be left empty. The field that we would like to populate is `version` (which has an ID of 15 in the schema).

elfeto: So the example above is the actual output if I set
```
          "version": project.version,
          "metadataVersion": project.version,
          "schemaVersion": project.version,
```

If I only set the version, I see NO version addition to the DOI. It's sent, but ignored by the API apparently.

elfeto: If you log into https://doi.test.datacite.org/, and check the draft DOI that ends with "g1hr-e130" you will see the word "Version" but its empty. 
That DOI was created with:
```
          "version": project.version,
```

bemoody:
Thanks for making these improvements, Felipe!

Some (mostly minor) comments:

> .env.example

DATACITE_URL and DATACITE_TEST_URL are not used, so please remove
them.

> physionet-django/console/publish_submission.html

Stray "{{datacite_is_enabled}}" ?

> physionet-django/console/utility.py

I still think it would be preferable to separate the functions of
"creating draft DOI" from "updating DOI with the current project
metadata".  And possibly the latter should be done automatically as
part of "publishing".

Please document what the return values of these functions mean.  It
looks like 'create_doi_draft' returns "" if disabled, or a DOI
(e.g. "10.13026/vc0e-th79") if succesful.  It looks like
'publish_doi_draft' returns "" if disabled, or None if enabled, which
seems odd.

The name "publish_doi_draft" is a bit odd - once this function is
called, it's no longer a "draft".

Existing DOIs (e.g. https://api.datacite.org/dois/10.13026/vc0e-th79)
have detailed metadata about each author:

    {
     "name": "Lee, Meredith",
     "nameType": "Personal",
     "givenName": "Meredith",
     "familyName": "Lee",
     "affiliation": [
      "Stanford University"
     ],
     "nameIdentifiers": []
    },

I don't know what "nameType" or "nameIdentifiers" mean exactly, but at
least "givenName" and "familyName" should be filled in.

What format should "name" be in?  All other things being equal, I'd
think 'author.get_full_name()' would be preferred, but if other
programs are going to try to parse it in silly ways then "Last, First"
might be better.  @tompollard, when you create a DOI by hand, do you
input the name as "Last, First" or is that done automatically from two
input fields?

IIRC, "url" is not required when creating a draft DOI, so should be
omitted (rather than using a dummy URL.)

What format should "description" be in?  "strip_tags" will kill tags
but is a poor substitute for actually converting to plain text, if
"plain text" is what is desired.

In publish_doi_draft:

        "id": doi.split('/')[1],

is setting off a lot of alarm bells.  Is that REALLY the correct
syntax to use?  What happens to the prefix?  Is the "id" actually
meaningful when posting to an existing DOI?

> physionet-django/console/views.py

The project URL should probably be determined by calling 'reverse'
rather than constructing it here.  (This is also incorrect, strictly
speaking, as it's missing the trailing slash.)

The 'messages.success' is not needed here.

I like that you invoke 'project.publish' before invoking
'publish_doi_draft', because if 'publish_doi_draft' fails, we do not
want to be left in a weird transient state (and if 'project.publish'
fails, we do not want to publish a DOI that is inaccurate.)  However,
I think 'publish_notify' should also be invoked before
'publish_doi_draft'.

> physionet-django/physionet/settings/base.py
> physionet-django/physionet/settings/production.py

Nothing should ever depend on "settings.PRODUCTION", just as nothing
should ever depend on whether "current_site == production_site" or
"'production' in settings.SETTINGS_MODULE" or any other way of saying
the same thing.  Please don't do it!

> physionet-django/physionet/settings/development/pgsql.py
> physionet-django/physionet/settings/development/sqlite.py

DATACITE_API_URL should be set the same way in both configurations.

> physionet-django/project/migrations/0036_auto_20200124_1053.py
> physionet-django/project/models.py

doi should be "blank=True" (and "null=True"?)

bemoody:
BTW, one other thought:

> I still think it would be preferable to separate the functions of
> "creating draft DOI" from "updating DOI with the current project
> metadata".  And possibly the latter should be done automatically as
> part of "publishing".

It is a good thing that the project metadata (apart from the URL) is
filled in when the draft DOI is created, because if for some reason
that gets rejected by api.datacite.org then it is good to catch the
problem early.  But we *also* need a way of programmatically updating
the metadata, either before or after publication.

elfeto:
>The name "publish_doi_draft" is a bit odd - once this function is
called, it's no longer a "draft".

But you are actually publishing what is a draft in the first place, and then once its published (function ends) it's no longer draft. (That was the thought under the name.) But I removed the draft part.

         "id": doi.split('/')[1],

was removed because it's not needed.
> It is a good thing that the project metadata (apart from the URL) is
filled in when the draft DOI is created, because if for some reason
that gets rejected by api.datacite.org then it is good to catch the
problem early.  But we *also* need a way of programmatically updating
the metadata, either before or after publication.

This is something that can be added in the future because it adds extra complexity and its not something required for publishing. @tompollard  can create an issue.


The rest of the things were addressed.

tompollard:
> What format should "description" be in?  "strip_tags" will kill tags but is a poor substitute for actually converting to plain text, if "plain text" is what is desired.

Is the source format HTML? In `project.models` we convert html to text with `html2text`:

```python
def dua_text_content(self):
        “”"
        Returns dua_html_content as plain text. Used when adding the DUA to
        plain text emails.
        “”"
        return html2text(self.dua_html_content)
```

bemoody:
html2text wraps lines, which might not be desirable.  But it's a good start.

Can we define this "plain text abstract" as a method/property of
project.models.Metadata, so we can use this for JSON-LD as well (issue
#866) ?

elfeto:
>Can we define this "plain text abstract" as a method/property of
project.models.Metadata, so we can use this for JSON-LD as well (issue
#866) ?

Done in the commit 672e665

bemoody:
I assume that what create_doi_draft returns is a string like
"10.13026/vc0e-th79".  But it might potentially mean something like
"doi:10.13026/vc0e-th79" or "http://dx.doi.org/10.13026/vc0e-th79" or
"https://doi.org/10.13026/vc0e-th79".  So can you please give an
explicit example in the doc string of create_doi_draft?

You should also check that the string does in fact match the format
you expect.  Make sure that "content['data']['id']" is a string and
matches an expected regular expression.  (We have
project.validators.validate_doi, although that function is broken in
several ways and currently unused!)


In commit 5cb019925cc7c76993574964ef93f6378b38fe59, you made some
significant changes to the operation of ActiveProject.publish.  It is
very difficult to read what you changed since you re-indented the bulk
of the function, and also changed several things, in a single commit.
Please try to break changes into separate commits that someone can
read.

However, making "publishing" into a transaction might be a good
change, but it is a *major* change that is unrelated to DOIs.  So
please undo this, and consider making this change in a separate pull
request.

That function is too complicated already.  It shouldn't be complicated
further by entangling it with external services.  Furthermore, this
gives us no control over the DOI that is used (or not used); this
would make it impossible to create a project using a DOI that has
already been assigned, not to mention creating a project *without* a
DOI.

elfeto:
All the comments were addressed

Where would it be ideal to add them?
I though the best place would be the project publish function.

**I left the create/publish DOI functions unused.**

bemoody:
On commit 0a8d07e1964d08c24e66f96a7c8ff10e0e3ae94d, this could be
helpful, but I didn't mean you needed to paste the complete JSON file.
I just meant for you to write a comment like this:

    def create_doi_draft(project):
        """
        Create a draft DOI with some basic information about the project.

        A POST request is done to set the base information.
        The assigned DOI is returned to be used in the template.

        When successful, this function returns a string like
        "10.13026/vc0e-th79".

        If DOI registration is disabled (settings.DATACITE_API_URL is
        None), this function returns an empty string ("").
        """

(except that the latter part of that isn't accurate with respect to
your latest version.)


Here's what would be ideal:

1. When the editor marks an ActiveProject as "Accepted":

   - If the ActiveProject doesn't already have a DOI, a draft DOI is
     automatically created for it.

   - If this is a new project (no published versions yet of the
     CoreProject), and the CoreProject doesn't already have a DOI, a
     draft DOI is automatically created for it.

   (Note that administrators have the ability to replace these
   automatically-assigned DOIs, or to set a project as "no DOI", if
   necessary.)

2. When the editor approves an ActiveProject for publication and sets
   the public URL:

   - If the ActiveProject has a DOI, the ActiveProject's DOI is updated
     with the current authors, title, description, URL, etc.

   - If the CoreProject has a DOI, and the ActiveProject is the latest
     version of this CoreProject, the CoreProject's DOI is updated
     with the current authors, title, description, URL, etc.

   - Project is officially published; project.publish is called,
     people are notified.

   - If the ActiveProject has a DOI, that DOI is switched to
     "findable" status.

   - If the CoreProject has a DOI, that DOI is switched to "findable"
     status.

3. A programmatic interface that we can use to update an existing DOI,
   for an existing PublishedProject or CoreProject, with the current
   authors, title, description, URL, etc.

   This is not something we want to do on a regular basis so we
   probably do not need a web interface for it.  However, we do need
   to allocate new DOIs, and reassign some existing DOIs, for existing
   PublishedProjects and CoreProjects.  Exactly how we want to do this
   is still up for discussion.

   Likewise, we will need to assign DOIs to any new ActiveProjects and
   CoreProjects that are in the pipeline when these changes are made.

elfeto:
Items 1 and 2 **SHOULD** be addressed.

The third item should be raised as an issue.

bemoody:
Thanks!

I like the idea of making auto_doi optional, but I don't think it
makes sense to make it a field of EditLog.

If I understand correctly, you are *only* updating the DOI metadata
(and "publishing" the DOI) if the DOI is currently in "draft" state.
Is this meant as a safety measure, or is it that "publish" is rejected
for already findable DOIs?

Note that when publishing a new version of a CoreProject, we do want
to update the existing (already findable) DOI.

Commit 2fd984fea2166207c93aada1e0e263a4b3354063 *is* essentially what
I was asking for with my third point above...

If we want to modify an existing DOI without changing its state, we
presumably want to omit the "event" attribute.  Keeping the same state
should presumably be the default behavior (e.g. event should default
to None and publish_doi should use event="publish".)

In generate_doi_info, what is "state" for?  That looks wrong and I
don't see it in the examples at
https://support.datacite.org/docs/api-create-dois

elfeto:
>I like the idea of making auto_doi optional, but I don't think it
makes sense to make it a field of EditLog.

I think is ideal to store the DOI decision on the editlog.
This can be helpful to debug problems.
If you feel strongly against it I can remove it. 

>If I understand correctly, you are *only* updating the DOI metadata
(and "publishing" the DOI) if the DOI is currently in "draft" state.
Is this meant as a safety measure, or is it that "publish" is rejected
for already findable DOIs?

Up to now I haven't tried to update with publish in state, but apparently it doesn't matter if you have it or not. It will overwrite all.
Right now I just removed it and I am only doing updates to publish.
This means that if it starts:
  - From draft to publish it will update the information, and set the state to publish.
  - From publish it will update the information keeping the same state.

>Note that when publishing a new version of a CoreProject, we do want
to update the existing (already findable) DOI.

That is done. It must meet 3 requirements.
 - The DOI must exists under the `project.core_project.doi`
 - The project being published must be the latest version
 - The DOI must be registered in DataCite.
 
The ` "event" attribute` apparently is unneeded. 

>In generate_doi_info, what is "state" for?  That looks wrong and I
don't see it in the examples at

Was ideally to be used as an update / publish, but I removed it since its unneeded.

bemoody:
> I think is ideal to store the DOI decision on the editlog.
> This can be helpful to debug problems.
> If you feel strongly against it I can remove it.

I guess this isn't displayed so it doesn't make much difference, just
seems a little weird to me.  But that's okay.

The actual design of the form could be better.  Instead of a drop-down
menu, it should be a checkbox, and should be checked by default, and
labelled something like "Create a DOI for this project when it is
accepted".


I'm having a little trouble understanding what you're saying.  I think
what you're saying is that if the DOI is already "findable", then
specifying "event":"publish" has no effect.  So with your latest
version, simply re-invoking publish_doi for an already-published
project should be enough to update the existing metadata.  Is that
correct?


I'm not sure get_doi_status here is doing what you intend.  If you
pass a non-existent DOI, then it will raise an exception.  On the
other hand, if you pass an existing findable DOI, then it will return
"findable", regardless of whether you're authorized to modify that
DOI.

(I don't know how permissions are managed, but I presume that our
account is only authorized to create and modify DOIs under our
allocated prefix.)

So... currently, AFAIK, get_doi_status will either return one of the
strings "draft", "registered", or "findable", or will raise an
exception.  It won't return a false value in any case (unless the API
behavior is changed in the future.)

elfeto:
>The actual design of the form could be better.  Instead of a drop-down
menu, it should be a checkbox, and should be checked by default, and
labelled something like "Create a DOI for this project when it is
accepted".

I agree, but since we are using a template for the form, the checkbox will be displayed in a newline like all the other inputs in the form and it looks weird.

What I meant was that having `"event":"publish` does nothing.
I can create without an event.
I can publish a draft without an event.
I can update a published DOI without an event.

Having `"state": "publish"` will work for both marking a draft as publish and updating a published DOI.

>I'm not sure get_doi_status here is doing what you intend.  If you
pass a non-existent DOI, then it will raise an exception.  On the
other hand, if you pass an existing findable DOI, then it will return
"findable", regardless of whether you're authorized to modify that
DOI.

That is what I intended. I just want to know that it exists. In theory since we would be publishing the dataset or software, then we would also be creating the DOI, in which it would be ours, and we could modify it.

>(I don't know how permissions are managed, but I presume that our
account is only authorized to create and modify DOIs under our
allocated prefix.)

Yes, we have 2 different prefixes, one for testing and one for production. Yes the permissions are create/modify. In theory we should be able to delete them also but I haven't tried it. BTW, there is only one account, so imagine this as a root account.

>So... currently, AFAIK, get_doi_status will either return one of the
strings "draft", "registered", or "findable", or will raise an
exception.  It won't return a false value in any case (unless the API
behavior is changed in the future.)

You are correct.

Note: This was intended as an initial try of the DOI, I expected to @tompollard  raise issues with modifications.

bemoody:
> I agree, but since we are using a template for the form, the checkbox will
> be displayed in a newline like all the other inputs in the form and it looks
> weird.

Hmm.... weird how?  We may be able to work with that.

> What I meant was that having `"event":"publish` does nothing.
> I can create without an event.
> I can publish a draft without an event.
> I can update a published DOI without an event.
>
> Having `"state": "publish"` will work for both marking a draft as publish
> and updating a published DOI.

That really doesn't seem right... is "state":"publish" documented anywhere?

>>I'm not sure get_doi_status here is doing what you intend.  If you
> pass a non-existent DOI, then it will raise an exception.  On the
> other hand, if you pass an existing findable DOI, then it will return
> "findable", regardless of whether you're authorized to modify that
> DOI.
>
> That is what I intended. I just want to know that it exists. In theory since
> we would be publishing the dataset or software, then we would also be
> creating the DOI, in which it would be ours, and we could modify it.

Yeah.  But you're calling get_doi_status in publish_doi and checking
the result as a boolean.  I'm wondering what the idea is there.

I'd be fine with just letting publish_doi raise an exception if an
invalid/nonexistent DOI is specified.

elfeto:
>Hmm.... weird how?  We may be able to work with that.

Maybe its just me....

Screen Shot 2020-02-11 at 8 18 16 PM


> That really doesn't seem right... is "state":"publish" documented anywhere?

I know but I tested it out earlier today. 
The documentation is a bit vague
https://support.datacite.org/docs/doi-states
https://support.datacite.org/docs/api-create-dois
https://support.datacite.org/docs/how-do-i-make-a-findable-doi-with-the-rest-api
In the third link they make use of the `"state": "publish"`.
I just tested them.
Using:
 - "state": "publish" -> can publish and update.
 - "event": "publish"-> can publish and update.
 - both -> can publish and update.
 - none -> can update

>Yeah.  But you're calling get_doi_status in publish_doi and checking
the result as a boolean.  I'm wondering what the idea is there.

The idea was that if it exists, it will return something, hence it will equals True.
I can do the "get_doi_status" and add a logger message saying that the DOI was published or updated if it makes more sense. because it should (I think) always return something unless it raises an error.

>I'd be fine with just letting publish_doi raise an exception if an
invalid/nonexistent DOI is specified.

Its a bit more complicated since there can be different errors.
 - Invalid field in payload
'{"errors":[{"source":"metadata","title":"Is invalid"},{"source":"metadata","title":"Is invalid"}]}'
'{"errors":[{"source":"url","title":"URL is not valid"}]}'
 - Invalid prefix in the Datacite URL
'{"errors":[{"status":"404","title":"The resource you are looking for doesn\'t exist."}]}'
- If the DOI doesnt exists, but the DOI is valid, then it will create it and publish it (no error code)
 - Invalid id in the Datacite URL (long id)
'{"errors":[{"status":"400","title":"You need to provide a payload following the JSONAPI spec"}]}'

The error codes I have seen are 400, 404, 422
If there is a change in the API and something breaks, it can raise a 500.
I haven't seen in the 3xx range yet, but that doesn't mean there wont be any.

bemoody:
...It took me a suprisingly long time to realize that was due to poor
CSS rather than anything wrong with the HTML.


diff --git a/physionet-django/static/custom/css/form-control-input.css
b/physionet-django/static/custom/css/form-control-input.css
index 4cb72612..63f48fae 100644
--- a/physionet-django/static/custom/css/form-control-input.css
+++ b/physionet-django/static/custom/css/form-control-input.css
@@ -16,6 +16,11 @@ input, select, django-ckeditor-widget, textarea {
   outline: none;
 }

+input[type=checkbox] {
+  display: inline;
+  width: auto;
+}
+
 form li label input{
   display: inline-block;
   width: unset;
diff --git a/physionet-django/console/forms.py
b/physionet-django/console/forms.py
index eb830a5c..5874b4a1 100644
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -86,16 +86,19 @@ class Meta:
             'no_phi': forms.Select(choices=YES_NO_UNDETERMINED),
             'pn_suitable': forms.Select(choices=YES_NO_UNDETERMINED),
             'editor_comments': forms.Textarea(),
-            'auto_doi': forms.Select(choices=YES_NO),
             'decision': forms.Select(choices=SUBMISSION_RESPONSE_CHOICES)
         }

-    def __init__(self, resource_type, *args, **kwargs):
+    def __init__(self, resource_type, initial={}, **kwargs):
         """
         Set the appropriate fields/labels for the given resource type,
         and make them required. Remove irrelevant fields.
         """
-        super().__init__(*args, **kwargs)
+        initial = {
+            'auto_doi': True,
+            **initial
+        }
+        super().__init__(initial=initial, **kwargs)
         self.resource_type = resource_type

         # This will be used in clean
diff --git a/physionet-django/project/models.py b/physionet-django/project/model
s.py
index e6181632..e089e614 100644
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -1751,7 +1751,7 @@ class EditLog(models.Model):
         'editor_comments': 'Comments to authors',
         'no_phi': 'Is the project free of protected health information?',
         'data_machine_readable': 'Are all files machine-readable?',
-        'auto_doi': 'Automatically assign a new doi?',
+        'auto_doi': 'Automatically assign a DOI when this project is accepted',
     }

     LABELS = (


(On a separate note, the <label>s are borked - they should apparently
be using '{{field.id_for_label}}' instead of '{{field.html_name}}'.)

bemoody:
> The documentation is a bit vague
> https://support.datacite.org/docs/doi-states
> https://support.datacite.org/docs/api-create-dois
> https://support.datacite.org/docs/how-do-i-make-a-findable-doi-with-the-rest-api
> In the third link they make use of the `"state": "publish"`.

I see "event":"publish" but not "state":"publish".

> I just tested them.
> Using:
>  - "state": "publish" -> can publish and update.
>  - "event": "publish"-> can publish and update.
>  - both -> can publish and update.
>  - none -> can update

I have a feeling this is either a mistake on DataCite's part, or a
kludge meant to interoperate with some other broken piece of software
(which is not meant to be supported and may randomly stop working.)  I
would hesitate to rely on an undocumented feature without direct
confirmation from DataCite.  "event":"publish" seems to be the
intended API.


> If the DOI doesnt exists, but the DOI is valid, then it will create
> it and publish it (no error code)

That seems undesirable, so for that reason it makes sense to check the
status before trying to publish.

I would either:

- have get_doi_status raise a specific exception
  (e.g. "DOINotFoundError") if the DOI doesn't exist (status = 404)

- have get_doi_status return None if the DOI doesn't exist, and have
  publish_doi raise an exception in that case

Either way, get_doi_status should also raise an exception if
content['data']['attributes']['state'] is not a string.

elfeto:
Added the event publish and 404 exception with the state string check exception.

bemoody:
Nice!  But you're still treating the result of get_doi_status as a
boolean.  It's not!

Don't write an 'if' statement (or, in this case, an 'and'
subexpression) unless both branches are conceivable possibilities.

elfeto:
Added a "if" subexpression. Since there can be two possibilities, either a the DOI is a draft and we want to publish, or is not a draft and we want to update. 
Only the draft option needs the publish event.

bemoody:
If the current status is "registered" (perhaps because the initial
release was severely broken, the project was temporarily removed and
is now being re-published under the same DOI), then it should be
published too.


But what I really don't like is that you're *hiding the logic* here.

You thought (and I agree) that it would be a good idea to call
get_doi_status before publishing a DOI.

I think (and I'm not sure, but I think you agree with me about this)
that the primary reason for doing that is so that get_doi_status will
raise an exception if something is wrong (e.g., the DOI does not
exist.)

But the way you have written the publish_doi function makes it look
like the intent is something completely different (that the behavior
should be different *depending on what get_doi_status returns.*) That
is *not true* - the previous version of the code, which ignored the
result of get_doi_status, would have worked just as well.

I don't feel strongly about whether you should or should not use
"event":"publish" for already-findable DOIs.  But please, just add a
comment to publish_doi saying:

    # get_doi_status will raise an exception if the DOI does not exist.

This makes it clear that raising an exception is the *intended
behavior*.

elfeto:
>But what I really don't like is that you're \*hiding the logic* here.

What do you mean in \*hiding the logic* ?

>But the way you have written the publish_doi function makes it look
like the intent is something completely different (that the behavior
should be different *depending on what get_doi_status returns.*) That
is *not true* - the previous version of the code, which ignored the
result of get_doi_status, would have worked just as well.

get_doi_status has two options from the beginning. Either returns something or it fails with an exception. In the previous version of the code for example 782d27e the state was ignored because I just wanted to be sure the DOI exist, hence returning the state (because of the name of the function.
The same function could had been replaced at the time to check_doi_exists and set its return value to bool leaving the state to publish.

>This makes it clear that raising an exception is the *intended
behavior*.

I thought it clear when it's written in the code the exception specifically for that, but it doesn't harm anyone having it in a comment or docstrings.

bemoody:
The source code should reflect the programmer's intentions.  Someone
reading the source code should be able to understand what it is doing
and why.

If you write a statement like:

    print("Hello world")

it is obvious what your intention is: you want the side effect of
displaying a message on the screen.

If you write a statement like:

    if foo() == "bar":
        print("Hello world")

that tells the reader something different.  Three things, actually:
- it is possible for foo to return a value equal to "bar"
- it is possible for foo to return a value not equal to "bar"
- the message should be printed in situations where foo() equals "bar".

It tells the reader that you care about the side effects of 'print',
but it does *not* say whether or not you care about the side effects
of 'foo'.

If you write a statement like:

    get_doi_status(xyz)

then it is clear to the reader that you do not care what the function
returns and are only calling it for its side effects.

If you write a statement like:

    if get_doi_status(xyz):
        print("Hello world")

then it is not at all obvious that you care about the function's side
effects.  But you *are* indicating that you care about the function's
return value - otherwise, you wouldn't have written the "if".

> get_doi_status has two options from the beginning. Either returns
> something or it fails with an exception.

Raising an exception is not a return value.  If get_doi_status raises
an exception, then the 'if' statement is never evaluated.  So, at the
time the 'if' statement is evaluated, there is only *one* possibility
- that get_doi_status returned something true.

bemoody:
This discussion has gotten quite long, so to briefly summarize, here
are (I think) the remaining issues:

- publish_notify should be invoked even if publish_doi fails

- 'settings.PRODUCTION' should be removed

- the 'doi' field (in UnpublishedProject, PublishedProject, and
  CoreProject) should have the attributes "blank=True" and "null=True"

- the 'auto_doi' form field should be True by default

- the 'auto_doi' form field should be displayed as a check box

- the label for 'auto_doi' should mention that it only takes effect
  when the project is accepted

- the label for 'auto_doi' should capitalize "DOI"

- if settings.DATACITE_PREFIX is not defined, 'auto_doi' should be
  disabled

There are also some issues with the actual DOI metadata (creator
names, affiliations, resourceTypeGeneral, publicationYear, version,
relatedIdentifiers, ...) but I'm ignoring those for now.

elfeto:
> \- publish_notify should be invoked even if publish_doi fails
> \- 'settings.PRODUCTION' should be removed
> \- the 'doi' field (in UnpublishedProject, PublishedProject, and
  CoreProject) should have the attributes "blank=True" and "null=True"
> \- the 'auto_doi' form field should be True by default
> \- the 'auto_doi' form field should be displayed as a check box
> \- the label for 'auto_doi' should mention that it only takes effect
  when the project is accepted
> \- the label for 'auto_doi' should capitalize "DOI"
> \- if settings.DATACITE_PREFIX is not defined, 'auto_doi' should be
  disabled

All done

> There are also some issues with the actual DOI metadata (creator
names, affiliations, resourceTypeGeneral, publicationYear, version,
relatedIdentifiers, ...) but I'm ignoring those for now.

You should list them for later changes to verify them here easier since its fresh on your mind, and would be easier to work with if they are known.

bemoody:
Looks good.  I take your word that it'll actually work on
Datacite's system :)

These are minor issues that stick out to me in the DOI metadata
itself:

- publicationYear should be the year of publication, not the
  current year

- authors' names should be taken from PublishedAuthor, not the
  user profile

- authors' "full names" should be in "Last, First" format

- I don't even know what format "description" should be
  in (perhaps nobody does :P)

- resourceType and resourceTypeGeneral should be defined at a
  higher level (perhaps in ProjectType)

and some additional fields that would be nice to add:

- affiliations from PublishedAffiliation

- nameType = "Personal"

- date (not just year) of publication

- project version

- cross-references between versioned DOIs and core DOIs

I don't think we necessarily need to resolve these issues before
merging this branch, but @tompollard is the expert.

elfeto:
>\- authors' "full names" should be in "Last, First" format
\- authors' names should be taken from PublishedAuthor, not the
  user profile

Done

elfeto:
This is just me merging with latest dev and fixing conflicts.  NOTHING new was added.

tompollard:
Okay, after discussion, merging then following up with some updates!

tompollard:
Thanks for setting this up Felipe :) Will be nice not to have to complete the form each time!

PR #858, opened 2020-01-29
Credentialing Application order by decision

elfeto:
Set past applications order by decision date not application date.


tompollard:
My understanding is that this change is to help people quickly find the most recently processed application (e.g. if a decision needs to be reversed). This seems to be a reasonably common task (for Ken at least), so makes sense.

Issue #658, opened 2019-08-06
References should usually be hyperlinks


bemoody:
Citations shown in project pages should contain a link to the resource in question, if it's available online.

Either the entire text or just the title of the work should be a link.

Obviously it's up to the author and/or editor to supply an appropriate stable URL (though a DOI would be preferable and could be auto-detected.)

PR #860, opened 2020-01-30
Follow IDF recommendation for displaying DOIs

tompollard:
Follow the current recommendations of the International DOI Foundation to format DOIs in the reference list, which as of this publication is as follows:

```
https://doi.org/xxxxx
```

See for example:
- https://www.crossref.org/display-guidelines/
- https://apastyle.apa.org/style-grammar-guidelines/references/dois-urls

### DIFF
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -736,7 +736,7 @@ class UnpublishedProject(models.Model):
         authors = self.authors.all().order_by('display_order')
         year = timezone.now().year
         doi = '10.13026/*****'
-        return '{} ({}). {}. PhysioNet. doi:{}'.format(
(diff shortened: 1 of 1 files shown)


bemoody:
Looks like this is just for active projects?  We presumably want to do
same for published projects.

For published projects, it'd be nice to make it a hyperlink.

Tangentially related: issues #688, #658.

tompollard:
Good points (and oops on missing the published project!). I can't see a quick way of adding the hyperlink so would like to leave this here for now if that's okay with you?

It would be good to do more on this when we have time (and agree with https://github.com/MIT-LCP/physionet-build/issues/688 that allowing people to download bibtex etc is important).

PR #867, opened 2020-02-07
Set ActiveProject publish as atomic

elfeto:
Convert the ActiveProject publish function as atomic, either all passes, or all fails.

### DIFF
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -1155,106 +1155,107 @@ class ActiveProject(Metadata, UnpublishedProject, SubmissionInfo):
         if not self.is_publishable():
             raise Exception('The project is not publishable')
 
-        # If this is a new version, previous fields need to be updated
-        # and slug needs to be carried over
-        if self.version_order:
-            previous_published_projects = self.core_project.publishedprojects.all()
-            previous_published_projects.update(is_latest_version=False)
+        with transaction.atomic():
+            # If this is a new version, previous fields need to be updated
+            # and slug needs to be carried over
+            if self.version_order:
+                previous_published_projects = self.core_project.publishedprojects.all()
+                previous_published_projects.update(is_latest_version=False)
 
-            slug = previous_published_projects.first().slug
-            title = previous_published_projects.first().title
+                slug = previous_published_projects.first().slug
+                title = previous_published_projects.first().title
 
-        published_project = PublishedProject(doi=doi,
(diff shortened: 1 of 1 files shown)


elfeto:
After the phone conversation, I moved all NON database related actions (file related) outside the atomic transaction.

bemoody:
On 2/18/20, Felipe Torres <notifications@github.com> wrote:
> After the phone conversation, I moved all NON database related actions (file
> related) outside the atomic transaction.

Why?  That doesn't sound like an improvement. :(

elfeto:
Up to my understanding of Django's atomic transaction, will only handle database transactions, and not files. Since most of the publish is DB related, but the end in which the files are managed, then I just added transaction atomic to all the DB related, and just bumped down to a if statement the files. 

That was the closest thing that I can think of to make publish an atomic transaction.

I'm open to ideas.

bemoody:
Well, it's true that the atomic transaction won't roll back changes to
the filesystem.

But, if the FS operations fail then we want publication to fail (no
changes made to the DB).  And if the DB operations fail then we want
the files to remain where they are.

So what are the possible failure modes here?

Maybe to be extra paranoid, it should be

   rename directory
   try
      with atomic transaction
         do stuff
   except
      rename directory back where it was

?

elfeto:
I just did the try except with the rename in the beginning.

bemoody:
(Looks like my reply was dropped somewhere?  Sorry if this message gets duplicated.)

Thanks!

The rmdir is probably unnecessary (and the isdir is definitely
unnecessary) but if you do want to do the rmdir then please update the
comment accordingly.

(Bear in mind that the rmdir will fail with ENOTEMPTY if the directory
isn't empty - i.e., there is already another published version.)

I would advise you to write:

    except:
        ...
        raise

rather than:

    except Exception as exception:
        ...
        raise exception

For a few reasons: (1) in cases like this, we want to catch all
exceptions (BaseException, not Exception); (2) "raise exception" adds
clutter to the stack trace whereas "raise" doesn't; (3) keep it short
and sweet.


It's probably not useful here, but I just learned about a Django
feature that may be useful in some cases:
'django.db.transaction.on_commit' to do something only after preceding
changes are successfully committed to the database.

elfeto:
> 'django.db.transaction.on_commit' to do something only after preceding

Interesting

bemoody:
Looks great - thanks!

PR #869, opened 2020-02-12
Reduce the width of tables in the credentialing forms

tompollard:
@kpierceHST has asked if we can make it easier to view the credentialing forms, which are currently a little too wide for his screen (specifically http://127.0.0.1:8000//console/past-credential-applications/unsuccessful). 

This is a couple of minor changes to column header text that I think help to address the issue:

- "Reference Contact Date" is now "Ref. Contact Date"
- datetimes are now dates
- "View Application" is now "View Details"

### DIFF
--- a/physionet-django/console/templates/console/past_credential_successful_user_list.html
+++ b/physionet-django/console/templates/console/past_credential_successful_user_list.html
@@ -6,8 +6,8 @@
           <th>Email</th>
           <th>Application Date</th>
           <th>Approval Date</th>
-          <th>Credentialed By</th>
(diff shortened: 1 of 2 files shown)


kpierceHST:
Hi again,

It's not really a matter of reducing screen width, but rather of handling overflow for the entire window.  Should probably be set to "auto", so scrollbar appears at bottom if necessary?

Ken

tompollard:
Thanks Ken, these changes are quick and should be an improvement anyway. We'll look at the CSS too.

PR #873, opened 2020-02-13
Fix some <label for='...'> attributes

bemoody:
I could have sworn that we had to try to fix this once before...

The 'for' attribute of a label must correspond to the 'id' of the target element, not its 'name'.  For example, in /register/, the "email" field has an 'id' of 'id_email' and a 'name' of 'email', so the corresponding label must have 'for="id_email"'.


elfeto:
Looks good

PR #875, opened 2020-02-15
Do a better job of marking project forms read-only

bemoody:
If you are not the submitting author of a project, numerous form
fields are not properly disabled, making you think you are
allowed to edit them, and potentially leading to serious
aggravation when you discover that the
slightly-paler-blue-than-normal submit button does nothing.

In particular, /projects/*/content/ did not correctly disable the
HTML input fields.

This is a bit messy, I did it in a bit of a hurry, so please
check carefully that it doesn't break anything.


elfeto:
Looks good.

The only minor thing if you want complete consistency:

The files section for the submitting author is different than for the other authors.
 - The submitting authors have the files panel but disabled and the files checkbox disabled
 - The Non-submitting author does not have the files panel and the checkbox are enabled.

Since the submit button is not there, it doesn't make a difference in the end, but it's just a difference in both ends. 

Apart from that, everything looks fine.

bemoody:
True.  We should hide the file manipulation buttons and checkboxes
when the project is not editable.

I also notice that the message shown to non-submitting authors is
"Only the submitting author may edit the metadata", which is a bit
strange.

bemoody:
Should be fixed now.

elfeto:
Yep, I was checking this out looks good.

Only odd thing NON PR related, the CSS for the messages in the contents is a bit wierd because the padding is overwritten to 0.

PR #878, opened 2020-02-17
Keep the wsgi and nginx logs for a longer time

elfeto:
Set the log rotation to remove files after 10 years of logs.
This is a temporary measure to not loose the logs until they are analyzed.


bemoody:
I hear what you're saying, but temporary measures tend not to be.  How
about setting a reasonable time frame (3 months? 6 months?), so that
we have a deadline for dealing with this?

elfeto:
I set it to 3 months, to force me work with this soon.

bemoody:
Thanks!

PR #893, opened 2020-02-26
Fix newline removal from credentialed serve_file

elfeto:
I just added a strip on the filename before its evaluated to serve a file. This will remove unwanted spaces and newline characters.

```
Internal Server Error: /files/eicu-crd/2.0/test.txt

BadHeaderError at /files/eicu-crd/2.0/test.txt
Header values can't contain newlines (got '/protected/published-projects/eicu-crd/2.0/test.txt\r')
...
```

### DIFF
--- a/physionet-django/physionet/utility.py
+++ b/physionet-django/physionet/utility.py
@@ -101,6 +101,7 @@ def serve_file(file_path, attach=True, allow_directory=False, sandbox=True):
     If allow_directory is true and file_path ends with a slash, serve
     a simple HTML directory listing.
     """
+    file_path = file_path.strip()
     accel_path = _file_x_accel_path(file_path)
     if accel_path:
         response = HttpResponse()


bemoody:
This is not a good idea.  We need to fix whatever is linking to
incorrect URLs, not add additional undocumented kludgy ad-hoc
behaviors to the server!

elfeto:
Then this url should not raise a "BadHeaderError" but return a 404 not found.

bemoody:
Yeah... this is a bit tricky actually because of all the layers of URL
encoding and URL decoding and path normalization going on.

I *think* what we want (to correctly handle non-ASCII) is to add a
layer of URL encoding in 'serve_file'.  Spaces should be encoded as
%20, even though as a matter of policy I don't want to allow filenames
with spaces.

But in addition, we may want to reject anything with control
characters in it and return a 404 in that case.

PR #895, opened 2020-02-27
open credential applications in new window

elfeto:
As requested from KP, open the view a specific application in a new window.


tompollard:
looks good to me, thanks

PR #897, opened 2020-03-02
Self managed approval project

izcram:
The changes introduce the concept of "self-managed access" projects, where the depositors of a project can grant and deny access based on a project description provided by the requester. This PR is mainly meant as preview to request some early feedback.

## Workflow for self-managed access projects

Instead of just acknowledging the data user agreement, the user needs to fill in a textfield describing the data use. The project submitter then gets notified via e-mail and can review the request, either accepting or rejecting along with the possibility to leave a comment. The requester can withdraw the request, while it is in a pending stage. The requester can re-apply whenever he/she withdrew the application or the request was denied. The requester gets notified via e-mail whenever the project submitter took a decision.

If the project submitter doesn't react within 2 weeks, the request gets automatically approved.

## Testing

The fixtures were extended with a project named "Self Managed Access Database Demo" submitted by the user `george`. The user `rgmark` can be used as requester to test the workflow (only credentialed user in the test fixture)

## What is currently missing

- [x] detail work on front-end (design/UX/wording)
- [x] cronjob to approve requests after 14 days
- extending project creation process, s.t. a project can be declared as "self-managed" (probably in separate PR)
- project submitter can appoint further approvers/project managers (probably in separate PR)


tompollard:
Nice, thanks @izcram. It looks like the test is failing on a missing migration. Please could you run the following command and then add the migration file/s to the pull request?

```
python manage.py makemigrations
```

bemoody:
Thanks, Marc!  What you have here is a great start, and although this will be a long message, most of the issues I see here are small ones that should hopefully be easy to address. :)

It looks like what you intend here is that a person should be able to submit a project access request, and submit a credentialing application, at the *same time*.  That wasn't something I had thought about before.

It certainly makes sense in terms of making the process speedier, although I hope it won't be confusing to the applicant.  It does mean we have to consider several different workflows and possible states that an applicant might be in:

- I might already be credentialed
- I might have submitted a credentialing application that was rejected
- I might have submitted a credentialing application that is pending
- I might not have submitted any credentialing application

- I might already have been approved for data access
- I might have submitted a data access request that was rejected
- I might have submitted a data access request that is pending
- I might not have submitted a data access request

which makes for at least 16 possible states.

Some specific comments on the code below...

----

In project/models.py:

    from django.contrib.auth import get_user_model

Appears to be unused.

In DataAccessRequest:

    user = models.ForeignKey('user.User', on_delete=models.CASCADE)

To be clear, I would change the name of this field to "applicant" or "requester" or something like that.

    # TODO: on delete: or no action?, assuming 1 is admin user
    responder = models.ForeignKey('user.User', null=True,
                                  related_name='data_access_request_user',
                                  on_delete=models.SET(1))

It likely doesn't matter, but on_delete should probably be SET_NULL.  We shouldn't assume anything about particular user IDs.

In PublishedProject.has_access:

        if self.access_policy:
            if self.access_policy == 2 and (
                not user.is_authenticated or not user.is_credentialed):
                return False

            if self.is_self_managed_access:
                records = DataAccessRequest.objects.filter(project=self,
                                                           user__id=user.id)
                return any(
                    r.status == DataAccessRequest.ACCEPT_REQUEST_VALUE for r in
                    records)
            else:
                return DUASignature.objects.filter(project=self,
                                                   user__id=user.id)

This could, I think, be written more efficiently and more succinctly as:

            if self.access_policy == 2 and not user.is_credentialed:
                return False

            if self.is_self_managed_access:
                return DataAccessRequest.objects.filter(
                    project=self, user=user,
                    status=DataAccessRequest.ACCEPT_REQUEST_VALUE).exists()
            else:
                return DUASignature.objects.filter(
                    project=self, user=user).exists()

If they aren't logged in, they can't possibly be credentialed.

----

In project/views.py:

    @login_required
    def request_data_access(request, project_slug, version):
        user = request.user
        proj = PublishedProject.objects.get(slug=project_slug, version=version)

        if not proj:
            raise Http404()

This won't work; you want to write instead:

    try:
        proj = PublishedProject.objects.get(slug=project_slug, version=version)
    except PublishedProject.DoesNotExist:
        raise Http404()

Please add a doc string to the start of each function describing what the view is for.  In particular, document which view(s) are to be used by the person applying for access, and which view(s) are to be used by the person reviewing the applications.

sign_dua should refuse requests for projects that have is_self_managed_access set.

request_data_access, data_access_request_status, and data_access_request_view should refuse requests for projects that do *not* have is_self_managed_access set.

    needs_credentialing = proj.access_policy == 2 and not bool(
        CredentialApplication.objects.get(user_id=user))

This won't work; again, get() raises an exception if the object doesn't exist.  And I think this needs further consideration of the different possible states for the credentialing application.  Legacy credentialed users, IIRC, will not have a CredentialApplication at all.

----

In search/urls.py:

    path('request-data-access/<project_slug>/<version>/',
         project_views.request_data_access,
         name='request_data_access'),
    path('data-access-request-status/<project_slug>/<version>/',
         project_views.data_access_request_status,
         name='data_access_request_status'),
    path('data-access-request-view/<project_id>/<user_id>',
         project_views.data_access_request_view,
         name='data_access_request_view'),

Better not to use the project "id" in the URL; use the project slug and version instead.

I'm not sure why you have separate 'request_data_access' and 'data_access_request_status' URLs, and I wonder if those could be consolidated into a single URL/view.

I would use something like "request-access/<project_slug>/\<version>/" for the URL that is used by the applicant, and "access-requests/<project_slug>/\<version>/<user_id>/" for the URL that is used by the author/manager.

izcram:
Thanks for your comments @bemoody ! I have addressed all of them except the last part regarding the URLs, I'll do this tomorrow.

Concerning parallel credentialing and access request processes: We initially discussed with Tom to be able to do this, however, it turns out we'd like to see some (verified) information collected during the credentialing process (e.g. affiliation). Also you are right, that handling all the different states could be quite involved. The intention hence is, that a user first needs to be credentialed and then can request access.

Regarding PublishedProject.has_access: If a user is logged in, then the 'is_credentialed' field is not set (AnonymousUser). So it seems I need to check whether somebody is logged in. I simplified the other parts according to your suggestions though.
  
Initially, I had `request_data_access` and `data_access_request_status` in one view, but then implementing the withdrawal process, I felt that it became a bit too complicated. Also, I think, that in some sense, these are two different things. I suggest to keep these separate, however, I can change the URLs as you suggested (in particular not using the project id)

izcram:
@bemoody and @tompollard , I've just addressed the last part about urls. Please have another look if you see things which need tweaking!

tompollard:
This looking good. One question that I have is how we approach the case of someone who wants to work with the data on a second project, with a different title and topic. e.g.:

- I request access with the title "Mortality prediction in the ICU".
- I complete my project, and now I want to work on "Effectiveness of intervention X".
- Do I need to submit a new access request?  

I only had a *very* brief play around, but at the moment I get a URL error after clicking "Submit request" at: http://127.0.0.1:8000/request-data-access/demoselfmanaged/1.0.0/

```
Request Method: POST
Request URL: http://127.0.0.1:8000/request-data-access/demoselfmanaged/1.0.0/

Template error:
In template /Users/tompollard/projects/physionet-build/physionet-django/notification/templates/notification/email/notify_owner_data_access_request.html, error at line 7
   Reverse for 'data_access_request_view' with arguments '('demoselfmanaged', '1.0.0', 2)' not found. 1 pattern(s) tried: ['data\\-access\\-request\\-view\\/(?P<project_id>[^/]+)\\/(?P<user_id>[^/]+)$']
   1 : Dear {{ user.get_full_name }},
   2 : 
   3 : There is a new data access request for the project
   4 : {{ data_access_request.project }}.
   5 : 
   6 : You can view and reply to the request following this link:
   7 : {{ request_protocol }}://{{ request_host}} {% url 'data_access_request_view' data_access_request.project.slug data_access_request.project.version data_access_request.requester.id %} 
   8 : 
   9 : {{ signature }}
   10 : 

Traceback:

File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/django/core/handlers/exception.py" in inner

File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response

File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response

File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/django/contrib/auth/decorators.py" in _wrapped_view

File "/Users/tompollard/projects/physionet-build/physionet-django/project/views.py" in request_data_access
  1645.                                                           request.get_host())

File "/Users/tompollard/projects/physionet-build/physionet-django/notification/utility.py" in notify_owner_data_access_request
  622.             'request_protocol': request_protocol

File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/django/template/loader.py" in render_to_string
  62.     return template.render(context, request)

File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/django/template/backends/django.py" in render
  61.             return self.template.render(context)

File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/django/template/base.py" in render
  171.                     return self._render(context)

File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/django/test/utils.py" in instrumented_test_render
  96.     return self.nodelist.render(context)

File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/django/template/base.py" in render
  937.                 bit = node.render_annotated(context)

File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/django/template/base.py" in render_annotated
  904.             return self.render(context)

File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/django/template/defaulttags.py" in render
  443.             url = reverse(view_name, args=args, kwargs=kwargs, current_app=current_app)

File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/django/urls/base.py" in reverse
  90.     return iri_to_uri(resolver._reverse_with_prefix(view, prefix, *args, **kwargs))

File "/Users/tompollard/projects/physionet-build/physionet-django/env/lib/python3.6/site-packages/django/urls/resolvers.py" in _reverse_with_prefix
  668.         raise NoReverseMatch(msg)

Exception Type: NoReverseMatch at /request-data-access/demoselfmanaged/1.0.0/
Exception Value: Reverse for 'data_access_request_view' with arguments '('demoselfmanaged', '1.0.0', 2)' not found. 1 pattern(s) tried: ['data\\-access\\-request\\-view\\/(?P<project_id>[^/]+)\\/(?P<user_id>[^/]+)$']
```

izcram:
Yes, good point about the second project! I'll ask for some input from our project team, what the requirement should be. 

One could in principle allow several requests per user for the same project. For the purpose of giving access to the data or not, only one needs to be accepted. Subsequent requests would be more of the sort of asking permission whether the data could be used for another purpose than stated in the first request.

How is this handled for credentialed users, if they pivot their research agenda?

For the URL: it should be http://127.0.0.1:8000/request-access/demoselfmanaged/1.0.0/ 
I changed this in 5a7209a . Where did you pick up the URL?

tompollard:
To generate the error above I run `resetdb` and `loaddemo` then:

- login as "rgmark"
- navigate to http://localhost:8000/content/demoselfmanaged/1.0.0/
- click "submit an access request"
- complete the form then click "submit request"

[update: just checked, my commit history is up to date - most recent commit is 5a7209accf7cc496cd359e4f05dcacb399d5a553)

tompollard:
> How is this handled for credentialed users, if they pivot their research agenda?

The credentialing process is just attempting to check whether this is a legitimate user. We don't restrict users to specific research agendas.

bemoody:
ed29def0a0d0ecc4de393a2b665c046382148b71

project.models.DataAccessRequest:

 - We should set a maximum length for the data_use_purpose.  I'm
   not sure we want to allow complex HTML there either, although
   maybe it doesn't hurt much.

project.views.request_data_access:

 - Better to use user.get_full_name() instead of
   f"{user.profile.first_names} {user.profile.last_name}"

project.views.data_access_request_status:

 - Typo in "Project doesn not exist"

project.views._user_allowed_viewing_requests:

 - I think it'd be better to make this a method of the
   PublishedProject class, and document what it's used for.  It
   looks like this controls whether the user is allowed to
   *respond* to access requests, in addition to simply viewing
   them.

 - This access is currently permitted only for the *submitting*
   author but it is the *corresponding* author who receives
   notifications.  Is this the right logic?

project.views.data_access_request_view:

 - CredentialApplication.objects.get(user_id=requester.id) will
   fail if they don't have exactly one credentialing application.
   Legitimate users could have zero (legacy credentialed) or more
   than one (submitted an incomplete application that was
   rejected and then submitted a second application that was
   accepted.)

   Also you can use "requester.credential_applications" instead
   of "CredentialApplication.objects.filter(user=requester)".

   I'm not really clear about what this specific info will be
   used for.  The application might be years old and thus
   misleading to the reviewer.

project.forms.DataAccessRequestForm:

 - What does this do?  Seems like this is being overridden
   somewhere else?

        widgets = {
            'data_use_purpose': forms.Textarea(attrs={'rows': 7}),
        }

 - Ideally, I think the form should be rejected if they don't
   modify the template at all.  But that might be tricky.

project/templates/project/request_data_access.html

 - The page should have a heading (h1) matching the page title.
   In particular, the reader should know what project they are
   applying for :)

 - If there are multiple published versions then (assuming that
   you need to apply for access to each version separately) there
   should probably be a warning if this is not the latest
   version.

 - "This dataset requires special permission by the depositor on
   a project by project basis. To proceed, please provide below a
   description of the research you aim to do with the data. The
   request will be reviewed within 14 days by the data
   depositor. You'll get notified via e-mail about their
   decision."

   => This could be worded better.  My suggestion:

      "This resource is available only for use by approved
      research projects.  Please provide a description of the
      specific research project that you aim to complete using
      the data.  The data depositor will review your request
      within 14 days.  You will be notified by email
      ({{ user.email }}) about their decision."

izcram:
Thanks for your comments! I addressed most of them, I believe.
 
How should legacy credentialing be handled? Are many users credentialed this way? In the current implementation, no credentialing information would be shown to the reviewer for legacy credentialing users. The idea of showing some information from the credentialing process is that the reviewer gets a better sense who is requesting access.

I briefly tried to verify that the template was edited by checking the `has_changed_data` attribute of the form. However, this didn't work, I think the HTML string gets modified by the browser, even though the field was not edited. I could try more, but I'm not sure whether it is worth the effort.

 I'm sorry for the ugly CSS hack to align the checkbox. I'm open to improvements, not knowing my way around CSS much.

bemoody:
> How should legacy credentialing be handled? Are many users
> credentialed this way?

I don't know the exact numbers, but yes, a great many users were
credentialed for access to MIMIC and/or eICU databases on the old
site (i.e. before roughly August last year.)

It looks like some, but not all, information from their original
credentialing application was saved in the LegacyCredential
model.  Looking at the model, this info is probably not really
useful here except perhaps 'country' and 'mimic_approval_date'.

The way you've handled it in commit
aa99fb92f9f5816176ac544c8adea89bc8ba2ac4 seems fine to me, as
long as you're okay with the limitations.  But you might consider
pulling information from the LegacyCredential if it exists.

> I briefly tried to verify that the template was edited by
> checking the has_changed_data attribute of the form. However,
> this didn't work, I think the HTML string gets modified by the
> browser, even though the field was not edited.

That's what I was afraid of.  If you don't see an easy way around
it then I wouldn't bother.

> I'm sorry for the ugly CSS hack to align the checkbox.

Don't worry about that.  We can clean it up later. :)


I'll try to do a little more testing this afternoon but this is
looking good to me.  @tompollard, do you see any issues I've
missed?

bemoody:
Some more small things:

The "Request Access" page (/request-access/demoselfmanaged/1.0.0/)
doesn't indicate which project you are applying for.  I think that
should be stated clearly.  (Compare /sign-dua/demoeicu/2.0.0/: the
project title/version are shown in the page heading and a link back to
the project is given below.)

In project/templates/project/project_home.html, the link {% url
'project_overview' da_request.project.slug %} is incorrect... you
probably meant to link to 'published_project' or
'data_access_requests_overview'.


The "related_name" specified in a ForeignKey field is the attribute of
the *other* class that is used to denote *this* class.  For example,
PublishedAuthor has:

    project = models.ForeignKey('project.PublishedProject',
        related_name='authors', db_index=True, on_delete=models.CASCADE)

which means that if "p" is a PublishedProject object, "p.authors" is
an iterable of PublishedAuthor objects.  Following this pattern, the
"project" foreign key in DataAccessRequest should have
related_name='data_access_requests'.


In search/urls.py, it'd be better to add a trailing slash to each of these:

    path('access-requests/<project_slug>/<version>/<user_id>',
         project_views.data_access_request_view,
         name='data_access_request_view'),
    path('access-requests/<project_slug>/<version>',
         project_views.data_access_requests_overview,
         name='data_access_requests_overview'),

izcram:
Thanks for your comments, I did another iteration based in them.

I added some code to show information for legacy credentialed users, however, I was not able to test it. Do you have a suggestion how I could add such a user to the fixtures? Do I just add a new user and add an appropriate row in the user_legacycredential table?  The 'migrated' flag, does it refer to the user having been migrated or the credentialing application?

Also, I think what is missing in that context, is a short phrase explaining to the reviewer why the displayed information is different. Do you have a suggestion?

bemoody:
If you register an account in the demo system, using the email address
admin@upr.edu, they should appear as "legacy credentialed".  That
account is used for tests, but we could certainly add another demo
user who already has legacy credentialed status.

I think that in that case, "migrated" should be true, "migrated_user"
should be the corresponding user account, and "migration_date" should
be the time that the user account was activated.

tompollard:
Nice work @izcram,  this is looking great! I'll take a more detailed look tomorrow, but it would be good to aim to merge on Friday if everyone is okay with that. I've added a couple of quick thoughts below:

- We'll need to finalize the License/DUA text. Ideally we would have a system for tracking versions of this text, but we can worry about that later.
- When a new data access request is made, it would be helpful if the person making the request received a notification email too (Dear X; You have requested access for project X; You will receive a decision by X; You can check the status of your request at X; ).
- There are some minor PEP8 styling issues that could be cleaned up (e.g. width <79 char).
- I think this is mentioned above, but the "I .. agree with the above terms" checkbox is awkwardly placed at the moment:

Screen Shot 2020-03-18 at 17 20 20

- Not sure off of the top of my head how to deal with `# TODO don't hardcode. how to get hostname to generate URL?` but I expect someone else here does. I think you can do something like `request.get_host()` or `Site.objects.get_current()`.

tompollard:
@izcram there are lots of "nice to haves" that aren't included here but should be addressed later. While things are fresh in your mind, please could you start making a note of them somewhere? We'll do the same and then we can raise new issues.

izcram:
@tompollard Merging soonish would work for me. I'll also open a bunch of issues with open points as you suggested.

- Regarding the license: The idea is that every project has its own license/DUA and would provide this during the creation of the project.
- requester also gets an email confirmation now
- I've tweaked the checkbox CSS, I hope it's in the right spot now ;) (tested in firefox and chrome)

tompollard:
@bemoody happy to merge?

bemoody:
Yes, looks good to me.

tompollard:
Really nice work @izcram, thanks!

Issue #883, opened 2020-02-18
Do not allow account activation without a password


bemoody:
You should not be able to activate an account without entering a password.

Best workflow, I think, would be:
- `/register/` does not include password fields.
- `/activate/*/*/` does not activate the account immediately; rather, it shows a form with "password" and "confirm password" fields which you must fill out and submit in order to activate the account.  (The same form could potentially also let you edit the data that you submitted in /register/.)
- Submitting the `/activate/*/*/` form should also log you in to the account.

Ideally, if you used the old /register/ form (and thus set a password) you shouldn't then be required to enter your password twice more.  However, I can't think of a good design for the activation form that would both:
- require you to enter your password at least once
- allow you to change your password if you can't remember what you entered before

PR #898, opened 2020-03-03
account activation with password

elfeto:
I am using the same method Django uses to reset the password but with the account validation.
Apparently the user token should be kept secret. 

Django uses active accounts by default, so they don't have a mechanism to hide the password on account activation.
I am using a similar or same code as in "django/contrib/auth/views.py".

This would hopefully open leeway to use the same or similar functions on adding an email and activating it.

### DIFF
--- a/physionet-django/user/forms.py
+++ b/physionet-django/user/forms.py
@@ -202,18 +202,13 @@ class RegistrationForm(forms.ModelForm):
     last_name = forms.CharField(max_length=50, label='Last Name',
                     widget=forms.TextInput(attrs={'class': 'form-control'}),
                     validators=[validate_name])
-    password1 = forms.CharField(label='Password',
-                    widget=forms.PasswordInput(attrs={'class': 'form-control'}))
-    password2 = forms.CharField(label='Password Confirmation',
-                    widget=forms.PasswordInput(attrs={'class': 'form-control'}))
 
     class Meta:
         model = User
         fields = ('email','username',)
         widgets = {
-            'email': forms.EmailInput(
-                attrs={'class': 'form-control dropemail'}),
-            'username':forms.TextInput(attrs={'class': 'form-control'}),
+            'email': forms.EmailInput(attrs={'class': 'form-control'}),
+            'username': forms.TextInput(attrs={'class': 'form-control'}),
         }
 
     def clean_username(self):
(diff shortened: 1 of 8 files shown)


elfeto:
This should close #883

bemoody:
Thanks!

You use the "readonly" attribute here rather than making the form
field "disabled".  Is that just because it looks better or is
there another reason?

In clean_password2, then, we probably want to use 'self.user',
*not* self.data['username'].

<http://127.0.0.1:8000/activate/NQ/5eh-4740b50ee8574c31ec3d/>
saves the token in a session variable and then redirects to
<http://127.0.0.1:8000/activate/NQ/user-activation/>.  The
redirection is a good idea, but it's a little confusing to reuse
the same view function this way.

The old activate_user function would also check whether
'user.is_active' is already set.  Probably if the token is valid
and the user is already active then this should redirect to the
login form.

Also, the new activation form should include the zxcvbn password
strength bar.

elfeto:
>You use the "readonly" attribute here rather than making the form
field "disabled".  Is that just because it looks better or is
there another reason?
 
No reason, I wanted the "disabled" usage. It was the first thing that came to mind.

>In clean_password2, then, we probably want to use 'self.user',
*not* self.data['username'].

Since its not a modelform it doesn't have a 'self.user', I tried making it a modelform to use that, but even providing the instance it tried to re-create the username and it gave me a duplicate error

>The old activate_user function would also check whether
'user.is_active' is already set.  Probably if the token is valid
and the user is already active then this should redirect to the
login form.

Then I will add a message with this saying that the account is already active (to not confuse the user)

>Also, the new activation form should include the zxcvbn password
strength bar.

I knew something was missing!!! ty :)

elfeto:
I changed:
 - From readonly to disabled
 - To self.user in clean_password2
 - Redirecting if the user is active with a message
 - Added  zxcvbn password strength bar

bemoody:
Hmm.  If you just set the disabled HTML attribute then it complains
the field is missing.  You want to set "disabled=True" as a parameter
for the CharField/EmailField.

activate_user should also use @sensitive_post_parameters so if it
crashes, passwords are not saved in the logs.

elfeto:
Fixed it and activated the account. :)

bemoody:
Tried to test this as many ways as I could think of and it looks good.

PR #901, opened 2020-03-04
open user profile in new window

elfeto:
Continuation of #895
Adding open username in new window as requested from KP (I missed one in the past PR).


tompollard:
Looks good to me!

Issue #884, opened 2020-02-18
Do not allow email address association without logging in


bemoody:
You should not be able to attach an email address to an account without being logged in to that account.

AFAICT, it should be straightforward to mark "verify_email" as "login_required" and require that "associated_email.user" equals the logged-in user.

PR #902, opened 2020-03-04
Do not allow email address association without logging in

elfeto:
Here I force the user that added the email to be logged in before
verifying the email address.

Addressing #884


bemoody:
I would redirect back to 'edit_emails' rather than to 'project_home'.
That is where the person probably wants to go next, e.g. to set the
new address as primary or public.

I would also change the message to "The email address abc@example.com
has been verified."

elfeto:
Done :)

Issue #733, opened 2019-09-06
"URL for anonymous access" should include https://


bemoody:
In /console/published-projects/\*/\*/, the hidden URL for anonymous
access is displayed.  It should explicitly include the https:// prefix
in case somebody copies and pastes the text (as opposed to
right-clicking and copying the link target) into an email or whatever.

You could use notification.utility.get_url_prefix here.

PR #903, opened 2020-03-04
Fixes URL display for anonymous access #733

Lucas-Mc:
Replace `request.get_host`  with `url_prefix` as generated from the `notification.utility.url_prefix` function. This adds the correct version of either `http` or `https` to the displayed anonymous URL.


tompollard:
Looks good, thanks!

PR #906, opened 2020-03-05
Scripts for installing and running server in a VM

bemoody:
These are the scripts I use for testing the server in a VM.

Hopefully they work, and can be useful to other people. ;)


Lucas-Mc:
I will be writing notes as I go through the process of running the VM:

1. Maybe specify in the README that the download should be in the form of `debian-10.*.*-amd64-netinst.iso` so that they know it is the correct one.

2. It fails if `physionet-test.img` already exists since it refuses to overwrite it, this occurs if someone tries to run `./kvm-pntest-create` without the required packages... maybe clarify this or allow overwrite privileges? 

3. The `./kvm-pntest-create` script will fail if virtualization is not enabled in the BIOS.. maybe specify this in the README? 

4. I was having issues connecting to Port 10022. It was refusing the connection to I am trying to manually allow it. I'm not sure if this is a normal behavior or just something local to myself, but still something to consider. Any ideas how to fix this? I'll keep trying.

bemoody:
Thanks for testing!  I'll try to improve the scripts and documentation.

> 3. The `./kvm-pntest-create` script will fail if virtualization is not
> enabled in the BIOS.. maybe specify this in the README?

Hmm, that's not something I've ever had to do and I'm not sure how to
check it.  Does qemu give you any useful error message in that case?

(There are several different levels of hardware virtualization
capabilities on both Intel and AMD, and I can't keep them straight.
But you don't need the fancy stuff just to run kvm - any x86_64 CPU
should do except for very very old ones.  I've run kvm on machines
that don't support Xen hardware VMs, for instance.)

One thing you might need to do is check permissions on /dev/kvm.  The
distro may and probably should give you permission automatically, but
you might need to add yourself to the 'kvm' group (and then log out /
log back in.)

> 4. I was having issues connecting to Port 10022. It was refusing the
> connection to I am trying to manually allow it. I'm not sure if this is a
> normal behavior or just something local to myself, but still something to
> consider. Any ideas how to fix this? I'll keep trying.

Very strange.  Firewall?  Apparmor?  What distro are you using?

Lucas-Mc:
Okay, I have some updates. When running `./kvm-pntest-create` without virtualization enabled, the initial error message will be "Could not access KVM kernel module: No such file or directory". The suggested fix for this is to run `sudo modprobe kvm-intel` which fails since the operation can not be supported due to virtualization not being enabled. This seems like a tough work-around, maybe just make an error message which refers to this as a possibility?

Some other issues I have run into, especially during the `./ssh-install-pn-test-server HEAD` command is:
1. "/sbin/useradd: No such file or directory"
2. "su: command not allowed"
3. "tee: /tmp/physionet-tmp-id.pub: Permission denied"

I managed to fix some of these though it seems like if I fix one issue then one of the other ones appears again. For example, if I adjust `/etc/ssh/sshd_config` to include Port 10022 instead of the default Port 22 then access is allowed for Port 10022 though `hostfwd=tcp::10022-:22` fails since the port is already being used. Further, there is no firewall (ufw) as it is currently disabled.

Hope this helps and I'll keep trying to fix the port issue and anything that appears and will let you know if I get it all to run. 

EDIT: I am using Ubuntu 18.04.4 LTS

bemoody:
Thanks!  It's really helpful to see where you're getting stuck.

There shouldn't be any tinkering with sshd_config involved.  It sounds
like perhaps you are SSHing back to your local machine, not to the VM
- that's not what you want!

If kvm-pntest-create succeeds, it should have created a disk image
'physionet-test.img'.  You should be able to boot this image by
running kvm-pntest, and this should display a login prompt:

    Debian GNU/Linux 10 test-server tty1

    test-server login:

You should be able to log in here using your local username and the
password 'asdf'.

Next step: you should also be able to SSH to the VM.  Open a second
terminal and run 'ssh -p 10022 localhost'.  Again, you should be able
to log in using the password 'asdf'.

Once you've gotten that to work, THEN try running ssh-install-pn-test-server.

Lucas-Mc:
Ahhhhhh, this helped a lot! After reading that, I realized the error was occurring earlier than I had originally expected. After running `./kvm-pntest`, I realized this was actually an error instead of a warning: "qemu-system-x86_64: at most 2047 MB RAM can be simulated". This was preventing my system from running the server in the first place. This is because I am on a 32-bit system but I am currently in the process of fixing this to get it to run.

Is this something that should be added to the README? Like a section for 32-bit users? Or would it not be worth the extra effort?

Lucas-Mc:
UPDATE: I managed to get it to work by editing `-enable-kvm -smp 2 \ -m 2048` to `-no-kvm \ -m 2047` so maybe this can be added like in a `try/except` statement for 32-bit systems? Maybe it could be detected beforehand and the appropriate execution could be run?

bemoody:
Fantastic!  Must confess I haven't tried virtualization of any
kind on a 32-bit host in many years. :)

Decreasing the RAM is certainly no problem, I just find qemu's
default to be far too small most of the time.

Lucas-Mc:
It is nice that it works but I definitely wouldn't suggest it. Without kvm it is extremely slow and probably not worth the hassle but still a nice discovery!

bemoody:
Made some small fixes and improvements.  Thanks for your suggestions!

tompollard:
@Lucas-Mc if you are happy with the updates, please could you merge?

Issue #795, opened 2019-10-24
Purge unactivated accounts 


bemoody:
If an account is created and never activated, it should be deleted after a short time (let's say 7 days.)

@elfeto mentioned this in a comment to issue #761 but this is a separate issue.

PR #907, opened 2020-03-05
Purge unactivated accounts

elfeto:
A cronjob is created to run the command to purge the accounts.

I added a command to the Django management in which once runned
it will remove all unactivated accounts after 7 days of creation.

No notice to the user is given at this time.

Closing issue #795

### DIFF
--- a/deploy/production/etc/cron.d/physionet
+++ b/deploy/production/etc/cron.d/physionet
@@ -12,3 +12,4 @@
 
 # Remove expired Django sessions
 31 23 * * *  www-data  env DJANGO_SETTINGS_MODULE=physionet.settings.production /physionet/python-env/physionet/bin/python3 /physionet/physionet-build/physionet-django/manage.py clearsessions
+35 23 * * *  www-data  env DJANGO_SETTINGS_MODULE=physionet.settings.production /physionet/python-env/physionet/bin/python3 /physionet/physionet-build/physionet-django/manage.py purgeaccounts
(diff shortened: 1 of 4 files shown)


bemoody:
Sounds good!

Please update the doc string to say what this command does.

It would probably be better to filter the objects at the database level, e.g.

    today = date.today()
    limit = today - timedelta(days=7)
    user_list = User.objects.filter(is_active=False, join_date__lt=limit)
    for user in user_list:
        ...

I'm not sure about 'date.today()'.  Should it be
'timezone.now().date()' instead (from django.utils)?  At worst we're
off by one day, but still, we should try to be consistent :)

We should print out the list of deleted accounts using LOGGER.info()
or similar, rather than print().

elfeto:
I had print initially, and then I converted it to logger, I guess I missed one.

Also I fixed the doc string and set the date in the filter.

bemoody:
You added the condition to the filter, but you kept the "if"
statement, which is now redundant.

elfeto:
I should pay more attention to these things...

tompollard:
It will be nice to say goodbye to some of those spam users. As this involves deleting things, ideally it would be accompanied by some tests (e.g. the function should never delete active users; database integrity should still be good after deleting users; etc). I'm okay if these go into a new issue.

bemoody:
Agreed!  Does the test case in pull #910 look reasonable?

tompollard:
Yep, they look good to me! Thanks for doing this.

bemoody:
I'm satisfied.  OK with you @elfeto?

elfeto:
Could not have written a better test myself.

PR #908, opened 2020-03-05
Automatically run doctests

bemoody:
The doctest module provides a convenient way to write simple unit tests that also serve as API documentation (or documentation that also functions as a test case).

These changes add doctests (for the notification.utility and physionet.utility modules) to the Django testing framework so that 'manage.py' runs them automatically.

### DIFF
--- a/physionet-django/physionet/tests.py
+++ b/physionet-django/physionet/tests.py
@@ -1,6 +1,6 @@
 import doctest
 
-from notification import utility
+from physionet import utility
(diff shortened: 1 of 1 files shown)


tompollard:
I'd not come across [doctests](https://docs.python.org/3/library/doctest.html) before. This is great. It makes simple tests straightforward to add, and will hopefully encourage us to write better docstrings too.

PR #909, opened 2020-03-05
Display DOIs on the copyedit page of the console

tompollard:
This adds a tab to the copyedit page in the console that displays DOIs that have been registered:

Screen Shot 2020-03-04 at 16 59 25

### DIFF
--- a/physionet-django/console/templates/console/submission_info_card.html
+++ b/physionet-django/console/templates/console/submission_info_card.html
@@ -11,6 +11,11 @@
       <li class="nav-item">
         <a class="nav-link {% if passphrase %} active {% endif %}" id="anonymous-tab" data-toggle="tab" href="#anonymous" role="tab" aria-controls="anonymous" aria-selected="false">Anonymous Access</a>
       </li>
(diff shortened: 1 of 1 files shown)


elfeto:
Since you can toggle the auto DOI on and off, and set the DOI at the end after publish, you might want to do an `if project.core_project.doi. or project.doi` in the template to show the DOI tab, and the same with the DOI fields?  assuming one has a DOI and the other doesn't.

Here I am just speculating possible cases in which this might be empty. 

Screen Shot 2020-03-05 at 2 53 42 PM

tompollard:
Good suggestion, thanks. We now display a "DOI not available" message in those cases. e.g., if the core DOI exists but the project version DOI does not:

Screen Shot 2020-03-06 at 10 20 37

I'll make a new pull request to add a button for registering DOIs in cases where one doesn't already exist.

elfeto:
>I'll make a new pull request to add a button for registering DOIs in cases where one doesn't already exist.

You can add the DOI in the manage page for published projects. Dunno if you want to also add it here.

PR #910, opened 2020-03-05
Test case for pull #907

bemoody:
This provides a test case for the purgeaccounts command.


tompollard:
that was quick, thanks for doing this!

PR #912, opened 2020-03-05
Error handling when requesting project AWS access

### LINKED ISSUE #727: 'project_request_access' is not an active project view

elfeto:
There is/were an issue with the function to grant access to MIMIC using AWS cloud formation.

The reason is because of the AWS/China.

The only workaround given to us by AWS was to put a copy of MIMIC in a AWS owned by a Chinese collaborator. This fixes the error by ignoring it until a better solution is found.

Fixes #727

### DIFF
--- a/physionet-django/notification/templates/notification/email/notify_aws_access_request.html
+++ b/physionet-django/notification/templates/notification/email/notify_aws_access_request.html
@@ -1,10 +1,25 @@
 Dear {{ user.get_full_name }},
 
-You have requested access to use {{ project }} in AWS S3 Storage.
+You are receiving this email because you applied for access to {{ project }} on Amazon Web Services.
+
+{% if successful %}
+We are pleased to confirm that you have been granted access to use {{ project }} in AWS S3 Storage.
 
 The name of the bucket is: {{ data_access.location }}
 
 You can use the following command to copy the content of the bucket to your computer:
 aws s3 sync {{ data_access.location }} DESTINATION
+{% else %}
+We are sorry to say that we were unable to grant access to use {{ project }} in AWS S3 Storage at this time.
+
+Reasons might include:
+ - The service is not available in your location.
+ - An incorrect AWS ID was used.
+ - The service requested is no longer available.
+
+Please reapply for access if you are able to address the issue.
+
+If you think this was an error please contact {{ contact_email }}.
+{% endif %}
 
 {{ signature }}
--- a/physionet-django/notification/utility.py
+++ b/physionet-django/notification/utility.py
@@ -597,12 +597,14 @@ def notify_gcp_access_request(data_access, user, project):
     send_mail(subject, body, settings.DEFAULT_FROM_EMAIL,
               [email], fail_silently=False)
 
-def notify_aws_access_request(user, project, data_access):
+def notify_aws_access_request(user, project, data_access, successful):
     subject = 'PhysioNet Amazon Web Service storage access'
     body = loader.render_to_string(
         'notification/email/notify_aws_access_request.html', {
             'user': user,
             'project': project,
+            'successful': successful,
+            'contact_email': settings.CONTACT_EMAIL,
             'signature': email_signature(),
             'footer': email_footer(),
             'data_access': data_access
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -310,14 +310,14 @@
               {% if item.platform == 1 %} {# ID for AWS open data #}
                 <li>Access using AWS <a href="{{item.location}}">Open Data repository</a></li>
               {% elif item.platform == 2 %} {# ID for AWS cloud bucket #}
-                <li><a href="{% url 'project_request_access' project.slug project.version 2 %}">Request access</a> to the data using AWS command line tools: <pre class="shell-command">aws s3 sync {{item.location}} DESTINATION</pre></li>
+                <li><a href="{% url 'published_project_request_access' project.slug project.version 2 %}">Request access</a> to the data using AWS command line tools: <pre class="shell-command">aws s3 sync {{item.location}} DESTINATION</pre></li>
               {% elif item.platform == 3 %} {# ID for Google cloud bucket email #}
                 {% if project.gcp and project.gcp.sent_files %}
-                    <li><a href="{% url 'project_request_access' project.slug project.version 3 %}">Request access</a> to the files using the <a href="https://console.cloud.google.com/storage/browser/{{ project.gcp.bucket_name }}/">Google Cloud Storage Browser</a>. Login with a Google account is required.</li>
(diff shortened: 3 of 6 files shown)


elfeto:
The only thing I don't address from #727 is:
>should be located somewhere under /content/<project_slug>/, not /projects/<project_slug>/

All the `content` urls are in the search, and all the project urls by default contain the `/projects/`.
I can place it under the physionet/urls.py and there set the `/content/<project_slug>/`

bemoody:
You're going to need to be a lot more specific.  Please start from the beginning and assume I have no idea what a cloud formation is.


> Fix MIMIC AWS access grant using cloud formation
>
> This issue is because of AWS and support in China. There was an
> initial user that is using AWS in China and when he tried to access
> our project using AWS it gave an error.
>
> After we contacted our AWS contact they told us it was because of
> the reason above.

What reason?

elfeto:
I wrote wrong, its not  AWS cloud formation (that is the service inside AWS), but AWS OpenData.

AWS OpenData is the registry of data available in AWS.

https://registry.opendata.aws/mimiciii/

It is a requirement of China that the AWS accounts in the China regions be created by Chinese organizations. In order to make this dataset available in our AWS China regions, it would need to have a Chinese entity (e.g. a University) create an account in AWS China to be accessible in that region. 

There was a person from China trying to access MIMIC using his AWS ID which was made in China that raised an error because it's an invalid AWS for the project.

I looped @alistairewj into this because he knows more than me in this OpenData thing that we set up.

alistairewj:
Yeah it's fairly hazy to me as well as I don't want to spend eons learning all the bits of AWS, but Felipe has done a good job summarizing. Essentially we need a way of telling Chinese AWS users that they can't use the data via AWS. I reached out to MIT about this but never heard back so realistically we just won't be able to support Chinese AWS users in the foreseeable future.

bemoody:
Thanks for the explanation.  Nothing we can do about Amazon's /
Chinese government's policies other than to be as transparent as
possible.


This is rather messy:

    try:
        message = response.json()['message'].split(',')[0]
        # The message can differ if the ID is already there, or non-existent
        LOGGER.info("AWS message '{0}' for project {1}".format(
            response.json()['message'], project))
        return message
    except AttributeError:
        LOGGER.info("Invalid AWS ID: {0} on user: {1} for the current bucket "
            "policy.\nError: {2}".format(user.cloud_information.aws_id,
                                         user, response.json()['error']))
        message = "Access could not be granted."


I don't know what "message" is exactly, but I think it is probably not
a good idea to try to post-process it.  It's almost certainly not wise
to assume that:
- any AttributeError means the operation failed
- the lack of an AttributeError means the operation succeeded


Can you please give us an explicit example of the JSON document that
you receive:
- if the operation is successful, i.e. the user ID is added to the
  access list?
- if the user ID was already in the list?
- if the user ID doesn't exist in Amazon's system at all?
- if the user ID is not permitted to be added for political reasons?

Some links to documentation from Amazon would also be welcome. :)


As a general matter, when we make requests to an external service, we
need to validate the response - check that it matches the format we
expect for either a "success" or a "failure" case.

And if we are *not* able to validate the response, what I think we
ought to do is to save the complete request and response to a file
somewhere, that we can later look at to debug the problem.  I think it
would be wise to do this for DataCite and Google and any other
external services that we interact with.

elfeto:
Responses will defer....
> \- if the operation is successful, i.e. the user ID is added to the
  access list?

b'{"error": "None", "message": "Accounts [\'XXXYYYZZZ\'] have been added, accounts [] have been skipped since already exist, accounts [] have been deleted since policy is too large"}'
> \- if the user ID was already in the list?

b'{"error": "None", "message": "No new accounts to add"}'

OR if the limit was exceeded, then the first person in added will be removed and the new person appended to the list, so I will be GUESSING that the other response would be:
b'{"error": "None", "message": "Accounts [\'AAABBBCCC\'] have been added, accounts [] have been skipped since already exist, accounts [\'XXXYYYZZZ\'] have been deleted since policy is too large"}'
> \- if the user ID doesn't exist in Amazon's system at all?

Can be either one of these depending on the ID
b'{"error": "Unexpected error: An error occurred (MalformedPolicy) when calling the PutBucketPolicy operation: Invalid principal in policy", "message": null}'
b'{"error": "None", "message": "No new accounts to add"}'
> \- if the user ID is not permitted to be added for political reasons?

Only one I have seen so fat is
b'{"error": "Unexpected error: An error occurred (MalformedPolicy) when calling the PutBucketPolicy operation: Invalid principal in policy", "message": null}'
But that doesn't mean there aren't any others.

bemoody:
Oh wow

bemoody:
Is there another URL we could poll to check whether a particular user
ID has access?

elfeto:
There is a way, but it would be adding more dependencies and installing the AWS SDK called Boto3 and more credentials.

Also up to my understanding there is no easy easy way to validate or check if an account exists.

elfeto:
ping on this? :)

bemoody:
This looks better than what we had before, thanks!

However, it looks like the return value from
grant_aws_open_data_access will be None if there is an unknown error.

Then, the message returned is always displayed as a "success" message
and the email notification is always sent, regardless of whether it
was successful or not. :/  I notice that the email template itself is
vague, it tells you that you have "requested access" but not whether
it was granted.

The "open_data_access" part means this only applies to open access
projects, correct?

elfeto:
> The "open_data_access" part means this only applies to open access
projects, correct?

Yes, The probability is that there wont exist many.

bemoody:
>> The "open_data_access" part means this only applies to open access
>> projects, correct?
>
> Yes, The probability is that there wont exist many.

Wait a minute, looking back at your original message, you said this
was about MIMIC.  MIMIC isn't open access.

That makes a difference because I am not comfortable about giving
people access to restricted data via a third-party service where we
have no way of reliably tracking who has been given access.

elfeto:
>That makes a difference because I am not comfortable about giving
people access to restricted data via a third-party service where we
have no way of reliably tracking who has been given access.

We can verify who has access at any time we want checking the online page or with the AWS API.
Only credentialed people can request this access.

bemoody:
What does the API tell you?  It looks to me like people can put
anything they want (up to 60 characters) in their "AWS ID" field, they
can change it whenever they want, request access as often as they
want, and there is no way to tie an "AWS ID" back to the credentialed
user.

tompollard:
@bemoody As far as I can see these changes are an improvement on the current state (i.e. better email; better error handling), so are you good to merge? Outside this thread, we will need to discuss the idea of AWS integration more generally.

PR #913, opened 2020-03-06
Add DOI validation on published projects

### LINKED ISSUE #911: Set DOI in published project bypasses validation
You can set a DOI in `http://localhost:8000/console/published-projects/demobsn/1.0/` and it bypasses all validations.

I can create: https://doi.org/132121212121212

elfeto:
The DOI in published project is not validated any anything can be entered as a DOI at the moment.

Here I use the same validator used across all DOI related usage.
Closes #911


tompollard:
thanks, looks good to me.

Issue #777, opened 2019-10-09
"Load more" paging for news items displayed at the top of projects


tompollard:
Some projects have lots of associated news items (e.g. https://physionet.org/content/challenge-2019/1.0.0/). It is awkward to have to scroll through so many news items just to get to the content itself, so it may be useful to add a "Load more" button after the first three or four items.

PR #915, opened 2020-03-06
Adds 'Load More' button for news #777

Lucas-Mc:
Adds button which gives users the option to view more news or collapse after finishing reading. This prevents the entire news section from being displayed without the user's preference and therefore reduces the starting page length. The button is first a dropdown which says "View all announcements" then converts to a dropup which says "View less" after clicked. If the "View less" dropup is clicked, then the button will convert back the initial dropdown.

### DIFF
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -53,13 +53,26 @@
 
   <!-- Latest news and announcements -->
   {% if news %}
-  <div class="alert alert-primary" role="alert">
-    {% for new in news %}
-      <strong>{{ new.title|safe }}</strong>
-      <em>({{ new.publish_datetime }})</em>
-      {{ new.content|safe }}
-    {% endfor %}
-  </div>
+    <div class="alert alert-primary" role="alert">
+      {% for new in news %}
+        {% if news|length > 2 %}
+          {% if forloop.counter == 3 %}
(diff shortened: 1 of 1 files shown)


elfeto:
Looks good, there is only one minor thing.

If you have 1 or 2 news, then the button is not needed. Since its there, it implies there are more news to load.

Lucas-Mc:
Ahhhhhh true, nice catch! I was only focused on the current issue case. I will make the edits and commit the new code for an updated review.

bemoody:
I'd think it would be better done as a <details>/<summary>.

    <p>Item 1</p>
    <p>Item 2</p>
    <details>
      <summary>More news</summary>
      <p>Item 3</p>
      <p>Item 4</p>
    </details>

Lucas-Mc:
Okay, thanks @bemoody for the suggestion. After a quick fix, here is the updated code. I think the `details`/`summary` tags look nice and are way less bulky than the original button.

elfeto:
On 24c0ce9 you removed the css and js that are in 23e061b.

I dont know if it was intentional or not because you left the 
`<script src="{% static 'custom/js/dropdown-dropup.js' %}"></script>`

It still works tho.

Lucas-Mc:
The script tags should be removed now since it was for the button. I'll remove it and push again.

PR #916, opened 2020-03-09
Site has been moved from alpha.physionet.org to physionet.org

shulim:
Hi, this is my first pull request, so I'm keeping it simple!

This is a small change to remove "alpha" from the physionet URL in the readme.


tompollard:
The Shippable build wasn't triggered for reasons that aren't clear to me (perhaps something to do with @shulim being only recently added to the permissions group?). I pushed an empty commit to trigger a build.

PR #917, opened 2020-03-09
Version logic

elfeto:
Here I fix the version logic comparison on publishing the project.

Also, since the version order would be messed up, I am sorting all and re-setting all the versions.

### DIFF
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -1630,7 +1630,8 @@ class PublishedProject(Metadata, SubmissionInfo):
 
     def set_version_order(self):
         """
-        Order the versions by number and set a correct version order.
+        Order the versions by number.
+        Then it set a correct version order and a correct latest version
         """
         published_projects = self.core_project.get_published_versions()
         project_versions = []
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -8,6 +8,7 @@ import pdb
 import pytz
 import stat
 import logging
+from distutils.version import StrictVersion
 
 import bleach
 import ckeditor.fields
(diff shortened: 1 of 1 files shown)


bemoody:
Sounds good; this basically agrees with my feeling about how version
numbering ought to be handled.


If I understand correctly, this sets the version_order at the time the
new project version is created (by the author submitting the "new
version" form).  But that might be wrong, because the new project's
version number might be changed before publication.  Or the new
project might be deleted without being published.

My first thought would be, is it actually *necessary* for an
ActiveProject to have a version_order field at all?  We could just
leave that field at zero, or remove the field entirely, until the
project is published.

But even if we do want an ActiveProject to have a version_order field,
I think it would be best to avoid *modifying* PublishedProject objects
in any way, until the new version is published.  One way to do this
would be to assign only odd numbers to published projects:

- published version "1.0.0" has version_order = 1
- published version "1.1.0" has version_order = 3
- published version "2.0.0" has version_order = 5

Then an active project with version "1.2.0" would have version_order =
4 and one with version "3.0.0" would have version_order = 6.


However we go about this, I think we probably want to have a single
function, to be called at publication time, that sets both the
version_order and is_latest_version fields for PublishedProjects.

elfeto:
While I was doing this PR I thought so, and I was doing to do an issue to then create the thing that you mention...

Now I am keeping the same logic we had in the creation of a new version. (ignoring it since it doesn't make a difference in the end.

Now I am only doing the alter once the project has been published.
I also moved the is latest version into the function to properly say what is the latest version.

bemoody:
The 'has_other_versions' logic seems a bit strange too.

It looks like has_other_versions is being set for all *except* the
newly-published version.

>>> prj = project.models.PublishedProject.objects.filter(slug='wctecgdb')
>>> [(p.version, p.version_order, p.is_latest_version, p.has_other_versions) for p in prj]
[('1.0.0', 0, False, True), ('1.0.1', 1, True, False)]

Presumably it should be set for all versions, unless there is only
one.  Then we can do away with logic like "if project.version_order or
project.has_other_versions".

elfeto:
Fixed :)

PR #918, opened 2020-03-09
Secure token generation for AssociatedEmail

bemoody:
Note: this is a security fix which is already deployed on the production server (i.e. this pull request is to bring the dev branch into sync with production/staging.)  But I'm happy to make further changes if there are any issues before merging into dev.

I'll make a couple of followup pull requests to deal with related issues:

- changing the token generation for *initial* email verification, to avoid weird corner cases with password resetting

- automatically removing AssociatedEmails once the time limit has expired

----

To "verify" a new AssociatedEmail address, the person needs to be able to prove that they actually are able to read email sent to that address.

What we are currently doing, however, is sending a *password reset* token to that email address.

Why is that wrong?

- Well, for one thing, a password-reset token is a magic token that anyone can use to reset that person's password.  That's what it's for, and since there is no salting, the same token can also be used for /reset-password/.  So if you mistype an email address, you have just given somebody else the ability to take control of your account.

- Password-reset tokens become invalid whenever you log in or change your password, so if you do either of those things after "adding" an AssociatedEmail and before "verifying" it, you'll find the link mysteriously fails to work.

- Password-reset tokens are completely independent of the AssociatedEmail (again, no salting), so they don't do anything to prove ownership of an address.

Thus, at present, anyone can add any email address they want to their PhysioNet account.  All they need to do is:

1. Create an account with a valid email address.

2. Go to /settings/emails/ and add two new email addresses, one real (bob@example.com) and one fake (bob@example.org).

3. Two emails will be sent - one to bob@example.com and one to bob@example.org.  For example:

    Subject: PhysioNet Email Verification
    To: bob@example.com
    ...
    http://localhost:8000/verify/OA/5ek-bcc31dd732fee419500d/

    Subject: PhysioNet Email Verification
    To: bob@example.org
    ...
    http://localhost:8000/verify/OQ/5ek-bcc31dd732fee419500d/

4. Given the first URL, it's easy to guess the second - AssociatedEmail ids are assigned sequentially.

We could try to come up with a cryptographic token based on a modified version of django.contrib.auth.tokens, or using django.core.signing... or we could just generate a random string and stick it in the database.

The latter seems to be what other people do (e.g. https://pypi.org/project/django-simple-email-confirmation/) and generally seems like a better idea than trying to roll one's own crypto.


elfeto:
Looks good.

Issue #926, opened 2020-03-12
Add zipcode to credentialing email sent to applicant


tompollard:
Apparently the email sent to applicants who request credentialed access is missing the zipcode:

> - make sure that the e-mail message in the "Send e-mail to applicant" mailto link includes applicant's zip/postcode.  Somehow that got left out.

Issue #863, opened 2020-02-04
Bad license URL in schema metadata


tompollard:
While looking at https://github.com/MIT-LCP/physionet-build/issues/862, I noticed that there is a bad URL listed for the license in our schema.org metadata. Correct URLs are listed at: https://physionet.org/about/publish/#licenses (licenses and DUAs need work more generally, but these are good for now).

```json
<!-- https://schema.org/ metadata for discovery -->
<script type="application/ld+json">
{
  "@context": "https://schema.org/",
  "@type": "Dataset",
  "name": "Paediatric Intensive Care database",
  "description": "The Paediatric Intensive Care (PIC) database is a large paediatric\u002Dspecific, single\u002Dcenter, bilingual database comprising information relating to children admitted to critical care units at a large children\u0026rsquo\u003Bs hospital in China. The database is de\u002Didentified and includes vital sign measurements, medications, laboratory measurements, fluid balance recordings, diagnostic codes, demographic information, and more. The data are publicly available after\u0026nbsp\u003Bcredentialing, which includes completion of a training course on research with human subjects and signing of a data use agreement mandating responsible handling of the data and adherence to the principle of collaborative research. The PIC database builds upon the success\u0026nbsp\u003Bof the\u0026nbsp\u003Bwidely used MIMIC (Medical Information Mart for Intensive Care) database, and extends the approach into the field of paediatric critical care. The database has many unique\u0026nbsp\u003Bcharacteristics which\u0026nbsp\u003Bsupport academic and industrial research including the development of\u0026nbsp\u003Bmachine learning algorithms, clinical decision support tools, quality improvement initiatives, and international data sharing.",
  "version": "1.0.0",
  "license": "https://github.com/MIT\u002DLCP/license\u002Dand\u002Ddua/tree/master/drafts",
  "datePublished" : "Dec. 1, 2019",
  "url": "https://physionet.org/content/picdb/1.0.0/",
  
  "identifier": "https://doi.org/10.13026/bcbd\u002D4t11",
  
  "creator": [
  
    {
      "@type": "Person",
      "givenName": "Haomin",
      "familyName": "Li",
      "name": "Haomin Li"
    },
  
    {
      "@type": "Person",
      "givenName": "Xian",
      "familyName": "Zeng",
      "name": "Xian Zeng"
    },
  
    {
      "@type": "Person",
      "givenName": "Gang",
      "familyName": "Yu",
      "name": "Gang Yu"
    }
  
    ],
[...]

PR #930, opened 2020-03-12
Adds zipcode to credentialing email #926

Lucas-Mc:
This fixes the zipcode not showing up when automatically generating the credentialing email to be sent to applicants. After further investigation, this was due to the zipcode not being included in the `demo-user.json` file as well as the zipcode field not being included in the `mailto_contact_applicant.html` code.

### DIFF
--- a/physionet-django/notification/templates/notification/email/mailto_contact_applicant.html
+++ b/physionet-django/notification/templates/notification/email/mailto_contact_applicant.html
@@ -10,6 +10,7 @@ Organization Name: {{ application.organization_name }}
 Job title or position: {{ application.job_title }}
 City: {{ application.city }}
 State/Province: {{ application.state_province }}
+ZIP/postal code: {{ application.zip_code }}
 Country: {{ application.get_country_display }}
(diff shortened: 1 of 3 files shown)


tompollard:
Thanks!   The demo user shouldn't really matter (except for testing), but I don't think it hurts to include the zip. 

It looks like the PR includes some changes to `physionet-django/project/templates/project/schema_metadata.json`. Is this intentional?

Lucas-Mc:
Hey @tompollard , no this was not intentional, the branch must have been created on top of un-merged code from that previous pull request #863 . I am looking at a way to fix it now.

bemoody:
Lucas, it looks like you accidentally made this change on top of
another feature branch.

The best way to fix this is to do something like:

git fetch origin
git checkout origin/dev
git checkout -b fix_applicant_email-v2
git cherry-pick 795cb207974e939926a14f5a240d39855ec2ce5a

and then push your new fix_applicant_email-v2, which ONLY includes the
change from commit 795cb207974e939926a14f5a240d39855ec2ce5a.

PR #931, opened 2020-03-13
Adds zipcode to credentialing email (*Updated*) #926

Lucas-Mc:
This fixes the zipcode not showing up when automatically generating the credentialing email to be sent to applicants.


tompollard:
thanks!

Issue #927, opened 2020-03-12
Minor updates to the credentialing management page


tompollard:
Request for a couple of changes to KP's credentialing tool (https://physionet.org/console/complete-credential-applications/):

- do not show the "contact reference" link/button if there is no reference listed.  Or, maybe better in terms of page display: show the link/button but make it non-functional.
- only for applications where the supervisor is "*@udacity.com", append a line to the e-mail message to supervisor that gives the applicant's e-mail address, as in the example below.

PR #932, opened 2020-03-13
Disables contact reference button #927

Lucas-Mc:
This disbales the contact reference button if no reference is known or given at the time. The button will still appear but will be unable to be clicked.

### DIFF
--- a/physionet-django/console/templates/console/complete_application_display.html
+++ b/physionet-django/console/templates/console/complete_application_display.html
@@ -10,9 +10,15 @@
     <form action="" method="post" class="form-signin"> <!-- empty action generates error! -->
       <p>
 	{% csrf_token %}
-	<button name="contact_reference" value="{{ application.id }}" type="submit" style="margin-right: 1em;">
-	  Contact Reference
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks Lucas, it sounded like Ken would like to \*hide\* the button (rather than disable) when in cases where a reference isn't required, so the else clause should perhaps be empty.

My understanding of `known_ref` is that it indicates a reference who has been contacted before. I think one of the other `CredentialApplication` fields may be better (maybe reference_email or reference_category. preferably one that is always populated if a reference is provided).

Lucas-Mc:
Thanks for the review @tompollard! This is what he said in that email: "Or, maybe better in terms of page display: show the link/button but make it non-functional.". Also, thanks for clarifying, I wasn't totally sure so it sounds like the reference_email may just be the best way to go.

tompollard:
> "Or, maybe better in terms of page display: show the link/button but make it non-functional."

Okay, cool, missed that!

tompollard:
looks good, thanks :)

PR #935, opened 2020-03-13
Fix style of checkbox and other non-text <input>

bemoody:
CSS for input elements is badly broken - elements such as checkboxes are made into "display:block;width:100%" - and rather than fixing the broken rule, I see folks trying to work around it (see pull #897) or avoiding using checkboxes entirely (see pull #855).

Our CSS is awful but let's try to make it better!

### DIFF
--- a/physionet-django/static/custom/css/form-control-input.css
+++ b/physionet-django/static/custom/css/form-control-input.css
@@ -1,6 +1,13 @@
 /* Styling of inputs to have the same bootstrap look as form-control elements */
 
-input, select, django-ckeditor-widget, textarea {
+input[type=email],
+input[type=file],
+input[type=password],
+input[type=text],
+input[type=url],
+select,
+django-ckeditor-widget,
+textarea {
   display: block;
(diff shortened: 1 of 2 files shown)


tompollard:
A couple of minor things that I noticed when testing this:

1. When using the console (e.g. http://127.0.0.1:8000/console/users/active/), the icon to the right of the search bar is shifted onto the following line:

Screen Shot 2020-03-13 at 19 23 38

2. The search box has changed in the console pages (not a big issue, but I think it looked a little cleaner before than after):

Before:

Screen Shot 2020-03-13 at 19 25 32

After:

Screen Shot 2020-03-13 at 19 25 36

Other than that, seems good to me!

bemoody:
Thanks for checking.

I now see that the nav bar is broken for me at > 1100px but I think
that is unrelated.  These two commits should fix the issues you're
seeing, let me know if they don't.

tompollard:
Thanks, looks good!

PR #936, opened 2020-03-16
Fix check for gcp access

elfeto:
Right now the check will always be a return False because it's doing a comparison of 3 and 4 against an object.

Also, there can be multiple access types like BigQuery and Storage and the person is requesting one. Now I am checking to check if that specific one exists.

### DIFF
--- a/physionet-django/project/utility.py
+++ b/physionet-django/project/utility.py
@@ -304,12 +304,6 @@ def grant_gcp_group_access(user, project, data_access):
 
     try:
         group_members = service.members()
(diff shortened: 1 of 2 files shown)


tompollard:
Felipe pointed out that the API call for group members appears to be limited to 200 results:

https://developers.google.com/admin-sdk/directory/v1/reference/groups/list

> maxResults | integer | Maximum number of results to return. Acceptable values are 1 to 200, inclusive. (Default: 200)

PR #939, opened 2020-03-16
Test cases for publishing projects

bemoody:
This adds some test cases for the "create project" and "new project
version" forms, which were not being tested at all AFAICT.

It adds some test cases for the publication workflow when publishing a
*new version* of an existing project.  It tests the version ordering
logic that is now implemented by pull #917.  Unfortunately, this is
currently broken due to the issue I raised in pull #938.  After
merging that fix, these tests should pass.

### DIFF
--- a/physionet-django/console/test_views.py
+++ b/physionet-django/console/test_views.py
@@ -23,6 +23,14 @@ class TestState(TestMixin):
 
     """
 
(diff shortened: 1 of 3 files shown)


tompollard:
restarted tests...

tompollard:
thanks, looks good to me.

PR #942, opened 2020-03-17
Add an email notification for failed tasks

### LINKED ISSUE #937: Notify about background task errors
If a background task fails it gets automatically retried.

If a task fails permanently (after 25 tries?), or perhaps earlier, it would be good to send an email so we know there is a problem.

elfeto:
When a task fails it sends a signal.
There are two signals, rescheduled and failed.

Here I capture both signals and send emails on each one.

Also, the default MAX_ATTEMPTS for this is by default 25.
I am reducing this to 5. In theory, if it fails once it will fail
all 5 times because the code should not change that fast.

Closes #937

### DIFF
--- a/physionet-django/console/tasks.py
+++ b/physionet-django/console/tasks.py
@@ -1,8 +1,10 @@
 import inspect
 import json
 
-from background_task.models import Task
+from background_task.models import Task, task_failed, task_rescheduled
+from django.dispatch import receiver
 
+import notification.utility as notification
 
 _model_tasks = {}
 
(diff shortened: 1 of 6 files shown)


elfeto:
To test this, I altered the function "make_checksum_background", and added pdb.set_trace().
In the PDB, I just exited it and it raised an error (can be done by typing 'q' or ctrl+c). Then I  waited.

bemoody:
Great, thanks for doing this!  Some small things:

- task_rescheduled_handler does not need to call pdb.set_trace

- send mail to the admins (technical@); you can use django.core.mail.mail_admins

- Typo in notify_failed_task.html: "The task has he been marked as complete"

This doesn't really belong in console/views.py, maybe it could be
placed in console/tasks.py instead.

elfeto:
Didn't notice I left a stranded pdb there... ty :) 

I remove the pdb, am using mail_admins, fixed the typo and move the files into tasks.py

bemoody:
Thanks!

PR #944, opened 2020-03-17
All published projects should have a DOI

tompollard:
All published projects should have a DOI, so this change: 

- Removes the "Auto DOI" flag from the editorial decision form.
- Sets the `auto_doi` flag to True unless `DATACITE_PREFIX` is not configured in the settings.

As far as I can see, it isn't currently possible to select the auto DOI field using the form anyway, so this pull request shouldn't change anything functionally.

I'm unclear about the purpose of the `auto_doi` flag, so we may want to clarify this with a comment in the model. I don't see any issues with keeping it for now.

As @elfeto mentioned, there may be situations where we want to manually assign a DOI (e.g. if a project with an existing DOI is migrated). The field might be helpful in those cases (it can be set to False using the shell for now).

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -77,6 +77,8 @@ class EditSubmissionForm(forms.ModelForm):
 
         labels = EditLog.COMMON_LABELS
 
+        auto_doi = forms.BooleanField(required=False, initial=True)
+
         widgets = {
             'soundly_produced': forms.Select(choices=YES_NO_UNDETERMINED),
             'well_described': forms.Select(choices=YES_NO_UNDETERMINED),
@@ -86,7 +88,8 @@ class EditSubmissionForm(forms.ModelForm):
             'no_phi': forms.Select(choices=YES_NO_UNDETERMINED),
             'pn_suitable': forms.Select(choices=YES_NO_UNDETERMINED),
             'editor_comments': forms.Textarea(),
-            'decision': forms.Select(choices=SUBMISSION_RESPONSE_CHOICES)
+            'decision': forms.Select(choices=SUBMISSION_RESPONSE_CHOICES),
+            'auto_doi': forms.HiddenInput()
         }
 
     def __init__(self, resource_type, *args, **kwargs):
@@ -97,8 +100,9 @@ class EditSubmissionForm(forms.ModelForm):
         super().__init__(*args, **kwargs)
         self.resource_type = resource_type
 
+        self.fields['auto_doi'].disabled = True
+
         if not settings.DATACITE_PREFIX:
-            self.fields['auto_doi'].disabled = True
             self.initial['auto_doi'] = False
 
         # This will be used in clean
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -1810,8 +1810,7 @@ class EditLog(models.Model):
         'pn_suitable': 'Is the content suitable for PhysioNet?',
         'editor_comments': 'Comments to authors',
         'no_phi': 'Is the project free of protected health information?',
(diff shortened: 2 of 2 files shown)


elfeto:
> As far as I can see, it isn't currently possible to select the auto DOI field using the form anyway, so these changes shouldn't change anything functionally.

Right now if I click on the checkbox, there is no DOI created on publish.

>As @elfeto mentioned, there may be situations where we want to manually assign a DOI (e.g. if a project with an existing DOI is migrated). The field might be helpful in those cases (it can be set to False using the shell for now).

I don't think its a good idea for an editor to commit something, to then change it in the background if the possibility exists to do it in the browser.

>Sets the auto_doi flag to True unless DATACITE_PREFIX is not configured in the settings.

This would mean that after this PR, I need to change my environmental file if I don't want to automatically create DOI while testing.

tompollard:
> Right now if I click on the checkbox, there is no DOI created on publish.

Are you able to click on the checkbox? It doesn't appear to be active.

> I don't think its a good idea for an editor to commit something, to then change it in the background if the possibility exists to do it in the browser.

If the situation comes up where we need to set the flag to False, then we are doing something out of the ordinary. I agree it's not ideal using the shell, and we can add a button to the console later if necessary.

> This would mean that after this PR, I need to change my environmental file if I don't want to automatically create DOI while testing.

I guess so, if you wanted to publish a project without registering a test DOI. As the live system is dependent on the API too, is this an issue?

elfeto:
> Are you able to click on the checkbox? It doesn't appear to be active.

I am, and I do. If it doesn't appear to be active, then u have an incomplete .env

>If the situation comes up where we need to set the flag to False, then we are doing something out of ordinary. I agree it's not ideal using the shell, and we can add a button to the console later if it is needed.

That might be the case, but an example not out of the ordinary: testing.
In theory, one should not use the shell. The only reasons I use it is to debug something that can't be debugged otherwise. 

>I guess so, if you wanted to publish a project without registering a test DOI. As the live system is dependent on the API, is this an issue?

If I understand ur sentence correctly...
I would say a better statement for what you wrote is, the live system is NOT and should not be dependent on any external service including this API.

Altering the env file an issue? no, but maybe a nuisance to have to alter the env file to do X and edit back to do Y.

tompollard:
> Are you able to click on the checkbox? It doesn't appear to be active.

I've tried with several different browsers (chrome; firefox; safari) and it wasn't clickable. @Lucas-Mc seemed to have the same issue too. I'll message you separately to see if we can work out what's going on.

elfeto:
>Are you able to click on the checkbox? It doesn't appear to be active.

The problem was that the variables were missing on the live site. Hence the checkbox was disabled and set to false. I just added them.

tompollard:
Thanks for sorting out the checkbox.

Essentially we need \*all\* published projects to have a DOI, so we shouldn't be allowing editors to switch this off.

If there is a scenario where `auto_doi` needs to be switched to False, it will be a technical issue (not an editorial issue) that can dealt with as appropriate. 

Wanting to be able to test the system with `auto_doi = False` is not a good reason to display the field in the editorial form. If you have a suggestion for testing that would work for you, let me know!

elfeto:
Well, if you say it like that, then this works.

I guess I can alter the .env file. Hopefully there wont be many DOI branches :)

tompollard:
Thanks, and I appreciate your flexibility on this :) If it becomes bothersome for testing, we can come with a different approach for switching it off.

> Hopefully there wont be many DOI branches :)

Just one more for now! It would be good to add DOI management buttons (register/update) to the management form at: http://localhost:8000/console/published-projects/demobsn/

Issue #510, opened 2019-05-20
Caching of static resources


bemoody:
"Static" resources (css, js, images) may change from time to time.

We *want* these to be cached - we don't want to slow down page loading when the file hasn't changed.

But sometimes we do need to make changes, & in some cases we may need to make changes to HTML and CSS simultaneously.

Is there a good and easy way with Django to ensure that clients get an up-to-date version when necessary?

lbulgarelli:
The way this is usually done is by versioning the files.

I’ve looked into it a while ago and found this:

https://docs.djangoproject.com/en/dev/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage

Which seems the proper django way to do it.

Lucas-Mc:
I just implemented a cache on [the waveform annotator](https://github.com/MIT-LCP/waveform-annotation) using [Redis](https://redis.io) which works great with Django and not too hard to set up if anyone is interested?

Haven't messed around with its view-specific caching but I know it's very flexible and seems intuitive. It would require a Redis service to run on the server just so Django can communicate with it (usually `localhost:6379`) if anyone is opposed?

PR #945, opened 2020-03-17
Adds ability to cache static files  #510

Lucas-Mc:
Utilizes the `ManifestStaticFilesStorage` Django subclass to enable caching of static files. This works by appending a MD5 hash to the static file name in order to properly version each instance of the file.

### DIFF
--- a/physionet-django/physionet/settings/production.py
+++ b/physionet-django/physionet/settings/production.py
@@ -56,3 +56,5 @@ STATIC_ROOT = '/data/pn-static'
 if RUNNING_TEST_SUITE:
     MEDIA_ROOT = os.path.join(MEDIA_ROOT, 'test')
     STATIC_ROOT = os.path.join(STATIC_ROOT, 'test')
+
+STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.ManifestStaticFilesStorage'


bemoody:
Sounds nice.  What does it do exactly?  Please give some specific
examples, before and after: what path is a file installed under, what
URL is presented to the client?

When are these hashes calculated, and of what data exactly?  How does,
for example, the "{%static%}" tag know what hash to use?

What needs to change in our deployment process, to handle existing
static files, new static files, old static files that are no longer
used?

What needs to be done to test this locally?

elfeto:
So I went the lazy way out and tested in on staging.

I set:
 - DEBUG to False
 - STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.ManifestStaticFilesStorage' 

Then when I ran `python manage.py collectstatic --noinput` it kinda exploted.


```
(physionet) pn@physionet-staging:/physionet/physionet-build/physionet-django$ python manage.py collectstatic --noinput
Post-processing 'lightwave/css/jquery-ui.min.css' failed!

  File "manage.py", line 22, in <module>
    execute_from_command_line(sys.argv)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 188, in handle
    collected = self.collect()
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 134, in collect
    raise processed
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/contrib/staticfiles/storage.py", line 294, in _post_process
    content = pattern.sub(converter, content)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/contrib/staticfiles/storage.py", line 195, in converter
    force=True, hashed_files=hashed_files,
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/contrib/staticfiles/storage.py", line 132, in _url
    hashed_name = hashed_name_func(*args)
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/contrib/staticfiles/storage.py", line 343, in _stored_name
    cache_name = self.clean_name(self.hashed_name(name))
  File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/contrib/staticfiles/storage.py", line 93, in hashed_name
    raise ValueError("The file '%s' could not be found with %r." % (filename, self))
ValueError: The file 'lightwave/css/"images/ui-icons_555555_256x240.png"' could not be found with <django.contrib.staticfiles.storage.ManifestStaticFilesStorage object at 0x7fabeca5e6d8>.
```

Obviously, I set things the way they were and all is working.

tompollard:
@elfeto not lazy, efficient! seems like a good idea to try something like this on staging.

Lucas-Mc:
I found the bug. See here:
Screen Shot 2020-03-18 at 9 17 05 AM
It appears that it is parsing the commented `images/ui-icons_555555_256x240.png` file as seen in the comment in Line 4 of `jquery-ui.min.css` as seen here. https://github.com/MIT-LCP/physionet-build/blob/c1af3d66ba16da8755e67085b8cf093b9a403599/physionet-django/lightwave/static/lightwave/css/jquery-ui.min.css#L4 So the solution must be to fix the regex as noted in the attached image, but I still need to look further into which one it is exactly. 

Note: The full thread can be seen [here](https://code.djangoproject.com/ticket/21080).

Lucas-Mc:
Closing this for now until a working solution is found and more investigation has been done about caching files in Django.

PR #947, opened 2020-03-19
Fix legacy publish function

elfeto:
At some point in time the resource type was a number and it was
changed to a ForeignKey to ProjectType.

The chages were done in all places but the project legacy model


elfeto:
To test this run the following:
```
./manage.py resetdb && ./manage.py loaddemo && ./manage.py loaddata pbank

./manage.py shell

for item in LegacyProject.objects.all(): item.publish()
```

If no errors were raised, then there is no problems...

bemoody:
Thanks!  For some reason that takes a suprisingly long time to run,
but it works.

PR #948, opened 2020-03-19
remove all explicit relative imports

elfeto:
Convert explicit relative imports into more readable imports


tompollard:
nice, thanks Felipe.

PR #949, opened 2020-03-19
Raise 404 if the object is not found in the console views

### LINKED ISSUE #849: Error if "ActiveProject matching query does not exist" (user clicks an expired link)
If a user saves a link to a project that is in submission, and then clicks this link after the project has been processed (accepted/rejected/etc), then the link generates an error.

```
Internal Server Error: /console/submitted-projects/SLUG/info/

DoesNotExist at /console/submitted-projects/SLUG/info/
ActiveProject matching query does not exist.

Request Method: GET
Request URL: https://www.physionet.org/console/submitted-projects/SLUG/info/

Traceback:

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/handlers/exception.py" in inner

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/contrib/auth/decorators.py" in _wrapped_view

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/contrib/auth/decorators.py" in _wrapped_view

File "./console/views.py" in submission_info
 216.     project = ActiveProject.objects.get(slug=project_slug)

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/db/models/manager.py" in manager_method
 82.                 return getattr(self.get_queryset(), name)(*args, **kwargs)

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/db/models/query.py" in get
 408.                 self.model._meta.object_name

Exception Type: DoesNotExist at /console/submitted-projects/SLUG/info/
Exception Value: ActiveProject matching query does not exist.
```

elfeto:
This might be a band-aid on a bigger problem.

We treat most models like they exist, but the information can change.
To fix this we either do this in most get functions or we create
a method that returns said object or raises 404.

Closes #849

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -75,10 +75,13 @@ def handling_editor(base_view):
     @login_required
     def handling_view(request, *args, **kwargs):
         user = request.user
-        project = ActiveProject.objects.get(slug=kwargs['project_slug'])
-        if user.is_admin and user == project.editor:
-            kwargs['project'] = project
-            return base_view(request, *args, **kwargs)
+        try:
+            project = ActiveProject.objects.get(slug=kwargs['project_slug'])
+            if user.is_admin and user == project.editor:
+                kwargs['project'] = project
+                return base_view(request, *args, **kwargs)
+        except ActiveProject.DoesNotExist:
+            raise Http404()
         raise Http404('Unable to access page')
     return handling_view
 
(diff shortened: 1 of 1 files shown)


tompollard:
Seems like a good idea to me. @bemoody are you okay with this?

bemoody:
Yeah, haven't tested it but it looks good!

PR #950, opened 2020-03-19
Allow titles to start with a number

### LINKED ISSUE #850: New validation rules do not allow title to begin with a number
It looks like the validation rules added in https://github.com/MIT-LCP/physionet-build/pull/832 do not allow a project title to begin with a number. Is this intentional?

Screen Shot 2020-01-13 at 16 23 21

There are existing projects that begin with a number (e.g. https://physionet.org/content/bhi-2018-challenge/), so I would prefer to stick with the original behaviour if possible.

If we do decide to prevent titles that begin with a number, this rule should be made clear in the error text.

elfeto:
This closes #850

### DIFF
--- a/physionet-django/project/validators.py
+++ b/physionet-django/project/validators.py
@@ -111,11 +111,11 @@ def validate_slug(value):
 
 def validate_title(value):
     """
-    Validate titles that start with an alphabetical character followed by
+    Validate titles that start with an alphanumeric character followed by
     characters marked as letters in Unicode along side with the following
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks, looks good to me. I can't think of good reason why we would have intentionally prevented numbers at the beginning of titles. @bemoody okay with you to merge?

bemoody:
Yeah, that seems fine.

tompollard:
thanks

Issue #925, opened 2020-03-12
Add explanatory text to the cloud management section of the project management console


tompollard:
The console includes a page for managing settings for published projects (e.g. http://localhost:8000/console/published-projects/demobsn/1.0/).  This page includes a section for managing cloud settings:

<img width="885" alt="Screen Shot 2020-03-11 at 18 38 38" src="https://user-images.githubusercontent.com/822601/76527131-153c0f00-6445-11ea-93bd-0268a35e63a3.png">

Clicking "Remove" will remove the option for requesting cloud access that appears in the files section of a project:

![Screen Shot 2020-03-12 at 09 40 29](https://user-images.githubusercontent.com/822601/76527384-85e32b80-6445-11ea-9d02-bd7c0eca2b83.png)

It will not (1) delete/deactivate the bucket or (2) remove access for users who are already using the bucket. It would be good to add some explanatory text to the project management page to make this clear. 

See  https://github.com/MIT-LCP/physionet-build/pull/755 for more discussion.

PR #951, opened 2020-03-19
Adds explanatory text to remove button #925

Lucas-Mc:
Adds a simple question mark next to the remove button which, upon hover, reveals an explanation as to what the exact functionality of the remove button is.

### DIFF
--- a/physionet-django/console/templates/console/manage_published_project.html
+++ b/physionet-django/console/templates/console/manage_published_project.html
@@ -214,7 +214,9 @@
       <li class="list-group-item">
         <h5 class="card-title mt-0 mb-1">Storage location</h5>
         <p>Add and remove storage options.</p>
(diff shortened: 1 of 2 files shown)


tompollard:
Not tested, but sounds like this might be overkill. The other sections show text in `<p>` `</p>`. Is there a good reason to do this differently?

Lucas-Mc:
@tompollard  I just figured if someone had a question regarding the remove button then it should be placed right there so they don't have to dig for it. Where do you think it should go? Maybe right above the table?

tompollard:
@Lucas-Mc  My original thought was that we would add a bit more detail where it says "Add and remove storage options." (should have made it clearer in the issue, sorry!). If we want to add location/service specific details, then these could go into a new column perhaps?

tompollard:
thanks, looks good to me

PR #952, opened 2020-03-19
Check quota and file size using JS before upload

### LINKED ISSUE #940: Check storage quota before uploading
It would be nice for the project file upload form to check *before
uploading* whether the files are likely to exceed the project's
storage quota, so that people don't waste their time and bandwidth
uploading files that the server will then reject.

Felipe suggested in issue #919 that it might be possible for
JavaScript to determine the number and sizes of files before
submitting the form - at which point we could show an error message
and prevent the form submission if the total is too large.

It would be good if we could prevent too-large uploads at the HTTP
level - HTTP is designed to allow this, although browser support is as
always a mixed bag.  I'm also not entirely sure how this works with
uwsgi/django, and I think it is probably not a good idea to make any
changes that would require *keeping a uwsgi worker thread busy* for
the entire time that the files are being uploaded.

Of course, nothing we do here can be foolproof, since the author might
be uploading *multiple files* simultaneously.

We may also want to put off implementing either of these until we have
an O(1) mechanism for measuring disk usage.  See also issue #518.

elfeto:
Here I am adding the size of all the files being uploaded after
the files were chosen but before the form was submitted.

If the files sizes exceeds the remaining quota a message will be
shown. If the form was submitted and the file sizes exceeds the
remaining quota then the message will be shown and the form will
be prevented from running.

This is an attempt to fix #940

### DIFF
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -86,7 +86,8 @@ class UploadFilesForm(ActiveProjectFilesForm):
     `subdir` is the project subdirectory relative to the file root.
     """
     file_field = forms.FileField(widget=forms.ClearableFileInput(
-        attrs={'multiple': True}), required=False, allow_empty_file=True)
+        attrs={'multiple': True, 'onchange': "check_size('upload');"}), required=False,
+        allow_empty_file=True)
 
     def clean_file_field(self):
         """
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -86,7 +86,7 @@ class UploadFilesForm(ActiveProjectFilesForm):
     `subdir` is the project subdirectory relative to the file root.
     """
     file_field = forms.FileField(widget=forms.ClearableFileInput(
-        attrs={'multiple': True, 'onchange': "check_size('upload');"}), required=False,
+        attrs={'multiple': True, 'onchange': "check_upload_size_limit('upload');"}), required=False,
         allow_empty_file=True)
 
     def clean_file_field(self):
--- a/physionet-django/project/templates/project/edit_files_panel.html
+++ b/physionet-django/project/templates/project/edit_files_panel.html
@@ -1,10 +1,9 @@
 {% load static %}
 
 <div id="forms-and-panel">
-  <form method="post" enctype="multipart/form-data" onsubmit="check_size('submit')">
+  <form method="post" enctype="multipart/form-data" onsubmit="check_upload_size_limit('submit')">
     {% csrf_token %}
     {% include "project/project_files_form.html" %}
-
     <div id="files-panel" class="card">
       <div class="card-header">
         Folder Navigation:
@@ -197,7 +196,7 @@
     document.getElementById("delete-items-message").innerHTML = "Are you sure you wish to delete the "+ n_selected + " selected item(s)?"
   }
 
-  function check_size(stage){
+  function check_upload_size_limit(stage){
     // Check the size of the files to be uploaded and stop the form if the file
     // size exceeds the maximum allowed.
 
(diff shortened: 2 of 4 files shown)


bemoody:
Cool!

281531263285c1bcc0558632dd44bf1caff7e056

>  function check_size(stage){

I would rename this function to make its purpose more clear.  Maybe
call it something like check_upload_size_limit.

>    if ({{ storage_info.remaining }} - size <= 0){

Not "size >= {{ storage_info.remaining }}" ?

>      // Minimize the current window
>      $('#upload-files-modal').modal('toggle');
>      // Show the message
>      $('#upload_files_quota').modal('toggle');

This is not great UX.  I'd think it would be better to keep the upload
field visible, but display an error message below it and disable the
submit button.  (And of course hide the message and re-enable the
submit button if the size is under the limit.)

>      if (stage == 'submit'){
>        // If the form is being submitted prevent the submission
>        $('form').submit(function (evt) {
>          evt.preventDefault();
>        });
>      }

My jquery fu is not great, but doesn't this *permanently* disable
submission until the page is reloaded?  I think what you want is to
pass evt as an argument to check_size, and call preventDefault if it
fails.

This raises another problem, however - "upload files" is part of the
same form as everything else on that page.  (Because of the check
boxes, "rename"/"move"/"delete" need to be part of the same form
although they have three separate submit buttons.  But I think
"upload" could be a separate HTML form.)  If the person starts to
select some files to upload, then changes their mind and decides to
rename an existing file instead, we *don't* want the selected files to
be uploaded.

elfeto:
I renamed the function to `check_upload_size_limit`. 
Changed the if statement to `size >= {{ storage_info.remaining }}`
Removed the `preventDefault` section and just set the submit button disabled if the size exceeds the quota, and back to enabled if the size does not.

>This raises another problem, however - "upload files" is part of the
same form as everything else on that page.  (Because of the check
boxes, "rename"/"move"/"delete" need to be part of the same form
although they have three separate submit buttons.  But I think
"upload" could be a separate HTML form.)  If the person starts to
select some files to upload, then changes their mind and decides to
rename an existing file instead, we *don't* want the selected files to
be uploaded.

I was able to create/delete folders while having the file upload form disabled due to quota, same as rename and move and request more allowance.

I can still try to figure out a way to separate the form.

bemoody:
> I was able to create/delete folders while having the file upload form
> disabled due to quota, same as rename and move and request more allowance.
>
> I can still try to figure out a way to separate the form.

This is a side issue, let's move this to issue #954.


This looks good apart from some small HTML things:

> <label id='error_data_size' style="display:none">

This should not be a <label>.  Maybe <div class="alert alert-danger"> ?

> The selected file/s exceed the capacity left in your storage
> quota. You can request more space by clicking "Request storage".
>
> The remaining size is {{ storage_info.readable_remaining }}. The
> total size of the file/s you are trying to upload is
> <a id='data_size'></a>.

This should not be an <a>.

I think this message is a little verbose and could be worded better.
We use the word "allowance" on the site rather than "quota" (or
"capacity") - I keep saying "quota" because I'm old school, but I
suspect people might find it confusing.  Best to be consistent.

I would rephrase this as

  The selected files are larger than your remaining storage allowance
  for this project.  You can request more space by clicking "Request
  Storage".

  The total size of the selected files is <span id="data_size"></span>.
  Your remaining storage allowance is {{ storage_info.readable_remaining }}.

elfeto:
Done. 

I didn't wanted to use the <a> tag there, but when I was inserting the content, I could not find any other tag that I visually liked. Thanks for the tags :) 

I also used ur wording.

bemoody:
Looks good to me!

PR #953, opened 2020-03-20
Register and update DOIs for published projects using the console

tompollard:
Add buttons to the management page for published projects to allow DOIs to be registered and updated.

Screen Shot 2020-03-20 at 10 51 51

Before merging, I'd like to work out how best to prevent multiple DOIs being registered when someone quickly clicks the "Create DOI" button several times in succession. Any suggestions? 

Some thoughts are:

- ~~adding a "wait for request response" step to `views.manage_doi_request()` or `utility.register_doi()`~~
- adding a "check whether a DOI already exists for the same title and version" step to `views.manage_doi_request()` or `utility.register_doi()`
- always register a draft first, and then move to findable state with `utility.update_doi()`. Doesn't really solve the underlying issue.

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -168,14 +168,12 @@ class EditSubmissionForm(forms.ModelForm):
                 if self.cleaned_data['auto_doi']:
                     # register draft DOIs
                     if not project.doi:
-                        payload = generate_doi_payload(project,
-                                                               event="draft")
+                        payload = generate_doi_payload(project, event="draft")
                         project.doi = register_doi(payload)
                         project.save()
                     if not project.core_project.doi:
-                        payload = generate_doi_payload(project,
-                                                               core_project=True,
-                                                               event="draft")
+                        payload = generate_doi_payload(project, event="draft",
+                                                       core_project=True)
                         project.core_project.doi = register_doi(payload)
                         project.core_project.save()
 
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -254,23 +254,6 @@ class PublishForm(forms.Form):
         return data
 
 
-class DOIForm(forms.ModelForm):
-    """
-    Form to edit the doi of a published project
-    """
-    class Meta:
-        model = PublishedProject
-        fields = ('doi',)
-        labels = {'doi':'DOI'}
-
-    def clean_doi(self):
-        data = self.cleaned_data['doi']
-        validate_doi(data)
-        if PublishedProject.objects.filter(doi=data).exclude(id=self.instance.id):
-            raise forms.ValidationError('Published project with DOI already exists.')
-        return data
-
-
 class TopicForm(forms.Form):
     """
     Form to set tags for a published project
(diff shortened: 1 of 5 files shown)


bemoody:
> Before merging, I&#39;d like to work out how best to prevent multiple DOIs
> being registered when someone quickly clicks the &quot;Create DOI&quot;
> button several times in succession. Any suggestions?

I guess what you mean is that register_doi can take a non-trivial
amount of time (perhaps because it requires coordination with multiple
servers to ensure the new DOI is globally unique.)

Maybe one way would be to do something like this:

    queryset = ActiveProject.objects.filter(id=project_id, doi=None)
    if queryset.update(doi='PENDING') != 1:
        raise DOIExistsError('Project already has a DOI')
    new_doi = None
    try:
        new_doi = utility.register_doi(payload)
    finally:
        queryset = ActiveProject.objects.filter(id=project_id, doi='PENDING')
        queryset.update(doi=new_doi)

This is not pretty.  And I feel like Django should have a better way
of doing this but I'm not sure it does.

tompollard:
Nice suggestion, thanks @bemoody. This implements your approach in `physionet-django/console/utility.py/register_doi()`. A few notes:

- `register_doi()` now takes `project` as an argument. We need to check whether the project is an `ActiveProject` or a `PublishedProject` to decide whether the filter should be `ActiveProject.objects.filter` or `PublishedProject.objects.filter`. I'm doing this with `type(project).objects.filter`. In case this is not clear, there is a comment to explain. 
- In `physionet-django/console/utility.py/register_doi()`, we check whether a DOI is already pending or assigned with `queryset = type(project).objects.filter(id=project.id, doi=None)`. Our test fixtures (`physionet-django/project/fixtures/demo-project.json`) set DOI fields as empty strings with `"DOI": ""`. As a result, our test projects are treated as having DOIs and in tests they raise a `DOIExistsError`. The solution is either (1) to update the filter with a `doi is None or ""` or to (2) update the DOI field in the fixtures to `"DOI": null`. Approach 2 seemed cleaner, so I went for that. 
- `register_doi()` now assigns the DOI to the doi field of the project itself, rather than returning a DOI for assignment. It was more straightforward to implement the pending state this way. One implication is that passing `project` to `register_doi` and then subsequently calling `project.save()` outside the function will overwrite the DOI assigned by `register_doi`. i.e. (1) the project object is created (2) `register_doi()` is called to register a DOI and then (3) `project.save()` overrides `register_doi` (see below for an example). I've moved `project.save()` before `register_doi()` to deal with this.

```
                project = edit_log.project

                # project.doi is null                
                ...
                if self.cleaned_data['auto_doi']:
                    # register draft DOIs
                    if not project.doi:
                        payload = generate_doi_payload(project, event="draft")
                        register_doi(payload, project)
                    if not project.core_project.doi:
                        payload = generate_doi_payload(project, event="draft",
                                                       core_project=True)
                        register_doi(payload, project.core_project)

                # project.doi is now 10.12345/123456

                # project save should not be called here 
                # without creating a new instance!
                project.save()

                # project.doi is now null

```

bemoody:
I haven't tested the functionality, but the code looks good!

tompollard:
Thanks @bemoody, I tested fairly heavily and I'm pretty confident that it is working as expected. Let me know if you would like details for https://doi.test.datacite.org/

At some point I'd like to update `generate_doi_payload` to remove the `core_project` argument, which is unnecessary and feels kind of clunky, but I'll leave that for another time!

PR #961, opened 2020-03-25
Fix check on protected projects

elfeto:
Protected projects dont check if the user is authenticated before
displaying the page.

There is a check done to see if the person has access to the files.
At the moment this check is only to credentialed projects and not
protected. I am adding the check to protected.


bemoody:
Thanks for this quick fix!  Please push to production ASAP.

Issue #251, opened 2019-02-28
Add cite button to published projects


tompollard:
Add prominent "cite" button to published projects that offers a quick way to get the citation for the project in common citation formats.

alistairewj:
Google scholar is awesome

![image](https://user-images.githubusercontent.com/1282676/54058627-68410d80-41c4-11e9-90b8-3545929989ac.png)

tompollard:
Looks like this overlaps with https://github.com/MIT-LCP/physionet-build/issues/688

PR #861, opened 2020-01-31
Improvements to recommended citation

bemoody:
This makes a few changes to the style of the recommended citation
at the top of each project page.  In particular, the version
number is included, which I think is essential.

Further typographical nitpicking is welcome.

Before:

<blockquote>Mark, R. G. (2020). Demo Limited Clinical Dataset. PhysioNet. https://doi.org/10.13026/*****</blockquote>

<blockquote>Mark, R. (2019). Demo 2018 BHI and BSN Data Challenge. PhysioNet.</blockquote>

<blockquote>Mark, R. G. (2018). Demo eICU Collaborative Research Database. PhysioNet. https://doi.org/10.13026/G2F309</blockquote>

After:

<blockquote>Mark, R. G. (2020). Demo Limited Clinical Dataset (version 1.0.0). <i>PhysioNet</i>. https://doi.org/10.13026/*****</blockquote>

<blockquote>Mark, R. (2019). Demo 2018 BHI and BSN Data Challenge (version 1.0). <i>PhysioNet</i>.</blockquote>

<blockquote>Mark, R. G. (2018). Demo eICU Collaborative Research Database (version 2.0.0). <i>PhysioNet</i>. <a href="https://doi.org/10.13026/G2F309">https://doi.org/10.13026/G2F309</a></blockquote>

PR #973, opened 2020-03-26
Adds citation tool #251

Lucas-Mc:
Adds option to cite project in five different citation styles (MLA, APA, Chicago, Harvard, and Vancouver). Appears as a clickable link which exposes a hover box with each of the available citation styles. Testing with current fixtures is limited so let me know if you find any bugs I haven't caught already!

### DIFF
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -758,6 +758,8 @@ class UnpublishedProject(models.Model):
         authors = self.authors.all().order_by('display_order')
         year = timezone.now().year
         doi = '10.13026/*****'
+
+        # TODO: Add other citation styles
         return format_html(
             '{authors} ({year}). {title} (version {version}). '
             '<i>PhysioNet</i>. https://doi.org/{doi}',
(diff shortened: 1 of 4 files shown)


Lucas-Mc:
I suppose the `citation_text()` function can also be removed now if this is accepted?

tompollard:
Thanks Lucas, I think the pop up box works well. I've added a couple of quick thoughts below:

- Until we work out how we want to reformat the content in the grey box, I'd suggest leaving it as it is (i.e. displaying the citation for the resource). This citation is the most important one, so we shouldn't give it less prominence than the others. The new link could say something like "View more citation formats".
- Including the style (Vancouver; Harvard; etc) as an argument to a citation method would be more flexible, and probably avoid code duplication

bemoody:
The function can be removed if nothing else refers to it and it's no
longer needed - it's up to you to figure out whether that's the case.
:)

But at the same time, it's best to avoid unnecessary "churn" in the
API since that will cause problems for merging and rebasing.  In this
case, you might want to keep 'citation_text' as an alias for
'citation_text_APA'.

Some things that could be done better here:

- I don't like the use of uppercase in function/method names.

- This is closely tied to the PublishedProject model.  We also need to
be able to generate citations for ActiveProjects (which differ from
PublishedProjects in that they don't have a publication date, and the
DOI is not yet assigned) and we may also want to generate citations in
other contexts.

- There's a bit of duplicated logic (is_legacy,
authors.all().order_by('display_order')).

So I think it would be better to have a single function that takes a
set of keyword arguments (citation_style, title, authors, year, doi,
...) and returns the HTML citation.  Another option might be to
construct a class that encapsulates all the relevant information and
could then have methods for generating various HTML citations (and
also BibTeX, etc.)

Lucas-Mc:
Thanks for the quick replies! I like the idea of condensing the multiple functions into just one single one; I think refactoring was going to be the next step after initial testing and approval but I will refactor it now and push the relevant commits. I can also fix the uppercase function names if that formats better to the current style guide (I see now PEP8 recommends all lowercase). I noticed how similar it was to "ActiveProjects" and figured that can be covered in a separate pull request maybe, unless it should also be covered here?

Lucas-Mc:
@tompollard The current `citation_text()` function didn't correspond to any of the existing citation styles (Though it is close to APA without the rules for number of authors). Should I continue to show it or revert to something like APA instead?

tompollard:
> @tompollard The current citation_text() function didn't correspond to any of the existing citation styles (Though it is close to APA without the rules for number of authors). Should I continue to show it or revert to something like APA instead?

The current style was added in https://github.com/MIT-LCP/physionet-build/pull/861 I believe. Adapting it to use one of the formal styles (e.g. APA) sounds good!

bemoody:
The author limit thing is a little funny, because for many years we
have listed that citation at the top of the page, for a paper with 10
authors.  If we limit the number of authors displayed in the project
citation, I think to be fair we should also limit the number of
authors displayed in the "Goldberger et al" citation.

Ideally I would want to display all three citations in the same style,
*and* have a popup with alternative citation styles, *and* show the
citation as bibtex.  Of course, doing that for the "project
publication" is a huge task itself and not something we can do
immediately.

physionet-django/project/models.py
```diff
@@ -1533,28 +1533,179 @@ def get_storage_info(self, force_calculate=True):
...
+
+        Returns
+        -------
+        citation_dict (dictionary) : dictionary of each citation style and
+            their respective citation for the requested project
+        """
+        # For if there are no authors or valuable metadata
         if self.is_legacy:
-            return ''
+            return {'APA':''}
```
tompollard: perhaps clearer to return `{}` or `None` here? The code seems clear, so no need for the comment.

physionet-django/project/models.py
```diff
@@ -1533,28 +1533,179 @@ def get_storage_info(self, force_calculate=True):
...
+                        "{authors} ({year}) '{title} (version {version}).', "
+                        "<i>PhysioNet</i>.",
+                        authors=temp_authors,
+                        year=self.publish_datetime.year,
+                        title=self.title,
+                        version=self.version)
+
+            elif (style == 'Vancouver'):
+
+                temp_authors = ', '.join(a.initialed_name().replace(',','').replace('.','') for a in authors)
```
tompollard: I think PEP8 states <80 characters in width.

physionet-django/project/models.py
```diff
@@ -1533,28 +1533,179 @@ def get_storage_info(self, force_calculate=True):
...
+                else:
+                    citation_format = format_html(
+                        '{authors}. "{title} (version {version})." '
+                        '<i>PhysioNet</i> ({year}).',
+                        authors=temp_authors,
+                        title=self.title,
+                        version=self.version,
+                        year=self.publish_datetime.year)
+
+            elif (style == 'APA'):
```
tompollard: no need for the parentheses in these clauses `style == 'APA:` is fine

physionet-django/project/models.py
```diff
@@ -1533,28 +1533,179 @@ def get_storage_info(self, force_calculate=True):
...
+                    temp_authors = authors[0].get_full_name(reverse=True)
+                elif (len(authors) == 2):
+                    first_author = authors[0].get_full_name(reverse=True)
+                    second_author = authors[1].get_full_name()
+                    temp_authors = first_author + ', and ' + second_author
+                else:
+                    temp_authors = ', '.join(authors[0].get_full_name().split())
+                    temp_authors += ', et al'
+
+                if self.doi:
```
tompollard: the only difference here appears to be presence/absence of the doi. Is there a way to avoid the code duplication (and ideally the if clause too?).

physionet-django/project/models.py
```diff
@@ -1533,28 +1533,179 @@ def get_storage_info(self, force_calculate=True):
...
+
+            elif (style == 'APA'):
+
+                if (len(authors) == 1):
+                    temp_authors = authors[0].initialed_name()
+                elif (len(authors) == 2):
+                    first_author = authors[0].initialed_name()
+                    second_author = authors[1].initialed_name()
+                    temp_authors = first_author + ', & ' + second_author
+                elif (len(authors) > 20):
```
tompollard: please could you explain the jump to 20 here?

physionet-django/project/models.py
```diff
@@ -1533,28 +1533,179 @@ def get_storage_info(self, force_calculate=True):
...
+                        '{authors}. "{title} (version {version})." '
+                        '<i>PhysioNet</i> ({year}).',
+                        authors=temp_authors,
+                        title=self.title,
+                        version=self.version,
+                        year=self.publish_datetime.year)
+
+            elif (style == 'APA'):
+
+                if (len(authors) == 1):
```
tompollard: is there a way to avoid the if clauses here? e.g. could the authors variable be populated with a list comprehension?

physionet-django/project/models.py
```diff
@@ -1533,28 +1533,179 @@ def get_storage_info(self, force_calculate=True):
...
+                    temp_authors = ', '.join(a.initialed_name() for a in authors[0:19])
+                    temp_authors += ',... ' + authors[len(authors)-1].initialed_name()
+                else:
+                    temp_authors = ', '.join(a.initialed_name() for a in authors[:(len(authors)-1)])
+                    temp_authors += ', & ' + authors[len(authors)-1].initialed_name()
+
+                if self.doi:
+                    citation_format = format_html(
+                        '{authors} ({year}). {title} (version {version}). '
+                        '<i>PhysioNet</i>. '
```
tompollard: Ideally we should avoid hardcoding `PhysioNet` here. It would be good if it could be taken from an environment variable.

tompollard (commented):
Thanks @Lucas-Mc, this is going in the right direction. I've added some quick thoughts. Give me a shout if you'd like to chat about any of these points.

physionet-django/project/models.py
Lucas-Mc: This is to resolve the single function refactor while still handling legacy projects. If I return `{}` then code will have to be changed in `views` since right now it is calling `citation_text.APA` as default but if I return `{}` instead then APA wouldn't be defined.

Lucas-Mc: I figured calling `citation_text.MLA` or whatever the style is would be easier than editing `views` and adding parameters... That reminds me, I think I need to re-edit the docstrings now.

physionet-django/project/models.py
Lucas-Mc: Good catch

physionet-django/project/models.py
Lucas-Mc: Sorry, just habit since it was always easier for me to read logically but I will change it.

physionet-django/project/models.py
Lucas-Mc: I noticed that as well and have been thinking about it for a while but I'll try to focus on it to see if I can come up with a solution... this issue is the formatting string would also have to change.

Lucas-Mc: Maybe append the DOI part after the initial template is created?

physionet-django/project/models.py
Lucas-Mc: This is just from the weird APA guidelines : [See here](https://owl.purdue.edu/owl/research_and_citation/apa_style/apa_formatting_and_style_guide/reference_list_author_authors.html).

physionet-django/project/models.py
Lucas-Mc: Will this ever change? Maybe it's in a fixture/JSON file somewhere?

physionet-django/project/models.py
Lucas-Mc: Hmmm, could you explain further? Each citation style has a particular formatting guideline.

Lucas-Mc: I see potential for further refactoring but it might make it harder to read.

physionet-django/project/models.py
tompollard: The idea is for the platform software to be reusable, so it may be different for other people. It probably needs to be added as a new variable in the .env file (oddly, I don't think it's defined at the moment!). Maybe `PUBLISHER_NAME`?

physionet-django/project/models.py
tompollard: Thanks, good to know!

physionet-django/project/models.py
tompollard: Most information seems consistent across styles, except perhaps the author string. Perhaps we could separate the structure from the content doing something like:

```
# get the structure
apa = "{author} {year} {title} {blah}"
harvard = "{year} {author} {blah}"

# get information that is shared for all styles
content = {"year": x,
    "title": title}
}

# get style specific strings
if harvard: auth_str = [",".join(x.get_full_name()) for x in authors]
elif apa auth_str = x
etc.

# unpack content with **
citation = apa.format(auth_str, **content)
```

Benjamin or Felipe may have a cleaner solution...

bemoody:
We need an implementation for both PublishedProject and
UnpublishedProject.  Notice that these two classes have different
implementations, but the same API: if you have either an
UnpublishedProject or a PublishedProject object,
'project.citation_text()' is an HTML string.

If you're going to define a new function that returns a dictionary,
please don't call it "citation_text".  Leave citation_text alone and
give the new function a new name.  Then, make sure that the new
function is implemented for both types of projects.

Also, think about how to write test cases for this function.  It's
impossible to test it just using the existing demo projects - none of
them have more than one author!

I think if you pause and try to figure out how to implement this in a
way that supports both PublishedProjects and UnpublishedProjects, and
also allows you to easily write a complete set of test cases, you'll
end up eliminating a lot of duplicated logic while also being easier
to extend in the future.


Typographic things:

MLA: Surely you shouldn't write "doi:https://doi.org/10.12345/678".

APA: There should be a space between "," and "...".

Harvard: Comma followed by "Available from:" looks very wrong.  A
quick web search suggests it should be period followed by "Available
at:".  Which is correct?

Vancouver: same as Harvard

I'd probably advise writing:

   {authors}. "{title}" (version {version}).

rather than:

   {authors}. "{title} (version {version})."

Please list specific references for the styles you're using (online
references if possible.)

physionet-django/project/models.py
Lucas-Mc: Let's make this another pull request?

Lucas-Mc:
I addressed most of the comments in the previous commit, now working on adding all styles for the `UnpublishedProject` class.

Lucas-Mc:
Here are my references for the citation styles:
1. MLA - https://owl.purdue.edu/owl/research_and_citation/mla_style/mla_formatting_and_style_guide/mla_formatting_and_style_guide.html
2. APA - https://owl.purdue.edu/owl/research_and_citation/apa_style/apa_style_introduction.html
3. Chicago - https://owl.purdue.edu/owl/research_and_citation/chicago_manual_17th_edition/cmos_formatting_and_style_guide/chicago_manual_of_style_17th_edition.html
4. Harvard - https://www.mendeley.com/guides/harvard-citation-guide
5. Vancouver - https://guides.lib.monash.edu/ld.php?content_id=14570618

bemoody:
This looks like an improvement, thanks!

You still have duplication between the published/unpublished versions,
though.  What would you need to do to avoid that, and make this into a
single function (perhaps a method defined in the Metadata class)?

Instead of taking the result of "initialed_name" and stripping out the
commas and periods, a better option might be to add optional arguments
to initialed_name so you can tell it not to include the commas and
periods in the first place.

Thanks for listing your references - please put them in the code!  If
it's not documented in the code, it's not documented. :)

physionet-django/project/models.py
tompollard: Yep, another PR would be fine.

physionet-django/project/models.py
```diff
@@ -858,18 +824,104 @@ def citation_text(self):
...
+                                'PhysioNet. {year}')
+
+                doi_format = ('. Available from: '
+                              '<a href="https://doi.org/{doi}">'
+                              'https://doi.org/{doi}</a>.')
+
                 all_authors = ', '.join(a.initialed_name()
-                                         .replace(',', '')
-                                         .replace('.', '') for a in authors)
+                                        .replace(',', '')
```
tompollard: Something that Benjamin suggested was to update the initialed_name function, so that it takes arguments (instead of fixing the string it returns).  I think this would be a good idea!

physionet-django/project/models.py
```diff
@@ -858,18 +824,104 @@ def citation_text(self):
...
+                citation_format = format_html(final_style,
+                                              author=all_authors,
+                                              **shared_content)
 
             citation_dict[style] = citation_format
 
         return citation_dict
 
+
+class SubmissionInfo(models.Model):
```
tompollard, line 874: Not clear why, but the diff suggests that these classes have been deleted and then re-added (when I think they've just been shifted down). Were they reformatted in some way? It's best to avoid this happening, because it makes it difficult for us to see what has changed.

physionet-django/project/models.py
```diff
@@ -683,128 +689,69 @@ def get_anonymous_url(self):
...
+        ----------
+        styles [list]:
+            ['MLA', 'APA', 'Chicago', 'Harvard', 'Vancouver']
 
-        This text resembles the final "citation text" as it will
-        appear once the project is published.  Since the project is
-        not yet published, the current year is used in place of the
-        publication year, and '*****' is used in place of the DOI
-        suffix.
+        Returns
```
tompollard, line 732: I think Benjamin suggested keep the returned value of citation_text as a string, rather than a dictionary? If so, perhaps create your dictionary in the view by iterating calls to citation_text() with each individual style that you would like to include?

physionet-django/project/models.py
Lucas-Mc: I'm not sure why it did that either... It's possible it was reformatted but I thought I just added the `citation_text()` function just above them.. Will look into it further!

physionet-django/project/models.py
Lucas-Mc: Okay, so have it loop in the view instead? Wasn't sure which would be more useful.

physionet-django/project/models.py
Lucas-Mc: I will do it, I think this is the only instance it is needed but I suppose it may be useful further down for something else.

physionet-django/project/models.py
tompollard: Yep, you could then iterate styles in your view to create a dictionary, or alternatively create another method on the model that returns all styles as a dictionary.

physionet-django/project/models.py
```diff
@@ -675,6 +689,169 @@ def get_anonymous_url(self):
...
+                                '<i>PhysioNet</i> ({year})')
+
+                doi_format = (', <a href="https://doi.org/{doi}">'
+                              'https://doi.org/{doi}</a>.')
+
+                if (len(authors) == 1):
+                    all_authors = authors[0].get_full_name(reverse=True)
+                elif (len(authors) == 2):
+                    first_author = authors[0].get_full_name(reverse=True)
+                    second_author = authors[1].get_full_name()
```
tompollard: Should this be reversed too?

Lucas-Mc: Actually no, the style guide wants the final name to be written in normal order like "Pollard T., Lucas McCullum" or something similar to that.

physionet-django/project/models.py
Lucas-Mc: Notice how if you go to "Files Changed" then it doesn't show up? It must've been something funky about the commit but I think it is good now.

bemoody:
In MLA style you have:

+            else:
+                all_authors = ', '.join(
+                    authors[0].get_full_name().split())
+                all_authors += ', et al'

I think you meant to use authors[0].get_full_name(reverse=True) here.

Lucas-Mc:
> In MLA style you have: + else: + all_authors = ', '.join( + authors[0].get_full_name().split()) + all_authors += ', et al' I think you meant to use authors[0].get_full_name(reverse=True) here.

Yeah, it ends up being the same logic so I just switched it over now.

tompollard:
I think this is an improvement and overall I'm happy to merge. Before we do, it would be good to tweak the way the link appears in the citation box. 

- there is a new line break before the citation which should be removed
- "Show More Citation Options" is floating, so it's not clear which paper it refers to. 
- Sentence case "Show more citation options" might work better for the link text.

Screen Shot 2020-04-07 at 13 16 08

Putting it at the end of the citation looks okay here, but it will look messy if the text is forced onto multiple lines:

Screen Shot 2020-04-07 at 13 30 11

So maybe something like this would work:

Screen Shot 2020-04-07 at 13 44 23

We'll want to improve the formatting of the section later, so this would do for now.

Lucas-Mc:
> I think this is an improvement and overall I'm happy to merge. Before we do, it would be good to tweak the way the link appears in the citation box.
> 
> * there is a new line break before the citation which should be removed
> * "Show More Citation Options" is floating, so it's not clear which paper it refers to.
> * Sentence case "Show more citation options" might work better for the link text.
> 
> Screen Shot 2020-04-07 at 13 16 08
> 
> Putting it at the end of the citation looks okay here, but it will look messy if the text is forced onto multiple lines:
> 
> Screen Shot 2020-04-07 at 13 30 11
> 
> So maybe something like this would work:
> 
> Screen Shot 2020-04-07 at 13 44 23
> 
> We'll want to improve the formatting of the section later, so this would do for now.

Just implemented your final design, let me know your thoughts!

tompollard:
Thanks @Lucas-Mc, looks good to me. @bemoody any comments before we merge?

bemoody:
Looks good!

tompollard:
Nice work @Lucas-Mc, thanks!

PR #976, opened 2020-03-27
Merge users

### LINKED ISSUE #857: Allow user accounts to be merged
It would be good to have a method for merging user accounts.  For example:

- the same person registers abc@test.com and abc@test2.com. 
- Each of these accounts has some associated information (e.g. projects, affiliations).
- Merging the accounts should link the information to a single account and resolve any conflicts (e.g. different affiliations on each account).

elfeto:
Merge user accounts closes #857

The person doing the merge cannot be part of the merge.
Both accounts need to be active to be eligible for a merge. (This check is done at a model and form level)

The project ownership is controlled by authorship.
Here I:
   - Move all associated emails.
       - If an associated email is primary set flag as False.
   - Move all credentialed applications.
       - If user is credentialed set the flags.
   - Move legacy credentialed if any.
   - Move cloud information if main user doesn’t have and second does.
   - Move published authorships.
   - Move active authorships.
   - Move author invitations.
       - If the main user is an author, delete authorship invitations.
       - If the main user already had an invitation to the project, delete
         newer ones.
   - Disable second user.


elfeto:
I created this branch before we merged the self-credentialing was added hence I was missing the published project from that PR. I rebased and fixed the tests.

bemoody:
Sorry, but I don't see that this approach could be maintainable.

We cannot have one massive function that attempts to somehow manage
every possible foreign key relationship involving a user object.

Rather, we would need to have something generic, so that each
individual model class has a way of handling the "merge" operation
independently (either by updating existing objects to point to the new
user object, or by combining two objects into one, or by raising an
exception to say "no, it is not possible to merge those two users at
all".)

I'm not going to say this would be easy, I think it would be rather
complicated, but I think if you want to have this feature then that is
what you'd need to do.

elfeto:
I can add in each model a "user_merge" function, but would me \*somewhat\* the same in the maintainability (not the same tho).... 

Every time we add a new model (hopefully we wont), we would have to add a merge function (if it touches the user, and then add the same function to the view.

It would be there (console.views), or we add a function to the physionet.utility or physionet.something to merge users. (cannot be done under the user models because one cannot import project.models inside it...

bemoody:
By generic, I mean that the User implementation does not need to know
what other models are referring to it.  If we later decided to add a
new model (say, lightwave.models.AnnotationSet for saving somebody's
annotations), it should be possible to do that without modifying
*anything* in the user package.

You can find examples on stackoverflow of how to do this in a "dumb"
way ("dumb" in the sense of "not requiring any complicated decisions
from the application code", not in the sense of "it would be a bad
idea to implement things this way".)  The dumb strategy is "change
every foreign key that points to object X to instead point to object
Y; and if there are any conflicts, that will raise a database error
and the operation fails."

That approach will fail when you have other objects that *do* conflict
with each other and need to be merged, or if there are requirements
that are not expressed as database-level constraints.  So you would
need some strategy for addressing that.  I don't know what that would
be.  It might mean defining a new API on the related models.  It might
mean defining a custom subclass of ForeignKey with a new API.  As I
said, it's not an easy problem.

elfeto:
To see if I understand what you say...

You are saying that would be more maintainable and acceptable to do this change separated across all the models, being model dependent? 

for example:
```
   .
   .
   .
   def merge_users(self, user1, user2)
      ....
```
and so on across all models?

bemoody:
Maybe.  What would merge_users do?  What function calls
merge_users?  What CredentialApplication object(s) is it called
on?

elfeto:
Merge users would merge the user information from one user onto the other.
For example:
I would call `CredentialApplication.merge_users(user1, user2)` to merge ALL the CredentialApplication related data.

elfeto:
And the same with ALL the other models that have the user object as a FK, one to one or w/e

bemoody:
So it would be a class method, then?  Yeah, that could work.

elfeto:
This would be an example:
```
    @staticmethod
    def merge_users(main_user, second_user):
        """
        Merge the credential aplications from two users.
        """
        logger.info("Merging the credential aplications of {0} onto {1}".format(
            second_user, main_user))

        # Move all credential applications and withdraw any active application
        credentialed_apps = second_user.credential_applications.all()
        logger.info('{0} credential application(s) were found in user {1}'.format(
            credentialed_apps.count(), second_user))
        for app in credentialed_apps:
            app.user = main_user
            if app.status == 0:
                app.status = 3
                logger.info("The credential application with ID #{0} was withdrawn.".format(
                    app.id))
            logger.info("The credential application with ID #{0} was moved to {1}".format(
                app.id, main_user))
            app.save()

        if second_user.is_credentialed and not main_user.is_credentialed:
            main_user.is_credentialed = True
            main_user.credential_datetime = second_user.credential_datetime
            main_user.save()
            logger.info("User {0} has been credentialed because user {1} "
                        "was credentialed.".format(main_user, second_user))
```
Would be used like this with the following output.:
```
>>> from user.models import *
>>> 
>>> user1 = User.objects.first()
>>> user2 = User.objects.get(username='rgmark')
>>> 
>>> CredentialApplication.merge_users(user1, user2)
INFO 2020-04-01 17:46:52,532 Merging the credential aplications of rgmark onto admin
INFO 2020-04-01 17:46:52,534 1 credential application(s) were found in user rgmark
INFO 2020-04-01 17:46:52,538 The credential application with ID #1 was moved to admin
INFO 2020-04-01 17:46:52,546 User admin has been credentialed because user rgmark was credentialed.
>>> 
```

elfeto:
I added the "related_name" to a couple of model objects that had the user as a FK but no related name. I use the related name to do the merge properly.

I have the user/models.py tests, now I will continue with the project/models.py tests.
Here I have a simple issue as I stated above. I cannot import ANYTHING from the project models to the user model. Hence I have to split the merge (for now) lets see how this goes.

As a preview pushing this as an \*incomplete\* PR just to have it here and if some1 has some comments they will be addressed.

bemoody:
> Here I have a simple issue as I stated above. I cannot import ANYTHING from
> the project models to the user model.

Do you mean because of 'from X import Y'?  That's a good reason not to
use 'from X import Y'. :)

But you shouldn't need to do that.  You just want to iterate over all
the related models by something like

    for field in User._meta.get_fields():
        model = field.related_model
        if model is not None and not field.many_to_one:
            if hasattr(model, 'merge_users'):
                model.merge_users(new_user, old_user)
            elif field.one_to_many:
                # might work, who knows?
                objs = getattr(old_user, field.name).select_for_update()
                remote_name = field.remote_field.name
                for obj in objs:
                    assert getattr(obj, remote_name) == old_user
                    setattr(obj, remote_name, new_user)
                    obj.save()
            else:
                raise Exception("don't know how to merge {}".format(field))

elfeto:
So....
I think this takes care of the user merge.
In summary, I added a function called `merge_users` to all the models that have the user object as a foreign key.
That merge function merges the information form two users within that model. This avoids one big function.
Then I took @bemoody approach of iterating through all the models and if the model has the function `merge_users`, then calling it.

I added fixtures with dummy data to do the tests, and since its quite big, i did a separate PR (#1004) with it as asked by @tompollard .

For future new models that have the user as a foreign key, the person adding the new model has to add the `merge_users` function too.

I think it would be nice to have for the future a place that is like a history for all the users/projects which lists the log information for them. There we could see all the messages from the merge and other things, this would be instead of going through the server logs.

elfeto:
I will close this PR until this is addressed in the issue.

PR #982, opened 2020-03-31
Test project content form

bemoody:
This adds a test case for project_content, which is also useful as a sanity check for bleach.


tompollard:
nice, thanks.

Issue #894, opened 2020-02-26
Project modified_datetime


bemoody:
We want to keep track of when an active project was most recently modified (for some definition of "modified".)

* We display a "Modified" date in /projects/.  This is ambiguous, but I would expect it to reflect any modification to the project *content* (files, description, authors, etc.), but not changes to the *status* of the project in the submission system (editor assigned, submission_status changed, etc.)

* We display a "Last Modified" timestamp in the project preview, if accessed via passphrase.  This timestamp specifically ought to reflect whether there have been any changes to the visible content of the preview (including the files), but should not reflect any invisible changes.

* It would also be useful, for various reasons (backups, file format checks), to keep track of when a project's files were last modified.  This could be coupled to the same timestamp as above, but it might be better to track it separately.

I started looking at this and quickly fell down a very deep rabbit hole.

My first thought for tracking the file modification time would be to touch the toplevel directory after any file operation.  But relying on the directory timestamp for anything important would likely lead to subtle bugs, and arguably we want this information in the relational DB anyway.

I note that 'project_files' does 'project.modified_datetime = timezone.now()' after any form is submitted that might modify the files.  However, it doesn't save the project afterwards, so this has no effect.

I also note that 'modified_datetime' is 'auto_now', which means that setting it explicitly has no effect; it's automatically set to the current time whenever the ActiveProject object is saved.  This may be both a good thing and a bad thing, considering the various uses and interpretations of modification time.

Of course, there are many other functions that modify the visible content of the project (e.g. the list of authors) but do not save the ActiveProject.  'project_discovery' is the only one that updates the timestamp explicitly.

bemoody:
There's room for improvement, but let's call this issue fixed.

PR #983, opened 2020-04-01
Track project modification time

bemoody:
Summary: Attempt to track the date and time when an active project was most recently modified (issue #894).  Update it when it should be updated, don't update it when it shouldn't be.

In particular:

1. A project should be considered "modified" when any files/directories are added, removed, or renamed.

2. A project should be considered "modified" when its description fields are changed.

3. A project should be considered "modified" when the publication or the list of references is changed.

4. A project should *not* be considered "modified" when the author submits it, the editor accepts or rejects it, authors approve it, etc.

5. A project should *not* be considered "modified" when the editor completes copyediting and doesn't change anything.  (This may be unreliable due to ckeditor and bleach making changes to the markup without changing the semantics.)

It's difficult to do this in any clean, consistent way.  (2), (4), and (5) all involve changing one or more fields of the ActiveProject object; (3) involves changing objects *other* than the ActiveProject; (1) doesn't involve changing the database at all.

The project model has always been divided into two abstract parent classes (Metadata and SubmissionInfo), although this distinction didn't really mean anything in the past.  However, for the most part, the Metadata fields are the ones that contain public project content (like "resource_type" and "abstract") and the SubmissionInfo fields are the ones that relate to the publication state (like "editor" and "submission_datetime".)  So to begin with, I've tried to formalize this, moving a few of the internal fields from Metadata to SubmissionInfo.

UnpublishedProject has the field "modified_datetime" which is updated automatically whenever the object is saved.  Here, I've changed this behavior: if the object was initially loaded from the database, and none of its Metadata fields were changed (!=) from their initial values, save() will skip saving those fields as well as modified_datetime.  In other cases, modified_datetime is updated as usual.

This is a kludge, but I'm not sure how to make it less kludgy, other than by splitting Metadata and SubmissionInfo into completely separate objects.  That may be what we want to do eventually, but that's a big disruptive change.

I've changed the save() and delete() of Topic, Reference, and Publication to force an update of the project's modified_datetime.  There may be a better way to do this!  Indirection through a GenericForeignKey is particularly unsettling.

As a result of these changes:

- project_content will change the modification timestamp if any changes were made to either the flat fields or the references.

- project_access will change the modification timestamp if any changes were made to the license or policy.

- project_discovery will always change the modification timestamp.  (This one deals with many-to-many fields, so it's not trivial to hook into save() in the same way.)

- project_files will change the modification timestamp when a "valid" file manipulation form is submitted (although it might not have had any effect.)

- Editing the relevant fields through /admin/ should also change the timestamp when appropriate.  (I haven't tested this very much.)

One thing this *doesn't* do is to change the modified_datetime when the list of authors changes.  There are some practical reasons:

- Any author can change their affiliation at any time before the project is submitted.  It feels slightly wrong for the "modification time" to be affected by non-submitting authors.

- Any author can change their displayed name at any time (even if the project is not editable!)  So if we really wanted to be sure that 'modified_datetime' reflects any change to the byline, that would require touching every active project you're coauthoring, whenever you change the spelling of your name.  Making that work would require breaking several layers of abstraction.

- When an author approves publication, that also touches the Author object, so it would be *another* class that needs to handle two distinct "save" conditions, and I just didn't want to go there.

### DIFF
--- a/physionet-django/console/test_views.py
+++ b/physionet-django/console/test_views.py
@@ -172,9 +172,22 @@ class TestState(TestMixin):
         Author approves publication
         """
         project = ActiveProject.objects.get(title='MIT-BIH Arrhythmia Database')
+
+        def get_project():
+            return ActiveProject.objects.get(id=project.id)
+
+        # The following steps should not alter the project timestamp,
+        # since project "Metadata" fields are not being changed (only
+        # "SubmissionInfo").
+        timestamp = project.modified_datetime
+
         project.submit(author_comments='')
+        self.assertEqual(get_project().modified_datetime, timestamp)
+
         editor = User.objects.get(username='admin')
         project.assign_editor(editor)
+        self.assertEqual(get_project().modified_datetime, timestamp)
+
         self.client.login(username='admin', password='Tester11!')
         # Accept submission
         response = self.client.post(reverse(
@@ -184,16 +197,22 @@ class TestState(TestMixin):
                 'pn_suitable': 1, 'editor_comments': 'Good.', 'decision': 2,
                 'auto_doi': 1
             })
+        self.assertEqual(get_project().modified_datetime, timestamp)
+
         # Complete copyedit
         response = self.client.post(reverse(
             'copyedit_submission', args=(project.slug,)),
             data={'complete_copyedit':'', 'made_changes':0})
+        self.assertEqual(get_project().modified_datetime, timestamp)
+
         # Approve publication
         self.assertFalse(ActiveProject.objects.get(id=project.id).is_publishable())
         self.client.login(username='rgmark', password='Tester11!')
         response = self.client.post(reverse(
             'project_submission', args=(project.slug,)),
             data={'approve_publication':''})
+        self.assertEqual(get_project().modified_datetime, timestamp)
+
         self.assertTrue(ActiveProject.objects.get(id=project.id).is_publishable())
 
     def test_publish(self):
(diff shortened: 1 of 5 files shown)


elfeto:
I saw what you mentioned about author changes but I dont know if the following should trigger a modification:
 - When an authorship invite is accepted.
 - Change of corresponding email
 - Change of corresponding author
 - Author removal
 - Storage allowance changes
 - Author order change ?

The copyedit saving even if nothing is changed might because of the discovery. 

I will continue checking this.

elfeto:
I believe I understand most of the code of the model functions to from_db and save that are only used to not overwrite the object if an empty save is submitted. 

If the actions above are not required to marked the project as modified, I believe this PR is ok :)

bemoody:
> The copyedit saving even if nothing is changed might because of the
> discovery.

Ah, that's a good point.  But I guess copyedit changes only happen
when you click "Save Content" - just clicking "Complete Copyedit"
doesn't save any of the other forms, right?

> I saw what you mentioned about author changes but I dont know if the
> following should trigger a modification:
>  - Storage allowance changes

nah

>  - When an authorship invite is accepted.
>  - Change of corresponding email
>  - Change of corresponding author
>  - Author removal
>  - Author order change ?

I have mixed feelings about these.  Even if we don't want to update
the timestamp automatically when saving the Author object, we could
still call content_modified in the various author-related forms.
Would that be better or would it overcomplicate things?

elfeto:
>Ah, that's a good point.  But I guess copyedit changes only happen
when you click "Save Content" - just clicking "Complete Copyedit"
doesn't save any of the other forms, right?

Yea it doesn't.

>Would that be better or would it overcomplicate things?

I don't have any strong feelings either way. I would refer to @tompollard here, if he doesnt mind I can merge :)

tompollard:
Thanks, looks good to me. Clarifying the distinction between `SubmissionInfo` and `Metadata` is helpful.

Issue #943, opened 2020-03-17
Allow handling editor to be reassigned


tompollard:
It should be possible to reassign a project to a different handling editor. i.e. if admin1 becomes handling editor for a project, they should be able to make admin2 handling editor instead. This is useful if someone needs help with handling a project, goes on vacation, moves to a new role, etc.

elfeto:
There are two options to do this:
Option 1. Add a button in the Editor home page, that will open a pop up window similar to the screenshots:
<img width="1505" alt="Screen Shot 2020-03-30 at 1 32 24 PM" src="https://user-images.githubusercontent.com/3857286/77948373-ef0ecf80-728a-11ea-931a-efa469ad3d6c.png">
<img width="488" alt="Screen Shot 2020-03-30 at 1 31 47 PM" src="https://user-images.githubusercontent.com/3857286/77948370-ee763900-728a-11ea-8d66-06f413bb0913.png">

Option 2. Add a tab in the project info page with the same information of the modal above.
<img width="1182" alt="Screen Shot 2020-03-30 at 1 33 05 PM" src="https://user-images.githubusercontent.com/3857286/77948443-06e65380-728b-11ea-87ba-65c83166b12b.png">

Note. The text on the modal is NOT updated, and will be changed with an explanation of what is happening. I took the screenshot of the Assign editor section.

elfeto:
Both options would be permanent. Not for a specific section. You are actually dumping your work onto another person.

tompollard:
Thanks Felipe! For me option 1 suggests that the assignment applies just for that section, so my preference is option 2.

PR #985, opened 2020-04-01
Allow handling editor to be reassigned

elfeto:
Allow the current project editor to reassign the project. This will send an email to both the new and old editor.

The form changes are handled in the submission_info function only.

Closing #943

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -58,6 +58,22 @@ class AssignEditorForm(forms.Form):
         return pid
 
 
+class ReassignEditorForm(forms.Form):
+    """
+    Assign an editor to a project under submission
+    """
+    editor = forms.ModelChoiceField(queryset=User.objects.filter(
+        is_admin=True), widget=forms.Select(attrs={'onchange': 'set_editor_text()'}))
+
+    def __init__(self, user, *args, **kwargs):
+        """
+        Set the appropriate queryset
+        """
+        super().__init__(*args, **kwargs)
+        self.fields['editor'].queryset = self.fields['editor'].queryset.exclude(
+            username=user.username)
+
+
 class EditSubmissionForm(forms.ModelForm):
     """
     For an editor to make a decision regarding a submission.
--- a/physionet-django/console/templates/console/awaiting_authors.html
+++ b/physionet-django/console/templates/console/awaiting_authors.html
@@ -75,4 +75,11 @@
 
 {% block local_js_bottom %}
 <script src="{% static 'custom/js/enable-popover.js' %}"></script>
+{% if project.editor == user %}
+<script>
+  function set_editor_text(){
+    $('#project_author').text($( "#id_editor option:selected" ).text());
+  }
+</script>
+{% endif %}
 {% endblock %}
--- a/physionet-django/console/templates/console/awaiting_authors.html
+++ b/physionet-django/console/templates/console/awaiting_authors.html
@@ -78,7 +78,7 @@
 {% if project.editor == user %}
 <script>
   function set_editor_text(){
-    $('#project_author').text($( "#id_editor option:selected" ).text());
+    $('#project_editor').text($( "#id_editor option:selected" ).text());
   }
 </script>
 {% endif %}
--- a/physionet-django/console/templates/console/copyedit_submission.html
+++ b/physionet-django/console/templates/console/copyedit_submission.html
@@ -97,4 +97,12 @@
 {% if project.has_other_versions or project.version_order %}
   <script src="{% static 'custom/js/disable-title.js' %}"></script>
 {% endif %}
+
+{% if project.editor == user %}
+<script>
+  function set_editor_text(){
+    $('#project_author').text($( "#id_editor option:selected" ).text());
+  }
+</script>
+{% endif %}
 {% endblock %}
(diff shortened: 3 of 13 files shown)


tompollard:
Thanks Felipe, this will be very useful. I haven't had a chance to review the code, but it works well from a user perspective. I've added a couple of quick thoughts below:

- Could you add an extra administrator to the fixtures to make this easy to test?
- Are there any simple tests that you could add to the testing framework?
- It looks like changes aren't logged. Will this be in a separate PR/issue?

elfeto:
> - Could you add an extra administrator to the fixtures to make this easy to test?

I added in the fixture another administrator
> - Are there any simple tests that you could add to the testing framework?

I added the test to the console tests
> - It looks like changes aren't logged. Will this be in a separate PR/issue?

I added a log message

tompollard:
Looks good to me. I'm happy to merge. @bemoody okay with you? A couple of minor things, indirectly related to this PR, that could be dealt with independently:

- It would be good to have a line break before "Regards" in the email sent to editors:

Screen Shot 2020-04-01 at 16 30 29

- On the project submission timeline (e.g. http://localhost:8000/projects/P4HIXQ3yZtNV11JoaUQw/submission/), I think we should remove the editor's name and email address (which solves the issue that the page displays the current editor, rather than the editor at the time it was assigned).

Screen Shot 2020-04-01 at 16 19 24

elfeto:
> On the project submission timeline (e.g. http://localhost:8000/projects/P4HIXQ3yZtNV11JoaUQw/submission/), I think we should remove the editor's name and email address (which solves the issue that the page displays the current editor, rather than the editor at the time it was assigned).

I can replace the text saying an editor has been assign, removing the name and email, but that will be overwritten (the date) if the editor is reassigned. (its just how its written)

elfeto:
@bemoody  can you take a look at this and the problems @tompollard mentions of the submission timeline.
>On the project submission timeline (e.g. http://localhost:8000/projects/P4HIXQ3yZtNV11JoaUQw/submission/), I think we should remove the editor's name and email address (which solves the issue that the page displays the current editor, rather than the editor at the time it was assigned).

The problem there is that there is that the editor is part of the "SubmissionInfo" which only allows one person and once its re-assign the `editor` and `editor_assignment_datetime` changes.

This will show omit the first editor and editor assignment date from the timeline. 

I believe @tompollard  also wants to remove the editor information from the timeline, but even by replacing it for `An editor has been assign to your submission` or the text you want to add, the dates will change if the project is re-assigned. 

This could be replaced with a new model of project logging. Assuming we log into a model all actions done in a project like:
 - Creation
 - Submission
 - Editor assignment and reassignment
 - Author asked to do changes
 - Copyedit changes / no changes
 - Author approval / Who approves
 - Publish

I might have missed one or 2, but that can be added in the proper code / PR.

The same could be done to users... to track changes in an easy way to find.

tompollard:
Thanks @elfeto, overall I think this is an improvement and I'd be happy to merge. 

As you mention, the issue that we create is this: currently we know who was responsible for all editorial decisions (acceptance, copyediting, publishing) because it is implicit, but when we implement the change we only know the current editor. 

I'm good with making a separate PR to introduce fixes to capture the changing state, but I'd like @bemoody to confirm that he is also happy.

bemoody:
Yeah, this is ugly.  I agree the EditLogs and CopyeditLogs and
hardcoded timestamps and such are not ideal.

Probably the easiest way for now would be:
- don't show the editor's name/address in the timeline
- don't change the editor_assignment_datetime when re-assigning

but it would certainly be good to think about having a more flexible
system for logging events in the timeline.

tompollard:
Thanks @bemoody. @elfeto, please could you make the suggested changes?

> - don't show the editor's name/address in the timeline
> - don't change the editor_assignment_datetime when re-assigning

Once these are done, let's merge and then try to address the state issue in a new PR.

elfeto:
There is no field in any models that mark a re-assignment.
> - don't change the editor_assignment_datetime when re-assigning

Would not be possible. This is a show or don't show scenario. 
Unless we add a field to the model `project_reassignment` as bool.

elfeto:
NVM on my comment, I didndt read the `don't change`. I read don't show :D

bemoody:
Is the idea that only the current editor is able to reassign the
project?  That doesn't appear to be enforced.

I think it would be clearer to add a new function "reassign_editor"
rather than adding arguments to "assign_editor".  Or else make the
behavior implicit: you update submission_status and
editor_assign_datetime *only if* submission_status is currently less
than 20.

What is this about?

+{% if project.editor == user %}
+<script>
+  function set_editor_text(){
+    $('#project_author').text($( "#id_editor option:selected" ).text());
+  }
+</script>
+{% endif %}

What does the project author have to do with anything?

elfeto:
>I think it would be clearer to add a new function "reassign_editor"
rather than adding arguments to "assign_editor".  Or else make the
behavior implicit: you update submission_status and
editor_assign_datetime *only if* submission_status is currently less
than 20.

The submission status is 10 at that point. This function before this PR was used only once and it was to alter the change of the submission status while assigning the editor.

tompollard:
Thanks everyone, I think the core issues are now addressed. Nice work @elfeto. @bemoody good spot on `project_author` and I agree that handling the reassignment in a new function looks cleaner!

Issue #975, opened 2020-03-27
Add and improve heading styles for project descriptions


tompollard:
Currently the text editor for project descriptions lists a single header "Heading 3":

![Screen Shot 2020-03-27 at 13 21 49](https://user-images.githubusercontent.com/822601/77782807-3a787200-702e-11ea-9c93-6f0cf2174a5e.png)

The style of "Heading 3" appears almost identical to the section header (which has an `<h2>` tag). e.g. in the example below, "Methods" is the section header and "Data Acquisition" is Heading 3.

![Screen Shot 2020-03-27 at 13 21 54](https://user-images.githubusercontent.com/822601/77783114-abb82500-702e-11ea-9cb0-56cf715e82de.png)

We should improve this at some point. Regardless of the HTML tags that are used in the document, we may want to consider section headers (e.g. in this case Abstract, Introduction, Methods) to be `Heading 1`, and then we should support at least `Heading 2` and `Heading 3` and `Heading 4`.

bemoody:
If you change format_tags to 'p;h3;h4;h5' then it will show "Heading
3", "Heading 4", "Heading 5".  This is implemented by the "format"
plugin and it doesn't look like there's a way to change the labels or
add custom options.

bemoody:
> This is implemented by the "format" plugin and it doesn't look
> like there's a way to change the labels or add custom options.

Weirdly enough, however, it is possible to change what *tags* it uses
(config.format_h1 etc.) so in theory we could have the "h1" option
actually generate an h3 tag.

tompollard:
> Weirdly enough, however, it is possible to change what *tags* it uses (config.format_h1 etc.) so in theory we could have the "h1" option actually generate an h3 tag.

Thanks for looking into this. Something like that would work. Kind of strange for us, but clear to the user. For reference, https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0216636 formats the article title as h1 (with id="artTitle"), then section headers as h2, subsection headers as h3 etc.

tompollard:
I think it might be worth reopening this. We have added support for additional headers (h2, h3 etc), but the issue of h2 and h3 headers looking very similar is still there:

![Screen Shot 2021-03-22 at 13 18 55](https://user-images.githubusercontent.com/822601/112031150-629b9380-8b11-11eb-96ee-a7adf1fffa93.png)

(the HTML for this chunk is:

```
<h2 id="description">Data Description</h2>
[...]

PR #986, opened 2020-04-01
More heading tags

bemoody:
This adds more heading tags to the ckeditor format menu.  It also renumbers them, so "Heading 2" is actually h3, "Heading 3" is actually h4, and "Heading 4" is actually h5.  See issue #975.

I'm not convinced that this is actually less confusing, but that's okay. :)

One concern I do have is that adding these options might result in authors choosing to use h4 rather than h3 because "it looks prettier", producing an inconsistent document structure.

### DIFF
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -305,7 +305,14 @@ CKEDITOR_CONFIGS = {
         'removeDialogTabs': 'link:advanced',
         'disableNativeSpellChecker': False,
         'width': '100%',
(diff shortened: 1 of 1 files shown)


tompollard:
This is a big improvement.

Screen Shot 2020-04-01 at 17 14 58

> One concern I do have is that adding these options might result in authors choosing to use h4 rather than h3 because "it looks prettier", producing an inconsistent document structure.

I agree this is an issue, but inconsistent use of headers is better than people inventing their own header styles (all caps, bold, italic, 17 point...)!

PR #988, opened 2020-04-02
Make published files read-only when loading demo

bemoody:
When projects are published, we make their files read-only.  For better testing, simulate this when running the "loaddemo" command as well.


elfeto:
Looks good

PR #993, opened 2020-04-03
Check if the legacy reference has been contacted

elfeto:
If the application has a reference from a legacy application and
that reference has been contacted in the current application, then
mark the reference as contacted not as known.


tompollard:
i'll take your word for it!

PR #994, opened 2020-04-03
Send "New project published" email to PhysioNet team

### LINKED ISSUE #974: Send "New project has been published email" to PhysioNet team
To keep everyone updated on new contributions, it would be nice if an email could be sent to the PhysioNet team upon publication (either copying the contributor email to the team, or sending a new email).

shulim:
Copies the contributor email to the team so that when a project is published, the PhysioNet team is notified with the same email as the contributor. This keeps everyone updated on new contributions and fixes #974.


elfeto:
You could use the bcc option for this.
https://docs.djangoproject.com/en/3.0/topics/email/#emailmessage-objects

elfeto:
NVM on my comment, I forgot to see the loop :)

bemoody:
Sounds great.  Thanks!

Some things that could be improved here: this sends the same message
(subject and body) that were sent to the last author.  ("We are
delighted to inform you that  your project has been published...")

Since this message is not directed at the authors, it would be better
not to say "your project" etc.  For people reading the message who
haven't heard of the project, it would be good to include the project
abstract in the message.

shulim:
> Since this message is not directed at the authors, it would be better not to say "your project" etc. For people reading the message who haven't heard of the project, it would be good to include the project abstract in the message.

Thank you for the feedback! I added a new email template that gets sent to the team, changed the name in the email so that it says "Dear Colleague" instead of the author name, and fixed the wording so the message is directed to the team instead of the authors. 

I like the idea of adding the project abstract in the message, but I am having some issues with formatting it in the email. If you don't mind, I'd like to work on that in a different issue.

shulim:
@elfeto Thank you for taking a look!

tompollard:
thanks @shulim. @elfeto and @bemoody are you happy to merge?

bemoody:
Haven't tested it but it looks fine to me.

Issue #969, opened 2020-03-26
Self managed credentialing - Appoint further reviewers


izcram:
Besides the corresponding author, other physionet users should be able to review requests.

The corresponding author should have the possibility to appoint other physionet users (not necessarily authors) to view and respond to access requests.

This would probably require an additional table `project_reviewer` with the following attributes:
  * user id
  * project id
  * appointing date
  * is revoked (corresponding author can revoke access)
  * revoke date
  
It should be possible to appoint again a revoked user.

Once a user is appointed, he/she is notified via email.

tompollard:
As part of this discussion, I'd like to clarify the difference between submitting author and corresponding author. I don't remember exactly how these two roles differ.

`project_reviewer` might not be quite right, because I think it will get confused with people who are invited as external peer reviewers for projects (forthcoming functionality!). Other than that, your suggestion sounds good to me.

PR #996, opened 2020-04-07
Invite further users to review access requests for self managed projects

izcram:
Addressing #969

Still using the term "Reviewer" as I couldn't come up with a better term, but qualified it to give "Data Access Request Reviewer".

So far, the handling of data access requests and inviting further reviewers can only be done by the corresponding author. However, I actually don't see any reason why the submitting author shouldn't be able to do so as well. Should this be changed?


Lucas-Mc:
I just got a chance to look at this and test everything and it looks great! Everything appears to be working as expected, meaning only the author and user with admin rights can respond to the data access request. 

It does have some funny glitches though.. the author can request access to themselves and then two duplicate projects will show up in Published Projects. Also, if I invite a reviewer and then try to back out of the page they will disappear in the Current Reviewers section after which I can re-add them and it says they already exist though their name reappears. The same thing happens for revoke also. Not sure if this is a big issue but something for others to discuss!

I think, for my preference, it would be nice to back out of the page after the invite without that middle step of the invited reviewer disappearing and the revoke box reappearing.

tompollard:
@Lucas-Mc thanks for looking. Please could you add some detail on how to reproduce these issues? I wasn't able to reproduce them using the steps below. 

> the author can request access to themselves and then two duplicate projects will show up in Published Projects.

- login as admin. accept credentialing request from george. logout as admin.
- login as george (author of http://localhost:8000/content/demoselfmanaged/1.0.0/)
- submit access request via http://localhost:8000/request-access/demoselfmanaged/1.0.0/
- accept the request at http://localhost:8000/access-requests/demoselfmanaged/1.0.0/

I only see "Self Managed Access Database Demo" listed once at: http://localhost:8000/projects/. Am I looking in the wrong place?

> The same thing happens for revoke also. 

- If I am at http://localhost:8000/manage-reviewers/demoselfmanaged/1.0.0/, I'm able to add and revoke users without any issue (e.g. I can invite "admin", then revoke admin, then invite admin etc without trouble).

> if I invite a reviewer and then try to back out of the page they will disappear in the Current Reviewers section after which I can re-add them and it says they already exist though their name reappears

- If I login as george, then click on the "manage reviewers" button at http://localhost:8000/projects/, I am taken to http://localhost:8000/manage-reviewers/demoselfmanaged/1.0.0/
- When I add a username (e..g "admin"), then click "invite reviewer" the user appears.
- By "backing up", do you mean clicking the backward arrow in your browser? If so, I end up at http://localhost:8000/projects/. 
- Clicking the "manage reviewers" button takes me back to http://localhost:8000/manage-reviewers/demoselfmanaged/1.0.0/, which includes an entry for the admin user. 

I agree that it would be good to have a clearer mechanism for navigating the site from the reviewer management pages. It shouldn't be necessary to use the backward arrow in the browser.

tompollard:
@izcram thanks, I agree, this is looking good. I've had a quick play around, and will look more closely later. A couple of quick things:

1. The new "manage reviewers" and "requests" buttons at: http://localhost:8000/projects/ could do with a tweak to the styling (perhaps display them on a new row?).

Screen Shot 2020-04-09 at 18 19 55

2. Not a big issue, but I wonder whether it would be better to invite reviewers by email address, rather than username? This makes it easier to invite someone who doesn't already have an account.

Screen Shot 2020-04-09 at 18 28 10

As a side issue, the way that accepted/rejected/withdrawn requests are displayed at http://localhost:8000/request-access/demoselfmanaged/1.0.0/ and http://localhost:8000/access-requests/demoselfmanaged/1.0.0/ could be improved. It can be a little confusing to make of sense of the outstanding applications at the moment. I think we already know this though, and it can be dealt with outside this pull request.

Lucas-Mc:
Here are the steps to re-create the issue:

1. Login as “george” and go down to “Self Managed Access Database Demo”
2. Click on “Manage Reviewers” and enter “george” into the “Physionet Username” box and click “Invite Reviewer"
3. Click on “Projects” at the top right of the screen next to “george” and scroll back down to Published Projects and you should see two “Self Managed Access Database Demo” projects 

Dup_Projects

tompollard:
@Lucas-Mc, ah interesting, well spotted!

izcram:
Thanks both @Lucas-Mc and @tompollard  for looking at it!

@Lucas-Mc : Nice catch with the self invitation! This is fixed in 7c58349 .
Regarding navigation: I've added a backlink to the project home. This doesn't address the confusion a user might have when using the browser navigation, I'm open for ideas if you think this should be fixed.

@tompollard : regarding the buttons: I could try to put "Manage Reviewers" and "Requests" button on a second line, but then I'm not sure how the alignment should work. If we keep the "float:right" then the two "button lines" will be aligned  on the right but not on the left, which looks a bit weird.
Another possibility could be to put all the buttons on a new line below "Version". That would use more vertical real estate. However, it may overall look a bit better on narrower screens, in particular mobile devices. Not sure how important these are to the project.
What option would you prefer?

I had similar thoughts regarding e-mail addresses, but then dropped it out of simplicity. The main difficulty  would be to link the e-mail addresses to the user accounts. That is, the person inviting would need to know which address was used to create the account (at ETH everyone has at least two aliases, people in general may have several e-mail addresses and may not remember which one they used to register on physionet...). So, if a person gets invited with a "wrong" address either that address has to be linked to the account or the corresponding author has to send out an invitation again....

elfeto:
I would like of there would be a button for reviewers to stop being a reviewer. 
- For example if `George` invites `rgmark` as a reviewer, then after X time `rgmark` decides he doesn't want to be a reviewer, he has to contact `George` and ask him to remove him. There is no way for `rgmark` to remove himself as a reviewer.

Under `DataAccessRequestReviewer` the field `invitation_date` should be renamed to `added_date` because an invitation can be rejected by the person, and once added you are forced to be a reviewer. 

I think `InviteDataAccessReviewerForm` should be a `model.form` of `DataAccessRequestReviewer`

Under `DataAccessRequestReviewer` you should be able to set a unique together
```
unique_together = (("project", "reviewer"),)
```
In the `Current Reviewers` page, you should list all reviewers even revoked ones, but instead of the revoke button, just display the date when they were revoked. Sorting them as Active, revoked newest-latest.

Under the model `DataAccessRequestReviewer`, please add a related_name to the reviewer.

izcram:
Thanks @elfeto  for your comments, I have addressed most of them in the last two commits.

Regarding the model form, I agree with you, however, see why it is not done using model forms in [this comment](https://github.com/MIT-LCP/physionet-build/pull/996/files#diff-b13d1a20cdeeac8dde800b8565ab3689R998)

As for the "self revocation"/"unsuscribe" functionality you mention, I can happily implement it by just adding a suitable button in the `data_access_requests_overview`. And the corresponding author should then also be notified via e-mail about this. However, I'm not quite sure how frequently this option would be used justifying the additional code for it. Your call.

elfeto:
Doing something like this would make use of the model form without the select form on the template, which will force you to write the username. (I'm not particularly keen overwriting the reviewer, but if I just set the widget as a textfield, it wont accept any text expecting a user object and it would receive a string)
```
class InviteDataAccessReviewerForm(forms.ModelForm):

    reviewer = forms.CharField(widget=forms.TextInput(
        attrs={'class': 'form-control'}),
        required=True, label='Physionet Username')

    class Meta:
        model = DataAccessRequestReviewer
        fields = ('reviewer',)

    def __init__(self, project, *args, **kwargs):
        super(InviteDataAccessReviewerForm, self).__init__(*args, **kwargs)
        self.project = project

    def clean_reviewer(self):
        reviewer_uname = self.cleaned_data['reviewer']
        try:
            reviewer = User.objects.get(username=reviewer_uname)
            if self.project.can_approve_requests(reviewer):
                raise forms.ValidationError(
                    f'User {reviewer_uname} is already allowed to review requests!')
        except User.DoesNotExist:
            raise forms.ValidationError(
                f'No user {reviewer_uname} found!', code='user_not_found')

        return reviewer

    def save(self):

        if self.errors: return

        invitation = super().save(commit=False)
        invitation.project = self.project
        invitation.save()
        return invitation
```

Also, how about instead of iterating to revoke something like this:
```
        reviewer_id = request.POST['revoke_reviewer']

        try:
            entries = DataAccessRequestReviewer.objects.get(
                project=project, reviewer_id=reviewer_id, is_revoked=False)
            entries.revoke()
        except DataAccessRequestReviewer.DoesNotExist:
            pass
```
That would have to be accompanied with a revoke method under the DataAccessRequestReviewer

I think its worth the extra code for the remove function. (but thats just my opinion)
I haven't checked your last changes, I will try to do that today.

izcram:
@elfeto  indeed, your `ModelForm` suggestion works! I've incorporated this now.

I also added the functionality, that a user can remove him/herself from the reviewers list by adding a button `data_access_requests_overview`. The owner(s) of the project (i.e. corresponding author and submitting author) get a notification via e-mail.
Note, that we may want to tweak the wording, I didn't spend much time on this.

elfeto:
try not to do bare excepts (line 1853)
```
    except PublishedProject.DoesNotExist:
```

This can be replaced
```
def is_data_access_reviewer(self, user):
    return DataAccessRequestReviewer.objects.filter(
        reviewer=user, is_revoked=False, project=self).exists()
```
with this
```
def is_data_access_reviewer(self, user):
    return self.data_access_request_reviewers.filter(reviewer=user, is_revoked=False).exists()
```

You can simplify the save function:
```
def save(self):
    if self.errors:
        return

    reviewer = self.cleaned_data['reviewer']

    invitation = DataAccessRequestReviewer()
    if DataAccessRequestReviewer.objects.filter(reviewer=reviewer,
                                                project=self.project).exists():
        # updating existing row in case a revoked user gets readded again
        invitation = DataAccessRequestReviewer.objects.get(
            reviewer=reviewer,
            project=self.project)
    else:
        invitation.reviewer = reviewer
        invitation.project = self.project

    invitation.is_revoked = False
    invitation.added_date = timezone.now()
    invitation.save()
    return invitation
```
to:
```
def save(self):
    if self.errors:
        return
    # If an object exists with `reviewer` and `project` it will update the defaults 
    # fields or it will create an object with all the fields specified.
    invitation, created = DataAccessRequestReviewer.objects.update_or_create(
        reviewer=self.cleaned_data['reviewer'], project=self.project,
        defaults={"is_revoked": False, 'added_date': timezone.now()})

    return invitation
```

No need to do any changes, but just as an FYI
The message inside the Http404 will only show if debug=True
If debug=False, then it's overwritten  by the template.

Apart from that, I don't see any other thing to be changed.

I didn't paid attention to the wording of the HTML text nor the email text since @tompollard  is better than me at that.

izcram:
Thanks @elfeto , with the `update_or_create` it looks much nicer!
I'll incorporate these suggestions together with suggestions @tompollard might have

physionet-django/notification/templates/notification/email/notify_owner_data_access_review_withdrawal.html
```diff
@@ -0,0 +1,7 @@
+Dear {{ owner.get_full_name }},
+
+This is to inform you, that the user {{ user.get_full_name }} wished to stop reviewing data access requests for the project {{ project }}.
```
tompollard: Minor suggestions: (1) no need for the comma in `inform you, that` (2) instead of `wished to`, `has decided to` sounds more natural.

physionet-django/notification/templates/notification/email/notify_user_invited_managing_requests.html
```diff
@@ -0,0 +1,9 @@
+Dear {{ invitation.reviewer.get_full_name }},
+
+You have been invited to review access requests for the project {{ invitation.project }}.
+
+You can view and respond to requests following this link:
```
tompollard: perhaps add "with" or "via" (e.g. `You can view and respond to requests with the following link:`).

tompollard (commented):
Just had a quick look at the emails and added a couple of very minor suggestions!

tompollard:
thanks @izcram, looks good to me (and thanks for the testing/reviewing @elfeto!)

PR #997, opened 2020-04-07
Remove sandbox for training report PDF

elfeto:
Due to the sandbox issue and chrome, the PDFs from the user training reports cant be rendered while I use google chrome.

https://bugs.chromium.org/p/chromium/issues/detail?id=413851

Since this is only for the PDF, I don't see a harm setting the sandbox and attach flags to False to render the PDF.

### DIFF
--- a/physionet-django/user/views.py
+++ b/physionet-django/user/views.py
@@ -486,7 +486,9 @@ def training_report(request, application_slug):
 
     if request.user == application.user or request.user.is_admin:
         try:
-            return utility.serve_file(application.training_completion_report.path, False)
+            return utility.serve_file(
+                application.training_completion_report.path, attach=False,
+                sandbox=False)
         except FileNotFoundError:
             raise Http404()
 


bemoody:
I'd say a better answer is don't use chrome!  Or download the PDF and
load it in a secure viewer.

Only way I could see this working is if you ran something to sanitize
the PDF first.

tompollard:
> I'd say a better answer is don't use chrome!  Or download the PDF and load it in a secure viewer.

Users have emailed to say that they were unclear whether or not their credentialing application had been submitted correctly because they were unable to view the PDF, so it would be good to find a fix.

~50% of people use Chrome, so not using Chrome isn't really a solution. 

It's also not clear how to download a PDF when it isn't rendered. You just see a blank screen and a URL without the .pdf filename. e.g.  https://physionet.org/credential-applications/ID_NUMBER/training-report/

bemoody:
Ah, gotcha, you're talking about viewing your OWN training report,
that you uploaded?  I didn't realize that was possible.

elfeto:
Me as an admin cannot view ANY training report that ANY person submits if I use chrome that is my default browser.
PS. The same happens with safari, and changing browsers to the only one that it works isn't a solution.

Also all users have the option of seeing their PDFs, and lots of users submit a wrong PDF. I assume they check which PDF they submitted to then submit a correct one.

elfeto:
If you feel strongly about the sandbox, a compromise would be to set the attachment flag to True to download the PDF, and hence we don't break the sandbox, and I get to see the PDF.

tompollard:
> Ah, gotcha, you're talking about viewing your OWN training report,
that you uploaded? 

Yes, that's right. Users are able to review their uploaded PDF when submitting an application for credentialed access.

bemoody:
> Me as an admin cannot view ANY training report that ANY person submits if I
> use chrome that is my default browser.

Right, that's the problem.  PDF files can contain malicious code and
should not be executed especially if you are logged in as an
administrator.

PR #999, opened 2020-04-08
Adds file naming suggestions

Lucas-Mc:
Adds more content to the File Naming section on the Share page to better inform the author about conventions that should be employed when naming files.


tompollard:
Thanks @Lucas-Mc, looks good! Couple of minor things:

- I know that it already appears this way, but please could you change "File Naming" to "File naming", so the case is consistent with other bullets?
- My understanding is that the following sentence refers to waveform labels rather than filenames, so I'm not sure this belongs in the file name section? "For signal names, upper case should be used where it improves readability (e.g. ABP_Sys is better than abp_sys).".

tompollard:
looks good to me, thanks!

Issue #1000, opened 2020-04-08
Increase word limit for "organization" on credentialing applications


tompollard:
Note from @kpierceHST 

> The 60-character limit (I think it's that) for "organization" is not sufficient for many applicants.  Generally it's just a truncated word, like "universi", but here there appears to be an entire word missing.

PR #1001, opened 2020-04-09
Increases organization character limit #1000

Lucas-Mc:
Increases the organization name character limit inside of the credentialing application from 60 characters to 200 characters.


tompollard:
Thanks Lucas, looks good to me. @bemoody I don't have a good handle on how https://github.com/MIT-LCP/physionet-build/pull/852 affects migrations. Is there anything needed here before merging? (the max_length length of the `CredentialApplication. organization_name` field has increased).

bemoody:
Good question.  In this case there is no issue because the format of
the data hasn't changed, either on the database side or on the Python
side.

You can test this using test-upgrade.sh, i.e.:

    git fetch origin
    git checkout origin/dev
    git merge --no-ff origin/incr_char_limit -m 'TEST MERGE'
    ./test-upgrade.sh origin/production

although I don't think there are any test cases for credentialing, so
this doesn't mean much :(

tompollard:
Thanks, that's helpful. I'm going to merge this for now, but we should add tests later.

PR #1003, opened 2020-04-10
Track disk usage of published vs. new files

bemoody:
As I think you all know, we save space on the server by creating hard links rather than copying files between versions.  If you have a 1 GB file in version 1.0, and you create a version 1.1 without modifying that file at all, there's only one copy of the data, and it only takes up 1 GB of space.

Currently, however, your storage allowance is simply a limit on the sum of the sizes of files within the active project directory.  This is simple for authors to understand, but it doesn't reflect reality - if you delete that 1 GB file and upload a new 1 GB file in its place, your visible "storage usage" hasn't changed but you're now using 1 GB more disk space than you were before.

With these changes, instead, the "usage" displayed will include the total disk space used by previously published versions of the same project.  This gives a more accurate picture of how much space your project is actually taking up.  (If you request 20 GB of storage and use 15 GB of it, and now you want to replace those files with 15 GB of new files, it doesn't seem unreasonable that you should need to submit a new storage request.)

This system, unlike the previous, is also compatible with the way filesystems measure disk usage, so it lays the groundwork for switching to kernel-enforced storage quotas, which I plan to do in a later pull request.

(To avoid counting files twice in this system, we look at the link count and the modification time: if the file has multiple links, and was last modified before the project was created, the file is considered to "belong" to an older version.  This is not rigorous but it should be good enough, and allows for testing without requiring any special configuration or privileges.)

Ultimately, these are prerequisites for efficiently managing large projects (on the scale of mimic3wdb or mimic-cxr) and for supporting some much-desired features (e.g. uploading via SFTP, or uploading as a ZIP archive.)


bemoody:
(Okay, I tried to tell github not to screw up the order of commits, by changing all the author dates to equal the committer dates, but clearly github managed to screw up the order anyway.  Sorry about that.)

tompollard:
:/ sorry, conflict with https://github.com/MIT-LCP/physionet-build/pull/1004

bemoody:
c'est la vie

bemoody:
Updated: further style improvements to the progress bar.

elfeto:
So, after some time reading this and going through this, it looks quite interesting. 

I have 2 questions:
Why is the "usable" model called `DemoQuotaManager`, I would assume that the word `Demo` would only be used for test and development not for production .
The `inodes_hard` and `inodes_soft` are always 0 for the time being. So shall I assume that this would be a temporary measure, until we can come up with some metrics?

Apart from that, I am looking forward to seeing this in staging.

bemoody:
Thanks for your review!

> I have 2 questions:
> Why is the "usable" model called `DemoQuotaManager`, I would assume that the
> word `Demo` would only be used for test and development not for production
> .

Well, I didn't finish implementing the other half of it :)

The idea was that the DemoQuotaManager would only be used in testing.
Like the current system, it's not efficient (need to read entire
directory tree on every page load) and it's not robust (limits can be
bypassed by somebody uploading many files quickly from multiple
connections.)

For production systems, the better option would be to use
kernel-enforced filesystem quotas.  This requires superuser
privileges, so my plan is to install a set of small scripts that are
then invoked using 'sudo'.  But, there are a number of different ways
to do that and I didn't want to overcomplicate this pull request.

(So, I'd be happy to change the name if you like, but that's why I picked it.)

Basically, this changes the model for how storage usage is measured.
So I wanted to let you all see how it would behave, and how it would
look from the author's perspective.  Would you find this intuitive?
Is there anything we could do to make it less confusing?

> The `inodes_hard` and `inodes_soft` are always 0 for the time being. So
> shall I assume that this would be a temporary measure, until we can come up
> with some metrics?

I've left open the option of setting inode limits... I do think there
should be a limit, but I think displaying both a byte limit and inode
limit might be too confusing, so my idea was "set the inode limit high
enough, as a fraction of the byte limit, that no one will ever hit it
unless they're doing something really outlandish."  Don't remember if
I implemented that in this pull request or not. :)

PR #1005, opened 2020-04-13
Escape paths correctly in /files/

bemoody:
Paths used in the X-Accel-Redirect header must be URL-escaped.  In particular, there's some buggy client out there that occasionally tries to download some files with %0D appended to the URL, which results in a 500 error because serve_file tries to set a response header with a control character in it.

In addition to control characters, this should allow correct handling of non-ASCII characters, as well as special ASCII characters like space, percent, and question-mark.  We don't currently allow any of those characters in published project files, but we should try to be robust.

### DIFF
--- a/physionet-django/physionet/utility.py
+++ b/physionet-django/physionet/utility.py
@@ -5,6 +5,7 @@ import shutil
 import struct
 import subprocess
 import tempfile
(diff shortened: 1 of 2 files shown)


elfeto:
Looks good.

PR #1004, opened 2020-04-10
Added fixture data for merge_user PR

elfeto:
This is just a fixture update with more information for the user merge PR #976

PR #1007, opened 2020-04-13
Revert "Added fixture data for merge_user PR"

elfeto:
Reverts MIT-LCP/physionet-build#1004 

I will do another PR adding the files, or just include them in the User merge PR.

### DIFF
--- a/physionet-django/project/fixtures/demo-project.json
+++ b/physionet-django/project/fixtures/demo-project.json
@@ -80,38 +80,6 @@
   }
 },
 {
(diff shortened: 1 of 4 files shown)


tompollard:
Does this need to be reverted? Better to fix the issues that were raised directly, if possible.

tompollard:
I'm closing this pull request because I think the PR was an improvement in general. It would be good to add the files as suggested by @bemoody. We can populate the abstract, background etc sections with more realistic data later.

PR #1008, opened 2020-04-13
Remove duplicated code for authors in UnpublishedProject

### LINKED ISSUE #984: Duplicated code for authors in UnpublishedProject
See lines 718 and 723. This may ultimately require a migration since it is part of the models but hopefully not! Still needs to be tested to be absolutely sure though.
https://github.com/MIT-LCP/physionet-build/blob/0301cfa6fd13d0a4c66f716a7b965e51a5111a82/physionet-django/project/models.py#L718-L723

shulim:
Deleted the repeated time "authors" is defined in UnpublishedProject and tested to make sure a migration isn't required, which fixes #984.

### DIFF
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -994,7 +994,6 @@ class UnpublishedProject(models.Model):
     publications = GenericRelation('project.Publication')
     topics = GenericRelation('project.Topic')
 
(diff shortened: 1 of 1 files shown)


bemoody:
Looks good.  Thanks for testing it!

PR #1010, opened 2020-04-14
Adds more citation styles for Physionet #990

Lucas-Mc:
Extends the functionality from #973 to include the standard Physionet citation. Further increases the width of the popup box to increase readability of lengthy citations.

### DIFF
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -608,9 +608,9 @@ class Metadata(models.Model):
             else:
                 return authors
 
-    def get_physionet_citation(self):
+    def get_platform_citation(self):
         """
-        Returns the information needed to generate the standard Phyionet
+        Returns the information needed to generate the standard platform
         citation in multiple formats (MLA, APA, Chicago, Harvard, and
         Vancouver).
 
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -608,6 +608,87 @@ class Metadata(models.Model):
             else:
                 return authors
 
+    def get_physionet_citation(self, style):
+        """
+        Returns the information needed to generate the standard Phyionet
+        citation in multiple formats (MLA, APA, Chicago, Harvard, and
+        Vancouver).
+
+        1. MLA (8th edition) [https://owl.purdue.edu/owl/research_and_citation/
+                              mla_style/mla_formatting_and_style_guide/
+                              mla_formatting_and_style_guide.html]
+        2. APA (7th edition) [https://owl.purdue.edu/owl/research_and_citation/
+                              apa_style/apa_style_introduction.html]
+        3. Chicago (17th edition) [https://owl.purdue.edu/owl/
+                                   research_and_citation/
+                                   chicago_manual_17th_edition/
+                                   cmos_formatting_and_style_guide/
+                                   chicago_manual_of_style_17th_edition.html]
+        4. Harvard [https://www.mendeley.com/guides/harvard-citation-guide]
+        5. Vancouver [https://guides.lib.monash.edu/ld.php?content_id=14570618]
+
+        Parameters
+        ----------
+        style [string]:
+            ['MLA', 'APA', 'Chicago', 'Harvard', 'Vancouver']
+
+        Returns
+        -------
+        citation_format [string]:
+            string containing the desired citation style
+        """
+        if style == 'MLA':
+
+            citation_format = ('Goldberger, A., et al. "PhysioBank, '
+                               'PhysioToolkit, and PhysioNet: Components of a '
+                               'new research resource for complex physiologic '
+                               'signals. Circulation [Online]. 101 (23), pp. '
+                               'e215–e220." (2000).')
+
+        elif style == 'APA':
+
+            citation_format = ('Goldberger, A., Amaral, L., Glass, L., '
+                               'Hausdorff, J., Ivanov, P. C., Mark, R., ... & '
+                               'Stanley, H. E. (2000). PhysioBank, '
+                               'PhysioToolkit, and PhysioNet: Components of a '
+                               'new research resource for complex physiologic '
+                               'signals. Circulation [Online]. 101 (23), pp. '
+                               'e215–e220.')
+
+        elif style == 'Chicago':
+
+            citation_format = ('Goldberger, A., L. Amaral, L. Glass, J. '
+                               'Hausdorff, P. C. Ivanov, R. Mark, J. E. '
+                               'Mietus, G. B. Moody, C. K. Peng, and H. E. '
+                               'Stanley. "PhysioBank, PhysioToolkit, and '
+                               'PhysioNet: Components of a new research '
+                               'resource for complex physiologic signals. '
+                               'Circulation [Online]. 101 (23), pp. '
+                               'e215–e220." (2000).')
+
+        elif style == 'Harvard':
+
+            citation_format = ('Goldberger, A., Amaral, L., Glass, L., '
+                               'Hausdorff, J., Ivanov, P.C., Mark, R., '
+                               'Mietus, J.E., Moody, G.B., Peng, C.K. and '
+                               'Stanley, H.E., 2000. PhysioBank, '
+                               'PhysioToolkit, and PhysioNet: Components of a '
+                               'new research resource for complex physiologic '
+                               'signals. Circulation [Online]. 101 (23), pp. '
+                               'e215–e220.')
+
+        elif style == 'Vancouver':
+
+            citation_format = ('Goldberger A, Amaral L, Glass L, Hausdorff J, '
+                               'Ivanov PC, Mark R, Mietus JE, Moody GB, Peng '
+                               'CK, Stanley HE. PhysioBank, PhysioToolkit, '
+                               'and PhysioNet: Components of a new research '
+                               'resource for complex physiologic signals. '
+                               'Circulation [Online]. 101 (23), pp. '
+                               'e215–e220.')
+
+        return citation_format
+
     def abstract_text_content(self):
         """
         Returns abstract as plain text.
@@ -911,12 +992,15 @@ class Metadata(models.Model):
 
         return citation_format
 
-    def citation_text_all(self):
+    def citation_text_all(self, physionet=False):
         styles = ['MLA', 'APA', 'Chicago', 'Harvard', 'Vancouver']
         citation_dict = {}
 
         for style in styles:
-            citation_dict[style] = self.citation_text(style)
+            if physionet:
+                citation_dict[style] = self.get_physionet_citation(style)
+            else:
+                citation_dict[style] = self.citation_text(style)
 
         return citation_dict
 
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -608,7 +608,7 @@ class Metadata(models.Model):
             else:
                 return authors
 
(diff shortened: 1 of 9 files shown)


tompollard:
Can you think of a way of using your `citation_text` method, rather than hardcoding each format? Whatever approach we use here is likely to be similar to our approach for the (silly-named) `PublishedPublication` objects.

Lucas-Mc:
@tompollard It would definitely be more bulky and computationally inefficient. It would have to compute the citation each time even though we already know what it should be. The issue now is that each method used (`initialed_name`, `get_full_name`, etc.) is based off an object which means the PhysioNet citation information would have to be included in that object somehow and I don't see the purpose if it is a standard for every publication.

tompollard:
@Lucas-Mc overall I think flexibility is more important for computational efficiency here. I'm not completely against hardcoding the formats, but it would be good to avoid if possible. It would be easier for us when modifying citation styles and updating the out-of-date citation, and easier for people repurposing the repo too. If generating the citations become an issue in terms of database/server load, then we can address that directly.

Lucas-Mc:
@tompollard So I guess that leaves two options (that I can think of now): (1) add the PhysioNet citation information to the PublishedProject (or Unpublished) object (actually it is confusing now I think those methods would still have to be changed since the object variables would be named different) or (2) modify the `initialed_name`, `get_full_name`, etc. methods to be compatible with raw text instead of objects. I prefer the second but what are your thoughts?

tompollard:
I guess whatever we do here should be consistent with how we handle citations for the `PublishedPublication` object. I'd need to give it some thought! @bemoody may have ideas...

tompollard:
If we want to temporarily hardcode the citations for now and postpone thoughts of generalizability (if you are okay with this @bemoody), then it might be worth:

1) leaving `citation_text_all` as it was, without the `physionet` argument.
2) letting `get_physionet_citation` return a dictionary, and call it directly from project/views.py (instead of `physionet_citations = project.citation_text_all(physionet=True)`)

This makes it easier to swap out the hardcoded chunk later. Might also be a good idea to put the modal in a separate, reusable template so it doesn't have to be duplicated. 

We should also use more general alternatives to `physionet` in the code where possible (thinking about moving towards a reusable platform).

Lucas-Mc:
@tompollard I pushed the suggestions you made in points 1. and 2. though I am not sure what exactly you want for the modal. Is there an example anywhere of a reusable HTML template? Or would it just be returned when writing a line such as `{% get_modal %}` in the HTML?

Should I just re-name "physionet_citation" to "platform_citation" and it can be changed as the platform does?

Lucas-Mc:
Commit history is ugly now because of the GitHub server being down in the middle of the push to this PR.

bemoody:
Sounds like a good opportunity to learn how to use 'git rebase -i'!  :)

Lucas-Mc:
> Sounds like a good opportunity to learn how to use 'git rebase -i'! :)

Done! 👍 (:

physionet-django/project/templates/project/published_project.html
```diff
@@ -87,7 +87,7 @@ <h1 class="form-signin-heading">{{ project.title }}</h1>
             <br><span>{{ citations.APA }}</span>
           </p>
           <div class="modal fade" id="citationModal" tabindex="-1" role="dialog" aria-labelledby="citationModalLabel" aria-hidden="true">
-            <div class="modal-dialog" role="document">
+            <div class="modal-dialog" role="document" style="max-width:800px">
```
tompollard: I think you'd planned on moving this style to the CSS?

physionet-django/static/bootstrap/css/bootstrap.css
```diff
@@ -4707,6 +4707,9 @@ button.close {
     max-width: 500px;
     margin: 30px auto;
   }
+  .modal-dialog.citation {
```
tompollard: It looks like you are editing the bootstrap.css, which would work but it would be overwritten if we update bootstrap. Could this go in one of the custom css files instead?

physionet-django/project/templates/project/published_project.html
Lucas-Mc: Hmmmm, I'm confused, I thought I fixed that with the last commit. Not sure where this branch is at now since it glitched out during the time github was down. I'll take a look at it.

physionet-django/static/bootstrap/css/bootstrap.css
Lucas-Mc: Ahhhh interesting, did not realize! I'll move it!

physionet-django/project/templates/project/published_project.html
Lucas-Mc: Ahhhh, maybe I forgot to change the published project as well, I didn't notice both of them the first look at the changes!

Lucas-Mc (commented):
The location of citation shouldn't matter, at least it didn't in testing, but I made this change to be consistent with the other `.modal-dialog.citation` CSS calls.

bemoody:
> 089e040bd863af623097751e5c9e8ca1b0c9eee3  Converts modal to template

I love that you're changing the modal stuff into a template - we
really need that!

I would rather make it a "global" template (i.e.
"physionet-django/templates/modal.html" instead of
"physionet-django/project/templates/project/modal_outline.html")
because it's something we should be using everywhere.

However, you should *never* disable autoescape in an HTML template.
Never, never, never.  Instead, ensure that the parameters are (when
appropriate) marked as "safe" - the best way to do this is to use
django.utils.html.format_html instead of str.format.

It's also a little annoying that this doesn't let you write the "modal
body" as part of the parent template.  One (somewhat messy) option
would be to define separate "modal_start.html" and "modal_end.html"
templates so you could write

{% include 'modal_start.html' with id_name='citationModal' title='Cite' %}
  <table><tbody>
  {% for style, citation in citations.items %}
    <tr>
      <th>{{ style }}</th>
      <td>{{ citation }}</td>
    </tr>
  {% endfor %}
  </tbody></table>
{% include 'modal_end.html' with id_name='citationModal' %}

... Or another option would be to define it as a custom tag so you
could write:

{% modal_box id_name='citationModal' title='Cite' %}
  <table><tbody>
  {% for style, citation in citations.items %}
    <tr>
      <th>{{ style }}</th>
      <td>{{ citation }}</td>
    </tr>
  {% endfor %}
  </tbody></table>
{% end_modal_box %}

... Or a third option would be to define each modal dialog as a
separate template file that inherits from modal.html, using the
{% block %} syntax.

Lucas-Mc:
Any thoughts on how to handle the original Physionet citation function? I know it would be better to not hard-code the citations in favor of using the existing citation style functions, however those functions work with objects. Currently, the original Physionet citation is not set up as an object so to make those functions work we'd have to find a way to make it an object. Any ideas?

tompollard:
> Any thoughts on how to handle the original Physionet citation function? I know it would be better to not hard-code the citations in favor of using the existing citation style functions, however those functions work with objects.

I think the hardcoded citation isn't too much to worry about for now, but it will create extra work if we make changes to any of the citation styles or if we add new ones. 

Currently `citation_text` generates the style strings (e.g. ```style_format = ('{author}. "{title}" (version {version}). ' '<i>PhysioNet</i> ({year})')```) and also gets project details from the project object itself.

If the task of generating the style strings was moved to a new function (e.g. `citation_format`) then you could reuse this component in your own `physionet_citation` function.

e.g. `physionet_citation` would assign values to `author`, `title`, etc and would then call `citation_format` to get the correct format. This would avoid the need to hardcode each individual style in the physionet citation function.

Lucas-Mc:
Nice idea @tompollard, I will work on this implementation and push an update once complete!

physionet-django/project/models.py
```diff
@@ -608,6 +611,88 @@ def get_author_info(self, separate_submitting=False, include_emails=False):
...
+            'Roger Mark',
+            'Joseph E Mietus',
+            'George B Moody',
+            'Chung Kang Peng',
+            'H Eugene Stanley'
+        ]
+
+        all_authors = []
+
+        for a in authors:
```
tompollard: if the author list is hardcoded (which seems necessary right now), what is the purpose of the loop?

physionet-django/project/models.py
```diff
@@ -608,6 +611,88 @@ def get_author_info(self, separate_submitting=False, include_emails=False):
...
+            all_authors.append(PublishedAuthor(
+                user=User.objects.get(username='admin'),
```
tompollard: `username='admin'`?

physionet-django/project/models.py
```diff
@@ -756,7 +841,8 @@ def get_anonymous_url(self):
 
         return anonymous.url
 
-    def citation_text(self, style):
+    def citation_text(self, style, platform='Physionet', authors=None,
```
tompollard: "Physionet" should be "PhysioNet", but I'm not sure we want this as a default argument (we want to avoid hardcoding this word where possible).

physionet-django/project/models.py
```diff
@@ -778,42 +864,56 @@ def citation_text(self, style):
...
-        else:
-            """
-            Since the project is not yet published, the current year is used in
-            place of the publication year, and '*****' is used in place of the
-            DOI suffix.
-            """
-            year = timezone.now().year
-            doi = '10.13026/*****'
+            else:
+                """
```
tompollard: `"""` isn't Python syntax for a comment as far as I'm aware (though might be wrong!). Generally best to use `#` at the start of a line.

physionet-django/project/models.py
Lucas-Mc: For some reason a user is required to create an instance of the `PublishedAuthor` class to use  its functions

physionet-django/project/models.py
Lucas-Mc: Yeah, I knew this would come up, has your global variable published yet? Oh wait, that would cause an error since I'm behind that push anyway

physionet-django/project/models.py
Lucas-Mc: That's used for a multi-line comment.. is there a PEP8 preference? It didn't show up in my linter

physionet-django/project/models.py
Lucas-Mc: The loop creates the `PublishedAuthor` instances from those names

Lucas-Mc: This is required to access the functions such as `initialed_name`, etc.

physionet-django/project/models.py
tompollard: It's a multiline string that hasn't been assigned to a variable, rather than a comment.

physionet-django/project/models.py
Lucas-Mc: I set it as a default since the platform would be used most often but I can remove it if that's better?

physionet-django/project/models.py
Lucas-Mc: Okay, I'll change it since I guess it's preferably used with docstrings, it was with the old code from whenever it was first developed

Lucas-Mc:
Reverted changes back to the hard-coded version since it was getting too messy to be worth the hassle.

tompollard:
Thanks Lucas, I think reverting those changes is best for now. Any thoughts on Benjamin's suggestions?

tompollard:
@Lucas-Mc and I just chatted. It looks like Benjamin's comments are dealt with, so merging.

PR #1012, opened 2020-04-14
Adds some basic Published Project usage stats

Lucas-Mc:
Adds the following data in tabular form for all published projects: Publish Date, Author Approval Date, Editor Assignment Date, and Submission Date. This also integrates the current charting feature for number of projects published throughout the year as well as its data usage. Further changes and improvements will provide some plotting of the average time taken in between review and final approval as well as additional requested information received here.


elfeto:
It's weird to see a time table of editors without editors, nor project.
Screen Shot 2020-04-16 at 9 00 17 AM

For the chars, it would be nice to click the chart item and see the projects under the # listed.
Screen Shot 2020-04-16 at 9 01 41 AM

Lucas-Mc:
@elfeto I can add the project title as the first column to the table if that sounds nice? Also, I'll see what I can do in reference to your other comment.. I hope it wouldn't make the tooltip too overwhelming.. or I suppose the tooltip wouldn't have to display it just once the bar is clicked on. For example, 2019 has 24 databases and that number is surely to grow.. where should all these project titles go?
Tooltip_24

tompollard:
We're really looking for summary stats, I think, rather than a list of submissions (imagine a weekly report for the team, giving us some indications about what's going well and where we can improve). Things like average time from submission to first decision; average time in author revisions; acceptance rate; etc would be good. Anything actionable is especially useful.

I'm not sure that we want to remove publicly accessible URLs unless completely necessary, even if they aren't being used to our knowledge. We created the public charts at https://physionet.org/charts/ because we were hoping to link to them from the about page. They just aren't at a point where we are ready to do this, so (in a separate PR) it would be good to improve them.

Lucas-Mc:
Thanks for the advice everyone! I am almost done making some improvements to it now and will post it when finished!

Lucas-Mc:
Just pushed some more edits to the current table. This includes extended table information such as time between project proposal events as well as graphs of this information. Currently, the time scale is in seconds on a log scale but I would like to find a better way to scale this. All thoughts and suggestions are welcome!

tompollard:
This feels like a couple of different pull requests in one. Let's get some basic details down as a starting point, and then we can move onto fancier things. 

How about displaying the following details on a page linked from a "Usage stats/Editorial" item in the menu? We probably need to ignore legacy projects for these stats to make sense. 

- Acceptance rate (what proportion of submissions are ultimately accepted?)
- Median days (not log seconds!) from submission to assignment of an editor.
- Median days from submission to first decision.
- Median days in each of the stages ("Awaiting editor assignment"; "Awaiting decision"; "Awaiting author revisions"; "Awaiting copyedit" etc).

Charts can come later. Improvements or changes to the current content at https://physionet.org/charts/ should be in a separate pull request.

Lucas-Mc:
Hey @tompollard, thanks for the suggestions! I haven't made any changes to `charts.html` just added charts to `usage_stats.html`. Should this still be separate? This would allow us to visualize the time between these project approval steps in a chronological way and the code for the graphs is already there. 

I can add a summary table of the values that you requested though I am hesitant to add time between revisions since not every project includes them, though if you want something more detailed than time between editor assignment and author approval then I can try.

tompollard:
> I haven't made any changes to charts.html just added charts to usage_stats.html. Should this still be separate? This would allow us to visualize the time between these project approval steps in a chronological way and the code for the graphs is already there.

The plots seem unnecessary, so I'd suggest leaving them for later and focusing on the basics. If we want to add plots later, these could be in a separate pull request. 

> I can add a summary table of the values that you requested though I am hesitant to add time between revisions since not every project includes them, though if you want something more detailed than time between editor assignment and author approval then I can try.

Either a summary table or a bullet point list would be fine. The suggestions above are just ideas, so do what you think makes sense. If it's difficult to calculate some of the stats, or some don't make sense, ignore them for now. Once we have the basic framework, we can start building on it.

Lucas-Mc:
@tompollard Okay, I'm working on a separate PR with only the basic stuff and I'll keep this branch for the advanced stuff since it already has the graphs. My idea with the graphs is that we could determine whether our response time during certain phases has either decreased or increased over time and we could attribute this to maybe increased/decreased workload or increased/decreased focus on project reviews or whatever it may be.

tompollard:
Thanks, I agree that some kind of time component would be good (e.g. allowing us to compare to the previous few months, previous year, or whatever). 

We can think about how best to do this once we have a few basic stats (perhaps raise an issue to gather ideas if there isn't one already?).

I'll close this pull request because otherwise we receive an email each time a commit is pushed if we are tracking the repo. It can be reopened later if needed.

PR #1014, opened 2020-04-16
fix error if resource_type is null

tompollard:
This fixes an error that is raised if resource_type is null in the content search string (e.g. http://localhost:8000/content/?orderby=blah-desc&topic=mimic&types=blah)

### DIFF
--- a/physionet-django/search/test_views.py
+++ b/physionet-django/search/test_views.py
@@ -1,0 +1,116 @@
+from django.test import TestCase
+from django.utils.html import escape
+from django.urls import reverse
+
+
+class TestProjectSearch(TestCase):
+    """
+    Tests for the project search engine.
+    """
+
+    def test_search_content(self):
+        """
+        Test the main content index.
+        """
+        url = reverse('content_index')
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+        self.assert_link(response, '/content/demobsn/1.0/')
+        self.assert_link(response, '/content/demoecg/10.5.24/')
+        self.assert_link(response, '/content/demoeicu/2.0.0/')
+        self.assert_link(response, '/content/demopsn/1.0/')
+        self.assert_link(response, '/content/demoselfmanaged/1.0.0/')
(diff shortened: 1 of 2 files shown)


elfeto:
This fails if there is no topic and the research type doesnt exists:
http://localhost:8080/content/?topic=&orderby=relevance-desc&types=5

I think we should re-do the search function up to some point like removing the check if its Postgres. (That should be done at a settings level.

bemoody:
Hmm, does this really work?  If you don't specify a topic (e.g.
http://localhost:8000/content/?types=fnord) it still seems to give a
TypeError.

tompollard:
This was intended as a quick fix to deal with the issue that I saw in the emails.

> I think we should re-do the search function up to some point like removing the check if its Postgres. 

I agree, the function could be improved.

tompollard:
If anyone has a better/alternative fix that can be implemented before the next round of emails, please make a PR!

bemoody:
The simple answer is that the bug is in content_index, it doesn't
handle the case where the form is not valid.

These tests should pass if you fix that :)

Something like:

--- a/physionet-django/search/views.py
+++ b/physionet-django/search/views.py
@@ -122,6 +122,8 @@ def content_index(request, resource_type=None):
         form_type = forms.ProjectTypeForm(request.GET)
         if form_type.is_valid():
             resource_type = [int(t) for t in form_type.cleaned_data['types']]
+        else:
+            resource_type = list(LABELS.keys())
     elif resource_type is None:
         resource_type = list(LABELS.keys())
         form_type = forms.ProjectTypeForm({'types': resource_type})

though as I mentioned, I don't like the fact that the same variable
"resource_type" is used first to hold a singular value, and later used
to hold a list of values.

elfeto:
I was going to suggest to convert the elif into an if statement. 

IF the form is NOT valid, then resource_type == None

tompollard:
alright this is a quick fix, i think we should merge to stop the emails! thanks for the tests @bemoody.

PR #995, opened 2020-04-06
Better handling of LICENSE/SHA256SUMS before publication

bemoody:
This fixes a couple of small issues related to publishing file content.

- LICENSE.txt should be visible to authors before approval (issue #520)

- Do not link SHA256SUMS.txt and LICENSE.txt from the previous project version (thought there was an issue about that but I can't find one.)

PR #1015, opened 2020-04-17
Don't regenerate LICENSE.txt at publication time

bemoody:
Followup to pull #995.

Now that all projects that were "awaiting author approval" or "awaiting publication" have been published, it is not necessary or desirable to regenerate the LICENSE.txt file at publication time.


elfeto:
Looks good.

PR #1017, opened 2020-04-17
Logger on database

elfeto:
Adds the logger app that records all logger messages to the database.
These are the same messages stored in the log files, but here we store them in the DB also.

In the DB model I added user and core_project.
For the future we can add things like expire.

This can help on stats.


bemoody:
Let's back up.  Define the problem you're trying to solve, and open an
issue so we can discuss it before deciding on a solution.

elfeto:
I created an issue for this #1025.

elfeto:
Ping on this?

tompollard:
@shulim will be wanting to use access logs (specifically views of project content) to generate public usage metrics. e.g. for a project like https://physionet.org/content/mimiciii/1.4/, we would want to display number of views of the project over time.

@bemoody if you think logging directly to the database is not a good idea (and I can see that it might not be), then the alternative is that we would would want to run a cron task to load relevant logging information into a database table. please could you add your thoughts to this thread, so we can make a decision on the pr?

bemoody:
Just to very quickly cover the issues that come to mind:

If what you are after is machine-readable long-term records of what
has happened, then dumping out human-readable text is, generally, no
good.

If what you are after is *debugging* then you want to display stuff in
a human-readable format, and you want to add new ad-hoc messages and
remove unneeded ones as your debugging needs change over time.

If you want to keep records long-term you also need to think carefully
about visitor privacy and what information actually needs to be
retained.

(And of course you need to think about performance and disk usage too. :))

Using Django to collect access statistics is problematic, too, because
we *want* to cut Django out of the loop whereever we can, because it's
a major bottleneck.  Access statistics (e.g. number of requests,
number of kilobytes) are best collected from the front-end (nginx).

So, think about what it is you're actually after, what problem you're
ultimately trying to solve.  Part of the solution *may* involve
recording more information in the database, but I don't think that
"dumping as much as we can in there and sorting it out later" is
likely to be a good strategy.

tompollard:
Okay, good points. So essentially I think we want to close this PR (unless there is anything you would like to push to keep @elfeto?)

PR #1019, opened 2020-04-17
Adds some basic editorial stats

Lucas-Mc:
Adds the ability to view the mean and median time taken between each step of the project submission and approval process. Further, the acceptance rate of submitted projects can also be seen.


elfeto:
Bare excepts are somewhat of a bad idea.

The try except is not needed for all of the fields in the dictionary.
These are published projects, if the project has one of the 4 fields of a revision it has all 4 fields.
One only have to check with an if/else or try/except only one of the following.
 - revision_request_datetime
 - editor_assignment_datetime
 - editor_accept_datetime
 - resubmission_datetime

The variables approved_projects and rejected_projects are unneeded in the template.

Then for example, what would this give us? with the dummy data of 4 published projects it just says acceptance rate of %100, and 0 for the rest of the information. (we did all in the same day?)
Screen Shot 2020-05-05 at 10 03 23 AM

physionet-django/console/views.py
```diff
@@ -1383,10 +1383,60 @@ def editorial_stats(request):
...
+        elapsed_times[project.title] = {}
+
+        if project.editor_assignment_datetime is not None:
+            elapsed_times[project.title]['time_sub_ed'] = \
+                abs((project.editor_assignment_datetime - \
+                project.submission_datetime).days)
+        else:
+            elapsed_times[project.title]['time_sub_ed'] = 0
+
+        if project.revision_request_datetime and project.editor_assignment_datetime is not None:
```
tompollard: Is `is not None` needed? (`None` evaluates to False)

Lucas-Mc: I suppose not, just being explicit, I can modify it

tompollard: Removing is consistent with the first clause (`if project.revision_request_datetime`) which does the same evaluation.

Lucas-Mc:
I modified the `demo-project` fixture to include more realistic revision dates and an `ArchivedProject` to adjust the acceptance rate.

Lucas-Mc:
Is all the information presented in the table useful? Would someone looking to investigate the average amount of days between certain steps of the publication process find all of these useful or are there some that I can cut out?

tompollard:
Just rebased on dev to fix the conflict on `console/urls.py`

tompollard:
Thanks for looking at this Lucas. I've listed a couple of thoughts below:

1) TypeError for legacy data (which presumably is missing some dates):

I began by loading the legacy data with:

```
# load the fixture
python manage.py loaddata project/fixtures/pbank.json 

# publish the projects
x = LegacyProject.objects.all()
for p in x:
    try:
        LegacyProject.objects.get(pk=p.id).publish()
    except:
        pass
```

After doing this, visiting http://127.0.0.1:8000/console/usage/editorial/stats/ raised a TypeError (unsupported operand type(s) for -: 'datetime.datetime' and 'NoneType').

2) the way the information is presented needs some thought. If just the demo data is loaded, the following webpage is displayed:

Screen Shot 2020-09-30 at 12 21 27

- in general, we are interested in using the metrics to keep an eye on how things are progressing over time. focusing on this point may help to determine what kind of information we want to include and how it is presented.
- clearly explaining how values are computed would make them more meaningful (e.g. does "acceptance rate" include all projects, or just projects that are accepted/rejected? does it include legacy content? does it include new versions of a project, or just the initial version? etc). 
- introducing more structure to the page would be an improvement. adding headers and explanatory text would help people to navigate the information. consider using separate tabs/pages if needed.
- what happens to the final table when there are several hundred projects? presumably it just creates a very large page. is that an issue?

tompollard:
How about simplifying to something similar to this as a starting point? 

Screen Shot 2020-09-30 at 13 51 23

physionet-django/console/templates/console/editorial.html
```diff
@@ -0,0 +1,33 @@
+{% extends "console/base_console.html" %}
```
tompollard: The filename of the template should indicate that this relates to stats (e.g. editorial_stats)

physionet-django/console/views.py
```diff
@@ -1402,6 +1403,75 @@ def guidelines_review(request):
         {'guidelines_review_nav': True})
 
 
+@login_required
+@user_passes_test(is_admin, redirect_field_name='project_home')
+def editorial_stats(request):
+    """
+    Editorial stats for reviewers.
+    """
+    approved_projects = PublishedProject.objects.all().order_by('-publish_datetime')
```
tompollard: "projects" is shorter and might avoid having to wrap across multiple lines below (not sure the "approved_" is necessary).

physionet-django/console/views.py
```diff
@@ -1402,6 +1403,75 @@ def guidelines_review(request):
...
+    # Filter out legacy projects
```
tompollard: the comments says "filter legacy projects" but the query is doing something different. you could drop these comments as the code is clear anyway.

physionet-django/console/views.py
```diff
@@ -1402,6 +1403,75 @@ def guidelines_review(request):
...
+    approved_projects = [i for i in approved_projects if i.editor_assignment_datetime != None]
```
tompollard: why not `projects = PublishedProject.objects.filter(is_legacy=False)` to filter legacy projects?

physionet-django/console/views.py
```diff
@@ -1402,6 +1403,75 @@ def guidelines_review(request):
...
+
+    # Create a set of years in which projects were published
```
tompollard: instead of creating a set of lists and zipping them together, it might be easier to populate a dictionary with year as the key? e.g. you could do:

```

years = [i.publish_datetime.year for i in projects]
stats = OrderedDict()

    # number published
    for y in sorted(set(years)):
        stats[y] = [years.count(y)]
```

physionet-django/console/views.py
```diff
@@ -1402,6 +1403,75 @@ def guidelines_review(request):
...
+    num_published = {}
+    for year in years_published:
+        num_published[str(year)] = 0
+
+    for project in approved_projects:
+        num_published[str(project.publish_datetime.year)] += 1
+
+    num_published = list(num_published.values())
+
+    # Determine the elapsed times between editorial events
```
tompollard: i find it difficult to follow a loop with lots of nesting like this. instead, the stats could be added in separate steps (which will make it easier to add new columns later). e.g.:

```
# submission to editor assigned
sub_ed = projects.annotate(tm=Cast(F('editor_assignment_datetime')-F('submission_datetime'),
                           DurationField())).values_list('tm', flat=True)

for y in stats:
    y_durations = sub_ed.filter(publish_datetime__year=y)
    days = [d.days for d in y_durations if d.days >= 0]
    try:
        stats[y].append(median(days))
    except StatisticsError:
        stats[y].append(None)
```

The query that I've suggested is pretty clunky, but I think it works!

tompollard: for submission to publication it should be:

```
# submission to publication
sub_pub = projects.annotate(tm=Cast(F('publish_datetime')-F('submission_datetime'),
                            DurationField())).values_list('tm', flat=True)

for y in stats:
    y_durations = sub_pub.filter(publish_datetime__year=y)
    days = [d.days for d in y_durations if d.days >= 0]
    try:
        stats[y].append(median(days))
    except StatisticsError:
        stats[y].append(None)
```

this adds some repetition, which isn't ideal. speed shouldn't be an issue as the page is only loaded internally.

physionet-django/console/views.py
```diff
@@ -1402,6 +1403,75 @@ def guidelines_review(request):
...
+        except statistics.StatisticsError:
+            # If there are no values to take the median of
+            yearly_sub_pub.append(0)
+
+    yearly_stats = zip(years_published,
+                       num_published,
+                       yearly_sub_ed,
+                       yearly_sub_pub)
+
+    return render(request, 'console/editorial.html',
```
tompollard: if the approach suggested above is used, the template would need to be changed slightly:

```
        {% for year, values in stats.items %}
          <tr>
            <td>{{ year }}</td>
            <td>{{ values.0 }}</td>
            <td>{{ values.1 }}</td>
            <td>{{ values.2 }}</td>
```

physionet-django/console/views.py
tompollard: (these need imports):

```
from django.db.models.functions import Cast
from django.db.models import F, DurationField
from statistics import median, StatisticsError
```

tompollard:
Looks good to me, thanks! Please could you squash the commits before we merge?

Lucas-Mc:
Done! Thanks @tompollard!

PR #1023, opened 2020-04-20
Pillow requirements for OSX

elfeto:
There is a missing hash on Pillow for OSX.


elfeto:
NVM this is a duplicate.

Issue #1024, opened 2020-04-20
Parse CITI verification links


bemoody:
When applying for credentialing, the applicant is asked to upload a PDF file documenting that they've taken a course in human subject research.

Valid training report files contain a link like this:

https://www.citiprogram.org/verify/?kaaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa-11111111

Training *certificate* files contain a link like this:

https://www.citiprogram.org/verify/?waaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa-11111111

(We don't accept the certificate for credentialing as it doesn't include enough detailed information.  This is a frequent source of confusion.)

These links point to a copy of the PDF file that anyone can download.  So as far as checking someone's training status is concerned, there's no need to look at the PDF file they uploaded, unless the verification service goes away for some reason.

For reviewing the application, the website could simply provide a link to citiprogram.org, avoiding the need for an extra step to check the report is authentic, and also avoiding many of the security problems of loading an untrusted PDF file from a trusted origin.

For the applicant, it would be good to have the form automatically rejected if they upload a "certificate" file.

It should be easy to do this since as far as I've seen, the original file is always PDF version 1.4 (hence, the URL is in plain text.)

tompollard:
The suggestion of parsing the CITI link is something that has been discussed in the past, so I'm surprised to see that this isn't captured in an existing issue. Seems like a good idea.

PR #1026, opened 2020-04-20
Improve handling of training report files

bemoody:
This makes a few related changes:

- When someone submits a credential application, try to extract the CITI verification URL from their training report (see issue #1024).

- When someone submits an application and it appears to contain a CITI "certificate", reject the submission.

- In the console pages for reviewing applications, link to the verification URL if one is found.  If a verification URL is not found, indicate this clearly (this should be a red flag for the application.)  A link to the original PDF is still provided, but...

- Serve the original PDF as an "attachment", forcing the browser to download it rather than displaying it inline.  This avoids showing a blank screen for (buggy) browsers that attempt to display PDF files inline but don't support sandboxing.

- Enforce a size limit on the completion report, because currently there doesn't seem to be one!

I believe this should work, but it could use further testing.

- Need to verify that legitimate CITI completion reports always have a "k" code and certificates always have a "w" code, and that both types are recognized by this function.

- Test this migration using a copy of the live database (and PDFs) to be sure it won't break anything.

- Test that the web interface behaves properly for applicants and for administrators.


elfeto:
The page `http://localhost:8000/console/complete-credential-applications/` is customized specifically for KP.

For my understanding he wants to open the PDF in another window not download the PDF.
There exists another page that is not used at the moment, you can set the download instead of opening in another window. (https://localhost:8000/console/credential-applications/)

In the cleaning I would add a check:
 - To see if the ` HIPAA and Human Subjects Research` course was taken and passed
 - That the dates are ALL valid (not expired)

but this is something that KP might have more insight in.

Shouldn't the clean raise that raises `Please upload the "Completion Report" file, not the "Completion Certificate"` be inside `TrainingCAF` rather than `CredentialApplicationForm`?

@kpierceHST  might be able to provide more insight into this PR, but he might want to move the internal mechanics to an email.

bemoody:
Thanks for your comments.  Just to be clear:

- If the person uploaded a real CITI completion report, then there
would be a link to view that document, in a separate window/tab, just
as there is currently.

- If the person did not upload a real CITI completion report then
there would only the link to download it.

We could change the form to require that you *must* provide a CITI
completion report, all other PDF files are rejected.  I was under the
impression that in some (unusual) cases, people could be approved
based on other equivalent training courses.  But if we simply say that
you *must* complete a course from CITI and you *must* upload the PDF
completion report that they sent you, it would make things much
simpler.

(Though then we'd get emails from people saying "I uploaded [a PDF
file that I printed from an MS Word document containing screenshots
of] the CITI training report, why isn't it working?")


If we do want to continue letting people upload whatever junk PDF
files they like, then yes, it'd be preferable to be able to view them
with one click rather than download-and-then-open.  I do think that
there should be a visible distinction between "you're viewing a PDF
file that the applicant uploaded" and "you're viewing a verified
training report from citiprogram.org".  Maybe using an iframe?


Parsing the text of the PDF file (let alone trying to understand the
structure of the document, "courses passed" etc.) is orders of
magnitude more complicated than extracting the URLs from link
annotations.  So I'm not proposing to do that. :)

You are absolutely right about the validation error, I was being lazy
and I'll fix that.

elfeto:
>Thanks for your comments.  Just to be clear:
>- If the person uploaded a real CITI completion report, then there
would be a link to view that document, in a separate window/tab, just
as there is currently.
>- If the person did not upload a real CITI completion report then
there would only the link to download it.

I apologize, but I don't seem to understand what you mean....

On this page `http://localhost:8888/console/complete-credential-applications/` the training should always be opened in a new tab.

On these pages:
- `http://localhost:8888/console/credential-applications/`
- `http://localhost:8888/console/credential-applications/qdOWz3Bn0k8FkqtcLpdu/process/`
I really don't mind that much the action on the credential report. ( I would say download because my browser is not ideal, but then I will have a lot of files in my downloads.... So anything would be ok)

>We could change the form to require that you *must* provide a CITI
completion report, all other PDF files are rejected.  I was under the
impression that in some (unusual) cases, people could be approved
based on other equivalent training courses.  But if we simply say that
you *must* complete a course from CITI and you *must* upload the PDF
completion report that they sent you, it would make things much
simpler.

From my understanding, that has happened but it would be unusual. But, they \*must\* upload the completion report they provide.

>(Though then we'd get emails from people saying "I uploaded [a PDF
file that I printed from an MS Word document containing screenshots
of] the CITI training report, why isn't it working?")

If we auto-reject because the training report is inadequate, we can specify and provide an example in the email and how to get the proper file.

If they want to work with a credentialed databases, then they should at least provide the required training correctly.

>If we do want to continue letting people upload whatever junk PDF
files they like, then yes, it'd be preferable to be able to view them
with one click rather than download-and-then-open.  I do think that
there should be a visible distinction between "you're viewing a PDF
file that the applicant uploaded" and "you're viewing a verified
training report from citiprogram.org".  Maybe using an iframe?

Since you mention about the parsing and orders of magnitude, we can replace the PDF upload to a URL input that \*must\* start with `https://citiprogram.org/verify/`, then open that in an iframe in another window.
We can get that link from the PDF itself too but it's more complicated.
We could validate that checking response code 200 and that the URL starts with the string above.

>Parsing the text of the PDF file (let alone trying to understand the
structure of the document, "courses passed" etc.) is orders of
magnitude more complicated than extracting the URLs from link
annotations.  So I'm not proposing to do that. :)

To parse the PDFs we don't have to do it by hand, we could use a library like:
 - PDFMiner (which was updated last November) (I haven't read about it)

bemoody:
> I apologize, but I don't seem to understand what you mean....
>
> On this page
> `http://localhost:8888/console/complete-credential-applications/` the
> training should always be opened in a new tab.

With the changes I've suggested, there would be two possibilities.

If the PDF file is a CITI training report, then
complete-credential-applications would display:

    Training Completion Report: CITI (download)

Clicking on "Training Completion Report: CITI" would display the
training report in a new tab.  Clicking on "download" would download
the file.

If the PDF file is *not* a CITI training report, then
complete-credential-applications would display:

    Training Completion Report: Unknown (download)

Clicking on "download" would download the file.

So what I meant to say is: if the applicant did the correct and
sensible thing, the administrator workflow would be no different than
it is at present.  When you click on the link, it opens the file in a
new tab.

The second case, however, needs improvement.

> If we auto-reject because the training report is inadequate, we can specify
> and provide an example in the email and how to get the proper file.
>
> If they want to work with a credentialed databases, then they should at
> least provide the required training correctly.

Indeed... but people applying for access may not be fluent in English.
And some of them are clinicians rather than programmers, and while
they undoubtedly know more than I ever will about physiology, the
difference between a DOC file and a PDF file is completely lost on
them. :)

> Since you mention about the parsing and orders of magnitude, we can replace
> the PDF upload to a URL input that \*must\* start with
> `https://citiprogram.org/verify/`, then open that in an iframe in another
> window.
> We can get that link from the PDF itself too but it's more complicated.

Not a *lot* more complicated, but you're right.

I don't know if it's better to ask for the PDF file, or the
verification URL.  I suspect the former is less confusing, but I don't
know.

kpierceHST:
Hi,

Just a couple of thoughts about all this. The two of you seem to be doing well at sorting through details.

- I try to be accommodating.  If someone sends a png, or pdf that's clearly cobbled together from a screen shot and a doc file, I'll take it as long as I can read report.
- I don't want a verification url.  (In general, I don't want to click on a link that doesn't come from PhysioNet.)  I want to load the completion report in a separate browser window, scan it to see whether it's acceptable, and close the window.
- I do sometime make judgement calls about training dates.
- The completion report sometimes has two pages, a green one showing initial training, and a blue one showing latest dates of completion for each module.

All for what it's worth.

Ken

kpierceHST:
Hi,

One further comment about this.  Felipe (I think...) writes

> If they want to work with a credentialed databases, then they should at least provide the required training correctly.

Which is true I suppose, but it's also true that the CITI site doesn't help matters at all, by offering pdfs of both "completion reports" and "completion certificates", the latter being what you'd typicially want to present to someone or hang on your wall.  Some of the researchers in the lcp have submitted certificates on the first pass; it's a common and understandable error.

Ken

tompollard:
@bemoody this is useful work that I think will improve efficiency (e.g. avoiding submissions that are clearly inadequate) so I'd like to pick it back up and merge soonish if possible. Do you know what needs to be done to finish things up? 

If Ken is not keen on some of the features (e.g. including a verification link) then we can exclude them from his management page (https://physionet.org/console/complete-credential-applications/), while still displaying them on the general credentialing pages.

tompollard:
@bemoody these changes look great to me. I have reverted two of the commits for now because they would affect two pages that are being actively used for credentialing. I would like to merge at this point, unless you have any objections. This will avoid having to refactor when @Lucas-Mc addresses https://github.com/MIT-LCP/physionet-build/issues/1179

tompollard:
(migrations need updating)

bemoody:
Rebased for latest migrations (632c7607 -> 020986c5).

Please test and merge if you're happy.  The data migration shouldn't
cause any problems, but it might take a while to run when you push it to
production.  It'd be prudent to test the migration on staging (after
rsyncing the credential-applications) before doing the same on
production.  Let me know if you'd like me to help with any of that. :)

bemoody:
> It'd be prudent to test the migration on staging (after rsyncing the
> credential-applications) before doing the same on production.  Let me
> know if you'd like me to help with any of that. :)

I've just realized this is more complicated than it sounds because the
database on staging is based on a very old clone of production.  Ideally
we would first transfer over a recent database dump in order to include
recent credential applications.  But again, it *should* be safe either
way...

This is why I'd like to have the DBs automatically synchronized.

tompollard:
thanks! @elfeto is looking at doing a sync of the databases now.

tompollard:
@elfeto synced the staging database with the live site (thanks!). 

The change was pushed to the live server.

PR #1027, opened 2020-04-22
Add credentialing files for user fixture

elfeto:
On #1004 X amount of credentialing applications were introduced
to facilitate the testings of credential applications.

The applications were all made in the console ignoring the file
upload. Here I also add to the demo information a soft link to the
file. This is the same file we used since the beginning of time.

I also fix the fixture with the file location.

A way to test this is by going to:
http://localhost:8888/credential-applications/tKoY8VSB3Sq1kxW19gcc/training-report/
It will fail on dev with a 500, and work on this branch.

A resetdb and loaddemo has to be done after swapping branches.

Just as an FYI, the file we used is a text file saying "report", so the page will be empty.


elfeto:
Ping ont his ?

tompollard:
Thanks Felipe, looks good to me. For later, should we replace the file with a valid PDF?

PR #1030, opened 2020-04-23
Define user views in user.views

bemoody:
Instead of defining view functions in user.urls (and customizing them using keyword arguments to as_view), define them in user.views (and customize them by subclassing.)

This makes it more clear how to extend them, and separates the URL patterns from the functions that handle them.


bemoody:
These are just style changes that would be helpful for implementing site maintenance mode, and maybe other things in the future.

elfeto:
Looks good!

Issue #674, opened 2019-08-08
Maintenance mode


bemoody:
We need to be able to switch seamlessly from one web server to another, or to temporarily switch to a standby server for mainentance.

* Case 1: planned migration from one master server to another.  Everything must eventually be transferred, and public resources must be continuously available.

* Case 2: unplanned failure of master server, requiring us to send clients to a backup server.  Public resources must remain available but might be out-of-date.

If a user uploads or deletes a file through the web interface, it must not mysteriously disappear/reappear later.  Likewise if they create projects, edit descriptions, etc.

With vast resources and a lot of effort we could automatically replicate everything in multiple places and keep services continuously available.

Next best thing, I think, is to be able to stop the world:
* freeze the system on server A
* copy everything from server A to server B
* move clients over to server B
* unfreeze server B

"Freezing" means that nobody is allowed to make any changes to the filesystem or to the relational DB.  In practice, I think the easiest way to do this is to *treat all users as logged-out* (since logged-out users should never be able to make any changes anyway), and hide the 'login'/'logout'/'register' pages

(A side effect being that restricted projects would be completely unavailable.)

Questions to resolve:

1. How can we get Django to treat all users as logged-out?

2. What is the fastest way to move the relational DB from one server to another?  (Assuming that the system is frozen so we don't care about propagating changes; but we do need to have a complete and internally consistent copy, and the web server continues to access the database in the meantime.)

bemoody:
Trying to work through this, I'm realizing that we actually should support a few different maintenance modes:

- "no-upload" mode: Everything works normally except that active projects can't be modified.

- "no-changes" mode: Visitors can log in and out (and download protected databases) but changes can't be saved (e.g. you can't create/edit projects.)

- "no-auth" mode: No ability to log in.  Protected databases are inaccessible.


From my understanding, it should be possible to make a graceful transition from one postgres server to another in a matter of seconds, which means a well executed transition should only require us to sit in "no-changes" mode for a few seconds.  But transferring active project files could take minutes or hours, during which time we want to be in "no-upload" mode to prevent loss of data.


I'm now looking at how to implement this using middleware.  It seems like it should be quite feasible.


I still need to understand how exactly this "log shipping" stuff is supposed to work.

elfeto:
> How can we get Django to treat all users as logged-out?

We can force log them out.

>"no-changes" mode: Visitors can log in and out (and download protected databases) but changes can't be saved (e.g. you can't create/edit projects.)

[...]

PR #1033, opened 2020-04-27
Support maintenance mode

bemoody:
This adds several new options that can be set in .env.

SYSTEM_MAINTENANCE_NO_CHANGES means that generally, nothing is allowed to make persistent changes.  For example, no one is allowed to register an account, change their password, edit project descriptions, or upload files.  However, this doesn't prevent people from logging in to view protected/unpublished projects.

SYSTEM_MAINTENANCE_NO_UPLOAD means that project files cannot be manipulated.  For example, new projects can't be created, files can't be uploaded, and active projects can't be published.  However, project descriptions can be edited and most other site functions work normally.

SYSTEM_MAINTENANCE_MESSAGE optionally defines the error message that is displayed.

For further background and discussion, see issue #674.

The "no-changes" mode is enforced by middleware.  GET requests should not modify server-side state, so are generally allowed.  POST requests can modify server-side state, so are generally disallowed.  There are some exceptions:

- /login/ uses POST, and it saves session information on the server side, but keeping login functional is more important than retaining the information, so this is allowed in "no-changes" mode.

- /reset/X/Y/ and /activate/X/Y/ rely on proper operation of sessions (and of course submitting the form makes changes to the user tables), so these are disallowed in "no-changes" mode.

- /verify/X/Y/ makes changes even on a GET request, so this is disallowed in "no-changes" mode.

- /about/ (yeah, that's the contact form) uses POST, but doesn't save information to the database.  It merely sends a message to the mail server, which doesn't care about any of this.

"No-upload" mode is enforced by individual views:

- /projects/X/files/: in no-upload mode, controls should be disabled and uploads should be refused.

- /projects/create/: all access denied.

- /projects/new-version/X/: all access denied.

- /console/submitted-projects/X/copyedit/: should allow editing description but not files.

- /console/submitted-projects/X/publish/: all access denied.

- /console/published-projects/X/Y/: should allow managing DOIs, uploading to GCP, but not changing files.

As you can see, there are a number of corner cases to consider, but I think this is better than a blanket approach that would make protected projects inaccessible for hours or days.  Please try to look for any other corner cases I've missed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,6 +3,11 @@
 SECRET_KEY=secret
 ALLOWED_HOSTS=[]
 
+# System maintenance mode
+#SYSTEM_MAINTENANCE_NO_CHANGES=1
+#SYSTEM_MAINTENANCE_NO_UPLOAD=1
+#SYSTEM_MAINTENANCE_MESSAGE=PhysioNet is undergoing maintenance, and projects cannot be edited. The site will be back online at 16:00 GMT.
+
 # GCP
 # USED to store ALL the published projects to GCP Buckets and BigQuery
 # The delegation email, might be possible to change in the:
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -22,6 +22,7 @@ from django.contrib.sites.models import Site
 
 from notification.models import News
 import notification.utility as notification
+from physionet.middleware.maintenance import ServiceUnavailable
 from physionet.utility import paginate
 import project.forms as project_forms
 from project.models import (ActiveProject, ArchivedProject, StorageRequest,
@@ -501,6 +502,8 @@ def publish_submission(request, project_slug, *args, **kwargs):
 
     if project.submission_status != 60:
         return redirect('editor_home')
+    if settings.SYSTEM_MAINTENANCE_NO_UPLOAD:
+        raise ServiceUnavailable()
 
     reassign_editor_form = forms.ReassignEditorForm(request.user)
     authors, author_emails, storage_info, edit_logs, copyedit_logs, latest_version = project.info_card()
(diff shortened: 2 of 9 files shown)


elfeto:
Since `SYSTEM_MAINTENANCE_NO_CHANGES` is server-wide used by the middleware, I don't think you missed any exceptions there. 

As for `SYSTEM_MAINTENANCE_NO_UPLOAD`:
 - One can still alter the profile picture.
 - One can still request credentialing (uploading the certificate)

Apart from that, I don't see any other thing you missed or anything wrong with the PR.

As a side note, since this is managed by the `.env` file, uwsgi has to be restarted or touched after the changes are done to see the maintenance.

tompollard:
Looks great to me, and I'm happy to merge once Felipe's comments above are dealt with. It  could be frustrating to discover that the site is in maintenance after completing a credentialing application or adding project content, so it would be good if setting a site-wide maintenance banner was also displayed. This can be dealt with later, in a separate pull request.

bemoody:
Good points all.

> As a side note, since this is managed by the .env file, uwsgi
> has to be restarted or touched after the changes are done to
> see the maintenance.

Yeah, I don't see any way to avoid that that wouldn't be horribly
flaky or add a performance burden to every request.

> It could be frustrating to discover that the site is in
> maintenance after completing a credentialing application or
> adding project content, so it would be good if setting a
> site-wide maintenance banner was also displayed

Yes, though making dynamic site-wide changes to templates is
Interesting as well :)

tompollard:
@bemoody We should merge this and deal with the minor issues mentioned above later. Please could you add a comment to the configuration file to note that `SYSTEM_MAINTENANCE_NO_UPLOAD` doesn't disable uploading of profile pictures or credentialing certificates?

bemoody:
I can do better than that, and handle those two cases :)

We can also consider adding a banner, but try this out and see what
you think.  For example, in "no-upload" mode it will show an error
when you click the "Create New Project" button, or the "apply for
access" link in credentialing.  You should be able to
browse existing project files, but the upload button should be hidden.

tompollard:
Looks good, thanks!

Issue #1028, opened 2020-04-23
mailto function breaks with special characters and comma


elfeto:
If the reference name of a credential application contains both a special character and a comma, the mailto function for external mail clients breaks the email.

Works fine:
 - a.reference_name = 'Saldaña Ramon'
 - a.reference_name = 'Ramon Saldana PHD'

Creates an empty 'To' mail at each comma:
 - a.reference_name = 'Ramon Saldaña, PHD'
 - a.reference_name = 'Saldaña Ramon, PHD'
 - a.reference_name = 'Saldaña, Ramon, PHD'

 
<img width="656" alt="Screen Shot 2020-04-23 at 9 18 18 AM" src="https://user-images.githubusercontent.com/3857286/80109793-8e01b100-8543-11ea-87dc-79667718c945.png">
<img width="652" alt="Screen Shot 2020-04-23 at 9 18 49 AM" src="https://user-images.githubusercontent.com/3857286/80109802-90640b00-8543-11ea-81e6-ac11e9ac71af.png">

elfeto:
At the moment this only happens using Thunderbird, but not the OSX mail client.

bemoody:
For clarity, could you give an example of the mailto link that it generates?

When you say "both a special character and a comma", do you mean that
it works correctly if the name contains a comma but no special
characters?  Your example of 'Ramon Saldana PHD' doesn't have a comma
in it.

elfeto:
```
a.reference_name = 'Saldaña Ramon'
mailto:%3D%3Futf-8%3Fq%3FSalda%3DC3%3DB1a_Ramon%3F%3D%20%3Cadmin2@mit.edu%3E?

a.reference_name = 'Ramon, Saldana, PHD'
mailto:%22Ramon%2C%20Saldana%2C%20PHD%22%20%3Cadmin2@mit.edu%3E?
```
```
a.reference_name = 'Ramon Saldaña, PHD'
mailto:%3D%3Futf-8%3Fq%3FRamon_Salda%3DC3%3DB1a%3D2C_PHD%3F%3D%20%3Cadmin2@mit.edu%3E?

a.reference_name = 'Saldaña Ramon, PHD'
mailto:%3D%3Futf-8%3Fq%3FSalda%3DC3%3DB1a_Ramon%3D2C_PHD%3F%3D%20%3Cadmin2@mit.edu%3E?

a.reference_name = 'Saldaña, Ramon, PHD'
[...]

PR #1035, opened 2020-04-30
Fixing issue in mailto function when the name contains a comma and special char

elfeto:
When a credentialed application adds the reference with a special character
ans a comma,  the mailto function breaks the 'TO' at the comma and adds
an empty email in the mailto.

Here I just remove all commas if any on the mailto 'TO' section.

closing #1028


tompollard:
looks good, thanks!

PR #1036, opened 2020-04-30
Fixing typo on function call

### LINKED ISSUE #1034: typo in method call (`submitting_author.set_display_infprevious_versiono()`)
Not seen anything bad happening as a result, but something strange is going on here:

https://github.com/MIT-LCP/physionet-build/blob/df26605229364590870f230bf7301e82242bf004/physionet-django/project/models.py#L590

elfeto:
Up to now, there is no case where separate_submitting is true. 

I removed the "previous_version" because there is no mention of version in that statement and kept "set_display_info" just as the else statement. 

closes #1034


Lucas-Mc:
Looks good to me!

Issue #1031, opened 2020-04-24
Incorrect link to plain language summary in open data license page


alistairewj:
The MIMIC-III demo uses the Open Data Commons Open Database License v1.0

https://physionet.org/content/mimiciii-demo/view-license/1.4/

The plain language summary link is incorrect:

* Incorrect link (current): http://www.opendatacommons.org/licenses/odbl/1-0/summary/
* Correct link: https://www.opendatacommons.org/licenses/odbl/summary/

tompollard:
thanks for picking this up. @Lucas-Mc please could you take a look at fixing this?

the whole license, dua thing needs sorting out properly...

PR #1037, opened 2020-04-30
Fixes ODBL link #1031

Lucas-Mc:
Updates the link used for the Open Data Commons Open Database License (ODBL) v1.0.


tompollard:
Thanks Lucas, looks good to me (slightly annoying that the diff displays the entire chunk of text, but I'm not sure there's anything we can do about that!). The fixture will need to be loaded to the database once pushed to the server, so I'll leave this open until someone is ready to deal with this at the same time.

tompollard:
Thanks @Lucas-Mc, this is now fixed on the database too, with:

```
from project.models import License

x = License.objects.get(name="Open Data Commons Open Database License v1.0")

x.html_content = x.html_content.replace("http://www.opendatacommons.org/licenses/odbl/1-0/summary/","https://www.opendatacommons.org/licenses/odbl/summary/")
```

Closes #1031

PR #1039, opened 2020-05-04
Adding associated emails to account purge

### LINKED ISSUE #1038: Purge unactivated associated emails
On continuation from #795 that deletes accounts that have not been activated, the same should be done for emails.

We should delete all emails that are associated to an active account, but the email has not been activated in 7 days.

Should an email notification be sent to the account primary email saying that the email has been removed since it has not been activated?

elfeto:
There is no purge on the associated emails added but not activated.
Here I add the associated emails to the purge.

If an active account has an associated email that hasn't been
activated, it will be deleted after 7 days.

Fixes #1038

### DIFF
--- a/physionet-django/user/test_views.py
+++ b/physionet-django/user/test_views.py
@@ -12,6 +12,7 @@ from django.core import mail
 from django.core.management import call_command
 from django.test import RequestFactory, TestCase
 from django.urls import reverse
+from django.utils import timezone
 
 from user.models import AssociatedEmail, Profile, User
 from user.views import (activate_user, edit_emails, edit_profile,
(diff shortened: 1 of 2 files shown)


bemoody:
Code looks good to me (not tested, but I see you added test cases - thanks!!)

As @tompollard suggested in issue #1038, it'd be good to mention in
the email message that the link will expire in 7 days.  However, we
should probably also do the same for register and maybe
password-reset, so maybe we should make that a separate pull request.

elfeto:
I just added the email message in #1046

physionet-django/user/test_views.py
```diff
@@ -266,6 +267,34 @@ def test_edit_emails(self):
...
+            'add_email': [''], 'email': 'tester2@mit.edu'})
+        self.assertIsNotNone(AssociatedEmail.objects.filter(email='tester2@mit.edu'))
+
+        # Test 2: add email to be removed
+        self.client.post(reverse('edit_emails'), data={
+            'add_email': [''], 'email': 'tester1@mit.edu'})
+        self.assertIsNotNone(AssociatedEmail.objects.filter(email='tester1@mit.edu'))
+
+        email1 = AssociatedEmail.objects.get(email='tester1@mit.edu')
+        email1.added_date -= timezone.timedelta(days=30)
```
tompollard, line 286: is the `-=` intentional?

elfeto: Either that or `email1.added_date = timezone.timedelta(days=-30)`  I had to set the added date 30 days into the past so that it would be removed.

tompollard: got it, thanks

tompollard:
thanks!

PR #1041, opened 2020-05-08
Prevents credential change of names

Lucas-Mc:
Prevents users from changing their name on new credential applications to prevent future confusion. This was a requested change of @kpierceHST.

### DIFF
--- a/physionet-django/user/forms.py
+++ b/physionet-django/user/forms.py
@@ -247,8 +247,10 @@ class PersonalCAF(forms.ModelForm):
             'organization_name', 'job_title', 'city', 'state_province',
             'zip_code', 'country', 'webpage')
         help_texts = {
-            'first_names': "First name(s).",
-            'last_name': "Last (family) name.",
+            'first_names': """First name(s). Your Physionet first name by default. This 
+                can be edited in your profile settings.""",
+            'last_name': """Last (family) name. Your Physionet last name by default. This 
+                can be edited in your profile settings.""",
             'suffix': """Please leave the suffix blank if your name does not 
                 include a suffix like "Jr." or "III". Do not list degrees. 
                 Do not put a prefix like "Mr" or "Ms". Do not put "not 
@@ -272,6 +274,8 @@ class PersonalCAF(forms.ModelForm):
         widgets = {
            'research_summary': forms.Textarea(attrs={'rows': 3}),
            'suffix': forms.TextInput(attrs={'autocomplete': 'off'}),
+           'first_names': forms.TextInput(attrs={'readonly':'readonly'}),
+           'last_name': forms.TextInput(attrs={'readonly':'readonly'}),
         }
         labels = {
             'state_province': 'State/Province',
--- a/physionet-django/user/forms.py
+++ b/physionet-django/user/forms.py
@@ -247,10 +247,10 @@ class PersonalCAF(forms.ModelForm):
             'organization_name', 'job_title', 'city', 'state_province',
             'zip_code', 'country', 'webpage')
         help_texts = {
-            'first_names': """First name(s). Your profile first name by default. This 
-                can be edited in your profile settings.""",
-            'last_name': """Last (family) name. Your profile last name by default. This 
-                can be edited in your profile settings.""",
+            'first_names': """You first name(s). This can be edited in your 
+                profile settings.""",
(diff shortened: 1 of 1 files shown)


bemoody:
Not sure what's going on here.  Please write comments and commit
messages to explain why you're doing what you're doing.

What do you mean by "new credential applications" and what do you mean
by "prevent users from changing their name"?

Setting the "readonly" attribute is usually not what you want, and
generally setting HTML attributes in a form class is a last resort.

Lucas-Mc:
Sorry, should have explained more for those who aren't on the e-mail chain. @kpierceHST was getting confused since the user name on [this](https://physionet.org/console/past-credential-applications/unsuccessful) link doesn't match what is on the "view" link on that same page for the submitted credential application. Therefore, to prevent this mismatch, Ken wanted us to remove the option for users to edit their name in the credential application and make it different than their profile name.

tompollard:
Just to add a bit more detail:

- Users can currently add their first name and last name at: http://localhost:8000/credential-application/
- Sometimes the names added to the form do not match the profile name that is added at: http://localhost:8000/settings/profile/, which is confusing.

What might be sensible is completely removing the first and last name fields from the credential application form altogether. One issue with doing this is that the suffix field ("Suffix, if applicable:") does not currently appear in the user profile.

physionet-django/user/forms.py
```diff
@@ -247,8 +247,10 @@ class Meta:
             'organization_name', 'job_title', 'city', 'state_province',
             'zip_code', 'country', 'webpage')
         help_texts = {
-            'first_names': "First name(s).",
-            'last_name': "Last (family) name.",
+            'first_names': """First name(s). Your profile first name by default. This
```
tompollard: Perhaps just "Your first name(s). This can be edited in your profile settings." would be clearer?

bemoody:
> Sorry, should have explained more for those who aren't on the e-mail chain.
> @kpierceHST was getting confused since the user name on
> [this](https://physionet.org/console/past-credential-applications/unsuccessful)
> link doesn't match what is on the "view" link on that same page for the
> submitted credential application. Therefore, to prevent this mismatch, Ken
> wanted us to remove the option for users to edit their name in the
> credential application and make it different than their profile name.

Mm.  Changing behavior of the application form doesn't prevent
mismatches, though.  If you say "the application must show the same
name that was in the person's profile when they submitted the
application", then they can still change their profile after
submitting the application, and then you have a mismatch.

If you say "the application must always show the name that is
currently in the person's profile", then the name on the application
today might not be the same as the name on the application tomorrow,
which is bad for other reasons.

physionet-django/user/forms.py
Lucas-Mc: I can change it, seems more concise too

Lucas-Mc:
Good point @bemoody, maybe just change which name is displayed on the previous applications webpage instead? Also remove the name field in the application all together?

tompollard:
Yeah, it's tough. I don't think there's any perfect solution. The benefit of the credentialing process is that it does go some way to ensuring that the name on record is a real name (so it's nice that with this PR the name ends up being added to the profile, not just the credential form).

It's true that someone might change their name again in the future (and there are legitimate reasons that they might want to do this). We can always consider logging changes so at least we have a record of names that are used.

kpierceHST:
If a person changes name, the person should submit a new data use agreement.  We have no way of verifying that a changed name applies to the same individual as before.  Much of the verification that goes on in credentialing is based on a person's name (google search, etc.).

Name should be a fixed attribute of a user's profile, and name in profile should be the same as name in data use agreement.

(The issue that started this whole discussion had to do with a name difference -- X, X and X are all the same person -- and it took me a couple of rounds of e-mail messages to sort things out.)

For what it's worth....

Ken

tompollard:
Thanks Ken and Benjamin, good points. There's clearly more to be done to deal with the issue of someone submitting a credentialing application and then subsequently changing their name in the profile. We will attempt to address this in a separate set of changes. 

@Lucas-Mc for this pull request I think we just need to fix the "readonly" attribute that we've discussed, and then we can merge.

tompollard:
Looks good to me, thanks Lucas. There are some points raised in this discussion that we should try to address later (i.e. what happens if someone changes their name after submitting a credentialing application).

PR #1043, opened 2020-05-11
Fix name and lastname disabled in cred apps

elfeto:
In #1041 disabled was introduced to name and last name for the credentialed applications. Since both fields are disabled they are not in the request.POST variables which makes this form to fail
with the following error:
 - {'first_names': ['This field is required.'], 'last_name': ['This field is required.']}

To be able to fix this, one has to disable the field in the form init so that the field isn't "cleaned" with an empty value and then initialize the field with the current name and last name.


tompollard:
nice, thanks!

PR #1044, opened 2020-05-11
Add "PLATFORM_NAME" variable to settings

tompollard:
This is a very minor change that adds a `PLATFORM_NAME` variable to the base settings and sets it to "PhysioNet". The platform name can then be displayed with 

```
settings.PLATFORM_NAME
```

I'm surprised that it didn't exist already, but I don't see it anywhere. The benefits are:

- makes the platform generablisable, by helping to reduce hardcoded "PhysioNet"s.
- helps to avoid misspellings (physionet, Physionet, etc).

Should we take this further and display different values for the different servers (e.g. PhysioNet-Dev; PhysioNet-Staging; PhysioNet)?

### DIFF
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -21,6 +21,8 @@ import logging.config
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
+# Customisable platform settings
+PLATFORM_NAME = 'PhysioNet'
 
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/1.11/howto/deployment/checklist/


bemoody:
It sounds reasonable, but I'm not entirely keen on "PLATFORM_NAME".
That sounds like it refers to, well, the platform (kernel? operating
system? python interpreter?)

I think "site" is the word you're looking for :)

> Should we take this further and display different values for the different
> servers (e.g. PhysioNet-Dev; PhysioNet-Staging; PhysioNet)?

Probably a good idea.

A possibly related thing is that I've often wished the
standby/development servers would display something recognizably
different in the navigation bar.  Maybe not the best idea to change
the text there, since the CSS is so flaky to begin with (though of
course that could be fixed :)) - another idea I had was just to change
the background color.

tompollard:
> A possibly related thing is that I've often wished the standby/development servers would display something recognizably different in the navigation bar.

Yeah, I agree this would be good. I think it would be worth replacing the site name in top left with PhysioNet-Dev or whatever (for me, the css issue isn't too much of a worry).

Felipe has pointed out that we can already grab the site name from the request object, with:

```
>>> from django.contrib.sites.shortcuts import get_current_site
>>> from django.http import HttpRequest
>>> request = HttpRequest()
>>> get_current_site(request)
<Site: localhost:8000>
>>> get_current_site(request).name
'PhysioNet-dev'
>>> 
```

So is there any reason not to close this pull request? The site names are in https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/fixtures/sites.json

bemoody:
Yeah, that works too.

tompollard:
I'll close this for now

PR #1046, opened 2020-05-11
Add 7 days notice in outgoing emails

elfeto:
Added a 7 days notice on the links that we send out to:
 - Register for new accounts
 - Reset passwords
 - Add new emails

I also updated a html template of a unused function.
The unverified_email_removal.html was part of a 'cron' app that
was removed X time ago. I am guessing this could/should be removed
but just in case I will leave that for another PR.


tompollard:
thanks felipe, looks good

PR #1048, opened 2020-05-13
Creates list of all published projects

Lucas-Mc:
Creates a page similar to [this](https://archive.physionet.org/physiobank/database/DBS) one on the old site. This will be used in WFDB-Python in the `get_dbs()` function to extract a list of all the currently available projects that are able to be analyzed. View original documentation here: https://github.com/MIT-LCP/wfdb-python/blob/f4ada092134ff062d836318042f373ab34abcaa3/wfdb/io/download.py#L209-L230

### DIFF
--- a/physionet-django/console/templates/console/all_projects.html
+++ b/physionet-django/console/templates/console/all_projects.html
@@ -1,3 +1,0 @@
-{% for project in all_projects.items %}
-  <p>{{ project.0 }}&emsp;{{ project.1 }}</p>
-{% endfor %}
--- a/physionet-django/console/templates/console/all_projects.html
+++ b/physionet-django/console/templates/console/all_projects.html
@@ -1,0 +1,3 @@
+{% for project in all_projects.items %}
+  <p>{{ project.0 }}&emsp;{{ project.1 }}</p>
+{% endfor %}
--- a/physionet-django/console/templates/console/console_navbar.html
+++ b/physionet-django/console/templates/console/console_navbar.html
@@ -174,7 +174,7 @@
             <a href="#">Not Implemented</a>
           </li>
           <li>
-            <a href="#">Not Implemented</a>
+            <a href="{% url 'all_projects' %}">All Projects</a>
           </li>
         </ul>
       </li>
--- a/physionet-django/console/templates/console/console_navbar.html
+++ b/physionet-django/console/templates/console/console_navbar.html
@@ -174,7 +174,7 @@
             <a href="#">Not Implemented</a>
           </li>
           <li>
-            <a href="{% url 'all_projects' %}">All Projects</a>
+            <a href="{% url 'all_projects' %}">Database List</a>
           </li>
         </ul>
       </li>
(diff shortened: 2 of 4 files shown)


bemoody:
Mmm.  If you're looking for a list of databases then I think what you
want is '/lightwave/server?action=dblist'.  Though it'd be nice to
have the traditional easy-to-parse DBS file too...

A pretty HTML list of projects is another matter...

Lucas-Mc:
Ahhh interesting, did not know about that page.. can definitely parse it but let's see what the consensus is on this one first.

Lucas-Mc:
Update: the Lightwave page can just be imported easily using `json.loads()`.. should this PR be continued but modified so that the display of projects looks nice, or is this functionality not needed right now?

tompollard:
I like the behavior of `'/lightwave/server?action=dblist`, but I think it would be better if the path didn't depend on lightwave. 

Could we start building out a more general API (e.g. `api.physionet.org` or `physionet.org/api`) to support requests like "give me all of the project slugs"?

Lucas-Mc:
Yeah @tompollard, that is what I was concerned about also.. what if Lightwave is removed after the annotation project is complete? Then the method to get all the projects would have to change as well

bemoody:
Yeah... we really should fix all this stuff

/physiobank/database/mitdb/1.0.0/100.hea should work for versioned files

/physiobank/database/DBS should give a tab separated list of
(versioned) databases

we could add a DBS.json if that would make people happy :)

tompollard:
> we could add a DBS.json if that would make people happy :)

I'd really like a REST API, but don't have any issues with a json file somewhere! Might be good to have a more informative name, e.g. `database_list.json` or something?

elfeto:
BTW, we have somewhat of a rest API (I know this might be out of the scope of the PR) but the app export has some information that you might want to look.

tompollard:
Just had a play around with this. It seems like we do need some way of offering up a list of databases (and other content types). A couple of quick thoughts:

- Is "http://physionet.org/console/usage/database_list.json" the right URL for serving the file?  The URL should be reasonably persistent if applications are going to be referencing the file. Up until now, "console" has not been used for public content.
- It feels slightly odd to have this under "Usage stats" in the menu, particularly when clicking the item drops the user into a place without any menus for navigation.

tompollard:
@elfeto good point, I'd managed to forget that https://physionet.org/rest/database-list/ existed! Still not a great URL, so maybe we should review that, but this seems to be roughly what you need Lucas?

Lucas-Mc:
@tompollard Yeah I think so, that's all published projects? I can just use that instead? I actually accidentally ran into this cite when I tried to set the URL to `database_list` for some reason though it didn't show all the results.

tompollard:
If a URL exists that contains the information that you need, then it makes sense to use it rather than to create something new.

> I tried to set the URL to database_list for some reason though it didn't show all the results.

It's very possible that the API output is buggy because it hasn't had much use yet, so some testing before making it a WFDB dependency would be good!

Lucas-Mc:
It looks like [the current list](https://physionet.org/rest/database-list/) will work nicely for my purposes though we can re-open this issue at another point if we feel like it isn't sufficient.

PR #1058, opened 2020-05-21
Show Anonymous access on non-open projects

elfeto:
At the moment the "Anonymous access" section is shown even on open projects under the manage published projects page.

Here I hide from the console the section of Anonymous access to projects that are open to the world.

### DIFF
--- a/physionet-django/console/templates/console/manage_published_project.html
+++ b/physionet-django/console/templates/console/manage_published_project.html
@@ -60,6 +60,7 @@
     </div>
 </div>
 
+{% if project.access_policy %}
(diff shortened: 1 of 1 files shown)


tompollard:
I have mixed feelings about this change. I kind of like consistency in the layout of pages like this, because makes it easier to get familiar where things are (even if some sections aren't relevant).

e.g. if I'm used to DOIs being below "anonymous access", then that can be helpful. If there were multiple rules for revealing/hiding content, you'd have to learn to navigate the page each time.

Saying this, it's only one chunk and I don't see any real reason not to merge, so...

PR #1059, opened 2020-05-21
Add a way to add new authors to legacy projects

### LINKED ISSUE #1057: Legacy authors
There should be a simple way to allow authors claim published legacy projects or to assign a legacy project to X authors.

I would suggest the workflow be that they send us an email and we assign them via an easy method under the manage public project page.

That way we can avoid issues like the one from https://physionet.org/content/earndb/1.0.0/

elfeto:
IF a published project is legacy, then another section will appear to add and only add new authors.

The order of the authors will be sequential as added.
The first author added will be the submitting author.

At the moment I left the corresponding to False as default.

To add the user object, I took a similar approach to the parent project.
closes #1057

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -7,11 +7,12 @@ from django.core.validators import validate_integer, validate_email, URLValidato
 from google.cloud import storage
 from django.db import transaction
 from django.conf import settings
+from dal import autocomplete
 
 from notification.models import News
 from project.models import (ActiveProject, EditLog, CopyeditLog, Contact,
-    PublishedProject, exists_project_slug, DataAccess)
-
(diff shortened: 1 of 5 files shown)


tompollard:
Looks good to me! I've rebased on dev because there were a couple of conflicts, and did some minor tidying.

I think it's important to limit the ability to edit authorship to the legacy projects only as you've done here. If/when the authorship for these projects is finalized, the ability to edit authors should be removed.

The autocomplete is a nice touch! It seems to be rendered in a slightly strange way - the input box jumps up from where I would expect it to be (see screenshot below) - but this isn't a major issue and can be dealt with later.

Screen Shot 2020-05-30 at 23 43 07

PR #1060, opened 2020-05-21
Adds a way to alter the contact information.

### LINKED ISSUE #1054: Alter the contact info of a published project
At the moment, there is no simple way to alter the contact person/info of a published project. Either the person is no longer available at X email, the contact changes or its handled by a group like the challenges. 

There should be a simple way (non shell) to alter that information.

The process would be, the submitting author sends us an email with that information request, and we make the changes.

elfeto:
Here I add a tab in the manage published project page to alter the contact information of the published project.

This is useful for challenges and legacy projects. 
closes #1054

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -9,8 +9,9 @@ from django.db import transaction
 from django.conf import settings
 
 from notification.models import News
-from project.models import (ActiveProject, EditLog, CopyeditLog,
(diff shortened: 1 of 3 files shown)


tompollard:
Looks good to me, thanks! This will be helpful for the legacy content. 

Side note, but the role of the corresponding author doesn't seem clear right now, and it would be good to try to address this at some point.

I'm not 100% clear of the relationship between "contact" and "corresponding_author", but it seems like they may be different words for basically the same thing?

PR #1061, opened 2020-05-28
Fixes references order #590

### LINKED ISSUE #590: References in unspecified order

Lucas-Mc:
Maintains the order of references originally submitted during the creation and editing of a new project. Fixes #590.

### DIFF
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -1701,7 +1701,7 @@ class ActiveProject(Metadata, UnpublishedProject, SubmissionInfo):
                     published_project.set_version_order()
 
                 # Same content, different objects.
-                for reference in self.references.all():
+                for reference in self.references.all().order_by('id'):
                     published_reference = PublishedReference.objects.create(
                         description=reference.description,
                         project=published_project)
--- a/physionet-django/project/views.py
+++ b/physionet-django/project/views.py
@@ -1103,7 +1103,7 @@ def project_preview(request, project_slug, subdir='', **kwargs):
     corresponding_author = authors.get(is_corresponding=True)
     corresponding_author.text_affiliations = ', '.join(a.name for a in corresponding_author.affiliations.all())
 
-    references = project.references.all()
+    references = project.references.all().order_by('id')
     publication = project.publications.all().first()
     topics = project.topics.all()
     parent_projects = project.parent_projects.all()
@@ -1530,7 +1530,7 @@ def published_project(request, project_slug, version, subdir=''):
     authors = project.authors.all().order_by('display_order')
     for a in authors:
         a.set_display_info()
-    references = project.references.all()
+    references = project.references.all().order_by('id')
     publication = project.publications.all().first()
     topics = project.topics.all()
     languages = project.programming_languages.all()


bemoody:
To be clear, this looks like it should work, so we should go ahead and
do this if it will make life easier for the time being, but I think
there are better ways to address the problem.

It would be worth checking that all references in published projects
are *currently* displayed in the way you expect.  I.e.: log in to the
production server, run "project.references.all()" for each published
project, and verify that the results are in increasing order of "id".
If that is not the case, then we need to investigate what the correct
order should be.

tompollard:
Just doing a bit of investigating...will post an update shortly!

tompollard:
The investigation was a bit more interesting than I think we would have liked, but not terrible.

```
from project.models import PublishedProject, ActiveProject
projects = PublishedProject.objects.all()
print(projects.count())
print(len([x for x in projects if x.references.count() > 1]))
```

- 214 published projects in total
- 31 published projects with 2 or more references

How many of these projects have references that aren't ordered by increasing ID?

```
# get the ID and title of every project whose references are not 
# listed in order of increasing ID
ouch = {p.id: p.title for p in projects 
        if list(p.references.all()) != list(p.references.all().order_by('id'))}
len(ouch)
```

- 30 projects are not ordered by increasing ID.

What are these projects?

```
for p in ouch: 
    print(p, ouch[p])
```

- 325 What's in a Note? Unpacking Predictive Value in Clinical Note Representations
- 326 Transformer models trained on MIMIC-III to generate synthetic patient notes
- 327 Clinical data from the MIMIC-II database for a case study on indwelling arterial catheters
- 296 MIMIC-CXR Database
- 222 MIMIC-III Clinical Database
...

Looking at these examples, the references seem to be in reverse order. e.g. https://physionet.org/content/mimiciii/1.4/ is clearly muddled, [1] in the text refers to ref 2 and [2] in the text refers to ref 1.

So running the same search as above, but looking at reversed ordering:

```
ouch2 = {p.id: p.title for p in projects 
        if list(p.references.all()) != list(p.references.all().order_by('-id'))}
len(ouch2)
```

- Only 1 project is not ordered by descending ID. This project is `{314: 'Cerebral Vasoregulation in Diabetes'}`. This project is ordered by ascending ID.

tompollard:
so thanks @Lucas-Mc, as far as i can see this fixes a major bug (that the reference section for most projects is displayed in totally the wrong order).

tompollard:
i'm testing on staging and it turns out that it isn't as simple as i'd thought. it seems that `list(p.references.all())` doesn't display the items in the order in which they are iterated, so the dict comprehensions above aren't doing what i thought they are.

i don't think we should revert this fix, but we may need to make some updates in the database to correct ordering for some specific projects. i'll spend a bit more time working this out.

tompollard:
The new ordering of references is now on the live server. I attempted to confirm that the references are being displayed correctly for published projects, fixing them where they were not. 

The following published projects should be broadly correct (though there may be minor issues that I didn't pick up):

```
- what-is-in-a-note
- mimic-cxr
- shareclefehealth2014task2
- wctecgdb
- phdsm
- chaos-heart-rate
- inipdmsa
- deidentifiedmedicaltext
- physiozoo
- hss
- mimiciii
- mednli-bionlp19
- bhi-2018-challenge
- mednli
- shareclefehealth2013
- ct-ich
- ptb-xl
- maternal-visceral-adipose
- picdb
- kinematic-actors-emotions
```

The citations in the following projects are not correct and, as far as I can see, never were:

```
- transformer-synthetic-note
- mimic-cxr-jpg
- phenotype-annotations-mimic
```

The following project is completely missing inline citations:

```
- cerebral-vasoreg-diabetes
```

I have kept screenshots of the original ordering in case these are helpful for reference. Some lists needed reversing after the update. To do this, I used the following approach:

```python
import copy

slug = "XXXX"
all_versions = PublishedProject.objects.filter(slug=slug)

# loop all versions
for p in all_versions: 
    original = {r.id:r for r in p.references.all()}
    new = copy.deepcopy(original)
    
    for k,k2 in zip(sorted(original), sorted(original, reverse=True)):
        # add . to avoid duplicate key issue
        new[k2].description = original[k].description + '.'
    
    # save the changes
    for k in new:
        new[k].save()
    
    # remove the extra '.'
    for k in new:
        new[k].description = new[k].description[:-1]
        new[k].save()
```

The change may have introduced ordering issues for active projects (i.e. ones in the system that are not published) so we all need to keep an eye out for this.

PR #1064, opened 2020-06-02
Remove Data Access section

elfeto:
There is an un-needed section of data access for self-credentialed projects.
Here I merge that section to the files section.

### DIFF
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -309,14 +309,6 @@
     <!-- /.sidebar -->
   </div>
 
-  {% if has_access and project.is_self_managed_access %}
-      <h2>Data Access</h2>
-      <div>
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks, this looks good. Here is a screenshot of what I now see:

Screen Shot 2020-06-02 at 12 02 26

- Not sure the `*` are needed around "new request"?
- As the text is related to data access, why not place under "Access the files"? It seems to fit better there than under the file size.
- I wonder whether we should cut "to this dataset" from the first sentence? It's possible that it will not always be a dataset, and I think the shorter message looks slightly cleaner.

elfeto:
Initially I added it under the "Files" sections not the "Access the files" because in theory this is the only place to see the history of requests because most people wont request access again for another project once they have access to the data. Also if the second request is declined and the person has access to the files, the access to the files stays.

I can move like this if you like:
Screen Shot 2020-06-03 at 9 12 08 AM

tompollard:
thanks!

PR #1065, opened 2020-06-02
Adds GraphQL API framework #1047

### LINKED ISSUE #1047: Develop API for integrating with annotation platforms
There are several existing platforms that could be used to gather useful annotations for PhysioNet datasets. This needs a lot more thought, but as a rough idea it would be good to develop a general API that:

1. allows an external platform to request a data file for annotation, perhaps along with associated metadata (details to be determined, but this might include existing annotations).
2. allows the external platform to submit a structured annotation back to PhysioNet. 

### Metadata profile for an annotation

The structure of the annotation will need to be developed. At minimum, the metadata should probably include:

- name/id of the annotation platform
- name/id of the user on the platform
- name/id of the record being annotated
- the annotation
- the time when the annotation was made.

### Metadata profile for an annotation task

One of the major challenges is understanding how the API can be made generalizable across PhysioNet, ideally to support multiple data types and modalities (images, waveforms, notes, etc). It feels like the annotation task will require a formal definition that would state things like:

- which files are to be annotated and the file type.
- a human-readable description of the annotation task (e.g. "The task is to label each R wave peak in the ECG recordings" or "The task is to draw a bounding box around pacemakers in the X-rays."
- a machine readable description of the annotation task that can be used by the annotation tool to render the task appropriately.

### Providing an interface for the annotation functionality

Annotation tasks may be driven by the research question, and there may be multiple annotation tasks for a single dataset. We need to come up with a simple way of allowing PhysioNet users to propose and implement an annotation task. My suggestion is that we do this with the use of a new "annotation" project type (see https://github.com/MIT-LCP/physionet-build/issues/1032). 

### Summary of tasks

So in summary, some good first steps might be to:

1. Design a metadata profile for a generalizable annotation. We can probably reuse or build on the format used by existing annotation platforms that we have looked at.
2. Design a metadata profile for an annotation task.
3. Review whether using the existing project functionality (https://github.com/MIT-LCP/physionet-build/issues/1032) could be modified to allow a new project type to be used for defining tasks and storing annotations.
4. Design the API!

Lucas-Mc:
Implements the GraphQL API interface to the website under the URL "https://physionet.org/graphql". Currently, the user is only able to query for published projects by title, version, slug, abstract, main_storage_size, and compressed_storage_size. Begins to fix #1047.

Currently, with the REST API, content is determined statically. For example, if I wanted a list of the database names I would have to go to [this](https://physionet.org/rest/database-list/) link, copy the contents, and then extract the data I want after the call to the server (GET). If we only wanted the project title, currently there would be no way to do so without creating another view and respective URL. 

With the GraphQL approach, data extraction is dynamic and the output data can be modified without creating separate views for each desired query. The whole idea of GraphQL is that everything is connected via a "graph" so all that has to be done on the development side (our side) is to provide it with the models and it will connect the data for us. 

If I wanted each published project along with its abstract, main storage size, and compressed storage size, all I would have to do it submit the following query:
Screen Shot 2020-06-02 at 4 28 07 PM

However, with the current REST API, we would have to make a call to the entire database and its sub-contents then extract what we wanted unless we wanted to make a separate view (and URL) for each possible database call:
Screen Shot 2020-06-08 at 1 51 41 PM


Lucas-Mc:
Just added published authors query shown below: 

Screen Shot 2020-06-03 at 8 11 44 AM

Lucas-Mc:
We need to think about how to separate the public and private fields. My thought is that we build two separate APIs (or maybe just one) but the `fields` description underneath the `Meta` class would change based of the user's credentials? See [here](https://docs.graphene-python.org/projects/django/en/latest/authorization/#user-based-queryset-filtering) for more. Thoughts anyone?

physionet-django/tests.py
```diff
@@ -0,0 +1,67 @@
+from schema import Query
```
tompollard, line 1: For consistency with other tests, should this go in the app subfolder, rather than the root folder?

physionet-django/tests.py
```diff
@@ -0,0 +1,67 @@
...
+from django.test.testcases import TestCase
+import graphene
+import pdb
+
+
+class AnExampleTest(TestCase):
```
tompollard, line 7: This should a more informative name.

physionet-django/tests.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def test_all_published_projects(self):
```
tompollard, line 9: It would be good to have some additional cases that demonstrate that the API is not giving access to content that we would like to keep private.

physionet-django/tests.py
```diff
@@ -0,0 +1,67 @@
...
+                        version
+                        slug
+                        abstract
+                        main_storage_size
+                        compressed_storage_size
+                    }
+                }
+            }
+        """
+        result = schema.execute(query)
```
tompollard, line 66: Can we check that the results are expected with a comparison to the known content?

physionet-django/tests.py
Lucas-Mc: So are you saying should the top-level `schema.py` file go into the app subfolder? If so, each app already has its own `schema.py` (or will after more than just the ones in `export` are added) and there should also be a project-level `schema.py` to wrap them all together. Do you think testing should be done for each app instead of just one testing file for all the QraphQL schemas together? Maybe only the `export` app folder should have the schemas and queries and tests?

physionet-django/tests.py
Lucas-Mc: Yes actually, we should be able to extract the content with `json.dumps(result.content)` I believe and compare it with either the fixtures or directly from the project objects.

physionet-django/tests.py
Lucas-Mc: That makes sense, I will add some tests which try to access content not described in the `fields` parameter.

physionet-django/tests.py
tompollard: the fixtures are loaded by `loaddemo` so comparisons can be made to the project object.

```py
project = PublishedProject.objects.get(title="thetitle")
assertequals(project.abstract,abstract)
```

physionet-django/tests.py
tompollard: sorry, it wasn't clear, but the comment doesn't refer to the import but to the module itself. most tests are in the associated app folder, but this test is in the project root. for consistency, perhaps move the test to `export` (which isn't a good name, but that's another story!).

Lucas-Mc:
With the removal of the interactive interface, queries can be made with the following syntax,
`http://localhost:8000/graphql?query=QUERY` where `QUERY` is the desired query, for example, 

`http://localhost:8000/graphql?query={all_published_projects{title}}` returns:

```
{"data":
  {"all_published_projects":
    [{"title":"Demo 2018 BHI and BSN Data Challenge"},
     {"title":"Demo ECG Signal Toolbox"},
     {"title":"Self Managed Access Database Demo"},
     {"title":"aa"},
     {"title":"aa"},
     {"title":"Demo eICU Collaborative Research Database"},
     {"title":"Demo software for parsing clinical notes"},{"title":"aa"}]
  }
}
```

EDIT: This is still more effective than a REST API since the endpoints are automatically generated by GraphQL based on the shape and elements of the objects it is given.

NOTE: We can also query to find all the possibilities of what to query using `query={__schema{types{fields{name}}}}`:
```
{"data":
  {"__schema":
    {"types":
      [{"fields":
        [{"name":"all_published_projects"},
         {"name":"all_published_authors"},
         {"name":"all_published_references"}]},
       {"fields":
        [{"name":"title"},
         {"name":"abstract"},
         {"name":"version"},
         {"name":"main_storage_size"},
         {"name":"compressed_storage_size"},
         {"name":"slug"}]},
       {"fields":null},
       {"fields":null},
       {"fields":
        [{"name":"first_names"},
         {"name":"last_name"},
         {"name":"corresponding_email"},
         {"name":"project"}]},
       {"fields":
        [{"name":"id"},
         {"name":"description"},
         {"name":"project"}]},
       {"fields":null},
       {"fields":
        [{"name":"types"},
         {"name":"queryType"},
         {"name":"mutationType"},
         {"name":"subscriptionType"},
         {"name":"directives"}]},
       {"fields":
        [{"name":"kind"},
         {"name":"name"},
         {"name":"description"},
         {"name":"fields"},
         {"name":"interfaces"},
         {"name":"possibleTypes"},
         {"name":"enumValues"},
         {"name":"inputFields"},
         {"name":"ofType"}]},
       {"fields":null},
       {"fields":null},
       {"fields":
        [{"name":"name"},
         {"name":"description"},
         {"name":"args"},
         {"name":"type"},
         {"name":"isDeprecated"},
         {"name":"deprecationReason"}]},
       {"fields":
        [{"name":"name"},
         {"name":"description"},
         {"name":"type"},
         {"name":"defaultValue"}]},
       {"fields":
        [{"name":"name"},
         {"name":"description"},
         {"name":"isDeprecated"},
         {"name":"deprecationReason"}]},
       {"fields":
        [{"name":"name"},
         {"name":"description"},
         {"name":"locations"},
         {"name":"args"}]},
       {"fields":null}]
    }
  }
}
```

Issue #1068, opened 2020-06-05
Provide citation details in "Your project was published" email to authors


tompollard:
There are benefits to people using formal citations when referring to content on PhysioNet (e.g. if people cite content using a DOI, then we can keep track of how content is being used, which is useful for measuring impact).

The email that is sent to authors currently says:

```
Project title: ert
Submission ID: TESTID
Submitting author: admin --- test@test.com

We are delighted to inform you that your project entitled "TEST" (version 0.1) has been published on PhysioNet. The project is available at https://physionet.org/content/TEST

We encourage you to share this link to increase your work's visibility within the research community. Thanks again for choosing to publish with PhysioNet, and we look forward to working with you again in the future.
```

It would be good to add the citation details to this email (and to give the DOI instead of the URL).

PR #1069, opened 2020-06-05
Add citation details to "project published" notification email

tompollard:
Minor change in response to #1068 to add citation details to the email that is sent to authors upon publication. The email now reads:

```
We are delighted to inform you that your project entitled "Test
database" (version 1.0) has been published on PhysioNet. The project
is available at: https://doi.org/10.7966/ymcx-6e03

The formal citation is:
Bot A. Test database (version 1.0). PhysioNet. 2020. Available from:
https://doi.org/10.7966/ymcx-6e03.

We encourage you to share these details to increase your work's
visibility in the research community. Thanks for choosing to
publish with PhysioNet.
```

### DIFF
--- a/physionet-django/notification/templates/notification/email/publish_notify.html
+++ b/physionet-django/notification/templates/notification/email/publish_notify.html
@@ -1,11 +1,14 @@
-{% load i18n %}{% autoescape off %}{% filter wordwrap:70 %}
+{% load i18n %}{% load project_templatetags %}{% autoescape off %}{% filter wordwrap:70 %}
 {{ project_info }}
 
 Dear {{ name }},
 
-We are delighted to inform you that your project entitled "{{ published_project.title }}" (version {{ published_project.version }}) has been published on PhysioNet. The project is available at {{ url_prefix }}{% url 'published_project' published_project.slug published_project.version %}.
+We are delighted to inform you that your project entitled "{{ published_project.title }}" (version {{ published_project.version }}) has been published on PhysioNet. The project is available at: {% if published_project.doi %}https://doi.org/{{ published_project.doi }}{% else %}{{ url_prefix }}{% url 'published_project' published_project.slug published_project.version %}{% endif %}
 
-We encourage you to share this link to increase your work's visibility within the research community. Thanks again for choosing to publish with PhysioNet, and we look forward to working with you again in the future.
+The formal citation is:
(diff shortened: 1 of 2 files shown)


tompollard:
Minor change after re-reading the email (final sentence has been removed). It now says:

```
We are delighted to inform you that your project entitled "Test
database" (version 1.0) has been published on PhysioNet. The project
is available at: https://doi.org/10.7966/ymcx-6e03

The formal citation is:
Bot A. Test database (version 1.0). PhysioNet. 2020. Available from:
https://doi.org/10.7966/ymcx-6e03.

We encourage you to share these details to increase your work's
visibility in the research community.
```

bemoody:
Do you have a particular reason for using the Vancouver style?  It
seems odd to use something different here from what is shown on the
site itself.

It's also awkward that you're changing the Python API here in order to
change the result at the template level.  I think you could write
instead {{project.citation_text_all.Vancouver}} if that's what you
want.

Simply using striptags will break apostrophes (as well as & " < > if
those were allowed.)  What we want is to also pass it through
html.unescape, which Django doesn't have a built-in filter for,
probably for good reason. :)  But you could do this:

    def citation_text_plain(self):
        citations = {}
        for style, html_text in self.citation_text_all().items():
            citations[style] = unescape(strip_tags(html_text))
        return citations

Or, maybe we just ought to define a custom "convert-to-plain-text"
template filter.

tompollard:
Thanks @bemoody. 

> Do you have a particular reason for using the Vancouver style? 

Not especially, other than it is the one that I'm most familiar with. Well spotted on the mismatch with the style that appears elsewhere. I've swapped to APA for consistency. 

> It's also awkward that you're changing the Python API here in order to change the result at the template level.

Having a default style in the function seems like a good idea because it helps to enforce consistency across the site (the alternative is that we state the style in templates, which may lead to mismatches). I've removed the default for now though.

> Simply using striptags will break apostrophes (as well as & " < > if those were allowed.) ... maybe we just ought to define a custom "convert-to-plain-text" template filter.

Good idea. I defined an html-to-text template filter as suggested.

bemoody:
Looks good, thanks!

bemoody:
One small style fix :)

tompollard:
> One small style fix :)

thanks, good catch!

PR #1072, opened 2020-06-15
Adds physionet.mit.edu to servers

Lucas-Mc:
Adds `physionet.mit.edu` to list of servers accepted by Django in the `nginx` configuration. This is needed since Tom and I have decided to redirect all traffic requests from `physionet.mit.edu` to `physionet.org`. The redirect has already been set up with the MIT Service Desk, we just need to give permission to it here and later update the certificates.

### DIFF
--- a/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
+++ b/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
@@ -205,3 +205,29 @@ server {
         return 301 https://$host$request_uri;
     }
 }
+
+server {
+    listen      80;
+    server_name physionet.mit.edu;
+    return      301 https://physionet.org$request_uri;
+}
+
+server {
+    listen      443 ssl; # managed by Certbot
+    ssl_certificate /etc/letsencrypt/live/alpha.physionet.org/fullchain.pem; # managed by Certbot
+    ssl_certificate_key /etc/letsencrypt/live/alpha.physionet.org/privkey.pem; # managed by Certbot
+    include /etc/letsencrypt/options-ssl-nginx.conf; # managed by Certbot
+    ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem; # managed by Certbot
+
+    # SSL stapling
+    ssl_stapling on;
+    ssl_stapling_verify on;
+    ssl_trusted_certificate /etc/letsencrypt/live/alpha.physionet.org/fullchain.pem;
+
+    # Security headers
+    add_header Strict-Transport-Security "max-age=2592000; includeSubDomains; preload";
+    add_header Accept-Ranges bytes;
+
+    server_name physionet.mit.edu;
+    return      301 https://physionet.org$request_uri;
+}


tompollard:
If someone visits https://physionet.mit.edu/content/mimiciii-demo/ in their browser, which of the following would you like to happen?

1. forward the user to https://physionet.org/content/mimiciii-demo/
2. serve the content at https://physionet.mit.edu/content/mimiciii-demo/
3. something else?

Lucas-Mc:
I would say the first one so we can try to migrate everything to `physionet.org`. Would this help with that?

tompollard:
In this case, `return` or `rewrite` may be needed somewhere?:

e.g.:

```
server_name     physionet.mit.edu;
return          301 http://www.physionet.org$request_uri;
```

or 

```
server_name physionet.mit.edu;
rewrite ^/(.*)$ http://physionet.org/$1 permanent;
```

bemoody:
Right, in order to *redirect* clients you want to add an additional
"server" block that just does a redirect.  (Maybe two additional
server blocks, I'm not sure if you can have one server that does both
SSL and non-SSL.)

We might need to add an option to the existing server blocks to mark
them as "default".

bemoody:
I'm not sure if this will work or not - does there need to be a
"listen" in there as well?

Either way we want to redirect to https://, not http://.

bemoody:
> I'm not sure if this will work or not - does there need to be a
> "listen" in there as well?

Unclear, seems like it responds to port 80 by default but not to port 443?

$ printf 'HEAD / HTTP/1.1\r\nHost: physionet.mit.edu\r\n\r\n' | nc
$server $httpport
HTTP/1.1 301 Moved Permanently
Server: nginx/1.14.2
Date: Wed, 17 Jun 2020 18:02:43 GMT
Content-Type: text/html
Content-Length: 185
Connection: keep-alive
Location: http://www.physionet.org/

$ printf 'HEAD / HTTP/1.1\r\nHost: physionet.mit.edu\r\n\r\n' |
openssl s_client -quiet -connect $server:$httpsport  2>/dev/null
HTTP/1.1 400 Bad Request
Server: nginx/1.14.2
Date: Wed, 17 Jun 2020 18:02:33 GMT
Content-Type: text/html; charset=utf-8
Connection: keep-alive

Lucas-Mc:
Update: I added `physionet.mit.edu` to the list of servers using `certbot certonly --expand -d www.physionet.org -d physionet.org -d mail.physionet.org -d alpha.physionet.org -d physionet.mit.edu`

`certbot certificates` now returns...
```
Certificate Name: alpha.physionet.org
    Domains: www.physionet.org alpha.physionet.org mail.physionet.org physionet.mit.edu physionet.org
```

tompollard:
Thanks Lucas, this looks good to me. @bemoody ?

bemoody:
Should it redirect to physionet.org rather than www.physionet.org?

tompollard:
Well spotted. Yes, I think that would be better.

PR #1074, opened 2020-06-18
Adds physionet.mit.edu to ALLOWED_HOSTS

Lucas-Mc:
Adds `physionet.mit.edu` to list of `ALLOWED_HOSTS` for the completion of the redirect to `physionet.org`.

### DIFF
--- a/physionet-django/physionet/settings/production.py
+++ b/physionet-django/physionet/settings/production.py
@@ -6,7 +6,7 @@ from physionet.settings.base import *
 
 DEBUG = False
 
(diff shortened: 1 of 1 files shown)


bemoody:
Why would that be necessary or desirable?  The whole point was that we
didn't want to do that.

tompollard:
This will mean that PhysioNet is also served from `https://physionet.mit.edu/`, which isn't really what we want.

Lucas-Mc:
The problem is we are still getting a 400 server error for an unknown reason

bemoody:
If there's an unknown reason, then start by figuring out the reason.

And please do not merge your own pull request.

PR #1075, opened 2020-06-18
Fixes 400 server error during physionet.mit.edu redirect

Lucas-Mc:
When trying to access physionet.mit.edu, a 400 server error is encountered with the current nginx config file. Adding this to `ALLOWED_HOSTS` solves the problem but the top URL reads `physionet.mit.edu` which isn't what we intend. Therefore, it appears like a rewrite is necessary to achieve our desired result. Suggestions are welcome though since this is my first time handling the servers on my own.

### DIFF
--- a/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
+++ b/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
@@ -209,7 +209,7 @@ server {
 server {
     listen      80;
     server_name physionet.mit.edu;
-    return      301 https://physionet.org$request_uri;
+    rewrite     ^/(.*)$ https://physionet.org/$1 permanent;
 }
 
 server {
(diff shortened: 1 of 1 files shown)


bemoody:
It looks like you have pushed but have not installed your previous
config changes.

When you push changes to the production server, it will tell you if
there are configuration files that are out of date, but it won't (and
cannot) update them automatically.

In this case, you would have to run:

  sudo install /physionet/physionet-build/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
/etc/nginx/sites-available/physionet_nginx.conf

  sudo nginx -t

  sudo systemctl reload nginx

But, if you're going to make changes, then experiment locally and
experiment on the staging server.  Don't touch the production server
until you are confident that it's not going to break anything.

PR #1080, opened 2020-06-22
Dynamically generate database content #1078

### LINKED ISSUE #1078: Replace static list of databases with dynamically generated content
I thought there was an issue for this, but I can't seem to find it. The front page has a prominent link to a list of databases: https://physionet.org/about/database/

The list is hardcoded and doesn't capture most of the content released in the past 6 months or so. If we want to have a single page listing all datasets, then it needs to be generated dynamically (same goes for software, challenges, etc).

The "classes" should probably just be removed, because I don't think they are particularly meaningful now. 

```
Class 1 - completed reference databases.
Class 2 - archival copies of raw data that support published research, contributed by authors or journals.
Class 3 - other contributed collections of data, including works in progress.
```

All published content is somewhere between "class 1" and "class 2". All datasets are contributed by the PhysioNet community. Limitations should be documented, and progress is captured through versioning.

Lucas-Mc:
Fixes the issue of the list of [databases](https://physionet.org/about/database/) not being dynamically generated. In the past, this database list was generated manually which is difficult to keep up with so a dynamic solution was requested. I first filter each project by their `access_policy` (open, restricted, credentialed) then sort the projects alphabetically by their lowercase title. Fixes #1078.


physionet-django/physionet/views.py
```diff
@@ -1,11 +1,13 @@
 from collections import OrderedDict
 from os import path
 from re import fullmatch
+import pdb
```
tompollard: is this needed?

physionet-django/physionet/views.py
```diff
@@ -126,7 +128,13 @@ def database_overview(request):
     """
     Temporary content overview
     """
-    return render(request, 'about/database_index.html')
+    open_projects = PublishedProject.objects.filter(access_policy=0).order_by(Lower('title'))
```
tompollard: the page should only display databases (this query gets all projects).

physionet-django/templates/about/database_content.html
```diff
@@ -23,785 +23,41 @@ <h2>Overview</h2>
...
-Each database is placed into a class according to the following specifications. Contributed data are placed in classes 2 and 3 on acceptance, and may be admitted to class 1 after review and a public comment period.</p>
-<ul>
-<li>Class 1 - completed reference databases.</li>
-<li>Class 2 - archival copies of raw data that support published research, contributed by authors or journals.</li>
-<li>Class 3 - other contributed collections of data, including works in progress.</li>
-</ul>
-
-<p>
-On this page, listings within each group are ordered by class, and then
+On this page, listings within each group are ordered by its access policy, and then
```
tompollard: This following sentence is difficult to follow:

> "On this page, listings within each group are ordered by its access policy, and then..."

Perhaps delete the sentence, and begin the overview with:

> This page displays a complete list of databases, grouped by access policy. To search for content, visit [the search page](https://physionet.org/content/).

physionet-django/templates/about/database_content.html
```diff
@@ -23,785 +23,41 @@ <h2>Overview</h2>
...
-      </li>
-    </ul></li>
-
-  <li>[Class 3] <a href="{% url 'published_project_latest' 'picdb' %}">Paediatric Intensive Care Database</a>:
-    The Paediatric Intensive Care (PIC) database is a large
-    paediatric-specific, single-center, bilingual database comprising
-    information relating to children admitted to critical care units
-    at a large children’s hospital in China.</li>
+{% for project in open_projects %}
+  <li><a href="{% url 'published_project_latest' project.slug %}">{{ project.title }}</a>: {% if project.short_description %}{{ project.short_description }}{% else %}{{ project.abstract|safe|truncatechars_html:200 }}{% endif %}</li>
```
tompollard: spacing between items in the list is inconsistent:

Screen Shot 2020-06-22 at 14 55 22

physionet-django/templates/about/database_content.html
```diff
@@ -23,785 +23,41 @@ <h2>Overview</h2>
...
-     most of which include beat annotations in addition to the original ECG
-     signals.</li>
-  <li> <a href="#cv">Other Cardiovascular Databases</a></li>
-  <li> <a href="#gait">Gait and Balance Databases</a></li>
-  <li> <a href="#neuro">Neuroelectric and Myoelectric Databases</a>: EEG, EHG, and more.</li>
-  <li> <a href="#synthetic">Synthetic Data</a></li>
-</ul>
-<li> <a href="#image">Image Databases</a></li>
-<li> <a href="#miscellaneous">Miscellaneous Databases</a></li>
+<li> <a href="#open">Open Access</a>: Data projects with open access.</li>
```
tompollard: These descriptions don't add much (e.g. "Open access: Data projects with open access"). If we include descriptions, they should be useful. 

Perhaps we could add something like: 
- "Open access datasets are shared under permissive licenses that allow broad reuse."
- "Restricted access datasets require a Data Use Agreement to be signed." etc

physionet-django/physionet/views.py
Lucas-Mc: No, but 18 other files have them and it would be nice to progress so all of them have it for debug purposes

physionet-django/templates/about/database_content.html
Lucas-Mc: Do you have more information for each? Is there a reference I can look at for a better description? Not exactly sure the differences

physionet-django/templates/about/database_content.html
Lucas-Mc: Hmmmm, does this have to do whether it's and abstract or short description? What did you submit here?

physionet-django/physionet/views.py
tompollard: Just because it's in our code already, doesn't mean it should be there! If you're using Python 3.7, `breakpoint()` should do what you need and it doesn't require an import.

physionet-django/templates/about/database_content.html
Lucas-Mc: Maybe because the abstract includes the `<p>` tag but the short description doesn't?

physionet-django/physionet/views.py
Lucas-Mc: Isn't the `venv` 3.6? I can remove it if it shouldn't be there

physionet-django/templates/about/database_content.html
tompollard: As far as I know, the only description is in the "Access" page for new submissions, but the description there isn't particularly clear. We can chat, but briefly:

- "Open access" means the content is shared under an open access license (e.g. MIT License; Creative Commons licenses).
- "Restricted access" means that users must (1) have a PhysioNet account and (2) sign a DUA.
- "Credentialed access" means that users must (1) have a PhysioNet account, (2) sign a DUA, and (3) complete the online CITI training.

physionet-django/templates/about/database_content.html
tompollard: Yep, I think it will be something like that.

physionet-django/physionet/views.py
tompollard: It's not a big deal, but I guess my question is why we would import pdb when it isn't used? If someone is debugging then they can `import pdb; pdb.set_trace()` wherever they want the breakpoint.

Lucas-Mc: I was just unsure of the consistency we want in the code since some files have it and some don't

physionet-django/templates/about/database_content.html
Lucas-Mc: Was there a reason the abstract includes the `<p>` tag? Just wondering

tompollard: From a quick look at the model, the abstract is HTML and the short description is a character field. I don't think there is a particularly strong reason for the difference, other than it may be helpful to constrain the short description because it is used for things like search listings, social media text, etc where HTML might be troublesome.

tompollard:
Meant to post the comment at: https://github.com/MIT-LCP/physionet-build/issues/1078 in this pull request!

tompollard:
Nice work Lucas. Please could you change the `<h1>` tags to `<h2>` for the subheaders so that we don't break [W3C validation](https://validator.w3.org/)? (e.g. `<h1 id="restricted">Restricted Access</h1>` should be `<h2 id="restricted">Restricted Access</h2>`). I'll then merge and we can test on the live site.

PR #1081, opened 2020-06-23
Redirects alpha.physionet.org to physionet.org #1079

### LINKED ISSUE #1079: alpha.physionet.org should redirect to physionet.org
alpha.physionet.org was intended as a temporary domain while the new PhysioNet website was under development.

We should now redirect alpha.physionet.org to physionet.org (in the same way that physionet.mit.edu redirects to physionet.org).

Lucas-Mc:
Redirects `alpha.physionet.org` to `physionet.org` in the same way that the recently added `physionet.mit.edu` redirects to `physionet.org`. The reason for this new redirect is that `alpha.physionet.org` was intended as a temporary domain while the new PhysioNet website was under development. Fixes #1079.

### DIFF
--- a/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
+++ b/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
@@ -6,8 +6,7 @@ upstream django {
 
 ## configuration of the server
 server {
-    server_name physionet.org alpha.physionet.org physionet-production.ecg.mit.edu;
-
+    server_name physionet.org physionet-production.ecg.mit.edu;
     charset     utf-8;
 
     # max upload size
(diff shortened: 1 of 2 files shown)


tompollard:
thanks lucas. i assume https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/settings/production.py also needs to be updated?

Lucas-Mc:
> thanks lucas. i assume https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/settings/production.py also needs to be updated?

Actually yes, I forgot that this one was already set up to conserve the URL, thanks!

bemoody:
Sounds good.

PR #1083, opened 2020-06-30
Fixes format and content in about/database_test #1080

Lucas-Mc:
Adds a missing line break and prevents previous versions of projects to be displayed. Currently all versions of a project are displayed on the test database page though we would only like the most recent version. Fixes both #1078 and #1080.


tompollard:
thanks, looks good to me.

PR #1087, opened 2020-07-01
Add "Full name" column to the list of credentialed users

tompollard:
This pull request adds a new column to the table displayed at: http://127.0.0.1:8000/console/complete-list-credentialed-people/. The change was requested by KP, the primary user of the table.

Before the pull request:

Screen Shot 2020-07-01 at 19 20 53

After the pull request:

Screen Shot 2020-07-01 at 19 20 18


Lucas-Mc:
I think this looks good and I like how you still kept both the first and last name fields!

tompollard:
Your positivity is appreciated but not required! Thanks for checking, and please merge if you are confident that I haven't broken anything.

PR #1088, opened 2020-07-03
Modifies about/database_test page #1078

### LINKED ISSUE #1078: Replace static list of databases with dynamically generated content

Lucas-Mc:
Modifies the `about/database_test` page to now display just a single list of all the latest versions of the databases sorted in alphabetical order. This replaces the previous layout which was separated by access policy. Additionally, more information is given on how to search for databases as well as the differences between the access policies for each database. Fixes #1078.

### DIFF
--- a/physionet-django/physionet/views.py
+++ b/physionet-django/physionet/views.py
@@ -134,19 +134,10 @@ def database_overview_test(request):
     """
     Temporary content overview
     """
-    open_projects = PublishedProject.objects.filter(
-        resource_type=0, access_policy=0,
-        is_latest_version=True).order_by(Lower('title'))
-    restricted_projects = PublishedProject.objects.filter(
-        resource_type=0, access_policy=1,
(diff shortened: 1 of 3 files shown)


tompollard:
Looks good to me, thanks Lucas.

PR #1091, opened 2020-07-03
Replaces about/database_test with about/database

Lucas-Mc:
Transfers the content from the `about/database_test` page to the normal `about/database` page which will now be linked to other parts of the PhysioNet website. This transfer comes after direct approval from Dr. Roger Mark after viewing the test page.


tompollard:
Thanks, looks good to me.

PR #1098, opened 2020-07-08
Replaces about/software_test with about/software

Lucas-Mc:
Transfers the content from the `about/software_test` page to the normal `about/software` page which will now be linked to other parts of the PhysioNet website. This transfer comes after direct approval from Dr. Roger Mark after viewing the test page.


physionet-django/templates/about/software_content.html
```diff
@@ -1,1468 +1,15 @@
...
-  Signals.
-  <i>Circulation</i> <b>101</b>(23):e215-e220 [Circulation Electronic Pages;
-  <a href="http://circ.ahajournals.org/content/101/23/e215.full"
-  target="other">http://circ.ahajournals.org/content/101/23/e215.full</a>];
-  2000 (June 13).</p>
-
+<section id="overview">
+    <h2>Overview</h2>
+  <p>This page displays an alphabetical list of all software projects on PhysioNet. To search content on PhysioNet, <a href="{% url 'content_index' %}">visit the search page</a>. Enter the search terms, add a filter for resource type if needed, and select how you would like the results to be ordered (for example, by relevance, by date, or by title).</p>
+  <p>Many of the time series datasets on PhysioNet (for example, ECGs) are shared in waveform-database (WFDB) compatible file formats.  For more information on these formats, see the <a href="https://www.physionet.org/physiotools/wag" />WFDB Applications Guide</a>. WFDB-formatted datasets can also be viewed online with <a href="{% url 'lightwave_home' %}">LightWAVE</a>.</p>
```
tompollard: please could you take another look at this URL? i think, at the least, relative would be better than absolute (more flexible, avoids breaking if the root changes).

tompollard:
looks good, thanks lucas.

PR #1099, opened 2020-07-08
Modifies search link to forward directly to desired content

Lucas-Mc:
Modifies both the database and software search links to forward directly to the search page again except now only the desired resource type is selected. For example, the redirect from the software page would take the user to the search page with only the software resource type selected.


tompollard:
thanks!

PR #1102, opened 2020-07-17
Fix funny cred app order

elfeto:
At the moment there is a funny credential application order where the order us by contacted reference and not by application date, an example is as follows:

Person 1
...Date of this agreement: May 1, 2020, 1:37 a.m.

Person 2
...Date of this agreement: May 5, 2020, 9:20 p.m.

Person 3
...Date of this agreement: May 3, 2020, 12:20 a.m.

In this PR after some conversation with KP we agreed on the following order:
- messages with reference not contacted, but with reference known, sorted by application date.
- messages with reference not contacted and reference unknown, sorted by application date.
- messages with reference contacted, sorted by application date.

The way i choosed to sort this is nor an unorthodox way because is a bit more complicated than what it looks.
In order to get all reference not contacted first, I have to sort by null in that way which can be done at the filter level, but the rest I believe it cant be done at that leve. hence the 3 way sort.

Apart from that KP wanted that all credential applications that have the reference as contacted, to not be able to alter that field. Here I check/alter if the reference contacted field is NONE (not contacted) and then and only then update the field.

### DIFF
--- a/physionet-django/console/templates/console/complete_application_display.html
+++ b/physionet-django/console/templates/console/complete_application_display.html
@@ -22,7 +22,7 @@
 	{{ application.reference_email }} 
 	{% if application.reference_contact_datetime %}
 		[The {{ application.reference_category|yesno:"reference,supervisor" }} was contacted on {{ application.reference_contact_datetime }}]
-	{% elif application.known_ref %}
+	{% elif application.ref_known_flag %}
         	[Reference is known]
 	{% endif %}
       </p>
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -964,8 +964,9 @@ def complete_credential_applications(request):
          request.POST['contact_reference'].isdigit():
             application_id = request.POST.get('contact_reference', '')
             application = CredentialApplication.objects.get(id=application_id)
-            application.reference_contact_datetime = timezone.now()
-            application.save()
+            if not application.reference_contact_datetime:
+                application.reference_contact_datetime = timezone.now()
+                application.save()
             # notification.contact_reference(request, application)
             if application.reference_category == 0:
                 mailto = notification.mailto_supervisor(request, application)
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -1006,40 +1006,26 @@ def complete_credential_applications(request):
     applications = applications.order_by(F('reference_contact_datetime').asc(
         nulls_first=True), 'application_datetime')
 
-    # Here we sort in 3 diferent ways:
+    # Group applications and sort by application date:
     # 1. reference not contacted, but with reference known
-    # 2. reference not contacted, but with reference unknown
-    # 3. reference contacted
-    # All 3 are sorted by application date
     known_ref_apps_not_contacted = []
+
+    # 2. reference not contacted, but with reference unknown
     unknown_ref_apps_not_contacted = []
+
+    # 3. reference contacted
     contacted_apps = []
-    # Do the proper sort
+
     for application in applications:
         application.mailto = notification.mailto_process_credential_complete(
             request, application, comments=False)
-        if CredentialApplication.objects.filter(
-            reference_email__iexact=application.reference_email,
-            reference_contact_datetime__isnull=False).exclude(
-                reference_email='') and \
-                application.reference_contact_datetime is None:
-            # If the reference has been contacted before, mark it so
-            application.known_ref = True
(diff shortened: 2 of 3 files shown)


bemoody:
This would be easier to read if you moved the "is this a known
reference" logic into a separate function.  Perhaps define it as a
method of CredentialApplication.

The sorting logic here seems overcomplicated even for the stated goal.
First you sort by "reference_contact_datetime, nulls first; then by
application_datetime", then you split this into three lists, then sort
each of those three lists in a somewhat roundabout way.

Is there a reason for that first complicated ordering?  Why not just
order the query set by application_datetime, then split into three
lists and concatenate them?

elfeto:
>The sorting logic here seems overcomplicated even for the stated goal.
First you sort by "reference_contact_datetime, nulls first; then by
application_datetime", then you split this into three lists, then sort
each of those three lists in a somewhat roundabout way.
> 
>Is there a reason for that first complicated ordering?  Why not just
order the query set by application_datetime, then split into three
lists and concatenate them?

It might be overcomplicated, but its the only way I could think of to sort it the way is desired.
 
The reasoning in the first order by reference_contact_datetime is to get the people whose reference has not been contacted first hence the `nulls_first ` section.

At that time we are ordering by reference_contact_datetime for the nulls first, but we don't want to order by reference_contact_datetime but with application_datetime.
(hence the unorthodox way that I just realized I wrote orthodox instead of unorthodox. I am correcting that)

As for the reason not sorting by application_datetime since the beginning is because then I would not be able to get the reference_contact_datetime nulls first. (or at least I don't know how to get them via that sort)

elfeto:
I can separate it into another function adding more comments, and I am also open to other ideas to achieve the same goal.

bemoody:
> As for the reason not sorting by application_datetime since the beginning is
> because then I would not be able to get the reference_contact_datetime nulls
> first. (or at least I don't know how to get them via that sort)

Isn't that exactly what you're doing when you iterate over the query
set?  You're scanning through all of the applications, putting the
ones that have reference_contact_datetime=null into one of the first
two lists and putting those with reference_contact_datetime!=null into
the third list.

elfeto:
> Isn't that exactly what you're doing when you iterate over the query
set?  You're scanning through all of the applications, putting the
ones that have reference_contact_datetime=null into one of the first
two lists and putting those with reference_contact_datetime!=null into
the third list.

In that if/else I am not checking that application by itself, but I am checking if there exist another application that has the same reference and if it has been contacted. I am not technically sorting in the loop, I am inserting them into the 3 sections to then be sorted.

The application_datetime is only sorted properly in the `t1 t2 t3`, not in the for loop nor the order_by.

Here is the scenario....

The setting the nulls first is done in the order by
```
applications = applications.order_by(F('reference_contact_datetime').asc(
        nulls_first=True), 'application_datetime')
```
but that only gives me the applications where `reference_contact_datetime` nulls first sorted by `reference_contact_datetime` and not `application_datetime`.

The important thing here is the `reference_contact_datetime` nulls first.

The iteration adding them into separate arrays is just to properly sort them in the 3way sort KP wants since I don't know how to achieve that in another way. still once is in in the `t0 t1 t2` is in the wrong order because is still sorted by `reference_contact_datetime`.

The 3way sort (`t1 t2 t3`) is to set the following order:

1. reference not contacted, but with reference known
2. reference not contacted, but with reference unknown
3. reference contacted

tompollard:
@bemoody the sorting isn't pretty, but the page is hacky anyway so I'm not worried about code being clean. As far as possible, our aim has been to keep this content independent from the main site, so it's easy to drop out later (e.g. when the full codebase is being reused). I haven't reviewed the pull request but if it works and doesn't have any major issues, my preference is to merge without spending too much time on refactoring.

tompollard:
I moved the "known reference" check to a method of CredentialApplication, as suggested. I think this makes the logic a little easier to follow. 

@bemoody please could you check and confirm that you are happy with this? The chunk could do with more work, but KP has been waiting for a few days so it would be good to avoid delaying further unless absolutely necessary. 

@elfeto I didn't find it easy to follow the logic, so please check that this still achieves what was intended!

elfeto:
@tompollard  I took some time to go over this because I acknowledge is not easy to read through... Maybe in the near future we should spend some time to simplify or convert it a bit more modular. 

As for the code, it works and looks better.

tompollard:
thanks for checking again @elfeto and thanks for making the change. @bemoody any reason why this shouldn't be merged?

bemoody:
Yeah, this looks clearer.  Thanks!

PR #1106, opened 2020-08-02
Adds waveform viewer [WIP]

Lucas-Mc:
Adds waveform viewer which allows the user to dynamically select the record and signal of the current project based on its slug and version. Currently, to access this waveform viewer, the user must go to a desired project where they will notice a `Visualize Waveforms (demo)` link which will send them to the app page. This app was developed using Dash / Plotly and includes a dropdown to select the desired record, and another dropdown to select the signal based on the desired record file's available signals. This app refreshes dynamically and changing the dropdown at any point will automatically refresh the displayed signal.

It should be noted that it takes a few seconds for the waveform to load, but once it is loaded it is very quick and responsive. I think there are several ways to increase the speed a bit.. either increase the `down_sample` of the signal (currently 3) or add a time range selector where the user selects their desired times which would allow the app to only generate a subset of the data thus increasing its load time, but anyways, currently it only takes a couple seconds which isn't too bad! (I also need to play around with the dev tools more to see which part takes the longest [I believe it's in the HTTP requests but not sure])

Additional details from the app HTML page itself:
```
      This is a demo waveform visualization built with Plotly/Dash. Select your desired
      record file and signal. The first 60 seconds are always displayed initially. You
      can zoom by clicking and dragging which range of times you would like to view. You
      can pan (move left and right) with the pan button on the graph figure as seen as a
      cross with arrows on each end in the top right corner upon hover. While you're there,
      you can generate a screenshot to be automatically downloaded, zoom in and out, and
      view the entire signal with the home button. You can also view the time and value of
      a waveform by hovering your mouse over a desired location. Another way to do this 
      would be to utilize the range-slider at the bottom of the figure; zooming can be done
      using the same technique as in the figure (click and drag). *Note: it may take a few
      seconds for the waveform to load depending on its size but this is a work in progress
      and there's much more optimization still left to do. :)
```


Lucas-Mc:
Most recent commit modifies this pull request to handle annotation project data. Now, the user selects their desired record and event and the plot automatically updates based on their choices. The user can then zoom and pan while each of the graphs remain aligned. Load times are much faster and no downsampling is required.

Screen Shot 2020-09-22 at 2 30 26 PM

Lucas-Mc:
TODO:
- [x] Add information about the event (VTACH, etc.)
- [x] Display the location of the event
- [x] Add place for user to submit feedback (True, False, Uncertain)
- [x] Clear the comment box each time a new signal is loaded

NOTE:
- Not in this PR, but later create some kind of API to handle the reviewer feedback

PR #1107, opened 2020-08-03
Bump pillow to 7.2.0

bemoody:
Bump pillow to 7.2.0.  This fixes some security issues.


bemoody:
(I don't think the existing test suite exercises Pillow at all, but the test cases from pull #1103 do.)

PR #1109, opened 2020-08-05
Fixes unsuccessful credential application decision dates #1108

### LINKED ISSUE #1108: Empty values for decision date for unsuccessful credentialed applications
This issue can be found on [this](https://physionet.org/console/past-credential-applications/unsuccessful) link where all applications are sorted by decision date, newest to oldest. Some of the applications have an empty value set to their decision date which is being pushed to the top of the sorted page. It appears that a `None` value may be the cause of this error but further investigation needs to be done to find the best way to fix it. https://github.com/MIT-LCP/physionet-build/blob/b1d24a04d7ad08ebe5d7217c4ab6bc2e2d201ea0/physionet-django/console/views.py#L1142-L1162

Lucas-Mc:
Fixes the logic of the template responsible for displaying the unsuccessful credential applications in order to display, if not null and in the order presented here, the: decision date, reference contact date, and application date. Fixes #1108.


elfeto:
This doesn't fix the problem, it just sets a random date time on the "Decision Date" table field.

Lucas-Mc:
I couldn't find the data in the database, is the decision date still there? In the meantime, I can stop it from setting the value to `None`. ... What if when we remove credentialing, the decision date gets updated? Like this:

```
            if request.POST['remove_credentialing'].isdigit():
                cid = request.POST['remove_credentialing']
                c_application = CredentialApplication.objects.filter(id=cid)
                if c_application:
                    c_application = c_application.get()
                    c_application.user.is_credentialed = False
                    c_application.user.credential_datetime = None
                    c_application.decision_datetime = timezone.now()
```

elfeto:
I feel we should keep the:
```
                    c_application.user.credential_datetime = None
```
The same (as in not set to None) because the user was credentialed at some point even if it was revoked, even though it can bring questions onto why is the credential_datetime set. @tompollard  any thoughts? 

As for the set of the decision_datetime, it should fix the problem of removing the credentialing, but also there is the legacy issue (its the same).

elfeto:
Here I take a slight different approach.
I add a revoke method and status option for  all applications and adds a revoke method for Legacy.

This will simplify the view function.

tompollard:
Thanks for looking at this! I like the idea of adding a "revoked" status, which deals with the underlying issue more directly.

It's important that we keep a historical log of (1) who signed a DUA and (2) the date the DUA was signed, even if the status is then revoked. e.g. in the following situation:

- person x signs DUA
- person x has access revoked for breaking DUA

In this case, we want to be able to demonstrate that the person both signed the DUA and broke the DUA. It looks like the new approach would remove this information in the following chunk:

```
        # Removes all DUA signatures
        with transaction.atomic():
            for dua in self.user.dua_signatures.all():
                logger.info('Removing DUA signature for project {0} and user \
                            {1}, this DUA was signed at {2}'.format(
                                dua.project, dua.user, dua.sign_datetime))
                dua.delete()
            self.user.save()
```

Is there a reason why we have to remove all of the dua_signatures? Are there obvious problems that it causes if we keep them there?

If they have to be removed, do we capture all of the information that we need in the log? If so, do we keep the logs long enough to be able to refer back to them 4 or 5 years after the DUA was first signed?

This kind of highlights a problem with the `DUASignature` model, which is that we should probably be recording exactly which DUA was signed (because it is possible that the DUA might be updated for a project), which we don't at the moment. We can save this issue for another day!

tompollard:
(i didn't intend to close the pull request...my mouse jumped!)

elfeto:
There is no need to remove the DUA because the access first check if the person is credentialed.

tompollard:
Thanks, looks good to me! @bemoody any feedback before merging?

bemoody:
I haven't followed this discussion and the five commits here bear no
resemblence to Lucas's original pull request.  Please open a new pull
request and describe the changes.

tompollard:
I think the overall target hasn't changed (fixing logic for displaying the unsuccessful credential applications) but I guess a new PR doesn't hurt!

PR #1112, opened 2020-08-05
Replace remove_credentialing for Revoke

elfeto:
Fixing issue #1108

Here I move the remove_credentialing inside the credential application itself creating a method for both Credential and Legacy Applications to revoke an existing approved application, also adding a revoked status.

For legacy I just set the user to not credentialed.
For Credentialed applications I set the status, date of change, and then remove the user credentialing.
All existing signed DUAs are kept.


tompollard:
Looks good to me! Essentially this adds a new "revoked" status, that explicitly records when a credentialing application has been revoked. @bemoody ?

bemoody:
Thanks.  This looks like a definite improvement.

I wonder, though, if we ought to keep track of when the application
was originally approved as well as when access was revoked.

What happens if an application is approved, the person applies for
access to a self-managed project, and then their credentialing is
revoked?

tompollard:
> I wonder, though, if we ought to keep track of when the application was originally approved as well as when access was revoked.

I would say yes, and there are definitely improvements that could be made to track state for both credentialing and signing of DUAs. I think these would be best in a new issue/PR unless you can think of a simple fix?

> What happens if an application is approved, the person applies for access to a self-managed project, and then their credentialing is revoked?

Good question. @elfeto, any idea? My assumption (without looking back at the code) is that the user is unable to access the data because they are no longer credentialed, but their DUA and self-approval status is unchanged.

elfeto:
It's a similar case of a signed DUA. The access would be lost because he is no longer credentialed.
He will be in the list of the project of people with access but he wont have access.

elfeto:
> I wonder, though, if we ought to keep track of when the application was originally approved as well as when access was revoked.

You are meaning to leave `decision_datetime` and jus add `revoked_datetime` ?

bemoody:
Looks good to me.  @tompollard?

tompollard:
Yeah, I'm happy with this. Still more work to do on improving the way we log credentialing and DUA history, but this is a step forward!

PR #1114, opened 2020-08-06
Don't allow physionet.org as email address in contact form

bemoody:
In the contact form (/about/#contact), require the submitter to supply an email address that is *not* on a subdomain of physionet.org, so that the message cannot appear to be a legitimate message "from" physionet.org even though it passes through the same mail servers.


tompollard:
good idea.

PR #1115, opened 2020-08-07
Add HTTP/2, and tidy up nginx configuration

bemoody:
This is an assortment of small fixes to the nginx configuration.  In particular, this enables the use of HTTP/2 over TLS, which should give better performance for browsers.

Note that the redirect server configuration is moved to a separate file, /etc/nginx/sites-available/physionet-redirect.conf.  In addition to installing this file, a symlink must be created: /etc/nginx/sites-enabled/physionet-redirect.conf -> ../sites-available/physionet-redirect.conf.

### DIFF
--- a/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
+++ b/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
@@ -7,6 +7,7 @@ upstream django {
 ## configuration of the server
 server {
     server_name physionet.org physionet-production.ecg.mit.edu;
+
     charset     utf-8;
 
     # max upload size
--- a/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
+++ b/deploy/production/etc/nginx/sites-available/physionet_nginx.conf
@@ -112,7 +112,7 @@ server {
         }
     }
 
-    listen 443 ssl; # managed by Certbot
+    listen 443 ssl default_server;
     ssl_certificate /etc/letsencrypt/live/alpha.physionet.org/fullchain.pem; # managed by Certbot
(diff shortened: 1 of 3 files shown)


tompollard:
@elfeto look okay to you? looks good to me, but not my area of expertise!

elfeto:
Not to be picky, but why one filename has a hyphen and the other an underscore?
Why did you add the ACME to the physionet-redirect.conf since is in the default config?

Apart from that, it's all good with me.

bemoody:
> Not to be picky, but why one filename has a hyphen and the other an
> underscore?

No reason at all.

> Why did you add the ACME to the physionet-redirect.conf since is in the
> default config?

It's handled in the default config for HTTP, not for HTTPS.

To verify the server, the CA needs to fetch a resource from
"http://physionet.mit.edu".  It's not totally clear to me, from
skimming RFC 8555, whether the verification *would* work if you have
"http://physionet.mit.edu/.well-known" redirect to
"https://physionet.org/.well-known".  The RFC does say that the
challenge "must be completed over HTTP, not HTTPS", but also says that
the CA "SHOULD" (not "MUST") follow redirections.  It then goes on to
talk about the security risks of doing that.

In any case, redirecting to "https://physionet.org" doesn't work if
for any reason you don't already have a certificate that the CA
recognizes.  So, better to limit the moving pieces here!

elfeto:
What I mean is that both the default config for HTTP and the redirect have the ACME and both have a redirect. 
As for the previous configuration, it just did a redirect without doing the ACME for "http://physionet.mit.edu/.well-known" and it appears the redirect to work well and the certificates granted.

Since in the RFC says SHOULD, it doesn't provide much confidence, so I guess having it repeated just in case wont do any harm.

PR #1119, opened 2020-08-20
Allow monitoring active uwsgi requests

bemoody:
A few times, I've logged in and noticed that some uwsgi process seemed to be running for an unusually long time (> 5 minutes.)  Generally a single request shouldn't ever take that long, and if it were multiple requests you would expect them to be distributed between multiple uwsgi workers.

But even if you see this happening, it's difficult to see what exactly is going on, because the request won't be *logged* (either by uwsgi or nginx) until after it finishes.

So here, we store the request path in the process command line, using the (undocumented?) uwsgi.setprocname function.  The path will then appear in the output of 'ps ax' or 'top -c'.

### DIFF
--- a/physionet-django/physionet/wsgi.py
+++ b/physionet-django/physionet/wsgi.py
@@ -9,8 +9,26 @@ https://docs.djangoproject.com/en/1.11/howto/deployment/wsgi/
 
 import os
 
+from django.core.signals import request_started, request_finished
 from django.core.wsgi import get_wsgi_application
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "physionet.settings.development")
 
 application = get_wsgi_application()
+
+# If we are running under uWSGI, then store the current request path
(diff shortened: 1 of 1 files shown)


elfeto:
I tried tested this in staging with a dummy 1gb file using `fallocate`, but all I could see in `ps ax` was the following:
```
www-data 18406  0.2  1.0 156460 83988 ?        S    15:19   0:00 uwsgi (idle)
```
If I used `top -c` a nginx worker would spwan
```
12454 www-data  20   0   83252  10620   7836 S   4.3   0.1   0:43.92 nginx: worker process                                                                                                                     
```
The same outcome for credentialed and public files, while the file was actually downloading.

I removed the file after the test and turned the config back to what it was.

bemoody:
Yeah, it's hard to see anything interesting when the system isn't under load. :)

Easiest way I found to test it was to stick in some dummy "sleep" statements.

PR #1121, opened 2020-08-20
Adds missing information on legacy author

### LINKED ISSUE #1120: Adding an author to a legacy project breaks project home
There is an error when an author is assigned via the console to a legacy project.

The author is successfully added but the object is created incomplete, and when the author goes to the project home it breaks with the following error:
```
DoesNotExist at /projects/
PublishedAuthor matching query does not exist.
```

The problem is that there are some fields missing (Defaulted to False or None)

elfeto:
Here I add the missing fields to the legacy author assignment.

Note: This is different than the contact information shown in the project page.

Fixes #1120

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -450,7 +450,7 @@ class PublishedProjectContactForm(forms.ModelForm):
 
 class CreateLegacyAuthorForm(forms.ModelForm):
     """
-    Merge two user accounts
+    Create an author for a legacy project.
     """
     author = forms.ModelChoiceField(
         queryset=User.objects.filter(is_active=True).order_by('username'),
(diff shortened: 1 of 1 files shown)


physionet-django/console/forms.py
```diff
@@ -482,13 +482,18 @@ def save(self):
         affiliation = super().save(commit=False)
         display_order = self.project.authors.count() + 1
         submitting = False
+        is_corresponding = False
+        corresponding_email = None
         if display_order == 1:
             submitting = True
+            is_corresponding = True
+            corresponding_email = author.email
```
tompollard, line 490: `corresponding_email` doesn't seem to get saved anywhere?

elfeto: Good catch :D I just added it to the published author creation

tompollard:
looks good to me, thanks!

PR #1122, opened 2020-08-24
CSV parsing infinite loop

bemoody:
If a CSV file is missing a line feed at the end, CSVFileView goes into an infinite loop.

I don't want to encourage authors to release such files, since they make life difficult for people trying to process the data... but anyway, the site shouldn't crash or freeze when attempting to display such a file. :)


tompollard:
good catch, looks good to me.

PR #1124, opened 2020-08-25
Known reference page

elfeto:
This page will show ALL known references ordered by contact date.
At the moment this is NOT a unique list since more than one applicant can have the same reference.

This was requested by KP via email to create something similar as what he had in the previos version on PN.
At the moment this will be used to remove the `known_ref` tag that is added when a reference is marked as contacted.

This will just set the `reference_contact_datetime` to None. 
AKA not known because it has not been contacted before (assuming there isn't another application where he has been contacted.)


physionet-django/console/views.py
```diff
@@ -1493,3 +1522,32 @@ def get_queryset(self):
             qs = qs.filter(username__icontains=self.q)
 
         return qs
+
+@login_required
+@user_passes_test(is_admin, redirect_field_name='project_home')
+def known_referencess(request):
```
tompollard: is the extra "s" in `known_referencess` intentional? it's used in the view and the template name.

physionet-django/console/views.py
```diff
@@ -1493,3 +1522,32 @@ def get_queryset(self):
...
+    """
+    List all known references witht he option of removing the contact date
+    """
+    user = request.user
+
+    if 'remove_known_ref' in request.POST and \
+     request.POST['remove_known_ref'].isdigit():
```
tompollard: this is formatted oddly (there is a single space on the second line, and i think it is best to avoid `\` to break lines. wrapping the clause with `(` .. `)` would work and i think conforms to pep8.

physionet-django/user/models.py
```diff
@@ -827,6 +827,13 @@ def revoke(self):
         logger.info('Credentialing for user {0} has been removed.'.format(
             self.user.email))
 
+    def remove_contact_reference(self):
+        """
+        Mark a reference as not contacted making the reference for this instance
```
tompollard: Could this be reworded slightly to:

``` 
Remove the date that indicates when the reference was contacted. Note that this may also affect the "known" status of the reference.
```

physionet-django/console/views.py
```diff
@@ -929,6 +929,35 @@ def users_search(request, group):
...
+
+        all_known_ref = applications.exclude(
+            reference_contact_datetime__isnull=True).order_by(
+            '-reference_contact_datetime')
+
+        if len(applications) == 0:
+            all_known_ref = paginate(request, applications, 50)
+
+        return render(request, 'console/application_list.html', {
+            'all_known_ref':all_known_ref})
```
tompollard: nitpick but pep8 wants a space after the `:`

physionet-django/console/templates/console/console_navbar.html
```diff
@@ -82,6 +82,9 @@
           <li class="nav-item {% if past_credentials_nav %}active{% endif %} ">
             <a id="nav_past_credential_applications" href="{% url 'past_credential_applications' 'successful' %}">Past Applications</a>
           </li>
+          <li class="nav-item {% if known_ref_nav %}active{% endif %} ">
+            <a id="nav_known_ref" href="{% url 'known_referencess' %}">Known references</a>
```
tompollard: Nitpick but the other items in this particular list are in title case, so it should perhaps be `Known References` for consistency. We are generally pretty inconsistent about casing and there is a mix in the menu.

physionet-django/console/templates/console/application_list.html
```diff
@@ -0,0 +1,34 @@
+<div id="searchitems">
```
tompollard: "application_list.html" doesn't give much information about the role of the template. It would be good if the name made it clear that this was a list of credentialing applications with known references.

physionet-django/console/templates/console/application_list.html
```diff
@@ -0,0 +1,34 @@
...
+        <td>{{ application.reference_contact_datetime | date }}</td>
+        <td>{{ application.reference_email }}</td>
+        <td>{{ application.reference_name }}</td>
+        <td>
+            <a href="{% url 'view_credential_application' application.slug %}" target="_blank">{{ application.id }} -- {{ application.user.get_full_name }}</a>
+        </td>
+        <td>
+            <form action="" method="post">
+            {% csrf_token %}
+              <button  name="remove_known_ref" value="{{ application.id }}" type="submit">
```
tompollard: the style of the button isn't consistent with the button style used elsewhere on the site (e.g. at http://localhost:8000/console/storage-requests/). Is this intentional? I think it would be best to be consistent if possible.

tompollard (commented):
Thanks Felipe, this looks good to me. I've made a few minor comments. Please could you respond to them before we merge?

Ideally "known reference" should be independent of whether or not an application has been contacted, so the logic is a bit of a kludge as we've discussed.

If we introduce a cleaner way of managing "known references" in future (e.g. tracking references in a new model) then this approach could be modified fairly easily, so I think this is a good stepping stone.

physionet-django/console/views.py
elfeto: Fixed in next commit :)

physionet-django/console/views.py
elfeto: For some reason my pep8 lint is broken... will have to see why...

physionet-django/user/models.py
elfeto: Fixed in next commit :)

physionet-django/console/views.py
elfeto: Fixed in next commit :)

physionet-django/console/templates/console/console_navbar.html
elfeto: Fixed in next commit :)

physionet-django/console/templates/console/application_list.html
elfeto: I changed the name of the file to known_references_list.html, that good ?

physionet-django/console/templates/console/application_list.html
elfeto: It is intensional, since KP prefers styles buttons. (http://localhost:8000/console/complete-credential-applications/)

I can add style if u want tho....

physionet-django/console/templates/console/application_list.html
tompollard: that sounds good, thanks!

physionet-django/console/templates/console/application_list.html
tompollard: Ok, let's leave it like this. I think we do want consistency, but looking around the site I don't think we are particularly consistent anyway, so this can be addressed independently.

physionet-django/console/views.py
elfeto: If using the (...) by pep8 that I fixed it I would have the following:
```
    if ('remove_known_ref' in request.POST and
       request.POST['remove_known_ref'].isdigit()):
        try:
```
VS 
```
    if 'remove_known_ref' in request.POST and \
       request.POST['remove_known_ref'].isdigit():
        try:
```
is the same really in terms of spaces, but I think we use more the \ than the (...) so I will add more spaces

physionet-django/console/templates/console/known_referencess.html
```diff
@@ -0,0 +1,27 @@
...
+{% block title %}PhysioNet Users{% endblock %}
+
+{% block local_css %}
+<link rel="stylesheet" type="text/css" href="{% static 'custom/css/pagination.css' %}">
+{% endblock %}
+
+{% block content %}
+<script src="{% static 'custom/js/cookie.js' %}"></script>
+<script src="{% static 'project/js/dynamic-formset.js' %}"></script>
+<script src="{% static 'console/js/search-console.js' %}"></script>
```
Lucas-Mc: Should these be in `{% block local_js_top %}` or does it not really matter? Only thing I can find, otherwise looks good! 👍

elfeto: I would say the search and cookie yes because of the search done, but the formset is prob not needed. I will remove it

tompollard:
looks good to me, thanks felipe.

PR #1125, opened 2020-08-25
Fixing typo and known ref search

elfeto:
Pagination for known ref is "broken" because of a variable typo, here I set the correct one.

The same happened in the search function.


tompollard:
thanks

Issue #1096, opened 2020-07-07
PhysioNet e-mail sent to spam folder


jraffa:
Received a non-critical e-mail about a physionet project that ended up in my spam folder.  Looks like it might be a known issue, but thought I'd fwd headers.  Only noticed because I was waiting for a different e-mail today, and gave that folder a look.

Maybe DKIM would help?

```
Received: from oc11expo22.exchange.mit.edu (18.9.4.84) by
 oc11expo22.exchange.mit.edu (18.9.4.84) with Microsoft SMTP Server (TLS) id
 15.0.1365.1 via Mailbox Transport; Tue, 30 Jun 2020 16:26:42 -0400
Received: from oc11exhyb6.exchange.mit.edu (18.9.1.111) by
 oc11expo22.exchange.mit.edu (18.9.4.84) with Microsoft SMTP Server (TLS) id
 15.0.1365.1; Tue, 30 Jun 2020 16:26:41 -0400
Received: from NAM02-CY1-obe.outbound.protection.outlook.com (104.47.37.53) by
 oc11exhyb6.exchange.mit.edu (18.9.1.111) with Microsoft SMTP Server (TLS) id
 15.0.1395.4 via Frontend Transport; Tue, 30 Jun 2020 16:26:41 -0400
Received: from CO2PR04CA0055.namprd04.prod.outlook.com (2603:10b6:102:1::23)
 by DM6PR01MB4476.prod.exchangelabs.com (2603:10b6:5:7a::20) with Microsoft
 SMTP Server (version=TLS1_2, cipher=TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384) id
 15.20.3131.25; Tue, 30 Jun 2020 20:26:40 +0000
Received: from CO1NAM03FT018.eop-NAM03.prod.protection.outlook.com
 (2603:10b6:102:1:cafe::cc) by CO2PR04CA0055.outlook.office365.com
 (2603:10b6:102:1::23) with Microsoft SMTP Server (version=TLS1_2,
 cipher=TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384) id 15.20.3131.21 via Frontend
 Transport; Tue, 30 Jun 2020 20:26:40 +0000
Authentication-Results: spf=pass (sender IP is 128.30.29.50)
 smtp.mailfrom=physionet.org; mit.edu; dkim=none (message not signed)
 header.d=none;mit.edu; dmarc=bestguesspass action=none
[...]

PR #1126, opened 2020-09-03
Remove contact form (temporarily?) and add a contact email address. May help to address #1096.

tompollard:
We currently display a contact form at: https://physionet.org/about/#contact. 

Displaying the form may contribute to the issue described in #1096 (physionet emails being labelled as spam), because spammers send us messages using the form with our address as the sender.

This pull request removes the form. It is intended to be a temporary fix while we investigate other solutions for managing emails sent by the form.

### DIFF
--- a/physionet-django/templates/about/contact.html
+++ b/physionet-django/templates/about/contact.html
@@ -1,31 +1,8 @@
 <section id="form">
   <div>
-    <form method="post" action="#contact">
-      {% csrf_token %}
 
-      {% for field in contact_form.visible_fields %}
-      <div class="row">
-        <div class="col-md-12">
-          <div class="form-group">
-            {{ field }}
-            {% for error in field.errors %}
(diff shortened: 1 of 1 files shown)


bemoody:
Seems fine to me.  This leaves the backend in place, but that's
probably fine when you just want to defeat dumb robots.  It would be
unpleasant if somebody was in the process of writing a message, then
submitted the form and nothing happened.

tompollard:
> This leaves the backend in place, but that's probably fine when you just want to defeat dumb robots.

Let me know if you'd like to remove the backend too. I figured this was just temporary so we could keep it, but I'm happy to remove.

> It would be unpleasant if somebody was in the process of writing a message, then submitted the form and nothing happened.

We could set the site to maintenance at a pre-specified time, but this seems like a minor issue. Annoying for one or two people at most!

PR #1127, opened 2020-09-04
Adds accessibility link #1123

### LINKED ISSUE #1123: add accessibility link
> every MIT.edu public web page (with limited exceptions) must have an Accessibility link by October 12, 2020. If you have footers on your site, consider adding the Accessibility link there. Complying is simple: the only thing you need to add is the word "Accessibility" and have it link to http://accessibility.mit.edu.

> So, all of you with lab/research pages, etc. – anything the public can access – must add an Accessibility link as described above.

Lucas-Mc:
Adds the accessibility link required by new MIT policy to the footer. This approach provides a very short description and adds a hyperlink to MIT's accessibility page. Fixes #1123.


tompollard:
Thanks, looks good to me. I'm not sure that the "more" is needed because we don't really provide accessibility options at the moment I don't think?. This can always be tweaked later if someone feels strongly about it!

PR #1140, opened 2020-10-14
Add "graduate student" category to the form used for requesting credentialed access

tompollard:
When a user is submitting a request for credentialed access, they are asked to self-assign themselves to one of the following categories:

```
    RESEARCHER_CATEGORIES = (
        (0, 'Student'),
        (1, 'Postdoc'),
        (2, 'Academic Researcher'),
        (3, 'Hospital Researcher'),
        (4, 'Industry Researcher'),
        (5, 'Government Researcher'),
        (6, 'Independent Researcher'),
    )
```

KP has asked for an additional 'Graduate Student' category to be added. This pull request adds the category with the same form behaviour as 'Student' (i.e. a supervisor is required as a reference). 

RESEARCHER_CATEGORIES is now a list to support ordering, and Graduate Student appears between Student and Postdoc (it would have been nice if the sequence 0-7 was spaced to allow insertions within the order).


bemoody:
"Student" vs "Graduate Student" is a little odd, but whatever.

What is the distinction between list and tuple?  That's not obvious.

Please explain your changes in your commit message, not just in your
pull request message.

tompollard:
> "Student" vs "Graduate Student" is a little odd, but whatever.

I agree, let me know if you'd like this changed in someway. Personally I would also remove "Postdoc" because I don't see it being different from "academic researcher". 

There are several things I'd like to change about the form/process really, but this didn't seem like the place to do it. 

> What is the distinction between list and tuple?

It was a totally unnecessary change which I've now reverted (lists are mutable, tuples are not).

> Please explain your changes in your commit message, not just in your pull request message.

Done!

PR #1147, opened 2020-10-29
Group database index by access policy. Relates to #1078.

tompollard:
KP would like a link that displays all projects requiring a Data Use Agreement. The database index (https://physionet.org/about/database/) seems like a good place to support this request. Relates to #1078.

After this change, http://physionet.org/about/database/#open would take users to a list of projects that require credentialed access.  

Groups are split by access type in the view and then rendered in the appropriate template. If the list of databases is empty for an access type, "No databases available." is displayed. e.g.

Screen Shot 2020-10-29 at 13 03 14

### DIFF
--- a/physionet-django/physionet/views.py
+++ b/physionet-django/physionet/views.py
@@ -127,10 +127,16 @@ def database_overview(request):
     """
     Temporary content overview
     """
-    all_projects = PublishedProject.objects.filter(
-        resource_type=0, is_latest_version=True).order_by(Lower('title'))
+    projects = {}
+    for i, policy in Metadata.ACCESS_POLICIES:
+        projects[i] = {}
+        projects[i]['policy'] = policy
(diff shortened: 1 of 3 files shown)


bemoody:
> physionet-django/physionet/views.py

"data" is not a good name for the template parameter.  "projects"
would be fine, or "projects_by_policy" or something similarly
descriptive.

> physionet-django/templates/about/database_content.html

Please replace tabs with spaces (lines 23-24) and remove the two blank
lines at the end of the file.

tompollard:
Thanks, those issues should now be addressed.

> "data" is not a good name for the template parameter.  "projects" would be fine, or "projects_by_policy" or something similarly descriptive.

I switched "data" to "projects".

> Please replace tabs with spaces (lines 23-24) and remove the two blank lines at the end of the file.

Great catch. Odd, I thought my editor was set to replace tabs with spaces.

PR #1151, opened 2020-11-02
Update nginx configuration for new certificate

bemoody:
I've generated the new certificate, but since I specified "physionet.org" as the first hostname in the list, certbot installed the files in /etc/letsencrypt/physionet.org instead of /etc/letsencrypt/alpha.physionet.org where it was before.  So this updates the nginx configuration to match.


tompollard:
Thanks, and it's good to be moving on from the alpha!

tompollard:
(pushed to live server).

PR #1153, opened 2020-11-07
Only display public profile for active accounts.

tompollard:
Currently we create a profile page for all user accounts, even those that are not yet active. This changes adds a requirement for the account to be active, because the majority of inactive accounts are spam.


Lucas-Mc:
Good idea! Merging!

PR #1154, opened 2020-11-07
Add management page for users to the console

tompollard:
We often have to use the shell to answer questions about individual users (for example, "which projects are associated with user x"?). This change adds an individual profile page for each user to the console (similar to the public profile page).

The page is intended to be a starting point for a user management tool that might be used for tasks such as:
- deactivating/deleting accounts
- checking which projects are associated with a user
- updating or merging accounts
- etc.

The page is accessed by finding the relevant user in the search tool (http://127.0.0.1:8000/console/users/all/) and then clicking the username.

Example profile:

Screen Shot 2020-11-07 at 04 24 15

### DIFF
--- a/physionet-django/console/templates/console/rejected_submissions.html
+++ b/physionet-django/console/templates/console/rejected_submissions.html
@@ -33,7 +33,11 @@
             <td>{{ project.title }}</td>
             <td><a href="{% url 'rejected_submission_history' project.slug %}">View History</a></td>
             <td>{{ project.version }}</td>
-            <td><a href="{% url 'public_profile' project.submitting_author.user.username %}">{{ project.submitting_author }}</td>
+            <td>{% if project.submitting_author %}
+              <a href="{% url 'user_management' project.submitting_author.user.username %}">{{ project.submitting_author }}</a>
+                {% else %}
+                  {{ project.submitting_author }}
+                {% endif %}</td>
             <td>{{ project.editor }}</td>
             <td>{{ project.creation_datetime|date }}</td>
             <td>{{ project.archive_datetime|date }}</td>
--- a/physionet-django/console/templates/console/user_management.html
+++ b/physionet-django/console/templates/console/user_management.html
@@ -1,0 +1,95 @@
+{% extends "console/base_console.html" %}
+
+{% load static %}
+
+{% block title %}User management{% endblock %}
+
+{% block content %}
+
+<div class="col-md-9 no-pd">
+  <h1>{{ subject.username }}</h1>
+  <hr>
+  <h3>Profile</h3>
+  <div class="row mb-1">
+    <div class="col-md-3">
+      Name:
+    </div>
+    <div class="col-md-9">
+      {{ subject.get_full_name }}
+    </div>
+  </div>
+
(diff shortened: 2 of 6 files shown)


Lucas-Mc:
I think this is a good idea and would help prevent possible mistakes while in the terminal... Looks good to me though I think @bemoody should see it also before it's merged.

bemoody:
Looks good.  Some small things:

    if User.objects.filter(username__iexact=username).exists():
        user = User.objects.get(username__iexact=username)
        ...
    else:
        raise Http404()

I think it's cleaner to do something like this:

    try:
        user = User.objects.get(username__iexact=username)
    except User.DoesNotExist:
        raise Http404()

although there are other patterns you could use, to avoid writing the
same query twice.

"user.associated_emails.filter(is_public=True).first()" looks
confusing.  I guess that a User is allowed to have exactly zero or one
public email address... but this ought to be wrapped up in a method or
property of the User model.

(It is maybe confusing that somebody submitting a project can have a
"primary" email address that will receive notifications, a
"corresponding" email address that will be shown on the published
project, and a "public" email address that will be shown on their
profile, and if I understand correctly, these might all be different.)

It seems like it would be better to take the part of
console/templates/console/user_management.html that is shared with
user/templates/user/public_profile.html, and factor that out into a
separate template file.

tompollard:
> I think it's cleaner to do something like this:

    try:
        user = User.objects.get(username__iexact=username)
    except User.DoesNotExist:
        raise Http404()

I agree that's cleaner. Maybe even better is to use the "get_or_404" functions that Django now provides: https://docs.djangoproject.com/en/3.1/topics/http/shortcuts/#get-object-or-404 so I'll see whether one of these is suitable.

> "user.associated_emails.filter(is_public=True).first()" looks confusing.  

I agree, this isn't very nice. I'll take a look at something cleaner. I'd basically copied code from the public profile page and intended this to be just a starting template. For the console, I think we would want to display all email addresses, not just public ones.

 > It seems like it would be better to take the part of console/templates/console/user_management.html that is shared with user/templates/user/public_profile.html, and factor that out into a separate template file.

This is a good point, but I would prefer to save this for later if it looks like there is enough overlap. I copied the public page as a template, but I'd intended the internal page to end up looking very different after we've added everything we want.

tompollard:
@bemoody I have done my best to respond to the points above. 

Instead of the `try .. except ...` block I used `get_object_404`. I'm now displaying all email addresses for the user, rather than just primary.

I haven't done anything about refactoring the public_profile HTML template, because as mentioned I think the layout will significantly diverge.

bemoody:
This doesn't work for active projects.  If I load the demo data, and
go to /console/user/manage/aewj/, I get

NoReverseMatch at /console/user/manage/aewj/

Reverse for 'published_project' with arguments
'('omAiNz9yC3MXvZTRe9sF', '')' not found. 1 pattern(s) tried:
['content/(?P<project_slug>[^/]+)/(?P<version>[^/]+)/$']

Probably we want to link to the submission info page.

tompollard:
Good catch, my bad. The projects should now link to a relevant page for their status (either a management page for the project or a submission info page).

Issue #1156, opened 2020-11-10
Display profiler console when in the development environment


tompollard:
[development/base.py](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/settings/development/base.py) includes a reference to a debug toolbar (`MIDDLEWARE += ['debug_toolbar.middleware.DebugToolbarMiddleware', ]`). 

We should update the dev environment to display the toolbar, unless anyone can see a reason not to do so.

It looks like displaying the toolbar requires the following chunk to be added to [base.py](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/settings/development/base.py):

```
INTERNAL_IPS = [
    '127.0.0.1',
]
```

...and [urls.py ](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/urls.py) needs to reference the toolbar URL:

```
import debug_toolbar

urlpatterns = [
    # ...
    path('__debug__/', include(debug_toolbar.urls)),
    # ...
```

Once this is done, the toolbar should appear:

![Screen Shot 2020-11-10 at 13 26 22](https://user-images.githubusercontent.com/822601/98715572-59770a80-2358-11eb-83e0-4191d0b453fb.png)

PR #1166, opened 2020-11-24
Add AJAX function to load less text in the credentialing page

elfeto:
Here I remove the templatetags that automatically generates the mailto links to
all the applicants, replacing it by an AJAX that will make a GET request once
the link is wanted in an attempt to load the page faster.


elfeto:
To test this, I disabled cache and checked the load speed using the browser network tab under "Inspect" and it was cut somewhat in half. Hopefully live will be better since the text loaded will differ.

This is using templatetags:
Normal png

This is using AJAX:
Ajax

tompollard:
Looks great, thanks Felipe. I was also going to say that the load time is roughly halved. Using the profiler described at: https://github.com/MIT-LCP/physionet-build/issues/1156:

Before:

Resource	| Value
--- | ---
User CPU time | 5295.893 msec
System CPU time | 3547.438 msec
Total CPU time | 8843.331 msec
Elapsed time | 8626.910 msec

After:

Resource	| Value
--- | ---
User CPU time | 2660.425 msec
System CPU time | 1805.672 msec
Total CPU time | 4466.097 msec
Elapsed time | 4284.539 msec

PR #1169, opened 2020-12-03
Undoes commit 180fc05 altering cred known ref sort

elfeto:
Here I am resetting the code as it was before the commit.
There is room for improvement but first we have to get it working.

Using the profiler described at: #1156:

Using Template tags broken:
Resource	| Value
--- | ---
User CPU time | 5295.893 msec
System CPU time | 3547.438 msec
Total CPU time | 8843.331 msec
Elapsed time | 8626.910 msec

Current broken:

Resource	| Value
--- | ---
User CPU time | 2660.425 msec
System CPU time | 1805.672 msec
Total CPU time | 4466.097 msec
Elapsed time | 4284.539 msec

Now:
Resource | Value
-- | --
User CPU time | 3790.943 msec
System CPU time | 1331.818 msec
Total CPU time | 5122.761 msec
Elapsed time | 4713.469 msec

### DIFF
--- a/physionet-django/console/templates/console/complete_application_display.html
+++ b/physionet-django/console/templates/console/complete_application_display.html
@@ -1,5 +1,3 @@
-{% load console_templatetags %}
-
 <div class="card mb-4" style="color: {{ color }}">
   <div class="card-header" id="application_{{ application.id }}">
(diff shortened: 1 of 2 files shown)


tompollard:
Let me take a quick look at this tonight, before completely undoing the commit. If I can't work out what's wrong, then I'll go ahead and merge this.

PR #1172, opened 2020-12-06
Display a list of signed data use agreements in the user profile menu

tompollard:
Users should have a straightforward way of viewing the data use agreements that they have signed. This change adds a new "Agreements" tab to the user profile, which lists all agreements that have been signed by the user. The list also links to printable copies of the agreements (added in response to a recent email request).

Example of the printable agreement:

Screen Shot 2020-12-05 at 23 58 25

There are improvements that could be made in the way DUAs are managed in the database (better logging, clearer provenance, etc), so this should be looked into at some point.

### DIFF
--- a/physionet-django/static/custom/css/settings.css
+++ b/physionet-django/static/custom/css/settings.css
@@ -102,3 +102,47 @@ input::placeholder {
     margin-top: -25px;
   }
 }
+
+/*
+ * Data Use Agreement
+ * --------------------------------------------------
+ */
+
+#agreement {
+  margin: 10px 10px 10px 10px;
+  font-family: Georgia, serif;
+  text-align: center;
+}
+
+#agreement p, #agreement ol {
+  text-align: left;
+}
+
+#outer-border {
+  width: 850px;
+  height: 950px;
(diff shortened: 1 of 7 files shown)


Lucas-Mc:
Hey @tompollard, I think this looks good and works in my testing though I just wanted ask whether or not you think the PhysioNet base HTML should be wrapped around the signed DUA after it's requested to be viewed? They could still print it though that might mess up what gets printed.. anyway, just my thoughts!

tompollard:
Thanks for taking a look!

> I just wanted ask whether or not you think the PhysioNet base HTML should be wrapped around the signed DUA after it's requested to be viewed?

Are you saying that you think that the agreement should be displayed within the standard HTML template (i.e. the template that includes the header, footer, menu, etc)? I agree this is worth considering, and we could definitely set things up so that clicking "print" ignores everything except the DUA itself.

For this pull request, I went for something quick and simple. I like that displaying the DUA by itself gives a kind of formal document feel, but there are definitely improvements that could be made (in terms of style, and various issues around the underlying data model).

PR #1176, opened 2020-12-23
Adds new demo project for waveforms

Lucas-Mc:
This change adds a new demo project that will be used to test future waveform viewers (separate pull request coming soon).   It includes 4 separate records (waveforms): 1 WAV format file (15 channels), 2 WFDB format files (2 and 10 channels respectively), and 1 EDF file (12 channels). One of the WFDB format files also includes some annotations which can also be used by the viewer to test its ability to render annotations. My hope was to include a wide variety of formats and number of channels to provide comprehensive testing for the viewer. Further files may be added in the future, but for now this should be sufficient while keeping the total storage allowance for the project low.


bemoody:
Sounds like a great idea.  wave_4.hea looks screwy though.  Is
wave_4.edf actually an EDF file?

If it *is* an EDF file then that header file is incorrect (it would
need to specify a byte offset for the signals) and you don't really
want a header file anyway, as WFDB can read the EDF header itself.

Lucas-Mc:
Thanks for the feedback @bemoody ! The file in question was interpolated from one of the WFDB-Python sample-data files so after I fix that issue I'll also update it in that repo too.

PR #1177, opened 2020-12-24
Adds new waveform viewer

Lucas-Mc:
This change adds a new waveform viewer which can be used alongside with LightWAVE. This is built using [Plotly](https://plotly.com) and [Dash](https://dash.plotly.com) which provides flexible and interactive plotting which can easily be modified with just a few lines of code. It also employs the WFDB-Python package for easy interactions with records in Python which can later be extended for more uses within PhysioNet. My proposed way to access this viewer is through a "Demo" link shown at the top of the LightWAVE home page where input records are chosen (http://localhost:8000/lightwave/demobsn/1.0?db=demobsn/1.0 for example). Also, testing for this viewer can be done after #1176 is implemented though `demobsn` should work also but is minimal.

I compiled a list of Pros and Cons for this viewer as compared to LightWAVE for those deciding which viewer to use.

Pros:
- Can be used on the development and staging environment
- Can be built separately easily to help authors with reviewing waveforms (no sandboxed version required)
- Displays all errors encountered (at least it should unless I missed a case)
- Auto-scaling of signals to prevent impossible to read signals (no external calibrations files: `wfdbcal`)
- Can pick a subset of signals
- Values on y-axis are shown
- Interactive hover / pan / zoom
- Can read and plot (*most*) annotations without `ANNOTATORS` file (displays warning if no `ANNOTATORS` file)
- Displays all EKG / ECG signals first, then BP, then the rest in alphabetical order
- Some waveforms can not be read by LightWAVE / take a long time to load
- Easier to switch between records using `Previous` and `Next` button

Cons:
- No play feature for automatic scrolling
- Only shows a maximum of a minute at a time due to performance concerns
- Markedly slower typically only when annotations are enabled
- ~~EDF is still a little sketchy (though almost fixed in WFDB-Python)~~ DONE: https://github.com/MIT-LCP/wfdb-python/pull/274

At the end of the day, I know there's probably some undiscovered errors here (it has been tested on all available databases on PhysioNet which is extensive though not comprehensive) which can be amended later though I think this is a strong start and an improvement for PhysioNet.


Lucas-Mc:
There is a conflict in the requirements for `pytz` which is being resolved here: https://github.com/MIT-LCP/wfdb-python/pull/275

requirements.txt
```diff
@@ -145,3 +149,82 @@ google-auth-httplib2==0.0.3 \
 psycopg2-binary==2.8.3 \
  --hash=sha256:d305313c5a9695f40c46294d4315ed3a07c7d2b55e48a9010dad7db7a66c8b7f\
  --hash=sha256:19d013e7b0817087517a4b3cab39c084d78898369e5c46258aab7be4f233d6a1
+
+# required for wfdb
+wfdb==3.2.0 \
+ --hash=sha256:0c28b2a74e4cb79209272d570b694f8509c54c9d4bea668804705576eb696823
+cycler==0.10.0 \
```
tompollard, line 173: There is no need to include packages that are dependencies of WFDB because they will be installed as needed. Removing the dependencies may help with the installation errors that you were getting.

Lucas-Mc:
This now should be rebased with the current `dev` branch so that the new demo project can be tested #1176

physionet-django/waveforms/dash_apps/finished_apps/waveform_vis.py
```diff
@@ -0,0 +1,872 @@
+# Project path configuration
+from django.conf import settings
```
tompollard: See https://www.python.org/dev/peps/pep-0008/#imports for PEP8 import convention. 

```
Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.
You should put a blank line between each group of imports.
```

physionet-django/waveforms/dash_apps/finished_apps/waveform_vis.py
```diff
@@ -0,0 +1,872 @@
...
+@app.callback(
+    [dash.dependencies.Output('dropdown_rec', 'options'),
+     dash.dependencies.Output('dropdown_rec', 'value'),
+     dash.dependencies.Output('error_text_rec', 'children')],
+    [dash.dependencies.Input('previous_annotation', 'n_clicks_timestamp'),
+     dash.dependencies.Input('next_annotation', 'n_clicks_timestamp'),
+     dash.dependencies.Input('set_slug', 'value')],
+    [dash.dependencies.State('set_record', 'value'),
+     dash.dependencies.State('set_version', 'value')])
+def get_records_options(click_previous, click_next, slug_value, record_value, version_value):
```
tompollard: Please add docstrings for functions that are missing them.

physionet-django/waveforms/dash_apps/finished_apps/waveform_vis.py
```diff
@@ -0,0 +1,872 @@
...
+    # Set the default error text
+    error_text = ['']
+    # Return default values if called without all the information
+    options_rec = []
+    return_record = None
+    if not slug_value and not version_value:
+        return options_rec, return_record, error_text
+
+    # Get the record file(s)
+    # TODO: Maybe make this more concrete
```
tompollard: It's okay including TODOs, I think, but if so then the task/meaning needs to be clear. If I'm reading this code in a year's time, I need to understand what is meant by "making this more concrete".

physionet-django/waveforms/dash_apps/finished_apps/waveform_vis.py
```diff
@@ -0,0 +1,872 @@
...
+PUBLIC_DBPATH = 'published-projects'
+# PUBLIC_ROOT: chroot directory for public databases
+if settings.STATIC_ROOT:
+    PUBLIC_ROOT = settings.STATIC_ROOT
+else:
+    PUBLIC_ROOT = os.path.join(settings.DEMO_FILE_ROOT, 'static')
+# All the project slug directories live here
+PROJECT_PATH = os.path.join(PUBLIC_ROOT, PUBLIC_DBPATH)
+# Formatting settings
+dropdown_width = '500px'
```
tompollard: It would be good to try avoid all of these global variables. Perhaps they could be dropped into a config/settings function?

physionet-django/waveforms/dash_apps/finished_apps/waveform_vis.py
```diff
@@ -0,0 +1,872 @@
...
+event_fontsize = '24px'
+# Maximum number of signals to display on the page
+max_display_sigs = 8
+# Set the error text font size and color
```
tompollard: I think it would be worth taking a look through these comments and thinking about whether they are necessary. In many cases they can be removed (e.g. it is clear from the code that `max_display_sigs = 8` is referring to the max number of signals, so the comment isn't necessary).

physionet-django/waveforms/dash_apps/finished_apps/waveform_vis.py
```diff
@@ -0,0 +1,872 @@
...
+    dcc.Input(id = 'set_version', type = 'hidden', value = ''),
+    dcc.Input(id = 'set_record', type = 'hidden', value = ''),
+])
+
+
+def window_signal(y_vals):
+    """
+    This uses the Coefficient of Variation (CV) approach to determine
+    significant changes in the signal then return the adjusted minimum
+    and maximum range... If a significant variation is signal is found
```
tompollard: Nitpicks, but no need for `...` and should be a `.` at the end of the final sentence. Perhaps not necessary here, but consider describing the args/output in the docstring.

physionet-django/waveforms/dash_apps/finished_apps/waveform_vis.py
```diff
@@ -0,0 +1,872 @@
...
+                config = plot_config
+            ),
+        ], style = {'display': 'inline-block'})
+    ], type = 'default'),
+    # Hidden div inside the app that stores the project slug, version, and record
+    dcc.Input(id = 'set_slug', type = 'hidden', value = ''),
+    dcc.Input(id = 'set_version', type = 'hidden', value = ''),
+    dcc.Input(id = 'set_record', type = 'hidden', value = ''),
+])
+
```
tompollard, line 135: PEP8 suggests one blank line between functions, two between classes.

physionet-django/waveforms/dash_apps/finished_apps/waveform_vis.py
```diff
@@ -0,0 +1,872 @@
...
+    try:
+        with open(records_file, 'r') as f:
+            all_records = f.read().splitlines()
+    except FileNotFoundError:
+        error_text.extend(['ERROR_REC: Record file not provided... {}'.format(records_file), html.Br()])
+        return options_rec, return_record, error_text
+    except Exception as e:
+        error_text.extend(['ERROR_REC: Record file incorrectly formatted... {}'.format(e), html.Br()])
+        return options_rec, return_record, error_text
+    # TODO: Probably should refactor this
```
tompollard: The spacing is a little inconsistent. Should be probably be an empty line here to separate the try sections. Same in a couple of other places.

physionet-django/waveforms/dash_apps/finished_apps/waveform_vis.py
```diff
@@ -0,0 +1,872 @@
...
+                'width': sig_thickness
+            },
+            'name': sig_name[r]
+        }), row = idx+1, col = 1)
+
+        # Display where the events are if any
+        if anns != [] and idx == 0:
+            for i,ann in enumerate(anns):
+                for a in anns_idx[i]:
+                    # Add only the label due to the hover feature... add
```
tompollard: Can this commented text be removed?

physionet-django/waveforms/dash_apps/finished_apps/waveform_vis.py
```diff
@@ -0,0 +1,872 @@
...
+    [dash.dependencies.Output('the_graph', 'figure'),
+     dash.dependencies.Output('error_text_graph', 'children')],
+    [dash.dependencies.Input('sig_name', 'value'),
+     dash.dependencies.Input('start_time', 'value'),
+     dash.dependencies.Input('annotation_status', 'value')],
+    [dash.dependencies.State('dropdown_rec', 'value'),
+     dash.dependencies.State('start_time', 'pattern'),
+     dash.dependencies.State('set_slug', 'value'),
+     dash.dependencies.State('set_version', 'value')])
+def update_graph(sig_name, start_time, annotation_status, dropdown_rec,
```
tompollard: This is a big function! Can it be broken down into easy to interpret sub-functions?

tompollard (commented):
The viewer looks good and it works nicely. The only real issue in terms of usability that I could see is that access/permissions need checking:

- currently, anyone (including anonymous users) can view waveforms for active (pre-publication) projects. Only admin and authors should be able to view at this stage. e.g. http://localhost:8000/waveforms/active_4riWnicqZRycqFZb6FUu/ can be viewed by an anonymous viewer
- are restricted/credentialed access restrictions applied, so that only authorised viewers can access the data?

The code could do with a clean up, especially physionet-django/waveforms/dash_apps/finished_apps/waveform_vis.py. 

I have added comments inline, but I think the main things are:

- try to follow PEP8 (install a linter if you don't have one!)
- add docstrings
- if adding TODOs, make sure that someone new to the code would understand what needs to be done and why
- there are a fair amount of comments that don't really add to readability, so could be removed.
- breaking down the big functions would make the code much easier to follow!

briangow:
This looks great!  Being able to select the signals you want to view is a nice feature.  I had a few thoughts around potential improvements, none of which should be required for an initial release of this waveform viewer:

- When saving to a .png, it may be nice to add a small title indicating the database and record number.
- When zooming in on the time axis the xlabel ticks disappear eventually, perhaps consider creating new finer grain ticks when zooming in.
- Autoscale and reset axis seem to do the same thing, which sounds like the expected behavior if a custom axes range isn't set (https://plotly.com/chart-studio-help/getting-to-know-the-plotly-modebar/).  Do you plan to make use of this custom axes option?  If not, is it possible to remove the reset axis button?
- Waveforms with annotations show the annotation at the top of the plot but since there isn't a line or dot that connects to the waveform itself, you can't tell exactly which sample the annotation applies to.  Consider adding a dot or line on the waveform to make this more precise.
- With 8 signals plotted the labels get very small and since they are also a gray color they become hard to read.  Consider making the label pure black, or bold.  You could also make the x-axis / time labels larger than the y-axis since there is space between ticks on the x-axis.  
- Consider adding a zoom out (and in) for the y-axis.  Some of the signals are clipped at the default zoom and users may want to see the peaks.

physionet-django/waveforms/dash_apps/finished_apps/waveform_vis.py
```diff
@@ -0,0 +1,801 @@
...
+    This uses the Coefficient of Variation (CV) approach to determine
+    significant changes in the signal then return the adjusted minimum
+    and maximum range... If a significant variation is signal is found
+    then filter out extrema using normal distribution
+    """
+    # Standard deviation signal range to window
+    std_range = 4
+    # Get parameters of the signal
+    temp_std = stats.median_absolute_deviation(y_vals, nan_policy='omit')
+    temp_std = np.nanstd(y_vals)
```
briangow: You assign `temp_std` to the median absolute deviation (MAD) and then overwrite it with standard deviation.  If you don't want to use MAD any longer, you can remove this line.  If you do want to use MAD, you may consider revising the comment above since Coefficient of Variation implies using standard deviation.

Lucas-Mc: Good catch! I thought I switched not sure how it got back... will fix!

Lucas-Mc:
Here's some initial thoughts in response @briangow :
> * When saving to a .png, it may be nice to add a small title indicating the database and record number.

Looks like I can edit the file name, but maybe not at the final level, only initially (maybe something like `demobsn.png`? hopefully I can get more specific) ... I'll have to take a look further

> * When zooming in on the time axis the xlabel ticks disappear eventually, perhaps consider creating new finer grain ticks when zooming in.

Hmmmm, might not be worth the effort since the x-axis ticks are set to be every 0.2 seconds and by then the downsampling would make it less than desirable... can you think of an application where I would need signals of that time scale?

> * Autoscale and reset axis seem to do the same thing, which sounds like the expected behavior if a custom axes range isn't set (https://plotly.com/chart-studio-help/getting-to-know-the-plotly-modebar/).  Do you plan to make use of this custom axes option?  If not, is it possible to remove the reset axis button?

This has been something on my to-do list for a while now but I can't seem to figure it out in the Python version only the JS version (this motivated me to seriously start considering a WFDB-JS if I have spare time)... It seems like a package specific issue that a lot of people have been complaining about but I think it's in the "we need more funding to get it done" stage unfortunately... Maybe I should just remove one of them for now so it's not confusing

> * Waveforms with annotations show the annotation at the top of the plot but since there isn't a line or dot that connects to the waveform itself, you can't tell exactly which sample the annotation applies to.  Consider adding a dot or line on the waveform to make this more precise.

This functionality was in the big commented blob in the annotation generator in `waveform_vis.py : def update_graph()` but I removed it since it seemed to be much slower for a large number of annotations and I figured the user could just hover over the graph and that vertical dashed line would show up and it lines it up while giving them specific values at the same time... I'll keep it on my radar though!

> * With 8 signals plotted the labels get very small and since they are also a gray color they become hard to read.  Consider making the label pure black, or bold.  You could also make the x-axis / time labels larger than the y-axis since there is space between ticks on the x-axis.

I see now... It looks like the color is set to `rgb(43,63,95)` by default so I'll look into changing it... I never noticed the x-axis labels getting smaller like that I think I accidentally made them a function of `n_sig` just like I did with the y-axis font size too so I'll switch that [**FIXED**]

> * Consider adding a zoom out (and in) for the y-axis.  Some of the signals are clipped at the default zoom and users may want to see the peaks.

This is a can of worms... How would I make it clean on a graph by graph basis if 8 signals are plotted at once? I think this would also make the manual zoom selection be a 2D rectangle with variable width and height instead of just width... Should I just relax the criteria for outliers or plot the whole signal with outliers and hope for the best?

briangow:
Just a few follow-up thoughts on some of the points @Lucas-Mc.  Again, I don't think any of these should be taken as necessary for the initial release of this waveform viewer.

- When zooming in on the time axis the xlabel ticks disappear eventually, perhaps consider creating new finer grain ticks when zooming in.
     - Hmmmm, might not be worth the effort since the x-axis ticks are set to be every 0.2 seconds and by then the downsampling would make it less than desirable... can you think of an application where I would need signals of that time
          - Are you setting the tick width based on the sampling rate?  That seems like a good idea.  Of course there are lots of signals where it'd be nice to have ticks closer than every 0.2 s.  EEG signals can be 30+ Hz for example.  Even with ECG's it could be convenient to have ticks a bit closer than 0.2 so you can quickly eyeball the time differences between the different peaks within a beat.
          
- Autoscale and reset axis seem to do the same thing, which sounds like the expected behavior if a custom axes range isn't set (https://plotly.com/chart-studio-help/getting-to-know-the-plotly-modebar/). Do you plan to make use of this custom axes option? If not, is it possible to remove the reset axis button?
     - This has been something on my to-do list for a while now but I can't seem to figure it out in the Python version only the JS version (this motivated me to seriously start considering a WFDB-JS if I have spare time)... It seems like a package specific issue that a lot of people have been complaining about but I think it's in the "we need more funding to get it done" stage unfortunately... Maybe I should just remove one of them for now so it's not confusing
          - Yeah, I think it would be better to remove the reset axes if it isn't going to be used to set custom axes. 

- Waveforms with annotations show the annotation at the top of the plot but since there isn't a line or dot that connects to the waveform itself, you can't tell exactly which sample the annotation applies to. Consider adding a dot or line on the waveform to make this more precise.
     - This functionality was in the big commented blob in the annotation generator in waveform_vis.py : def update_graph() but I removed it since it seemed to be much slower for a large number of annotations and I figured the user could just hover over the graph and that vertical dashed line would show up and it lines it up while giving them specific values at the same time... I'll keep it on my radar though!
          - I wonder if it's possible to do something where if the user clicks a certain spot (near annotation marker) or enters a certain key sequence that the annotation line will appear.  If it was on-demand, it wouldn't slow down plotting in general but could be accessed if needed. 

- Consider adding a zoom out (and in) for the y-axis. Some of the signals are clipped at the default zoom and users may want to see the peaks.
     - This is a can of worms... How would I make it clean on a graph by graph basis if 8 signals are plotted at once? I think this would also make the manual zoom selection be a 2D rectangle with variable width and height instead of just width... Should I just relax the criteria for outliers or plot the whole signal with outliers and hope for the best?
          - I wasn't proposing any changes to your default zoom, only the ability for the user to zoom out on the y-axis, especially for cases where the waveform is clipped.  I think it'd be fine to zoom out of all of the plotted waveforms by the same amount (percentage).

tompollard:
Just as a reminder, latest discussion around this pull request was asking what kind of load it would place on our servers. Needs some investigation!

tompollard:
closing this for now, but we should revisit later.

Issue #1185, opened 2021-01-06
Simple implementation of the credentialing workflow


tompollard:
We need to provide a framework for allowing the PhysioNet team to help process credentialing applications. This issue suggests a basic implementation that we could serve as a starting point.

1. Add a "Processing" page to the credentialing section of the console (https://physionet.org/console/credential-applications/). The template will be modeled on the project management tool (https://physionet.org/console/submitted-projects/).
2. Applications will have several statuses. (1) Initial review (2) Training checks (3) Personal information checks (4) Reference checks (5) [optional] Awaiting reference response (6) Final review and approval/not approved. 
3. Each status tab displays a list of people in that status. Clicking on a person will display a page with their profile. There will be a checklist (similar to the project review page) that needs to be checked to move the application to the next stage.

Notes:
- Do applications need to be assigned to a unique handler (in the same way as projects)? It would be good to avoid this if possible.
- Later issues can deal with enhancing this framework, e.g.: (a) Adding template emails and contact management; (b) Making the process less linear, e.g. allowing users to fix and issue with training without going back to the beginning of the process.

PR #1187, opened 2021-01-08
Adds "Processing" page under "Credentialing"

shulim:
This is the first of a series of changes to address #1185. To work toward a simple implementation of the credentialing workflow, I've added a "Processing" tab under the "Credentialing" dropdown menu (pictured below). For now, clicking on it goes to a page with a series of tabs across the top, similar to the project management tool, though the tabs are empty since the logic for displaying applications has not yet been implemented. I'd be open to feedback on what's currently here!

Screen Shot 2021-01-08 at 4 15 18 PM


Lucas-Mc:
This looks good and is a really great start! The next steps could be to slowly begin to try and fill out each section.. I can do "training checks" :)

Issue #1181, opened 2021-01-06
Add custom training for users. Ref #1179


tompollard:
Consider adding PhysioNet-specific training to the training section added in #1179. At a simple level this might involve:

- Creating a structured list of questions and answers (along with an example document that can be shared publicly, in a similar way to the environment files).
- Adding a form to the training page that asks a set of randomly sampled questions from the Q&A document. 

I have a list of questions that we could use as a starting point.

PR #1189, opened 2021-01-12
Adds Credentialing Training Page #1179 

### LINKED ISSUE #1179: Add a "Training" page to the user profile
Add a "Training" page to the user profile (https://physionet.org/settings/profile/). The training page should:

- provide guidance for taking the CITI training
- Include a form to upload certificate

The credentialing page (https://physionet.org/settings/credentialing/) will need to be updated to reference this new page. It should not be possible to submit an application without first completing the training step.

Lucas-Mc:
This change adds a training page for the credentialing application process. Currently, the training completion report and credentialing application are submitted at the same time. With this change, users will have to first submit a valid training completion report in order to gain access to the credentialing application form. This will help filter out early-stage failing credentialing applications as well as develop a foundation for creating our own custom training process in the future. See #1026 and #1181.

- User are allowed to submit multiple training completion reports before a pending credentialing application is in the queue, but are prevented once a credentialing application is submitted
- Training submission status will be returned if the application is revoked, withdrawn, or rejected. If the application is revoked the user must re-complete training
- Users are permitted to submit as many training completion reports as desired before submitting a credentialing application, though only the most recent submission will be used

Fixes #1179.


Lucas-Mc:
It seems like this test needs to be modified to handle the proposed new structure `user.test_integration.TestCredentialing.test_credential_application`

tompollard:
Thanks Lucas, I had a quick play around with this and I came across a few issues:

- If I navigate directly to the submission form (http://localhost:8000/credential-application/) without uploading my training report, then submit, I see the following error:

```
Traceback:

File "/Users/tompollard/projects/physionet-build/env/lib/python3.8/site-packages/django/core/handlers/exception.py" in inner

File "/Users/tompollard/projects/physionet-build/env/lib/python3.8/site-packages/django/core/handlers/base.py" in _get_response

File "/Users/tompollard/projects/physionet-build/env/lib/python3.8/site-packages/django/core/handlers/base.py" in _get_response

File "/Users/tompollard/projects/physionet-build/env/lib/python3.8/site-packages/django/contrib/auth/decorators.py" in _wrapped_view

File "/Users/tompollard/projects/physionet-build/physionet-django/user/views.py" in credential_application
  651.             application = form.save()

File "/Users/tompollard/projects/physionet-build/physionet-django/user/forms.py" in save
  575.         credential_application = CredentialApplication.objects.filter(

File "/Users/tompollard/projects/physionet-build/env/lib/python3.8/site-packages/django/db/models/query.py" in __getitem__
  309.         return qs._result_cache[0]

Exception Type: IndexError at /credential-application/
Exception Value: list index out of range
```

- If I withdraw a credentialing application and then upload a new training report, my application is mysteriously resubmitted. To reproduce:
    - login as a user with an open application (e.g. admin on a fresh db);
    - withdraw the application at http://localhost:8000/settings/credentialing/; 
    - upload a new training report at http://localhost:8000/settings/training/
    - go back to http://localhost:8000/settings/credentialing/

- I think the behaviour of the training page could be clearer. Sometimes it isn't possible to upload a new training report (the submit button is ghosted out) but a reason isn't given. I think this could end up confusing our users.

- A "CITI Completion Report" header is displayed in the credentialing form http://localhost:8000/credential-application/, but with no text under the header.

Lucas-Mc:
> * If I navigate directly to the submission form (http://localhost:8000/credential-application/) without uploading my training report, then submit, I see the following error:

I didn't think about the user having a link to the application already somehow, just the path to it, I'll try to fix that

> * If I withdraw a credentialing application and then upload a new training report, my application is mysteriously resubmitted. To reproduce:
>   
>   * login as a user with an open application (e.g. admin on a fresh db);
>   * withdraw the application at http://localhost:8000/settings/credentialing/;
>   * upload a new training report at http://localhost:8000/settings/training/
>   * go back to http://localhost:8000/settings/credentialing/

I hadn't noticed this before, but strange, let me see what's going on here

> * I think the behaviour of the training page could be clearer. Sometimes it isn't possible to upload a new training report (the submit button is ghosted out) but a reason isn't given. I think this could end up confusing our users.

The top part has text with a yellow question mark `You have a pending application. No further training is needed.`.... Should this be made more clear? How?

tompollard:
I just took another look at this and have added a few thoughts below:

1. I don't think the "Training Report Submission" header is necessary.
2. I'd suggest simplifying the opening text a little. Perhaps change this: 

> In order to become a credentialed PhysioNet user and access the restricted-access clinical databases, you must complete a suitable training program in human research subject protections and HIPAA regulations. We recommend the Collaborative Institutional Training Initiative (CITI) Program's “Data or Specimens Only Research” course.

to:

> You must complete the Collaborative Institutional Training Initiative (CITI) Program's “Data or Specimens Only Research” course to become a credentialed user. The course provides important guidance on human research and patient privacy.


3. The next paragraph could be put into an if statement with options like:
    - Please complete the CITI training and upload your completion report [+ upload button.]
    - We have received your training report. You may now submit an application for credentialed access or replace the training report if necessary. [+ replace button.]
    - We have received your training report and your application for credentialed access is under review. [No upload button.] 
    - We have received your training report and your application for credentialed access has been approved. [No upload button.] 

4. The yellow "?" icon to the left of "You have a pending application" is almost invisible on my screen. Using black would be consistent with other pages. The clock icon that is used on the credentialing page might be more appropriate for "Pending" than a question mark, or a bullet could also work.

5. Some text is duplicated, perhaps as a result of a separate pull request that moved help text (e.g. "Do NOT upload the completion certificate."; "Do not upload the completion certificate.")

Screenshot of the page proposed in this pull request for reference:

Screen Shot 2021-04-22 at 14 52 56

Lucas-Mc:
Thanks for the suggestions @tompollard! All are implemented now except for this one: 
> 3. Some text is duplicated, perhaps as a result of a separate pull request that moved help text (e.g. "Do NOT upload the completion certificate."; "Do not upload the completion certificate.")

I figured the duplicate text should be kept to emphasize our requirement that the user not upload the certificate. It's only stated twice: once right above the upload button and once in the CITI course instructions below. I did make them uniform, however, by removing the bold for one of them.

tompollard:
Thanks @Lucas-Mc:

- Has something strange happened to the commit history? Some of the new commits seem unrelated to this change.
- When I submit a training report, my (previously withdrawn) application for credentialed access is automatically submitted, which should not happen.
- As mentioned in my previous comment, could you hide the upload text, button etc when they are not needed?
- I don't think it is a good idea to repeat the same text twice in succession (e.g. "Do NOT upload the completion certificate." ... "Do not upload the completion certificate.)". We just need to present this point once, clearly.
- Didn't mention this before, but I think it would be cleaner to leave the CITI training instructions where they are now (http://localhost:8000/about/citi-course/) and instead link to these instructions.

Lucas-Mc:
Hey @tompollard, I fixed all of the issues you mentioned! Turns out there was a bigger issue at hand which I had to sort through about setting the correct submission status and doing the proper filtering.

Also, the commit history looks good to me, at least the same as it has been however strange it looks!

tompollard:
This change was incorporated in a later pull request: https://github.com/MIT-LCP/physionet-build/pull/1464

PR #1190, opened 2021-01-12
Adds credentialing application processing statuses

shulim:
To continue addressing #1185, these changes create submission statuses similar to those in project management (10: 'Initial Review', etc.). Each tab displays the ongoing applications with their respective status, and the Process page allows the user to approve the application's status to move to the next stage. Currently, the Process page only shows the box that corresponds to that stage (i.e. the "Initial Review" application Process pages only have the option to Approve Initial Review). The "Process Application" box that was there before now only appears in the Final Review stage.

Screen Shot 2021-01-12 at 5 21 30 PM
Screen Shot 2021-01-12 at 5 21 57 PM

I'll think more about the following:
- How the Accept and Reject statuses will play in (there's a preexisting 0, 1, 2, 3 status numbering system)
- If this single-status approach is acceptable, or if a non-linear approval process would be better
- What different information to display in the table view under each tab (currently all the tables are formatted the same way)

I'd also appreciate some feedback on what's here. Thank you!


Lucas-Mc:
This looks good! It appears like the time elapsed isn't being displayed and I think it's because the different application types aren't being assigned the `time_elapsed` attribute, see error here:
```
initial_applications = applications.filter(submission_status=10)
(Pdb) initial_applications[0].time_elapsed
*** AttributeError: 'CredentialApplication' object has no attribute 'time_elapsed'
```
Which is strange because dropping the debugger the step above shows it was assigned:
```
(Pdb) application.time_elapsed
19
(Pdb) [a.time_elapsed for a in applications]
[299, 299, 295, 295, 293, 292, 292, 292, 292, 292, 291, 291, 291, 291, 291, 291, 291, 291, 291, 291, 291, 291, 291, 290, 290, 290, 290, 290, 290, 290, 290, 290, 290, 290, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 289, 288, 288, 288, 287, 287, 287, 268, 237, 790, 317, 314, 311, 305, 280, 249, 243, 235, 214, 214, 214, 208, 206, 202, 201, 172, 157, 150, 134, 125, 122, 98, 92, 82, 70, 59, 58, 54, 32, 27, 26, 21, 19]
```
I think it have to be saved with `application.save()` after being changed but strangely this doesn't work either 🤔

tompollard:
`time_elapsed ` isn't something that can be called on`CredentialApplication` objects. 

It looks like there is a view that temporarily assigns this attribute at: https://github.com/MIT-LCP/physionet-build/blob/541ade3f8bbfc9fd5f3004650285b425242f1952/physionet-django/console/views.py#L1015

How about adding a `time_elapsed` method to the user.CredentialApplication model?

tompollard:
> I think it have to be saved with application.save() after being changed but strangely this doesn't work either 🤔

It doesn't work because the `application` is an object that maps to our database. There is no field called `time_elapsed` in our database, so we can't save it (and it wouldn't really make sense to save a dynamic value like this anyway).

Lucas-Mc:
> It doesn't work because the `application` is an object that maps to our database. There is no field called `time_elapsed` in our database, so we can't save it (and it wouldn't really make sense to save a dynamic value like this anyway).
>>
True! I didn't bother to check the model 👍 Seems like adding that attribute to the model would be the best way to go? Either that or pass the `time_elapsed` to the template as a variable and manipulate it there to display it for each application?

tompollard:
Either adding a method (rather than attribute, which would modify the database) to the model or dealing with this in the template would work. Not sure of the speed implications of each approach, but either is fine for now! [Update: you might be able to use this: https://docs.djangoproject.com/en/dev/ref/templates/builtins/#timesince]

shulim:
Thank you for the suggestions! I went with the approach of adding a method for time_elapsed to the CredentialApplication model. If we need to optimize for speed later and one of the other ways is faster, though, I can work on it!

PR #1191, opened 2021-01-13
Allow admin users to temporarily lock an account using the console

tompollard:
We occasionally need to remove a user, usually at the request of the user. It is unsafe doing this through the console, so it is helpful to have a tool available in the admin console. This pull request:

- adds a simple "deactivate_profile" method to the user model
- adds a "deactivate profile" button to the user management page in the console.

To test the functionality, click a username in: http://localhost:8000/console/users/active/ and then click the button on the user management page to deactivate the user.

One issue that I will address later (but not in this pull request for timeliness) is that a deactivated email address will appear publicly if the user is a corresponding author of a published project.

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -317,6 +317,13 @@ class DeprecateFilesForm(forms.Form):
     delete_files = forms.ChoiceField(choices=YES_NO)
 
 
+class DeactivateUserForm(forms.Form):
+    """
+    Deactivate a user profile from the management console.
+    """
+    deactivate_user_btn = forms.BooleanField()
+
+
 class ProcessCredentialForm(forms.ModelForm):
     """
     Form to respond to a credential application
--- a/physionet-django/console/templates/console/user_management.html
+++ b/physionet-django/console/templates/console/user_management.html
@@ -68,6 +68,16 @@
   {% endif %}
 
   <br />
+  <form action="" method="POST">
+    {% csrf_token %}
+    {% if subject.is_active %}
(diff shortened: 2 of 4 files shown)


bemoody:
First, what is it you actually *want*?  What does "deleting an account"
mean?  Your commit messages tell me what you're doing, they don't tell
me why.

You cannot simply set "is_active" to false without considering what
that means for the rest of the system.  In particular,
"is_active=False" means that the account is treated as "not yet
activated"... meaning that it CAN be activated.  Also,
"is_active=False" means the account will be purged when the activation
period expires.

tompollard:
> First, what is it you actually *want*?  What does "deleting an account"
mean?  Your commit messages tell me what you're doing, they don't tell
me why.

Firstly, I want a quick and simple way of "deactivating" an account. This means that the user can no longer login, and their account is hidden from the public. 

I started with the idea of deleting accounts (which is why you might see the word "delete" somewhere), but decided it would be safer to start with "deactivating".

> You cannot simply set "is_active" to false without considering what
that means for the rest of the system.

Yep, after some discussion with Felipe, I agree this is not a good idea. Might be easiest to discuss this, to work out exactly what we want.

tompollard:
After discussion, I think we agreed that we would like:

1) to provide all users with the ability to "delete" their account. This removes all personal information from the system. Care is needed to avoid causing data integrity issues, and to begin we may want to implement the functionality only for users who are not associated with project content.
2) [optional, this pull request] to provide admin users with the ability to temporarily "lock" an account. Locked accounts cannot be seen or accessed by non-admin users.

Issue #1183, opened 2021-01-06
Review and update credentialing form fields


tompollard:
We currently ask for very limited information in the credentialing form. Review the information that we ask for. Is it sufficient for our needs? Consider adding optional new fields (e.g. work email; office phone number; github account; researcher IDs like ORCID; NIH commons; etc).

PR #1193, opened 2021-01-20
Adds institutional e-mail to credential application #1183

Lucas-Mc:
Adds institutional e-mail to the credential application to provide more information which can be used to conduct personal checks for applicants. Examples of institutional e-mails include `@example.edu` and excludes `@gmail.com`, `@hotmail.com`, etc.  I am not sure if all institutional e-mails end in `.edu` so this check can be added later. Further checks can also be added to check for known non-institutional e-mails such as the ones described above. At the moment, this is only required for students, graduate students, and postdocs. Address #1183.


bemoody:
You're not doing anything to check that the email address actually
belongs to the person.  It seems to me like the obvious thing would be
to require them to actually add a .edu address to their PhysioNet
account.

> I am not sure if all institutional e-mails end in `.edu` so this
> check can be added later.

Wikipedia:
> Since 2001, new registrants for second-level domain names [under
> .edu] have been required to be United States–affiliated institutions
> of higher education.

So that wouldn't be enough.

Lucas-Mc:
I don't understand the logic here. The job of checking who it belongs to is up to the credential application reviewer; what stops them from putting the same e-mail address in their profile than in the application?

The reason it would be preferable in the application is so we can enforce that all student and postdoc applicants submit their institutional e-mail along with their application for extra identity checks in the review process. Why should it be in their profile? What are we going to do with it after it's added there? I certainly wouldn't want my institutional e-mail to also be my contact e-mail.

If it's only served for the purpose of credential applications, then why make it global? Further, we can't enforce users to submit their institutional e-mail from the Sign Up stage as not everyone has one since they don't specify they are a student or postdoc until the credential application. Also, if we make it optional, I know I wouldn't fill out optional fields during a Sign Up if I didn't have to.

bemoody:
A PhysioNet account can have many verified email addresses.  Requiring
an email address to be verified doesn't mean it needs to be their
primary/contact address.  It certainly doesn't need to be part of the
initial user registration form, and I agree that would not make sense.

What is it that you're trying to establish here?

(a) "PhysioNet user jsmith123 is John Smith"

(b) "John Smith is a student at Harvard University"

(c) "PhysioNet user jsmith123 is a student at Harvard University"

Which statement(s) are you trying to prove and how would this form field
be used to support that statement?

PR #1194, opened 2021-01-21
Adds intermediate credential processing forms

shulim:
These changes continue to address #1185 by adding credential review questions to the `CredentialApplication` model, so that there can be different forms displayed on a credential application's Processing page while it is in different stages. Every stage (e.g. Initial review, Training check) has its own form (e.g. `InitialCredentialForm`, `TrainingCredentialForm`), which is displayed conditionally based on the credential application's `review_status`. Currently, the forms need need to have all "Yes" answers to successfully approve an application to move on to the next stage, but in the future, I'll work on adding more nuances to some questions that may not be applicable to all applicants. Thanks for taking a look, and hope to hear some feedback!

Screen Shot 2021-01-20 at 6 24 53 PM


physionet-django/console/templates/console/process_credential_application.html
```diff
@@ -27,8 +27,38 @@ <h1>Process Credential Application - {{ app_user.username }}</h1>
...
+            {% if application.reference_response_datetime %}
+              <p><i class="fas fa-check" style="color:green"></i> The reference verified the applicant on {{ application.reference_response_datetime }}</p>
+              {% if application.reference_response_text %}
+              <h2 style='font-size:20px;'>Reference comments:</h2>
+              <p style='margin-left:2em'>{{ application.reference_response_text }}</p>
+              {% endif %}
+            {% endif %}
+        </div>
+      </div>
       <div class="col-sm-12">
```
tompollard: I think this whole div section (`<div class="col-sm-12">`) can be removed. Looks like the div is closed on line 153?

physionet-django/console/forms.py
```diff
@@ -360,6 +366,353 @@ def save(self):
...
+    decision = forms.ChoiceField()
+
+    class Meta:
+        model = CredentialApplication
+        fields = ('required_complete', 'not_frivolous', 'easily_translated',
+                  'responder_comments', 'decision')
+
+        labels = {
+            'required_complete': 'Are all of the required fields complete?',
+            'not_frivolous': 'Does the application clearly avoid frivolous text such as "aa"?',
```
tompollard: could we change this to: `"Does the application contain any obviously incorrect information?"` (and `not_frivolous` to something like `appears_correct`.  Saying an application is frivolous might seem unfriendly!

physionet-django/user/models.py
```diff
@@ -753,6 +752,38 @@ class CredentialApplication(models.Model):
         blank=True)
     revoked_datetime = models.DateTimeField(null=True)
 
+    # Initial review questions
```
tompollard: It would be helpful if these fields were easily recognizable as checks that happen during review. Perhaps `review_` or `check_` or `app_` could be appended to the start of each one?

I keep questioning whether we need them all in the model, but I guess it doesn't cause any harm and it may be helpful to have a record of each check.

physionet-django/console/templates/console/process_credential_application.html
```diff
@@ -27,8 +27,38 @@ <h1>Process Credential Application - {{ app_user.username }}</h1>
   </div>
   <div class="col-sm-6">
```
tompollard: I think if you switch all of the col-sm-6 to col-sm-5 then you will find the content sits in two columns. Take a look at for more guidance on bootstrap columns: https://getbootstrap.com/docs/4.1/layout/grid/

physionet-django/console/templates/console/process_credential_application.html
```diff
@@ -27,8 +27,38 @@ <h1>Process Credential Application - {{ app_user.username }}</h1>
...
     <div class="row">
+      <div class="card mb-4">
+        <div class="card-header">
+          Contact Reference
```
tompollard, line 32: I think the "Contact Reference" section is only needed for the "Reference check" status.

tompollard (commented):
This is beginning to come together :) I've added a few comments inline that it would be good to address before merging. Some other things that we can worry about later:

- It would be good to add "Search for X" buttons that run a Google search using fields from the form (e.g. for searching for an institution, or a user, or a reference).
- The page used to process an individual application should clearly display the stage at the top. e.g. Instead of "Process Credential Application - geraldinepennington", we should say "Training check" or "Initial review" etc.
- I don't think we need "Ongoing applications" and "Processing", so one of these could be dropped. Perhaps a better name would be "Under review" or similar?
- In http://localhost:8000/console/credential_processing/, applications should be sorted from oldest (top) to newest (bottom). No need for anything more complicated than this.
- when viewing the application page e.g. http://localhost:8000/console/credential-applications/qgRhw3V6PGBH7m5Qp5on/process/ "Processing" should be highlighted in the menu rather than "Ongoing applications" (or whatever the name for this section becomes).
- Some of the questions in the forms don't apply to everyone (e.g. "If the applicant is a student or postdoc: Is the reference..."). Should these questions only displayed if the applicant meets the requirement (e.g. in this case they have submitted as a student or postdoc)?
- The comments (I think) are saved to single field in the database, so they are overwritten each time the form is saved. This is not ideal.
- Emails should incorporate the form response (e.g. if x,y,z are highlighted during the review, the email should refer to x,y,z as issues).
- There is a fair amount of duplicated code but we can clean this up later.

Lucas-Mc:
Tom covered pretty much everything I just have a suggestion for the Google approach! An easy implementation for now would just be to create links of the form: `https://www.google.com/search?q="MIT"` ... This should probably open up a new browser tab, so all together:
```
<a href="https://www.google.com/search?q=\"MIT\"" target="_blank">Search for "MIT"</a>
```

tompollard:
Strange, tests pass for me locally. I'm unclear why they aren't passing on the server. I guess it must be either (1) a sqlite vs postgres issue or (2) a migration issue.

tompollard:
Big thanks to @bemoody for working out how to solve the issue with the migrations/fixtures. For the record, this was the fix:

```
Try this:

- delete the new migration
- ./manage.py resetdb
- ./manage.py loaddemo
- ./manage.py dumpdata project --indent 2 --natural-foreign
project/fixtures/demo-project.json
- ./manage.py dumpdata user --indent 2 --natural-foreign
user/fixtures/demo-user.json
- ./manage.py makemigrations
- ./manage.py resetdb
- ./manage.py loaddemo
- ./manage.py test

With --natural-foreign, generic foreign keys should appear as

 "content_type": ["project", "activeproject"]

instead of

 "content_type": 16
```

PR #1196, opened 2021-01-22
Creates DB sync between PN live and staging

elfeto:
There is a need to keep a synchronized database on Physionet between the live site and staging for proper testing.

Here I add a cronjob to be run weekly on Sundays to do a full backup dump and send it to staging.
Once in staging it will destroy the database and load the database from the full backup.

A log is kept from the database import.

For the database authentication this is using the recommended from Postgres for non-interactive authentication that is a .pgpass file in the user home directory.
I also created an RSA ssh key of 4096 bits for the scp command.


bemoody:
Sounds excellent.  Thanks for working on this.

Scheduling the task at midnight might not be the best, since a number
of other things are also scheduled to occur every day at midnight.

I think we only want the cron job to run on the production server, so:

    deploy/common/etc/cron.d/postgres-weekly-sync
    -> deploy/production/etc/cron.d/postgres-weekly-sync

And I'd be a tiny bit more comfortable if we only install the script
on the machine where it's needed:

    deploy/common/usr/local/bin/postgres-weekly-sync-recv
    -> deploy/staging/usr/local/bin/postgres-weekly-sync-recv

    deploy/common/usr/local/bin/postgres-weekly-sync-send
    -> deploy/production/usr/local/bin/postgres-weekly-sync-send

elfeto:
Thanks for checking it.

I moved the files as you mentioned it. (dunno know why I didn't created the folders in the first place.)
I modified the cronjob to 8AM.

Issue #870, opened 2020-02-12
Add option to provide ORCID in user profile


tompollard:
It should be possible for users to provide their [ORCID](https://orcid.org/) (unique researcher ID) in the user profile. This may be useful for several reasons (e.g. syncing projects with research profiles; identifying duplicate accounts; enabling login with ORCID credentials; providing additional information for credentialing applications, etc). More thought is needed about details of the implementation.

tompollard:
This may help with handling credentialing applications, so it would be good to prioritize. Details on the APIs are at: https://orcid.org/organizations/integrators

elfeto:
After checking the API, I found https://orcid.org/about/membership while requesting access to it https://orcid.org/content/register-client-application-0. I Either way requested a client application that has to be approved by them and I will get more information about it.

BUT it seems to be a paid service. I would recommend to see if MIT has some kind of connection with ORCID and see if there is a possibility to talk with a person that can give us more information.

The other option right now would be to just add an input box to ask for the individual orchid, but there is no way to corroborate that the ID added is actually the person's ID.

tompollard:
> BUT it seems to be a paid service. I would recommend to see if MIT has some kind of connection with ORCID and see if there is a possibility to talk with a person that can give us more information.

Great suggestion, let's contact MIT Libraries as a starting point [UPDATE: I checked a little while ago and it looked like there is a free service that does what we need it to do].

PR #1198, opened 2021-01-22
Add tab and button in profile for user to link to their ORCID account.

briangow:
It would be useful to be able to link PhysioNet user profiles to ORCID profiles.  See #870 for more details.  This change adds a tab in the user profile where the user can go to click on a button.  This will connect them to orcid.org where they can authorize connecting their ORCID iD to PhysioNet.


bemoody:
Thanks, Brian; this looks like a great start!


* Did you intend to set your name and email address to:

    briangow <4754434+briangow at users.noreply.github.com>

  Check your git configuration (~/.gitconfig).


* Please add ORCID_LOCAL_CLIENT_ID and ORCID_LOCAL_CLIENT_SECRET to
  .env.example.  (Better, rename these to "ORCID_CLIENT_ID" and
  "ORCID_CLIENT_SECRET" to match the names used in Python.)

  In .env.example, include a brief explanation of what these variables
  are for, and how to generate/register a client ID and client secret
  with orcid.org.


* If there is no ORCID ID available, then we simply shouldn't create
  an Orcid object; if the user asks to "unlink" it then it should
  simply be deleted.  Don't store values like "no iD linked" or "ORCID
  iD is not linked." in the database.


* Decide what the format of the orcid_id field should be.  Should it
  be "0000-0002-1825-0097"?  Should it be
  "https://orcid.org/0000-0002-1825-0097"?  Document the format with a
  comment in user/models.py.

  Additionally, you should write a validator function in
  user/validators.py that only accepts strings in the correct format,
  and use that as the validator for the CharField.


* The edit_orcid and auth_orcid functions in user/views.py must be
  marked with the @login_required decorator.  auth_orcid will also
  need to be marked with @disallow_during_maintenance.


* In edit_orcid, anyone is able to specify any ID and name they want,
  simply by going to a URL such as
  "https://physionet.org/settings/orcid/0000-0002-1825-0097/Frodo%20Baggins/".
  That's not what we want; the point of going through the oauth
  process is to obtain a verified ID and name from orcid.org.

  So auth_orcid should save the data as soon as it's retrieved, and
  then redirect to /settings/orcid/.


* oauth.fetch_token can presumably fail for a lot of reasons, and in
  particular it will fail if the "code" isn't valid or has expired.
  If this happens we want to display an error message and not crash.

  Also we should display an error message if the token doesn't include
  "orcid" and "name" elements or if the "orcid" value doesn't match
  the expected format.


* Don't use two different URL patterns ("settings/orcid/" and
  "settings/orcid/<orcid_id>/<orcid_name>/") with the same name and
  view function (edit_orcid).  We don't want the second one anyway
  (see above) but if we did want two different URL patterns they
  should be handled by different functions.

briangow:
Thanks @bemoody for taking the time to review this and providing clear feedback!  I did have a couple of questions about the .env variables.  I'm not sure if I fully understand how these are used in conjunction with base.py, staging.py, and production.py but this is what I was trying to accomplish:

- When testing on a local database make the user interact with the ORCID sandbox (sandbox.orcid.org).  Therefore, in base.py I use: 
```
ORCID_CLIENT_ID = config('ORCID_LOCAL_CLIENT_ID', default=False)
ORCID_CLIENT_SECRET = config('ORCID_LOCAL_CLIENT_SECRET', default=False)
```
where ORCID_LOCAL_CLIENT_ID and ORCID_LOCAL_CLIENT_SECRET are defined in the .env file and contain codes specific to sandbox.orcid.org.  Also in base.py we use a local redirect URI: http://127.0.0.1:8000/authorcid

- When running on physionet.org or staging.physionet.org we want the user to interact with the actual ORCID site, orcid.org.  Therefore these variables point to (\_PHYSIO_) codes in .env specific to interacting with the actual API:
```
ORCID_CLIENT_ID = config('ORCID_PHYSIO_CLIENT_ID', default=False)
ORCID_CLIENT_SECRET = config('ORCID_PHYSIO_CLIENT_SECRET', default=False)
```
In regards to adding something to .env.example, should I add an example which assumes the user is doing local testing?  Perhaps something along these lines:

These variables must be set through a registered account with ORCID to allow interaction with the ORCID API:
ORCID_LOCAL_CLIENT_ID=\<client ID from your ORCID account>
ORCID_LOCAL_CLIENT_SECRET=\<client secret from your ORCID account>
where \<client ID from your ORCID account> and \<client secret from your ORCID account> are obtained by registering for an account.  When doing development work it is advised to use an ORCID sandbox account, sandbox.orcid.org.  Under development tools you can setup a local redirect URI which points to http://127.0.0.1:8000/authorcid.  From the URL's provided thereafter, copy your client ID and client secret codes to your .env file.  

Given this, should I leave the \_LOCAL_ in the .env / .env.example variable names?  

For those of us who have access to physionet.org and staging.physionet.org should I add similar comments to the .env file?

briangow:
@bemoody I’ve made the updates you suggested above, thanks again for the tips! 

Regarding our discussion around storing the ORCID token.  I’m storing the token information in the database in separate fields:

```
    user = models.OneToOneField('user.User', related_name='orcid',
                                on_delete=models.CASCADE)
    orcid_id = models.CharField(max_length=50,
                                validators=[validators.validate_orcid_id],
                                default=('ORCID iD is not linked.'))
    orcid_name = models.CharField(max_length=50, default=('ORCID name is not linked.'))
    orcid_access_token = models.CharField(max_length=50,
                                          validators=[validators.validate_orcid_token],
                                          default=('ORCID access token is not linked.'))
    orcid_refresh_token = models.CharField(max_length=50,
                                           validators=[validators.validate_orcid_token],
                                           default=('ORCID refresh token is not linked.'))
    orcid_token_type = models.CharField(max_length=50, default=('ORCID token type is not linked.'))
    orcid_token_scope = models.CharField(max_length=50, default=('ORCID token scope is not linked.'))
    orcid_token_expiration = models.DecimalField(max_digits=50, decimal_places=40, default=0)
```

However, I could also store the original JSON token with all of its information in a single field.  Do you suggest I do this instead of the above, in addition to the above, or just stick with the current implementation?

Regarding exchanging the ORCID token for additional information about a given account holder.  I eventually found the correct URL for token exchanges on the public API (pub.sandbox.orcid.org / pub.orcid.org) in a forum.  The ORCID documentation only mentions the member API (api.sandbox.orcid.org / api.orcid.org).  Doing a token exchange on the public API isn’t common though because a token isn’t required to read a user’s public information.  I’m still trying to figure out with IS&T whether MIT has an institution account with ORCID.  Using this account would enable us to use the member API to add information to a given user’s ORCID account in the future if that is a feature we wanted to add to PhysioNet (Ex: add information about a dataset they published in a PhysioNet project to their ORCID profile).  

Lastly, I think it would be nice to have the users ORCID iD hyperlink to their profile as you implied.  Again, I’d need to distinguish whether to point to sandbox.orcid.org (when people are doing local development) or orcid.org.  Using variables from settings.py in a template appears to require a couple of extra steps: https://stackoverflow.com/questions/12166537/django-settings-py-variables-in-templates .  Would you recommend following this approach or instead storing a “base url” (sandbox.orcid.org or orcid.org) in the database alongside the iD?

bemoody:
Please combine the migrations into one, which you can do by deleting
0032_orcid.py and
0033_auto_20210126_1653_squashed_0038_auto_20210128_1216.py, and
re-running 'manage.py makemigrations'.  (After doing this you'll have
to reset your local database by running 'manage.py resetdb'.)

In the Orcid class, there's no need to have a method for retrieving
each attribute; for example, instead of 'foo.get_orcid_id()' you can
just use 'foo.orcid_id'.  Accessor functions like these are good style
in some other languages (C++, Java), but in Python they really don't
gain you anything.

Apparently the last character can be 'X' instead of a digit
(https://en.wikipedia.org/wiki/ORCID#Identifiers) so in
validate_orcid_id, I think the pattern should be
'^[0-9]{4}-[0-9]{4}-[0-9]{4}-[0-9]{3}[0-9X]$'.

We also want to run this validator before saving, which you could do
by calling it explicitly (as you're currently doing with
validate_orcid_token), or else by calling orcid_profile.full_clean()
to run all of the field validators.

In user/urls.py, you can use path('authorcid/', ...) instead of
re_path(r'^authorcid/$', ...).  I think 'path' should be preferred as
it's easier to read.

In user/views.py, please keep the "import" statements grouped and
sorted:

- first, imports from the Python standard library ("import logging",
  "import os")

- second, imports from external packages
  ("from django.core.exceptions import ValidationError",
  "from requests_oauthlib import OAuth2Session")

- third, imports from other modules within this package
  ("from user import validators")

(see https://www.python.org/dev/peps/pep-0008/#imports)

tompollard:
One quick style thing if you are regenerating the migrations: 

The leading `orcid_` in the model attributes and functions feel like they could be dropped, since the model is dedicated to ORCID. I'd imagine that any instances we create will be called "orcid" or similar, so "`orcid.access_token`" seems clear enough.

If we are thinking about changing the model later to capture additional identifiers (e.g. Researcher ID), then I guess this is an argument for keeping things as they are. I think we would just create a new model for these though?

tompollard:
This is now available on the staging server if anyone would like to test. Works well!

bemoody:
Thanks, Tom.  I haven't tested in detail but if you're satisfied then
feel free to merge.

tompollard:
Thanks Benjamin, as we don't have an on/off switch for the implementation (might be nice to add in the future) I'll hold off merging until Brian has heard back about adding us to the MIT ORCID membership account. My understanding is that we have been approved by MIT and we are now waiting for details from ORCID.

tompollard:
Looks great, thanks!

PR #1205, opened 2021-01-26
Add "contact reference" template email to the credentialing workflow (+ minor tweaks)

tompollard:
Builds on https://github.com/MIT-LCP/physionet-build/pull/1199 to add email templates to the credentialing workflow. This step adds a single "Contact reference" template as a pop-up modal. Also

- Fixes the sorting of applications on the credentialing pages (avoids overlaps between lists on the processing and management pages).
- Edits the text of the reference email.

@Lucas-Mc please could you take a look? Sorry about the messy commit history. We can use the same approach to add a pop up email to other pages of the workflow.

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -331,6 +331,37 @@ class DeprecateFilesForm(forms.Form):
     delete_files = forms.ChoiceField(choices=YES_NO)
 
 
+class ContactApplicantCredentialForm(forms.ModelForm):
+    """
+    Form to contact the applicant of a credential application
+    """
+
+    class Meta:
+        model = CredentialApplication
+        fields = ('app_contact_comments',)
+        labels = {
+            'app_contact_comments':'Comments (required to contact applicant)'
+        }
+        widgets = {
+            'app_contact_comments':forms.Textarea(attrs={'rows': 3}),
+        }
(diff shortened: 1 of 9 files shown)


Lucas-Mc:
I think this is a good start! In the future, maybe we should allow the reviewer to still have access to the "Contact Reference" button even after they've already sent an e-mail in case they wanted to remind them or submit another response. I also think it would help out if the default text box size was larger so more of the e-mail template can be seen without so much scrolling.

I think this is a separate issue, but when I reject an applicant it still brings me to the next step (so ID check if I rejected during training) without contacting them via e-mail (at least from what I can see in the terminal). This page can not be accessed again once I leave though.

tompollard:
Thanks @Lucas-Mc

> maybe we should allow the reviewer to still have access to the "Contact Reference" button even after they've already sent an e-mail 

Good point, I'll add this. Can this go into a new pull request?

> I think this is a separate issue, but when I reject an applicant it still brings me to the next step

Well spotted, I'll fix this. I think it is unrelated. Okay to do in another pull request too?

Lucas-Mc:
> Well spotted, I'll fix this. I think it is unrelated. Okay to do in another pull request too?

Sounds good!

PR #1208, opened 2021-01-26
Converts dropdown widget to radio buttons for credential processing

Lucas-Mc:
This change converts the current dropdown widget to radio buttons. The default value is being set to `True` so that no radio buttons are selected while the reviewer is processing the application. The final check page still has the dropdown since it is derived from the CredentialApplication model and I didn't want to mess with that, plus changing this one to a radio button wouldn't significantly impact speed while processing applications.

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -30,7 +30,6 @@ SUBMISSION_RESPONSE_CHOICES = (
 )
 
 REVIEW_RESPONSE_CHOICES = (
-    ('', '-----------'),
     (1, 'Pass to Next Stage'),
     (0, 'Reject'),
 )
@@ -49,8 +48,13 @@ YES_NO_UNDETERMINED = (
     (None, 'Undetermined')
 )
 
+YES_NO_UNDETERMINED_REVIEW = (
+    (1, 'Yes'),
+    (0, 'No'),
+    (None, 'Undetermined')
+)
+
 YES_NO_NA_UNDETERMINED = (
-    ('', '-----------'),
     (1, 'Yes'),
     (0, 'No'),
     (None, 'N/A or Undetermined')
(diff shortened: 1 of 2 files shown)


tompollard:
Thanks, and sorry for the delay in getting to this. As far as I can see, it should be enough to:

1) Remove the `('', '-----------')` option from the response labels. Perhaps by creating new response lists specifically for the CredentialReview class?
2) Change the `Select` widgets to `RadioSelect` widgets.

Unless I've missed something, I don't think the migrations or `default=True`s are needed. 

It would be good to clean up the layout a little before merging if possible. Currently it is shown as:

Screen Shot 2021-01-27 at 18 49 36

Can we remove the bullet points, and perhaps display the list in a row?

tompollard:
One other quick note. In https://github.com/MIT-LCP/physionet-build/pull/1215 we added the following response set:

```
YES_NO_NA_UNDETERMINED = (
    ('', '-----------'),
    (1, 'Yes'),
    (0, 'No'),
    (None, 'N/A or Undetermined')
```

So this set will need to be updated too.

Lucas-Mc:
Screen Shot 2021-01-28 at 11 14 19 AM

This is what happens if we don't change the model... thoughts?

tompollard:
> This is what happens if we don't change the model... thoughts?

Looks fine to me, and I think @kepaik said that we should not default to True. We can always add the default later if needed, but the priority is to switch from the dropdown to the buttons.

Lucas-Mc:
The updated appearance:

Screen Shot 2021-01-29 at 8 01 37 AM

Lucas-Mc:
Once we merge #1218 then we can convert the Final Review dropdown to radio buttons as well without messing with the current system.

tompollard:
Looks good, thanks Lucas! I think it would be good to try to play around with the formatting a little more - closing the vertical space would be nice - but we can worry about that later.

PR #1210, opened 2021-01-26
Adds credentialing email as bcc

Lucas-Mc:
This change adds the new credentialing PhysioNet e-mail as a Blind Carbon Copy (BCC) (not seen by the applicant) to begin processing applications under the new workflow. The decision for BCC over CC (seen by the applicant) was to prevent direct interference with the current workflow until the new workflow is finalized.

### DIFF
--- a/physionet-django/notification/utility.py
+++ b/physionet-django/notification/utility.py
@@ -649,8 +649,14 @@ def process_credential_complete(request, application, comments=True):
             'footer': email_footer()
         })
 
-    send_mail(subject, body, settings.DEFAULT_FROM_EMAIL,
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks Lucas, looks good. Quick thing, I notice that `headers` is set in `send_contact_message` but not here. Is there a reason why it isn't needed?

Lucas-Mc:
I just wanted to match the original as much as possible and it seemed like it based on the terminal output. Also, it seems like something else is going on in that function to customize the header. It didn't seem like a big deal, so I opted not to include it.

PR #1214, opened 2021-01-27
Changes redirect if rejecting credential application

Lucas-Mc:
Currently, if a credential application is rejected, the reviewer is redirected to the "Processing" page without being notified that the rejection was successful. Therefore, this change redirects the reviewer to an intermediate page which confirms the rejection.


tompollard:
Thanks, on a roll Lucas!

PR #1215, opened 2021-01-27
Fixes N/A logic in credential processing forms

shulim:
Previously, in the credentialing workflow (part of #1185), selecting "N/A" in a form would prevent a credentialing application from proceeding to the next stage. These changes modify three fields in the CredentialReview model so that they can take in integer inputs, which would allow for submission of both "Yes" and "N/A" to pass to the next stage, while "No" and "Undetermined" cannot.

Edit: undid the modifications to the CredentialReview model to avoid the need for migrations that could disrupt previous model field values, and instead I combined the "N/A" and "Undetermined" options to allow passing to the next stage.

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -53,10 +53,8 @@ YES_NO_NA_UNDETERMINED = (
     ('', '-----------'),
     (1, 'Yes'),
     (0, 'No'),
-    (2, 'N/A'),
-    (None, 'Undetermined')
+    (None, 'N/A or Undetermined')
 )
-
 
 class AssignEditorForm(forms.Form):
     """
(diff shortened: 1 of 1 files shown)


physionet-django/console/forms.py
```diff
@@ -51,10 +51,10 @@
 
 YES_NO_NA_UNDETERMINED = (
     ('', '-----------'),
-    (1, 'Yes'),
     (0, 'No'),
+    (1, 'Yes'),
```
tompollard, line 54: Changing the order here (or perhaps the change in the model) has affected the default order of responses in the form, so "no" appears before "yes". It would be better if these fields were consistent, with "yes" appearing first.

physionet-django/console/forms.py
```diff
@@ -51,10 +51,10 @@
 
 YES_NO_NA_UNDETERMINED = (
```
tompollard: Instead of having this variable being repeated, it might be better to reference the model (e.g. `CredentialReview.YES_NO_NA_UNDETERMINED.`) to ensure that they stay in sync?

physionet-django/console/forms.py
shulim: Thanks for taking a look, I switched the "No" and the "Yes"!

physionet-django/console/forms.py
shulim: Thanks for pointing that out, I removed the variable from the file so that the fields reference the model

PR #1218, opened 2021-01-27
Creates separate processing form for credential application review in stages

Lucas-Mc:
This change creates a new and separate credential application processing form for those reviewed in stages to remove its current ties to the current, non-stages, credential application review process.


tompollard:
Thanks Lucas, looks good. Seems odd that `ProcessCredentialReviewForm` gets assigned to both `process_credential_form` and `intermediate_credential_form` in the `process_credential_application` view? We can do a clean up later.

PR #1220, opened 2021-01-27
Add internal reviewer comments for credential application review #1219

### LINKED ISSUE #1219: Comment box for credentialing reviewers
It would be good to have a reviewer internal comment field, invisible to the applicant, where we can take notes or what not. For example, saving identity search results.

Lucas-Mc:
This change adds an internal reviewer comment field which can be used for storing intermediate thoughts from the reviewer while reviewing a credential application. These comments are persisted throughout each stage of the process where the reviewer can either add more comments or overwrite the current ones. External comments (to the applicant) are no longer persistent and are now only used while rejecting an application. Both internal and external comments can be seem by internal users when viewing previous applications and neither are shown to external users.

Thoughts need to be taken as to whether or not this is the best way to go about comments. Should they be stored separately for each stage? Can we edit them later?

Fixes #1219.


tompollard:
The ability to add internal notes was addressed in a subsequent PR.

PR #1227, opened 2021-01-28
Fixes application review filter #1226

### LINKED ISSUE #1226: Contacted references appearing in "Initial Review" tab
I am assuming these are from Ken Pierce's workflow.

Lucas-Mc:
Currently, the logic for filtering out credential applications in which the reference has been contacted is to use an `.exclude()` filter with multiple arguments which uses the logical AND operator (https://docs.djangoproject.com/en/3.1/ref/models/querysets/#exclude). Unfortunately, it appears that cases show up where we need the OR operator to perfectly exclude everything so this fixes that.

Fixes #1226.

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -1387,9 +1387,12 @@ def credential_processing(request):
 
     # TODO: Remove this step. If KP has contacted the reference, exclude the
     # application from our list. Avoid toes.
-    no_review = Q(credential_review__isnull=True)
+    review_not_underway = (Q(credential_review__status__lte=10) |
+                           Q(credential_review__isnull=True))
+
     ref_contacted = Q(reference_contact_datetime__isnull=False)
-    applications = applications.exclude(no_review, ref_contacted)
+
+    applications = applications.exclude(review_not_underway, ref_contacted)
 
     # Awaiting initial review
     initial_1 = Q(credential_review__isnull=True)


tompollard:
The Django filter/exclude syntax is kind of confusing (at least for me), but I'm pretty sure the change here is not what we want. As far as I can see, it now filters for applications that do not have a review: 

`applications = applications.filter(no_review)`

This means that we only display applications that are not under review. As soon as a review is started (i.e. the initial review is completed) the application will drop out of our workflow.

The current syntax is:

```
    no_review = Q(credential_review__isnull=True)
    ref_contacted = Q(reference_contact_datetime__isnull=False)
    applications = applications.exclude(no_review, ref_contacted)
```

... which I think is roughly what we want. The `exclude(no_review, ref_contacted)` is an `and` clause, saying that we will exclude applications where (1) `no_review` and (2) `ref_contacted` are both True

The issue is that filtering on `no_review` is too strict. The CredentialReview is created when "process" is clicked. Sometimes people click "process" then decide not to continue with their review.

Lucas-Mc:
So it sounds like the solution is to remove `no_review` from the `exclude` statement? Unless I'm confusing something?

Actually, I think it's harder than that....

tompollard:
> So it sounds like the solution is to remove no_review from the exclude statement? Unless I'm confusing something?

`applications = applications.exclude(ref_contacted)` would exclude all applications where the reference is contacted, which isn't what we want.

I think the solution is to replace the following chunk:

```py
no_review = Q(credential_review__isnull=True)
```

With something like this [TP: UPDATED...NOT TESTED!]:

```python
review_not_underway_list = [x[0] for x in CredentialReview.REVIEW_STATUS_LABELS if x[0] and x[0] <= 10]
review_not_underway = Q(credential_review__status__in=review_not_underway_list)
```

Then filter with:

```python
applications = applications.exclude(review_not_underway, ref_contacted)
```

tompollard:
Sorry, going to update the previous post (we actually want the opposite of "review_underway").

Lucas-Mc:
Fixed this by simply excluding applications whose reference have been contacted.. this worked in all the situations I tried during testing. The application `no_review` filter is already applied for the first tab and the other tabs are for review status >=20 which isn't possible without the first `initial review` being completed so this should cover everything.

tompollard:
`applications = applications.exclude(ref_contacted)` will not work, because it will exclude everyone who has been contacted. Try this:

- Pick an application at https://physionet.org/console/credential_processing/ that includes a reference.
- Move the application through the workflow until the "Reference response" step.
- Click "Contact Reference" and send the reference email.
- Go to https://physionet.org/console/credential_processing/ and try to find the user in the workflow.

Lucas-Mc:
Okay, sorry about that! This one should definitely work! (famous last words)🤞

tompollard:
This update looks like it will get applications where (1) there is no review OR (2) the reference has been contacted. Have you tested it by inviting someone using the management page to see if the application drops out of the processing page?

tompollard:
I just tested the latest approach by:

1. going to http://localhost:8000/console/credential_processing/ and clicking "process" for monicadavis
2. going to http://localhost:8000/console/complete-credential-applications/ and contacting the reference for this user
3. monicadavis should drop out of the list at: http://localhost:8000/console/credential_processing/, but she is still there (now with contact date).

tompollard:
If a review is not underway (i.e. does not exist or has not been started) AND the reference has been contacted, then it means that the application is being handled via KP's custom workflow. I see that there are less than ("lte") filter, so instead of the list comprehension we can do this:

```
review_not_underway = (Q(credential_review__status__lte=10) | Q(credential_review__isnull=True))
ref_contacted = Q(reference_contact_datetime__isnull=False)

applications = applications.exclude(review_not_underway, ref_contacted)
```

@Lucas-Mc I've added this fix. Please could you take a look and merge if you are happy?

PR #1229, opened 2021-01-28
Adds credentialing email as CC when contacting references #1224

### LINKED ISSUE #1224: Copy of email to references sent to admin distribution
I wonder if emails sent to references should be forwarded/copied to the admin distribution?
Particularly since that is one of the customizable messages so far.
Would be helpful in order to see whether the "we/us" might be referenced for clarification.

Lucas-Mc:
This change adds the admin credentialing email account as a CC to the email sent to the reference. The instructions in the email are changed from `reply` to `reply all` so that the reference knows to also include this account in their email response should they choose to do so.

Fixes #1224.

### DIFF
--- a/physionet-django/notification/utility.py
+++ b/physionet-django/notification/utility.py
@@ -475,7 +475,7 @@ def contact_applicant(request, application, comments):
 
 
 def contact_reference(request, application, send=True, wordwrap=True,
-                      subject="", body=""):
+                      subject="", body="", cc=None):
     """
     Request verification from a credentialing applicant's reference.
 
@@ -485,6 +485,7 @@ def contact_reference(request, application, send=True, wordwrap=True,
         wordwrap : If True, wraps body at 70 characters.
         subject : Subject line.
         body : Body text.
+        cc : The address to be CC'd to the email.
 
     Returns:
         dict : email name, subject, body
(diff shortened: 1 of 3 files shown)


physionet-django/notification/utility.py
```diff
@@ -475,7 +475,7 @@ def contact_applicant(request, application, comments):
 
 
 def contact_reference(request, application, send=True, wordwrap=True,
-                      subject="", body=""):
+                      subject="", body="", cc=""):
```
tompollard: My bad for doing this in the first place, but subject, body, cc should all really be `=None` rather than `=""` because this indicates they are unset.

physionet-django/notification/utility.py
```diff
@@ -509,8 +510,22 @@ def contact_reference(request, application, send=True, wordwrap=True,
...
     if send:
-        send_mail(subject, body, settings.DEFAULT_FROM_EMAIL,
-                  [application.reference_email], fail_silently=False)
+        if cc:
+            message = EmailMessage(
+                subject=subject,
+                body=body,
+                from_email=settings.DEFAULT_FROM_EMAIL,
+                to=[application.reference_email],
+                cc=[cc]
```
tompollard: Instead of the `if cc` clause, could you instead just pass the cc (which in some cases may be `None`)?

physionet-django/notification/utility.py
Lucas-Mc: The logic is the same but I can change it

physionet-django/notification/utility.py
Lucas-Mc: I can investigate what happens if we try to pass a `None` as the cc

Lucas-Mc: Looks like its `__init__` is `None` so it should work but I'll double-check with some tests:

https://github.com/django/django/blob/59e503b6708d41a44f2aa320272de3e2ecb5d65c/django/core/mail/message.py#L188-L196

Lucas-Mc: Also they initialize subject and body as `''` 🤔

PR #1230, opened 2021-01-28
Adds final reviewer when viewing previous credential applications #1223

### LINKED ISSUE #1223: Reviewer identity for unsuccessful/rejected applications
In the ">Past Applications > Unsuccessful Applications" table, it would be useful to be able to know who processed the application.

Lucas-Mc:
This change includes the final reviewer identity when looking at previous credential applications here: 

http://localhost:8000/console/past-credential-applications/successful
http://localhost:8000/console/past-credential-applications/unsuccessful

Fixes #1223.

### DIFF
--- a/physionet-django/console/templates/console/past_credential_unsuccessful_user_list.html
+++ b/physionet-django/console/templates/console/past_credential_unsuccessful_user_list.html
@@ -9,6 +9,7 @@
           <th>Decision Date</th>
           <th>Decision</th>
           <th>Comment</th>
+          <th>Reviewer</th>
           <th>View Details</th>
           <th>Manage</th>
         </tr>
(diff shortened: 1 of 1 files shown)


tompollard:
This change displays the reviewer's name to the applicant at http://localhost:8000/settings/credentialing/applications/, which I don't think we want to do.

It's a bit of a squeeze, but I think it would be best to put the user into the table at http://localhost:8000/console/past-credential-applications/unsuccessful.

e.g. this is how it is displayed on the "successful" page:

Screen Shot 2021-01-28 at 11 21 38

tompollard:
(Side issue, but we should really have separate public and private templates for displaying the credentialing history to help avoid this kind of thing happening).

Lucas-Mc:
> (Side issue, but we should really have separate public and private templates for displaying the credentialing history to help avoid this kind of thing happening).

AGREE

tompollard:
Thanks!

PR #1238, opened 2021-02-01
Adds search applicant feature for credential application processing

Lucas-Mc:
This change adds a "Search applicant" link which Google searches a credential applicant's personal information, specifically `search?q = first_names last_name organization_name`. This google search is opened in a new browser tab so the current page is not lost.


tompollard:
This is great! I think the link would be useful at all stages. Please could you move it somewhere in the profile section (e.g. in the example shown below).

Screen Shot 2021-02-01 at 18 22 23

tompollard:
Thanks!

PR #1250, opened 2021-02-03
Adds credential application tracker to admin console

Lucas-Mc:
This change adds a credential application tracker to the admin console under the "Usage Stats -> Credentialing" tab. This graph tracks the number of applications submitted, subtracts the number processed, and plots that difference over time (pending applications). This graph is created with D3 SVG and is semi-dynamic given a window size where is adjusts its window size `onload` but not `onchange`. I also fixed one of the entries in the demo data which had an application date after its approval date.


tompollard:
Credentaling has moved on since the PR was opened. Let's revisit this later.

PR #1252, opened 2021-02-04
Prevents user from adding themselves as a references

Lucas-Mc:
This change prevents the submitting user of a new credential application form from placing their reference email as one which is associated with their account. For this, I pass the `request.user` into the form and check that it's not in the `user.get_emails()` list.

### DIFF
--- a/physionet-django/user/forms.py
+++ b/physionet-django/user/forms.py
@@ -445,6 +445,12 @@ class ReferenceCAF(forms.ModelForm):
             'reference_title': 'Reference job title or position'
         }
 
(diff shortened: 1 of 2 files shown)


physionet-django/user/forms.py
```diff
@@ -444,7 +444,12 @@ class Meta:
         labels = {
             'reference_title': 'Reference job title or position'
         }
-
+    def __init__(self, user, *args, **kwargs):
```
tompollard, line 448: please could you add an empty separator row above __init__ ?

PR #1253, opened 2021-02-04
Adds indicator of known reference #1233

### LINKED ISSUE #1233: Identify whether a reference is a known reference and/or if they already have PhysioNet credentials
Maybe just a badge of some sort, to identify if a reference is an authenticated user.

Previously contacted references are less helpful, since we should still validate their identity, but is reassuring that they have already been cleared once at least.

Lucas-Mc:
This change adds an indicator of a known reference to a credential application review. This is done by simply adding a "Known Reference" highlighted text in the Reference section of the application display:
Screen Shot 2021-02-04 at 12 34 53 PM
Nothing is done if the reference is not known.
Fixes #1233.

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -1159,7 +1159,7 @@ def process_credential_application(request, application_slug):
     contact_cred_ref_form = forms.ContactCredentialRefForm(initial=ref_email)
 
     try:
-        ref = User.objects.get(associated_emails__email=application.reference_email.lower(),
+        ref = User.objects.get(associated_emails__email__iexact=application.reference_email,
                                associated_emails__is_verified=True)
         known_active = True
         known_cred = ref.is_credentialed
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -8,6 +8,7 @@ from itertools import chain
 from statistics import median, StatisticsError
 from collections import OrderedDict
 
+from django.core.exceptions import ObjectDoesNotExist
 from django.core.validators import validate_email
 from django.contrib import messages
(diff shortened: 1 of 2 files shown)


physionet-django/console/views.py
```diff
@@ -1155,6 +1155,18 @@ def process_credential_application(request, application_slug):
 
     ref_email = notification.contact_reference(request, application,
                                                send=False,  wordwrap=False)
+    all_active_users = User.objects.filter(is_active=True)
```
tompollard: Instead of iterating emails, you could query for the user object directly. e.g. 

```
from django.core.exceptions import ObjectDoesNotExist

try:
    ref = User.objects.get(associated_emails__email=email.lower(),
                           associated_emails__is_verified=True)
    known_active = True
    known_cred = ref.is_credentialed
except ObjectDoesNotExist:
    known_active = False
    known_cred = False
```

Lucas-Mc: Much cleaner thanks!

physionet-django/console/views.py
```diff
@@ -1157,6 +1158,15 @@ def process_credential_application(request, application_slug):
                                                send=False,  wordwrap=False)
     contact_cred_ref_form = forms.ContactCredentialRefForm(initial=ref_email)
 
+    try:
+        ref = User.objects.get(associated_emails__email=application.reference_email.lower(),
```
tompollard: Sorry, my bad, there is a case insensitive search term (`email__iexact="applications.reference_...`) which seems better than converting to lower()?

tompollard:
@Lucas-Mc I've added a case insensitive search on the email address. Please could you check and merge if you are happy?

PR #1257, opened 2021-02-08
Add reference organization field to credential application form

Lucas-Mc:
This change adds a new field to the credential application form where the user must submit an organization for the reference if one is required. The user may submit "None" if the reference happens to be an independent researcher (this can be changed if we think this could be a possible loophole). I also made changes to add this new field to the processing applications page for the credential application review workflow and for previous credential applications.

I further updated the demo user data by using regex to set the reference's organization to the same as the applicant's which should be suitable for now. Users without a reference have the reference's organization set to blank (`""`).


tompollard:
Looks good, thanks Lucas!

PR #1260, opened 2021-02-08
Adds previous credential applications to admin user profile

Lucas-Mc:
This change adds previous credential applications to the admin user profile in order to more effectively review new credential applications, specifically for those who must re-submit due to insufficient information.

### DIFF
--- a/physionet-django/user/views.py
+++ b/physionet-django/user/views.py
@@ -576,12 +576,16 @@ def edit_credentialing(request):
 
 
 @login_required
-def user_credential_applications(request):
+def user_credential_applications(request, user=None):
     """
     All the credential applications made by a user
     """
+    if user:
+        request_user = User.objects.filter(username=user)[0]
+    else:
+        request_user = request.user
     applications = CredentialApplication.objects.filter(
-        user=request.user).order_by('-application_datetime')
+        user=request_user).order_by('-application_datetime')
 
     return render(request, 'user/user_credential_applications.html',
         {'applications':applications})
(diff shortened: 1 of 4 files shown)


tompollard:
This successfully adds the previous applications to the user management page, but the way the information is displayed could be cleaner. Currently all previous applications get dropped into the user management page, which kind of dominates the page. Could the previous applications instead be displayed on a separate "application history page"?

Lucas-Mc:
This is possible, but wouldn't we still want that to be tied to the user management page so all the information regarding the user is in the same place? Or did you mean separate tabs on that page maybe for personal info, projects, and credential applications?

physionet-django/user/views.py
```diff
@@ -573,12 +573,16 @@ def edit_credentialing(request):
 
 
 @login_required
-def user_credential_applications(request):
+def user_credential_applications(request, user=None):
```
tompollard: I think it would be clearer to use `username` as the argument, if this is what you are expecting to receive.

physionet-django/user/views.py
```diff
@@ -573,12 +573,16 @@ def edit_credentialing(request):
...
     """
     All the credential applications made by a user
     """
+    if user:
+        request_user = User.objects.filter(username=user)[0]
```
tompollard: Instead of using a filter and indexing the first item, it seems cleaner to use `get` (we would expect usernames to be unique). [get_object_or_404](https://docs.djangoproject.com/en/3.1/topics/http/shortcuts/#get-object-or-404) might be even better.

tompollard:
This looks pretty good, but it allows **any** registered user to view someone's application history. e.g. try logging in as rgmark (a non-admin user) and viewing http://localhost:8000/settings/credentialing/applications/aewj/

physionet-django/user/views.py
Lucas-Mc: Nice catch!

PR #1261, opened 2021-02-09
Allows credential reviewer to contact reference multiple times

Lucas-Mc:
This change allows the credential application reviewer to contact the reference multiple times. Each time the reference is contacted, the `reference_contact_datetime` attribute will be updated to `timezone.now()` rather than preserving all previous contact times. I figured that storing all previous contact times was not worth the trouble to modify the model and respective views. I'm not sure the behavior of the temporary link it sends but from my testing so far I see that it sends the exact same link so no trouble should be encountered.

### DIFF
--- a/physionet-django/console/templates/console/process_credential_application.html
+++ b/physionet-django/console/templates/console/process_credential_application.html
@@ -35,7 +35,7 @@
           </div>
           <div class="card-body">
             {% if application.reference_email|length < 1 %}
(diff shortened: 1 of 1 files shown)


tompollard:
We may need to make some amendments to this later (e.g. making sure that the link continues to be valid; capturing both the "first contacted" date, as well as "last contacted"), but I think this is an improvement.

PR #1267, opened 2021-02-17
Adds pending applications to Past Applications page

Lucas-Mc:
This change adds processing applications to the Past Applications page (http://localhost:8000/console/past-credential-applications/processing) where general information (user, email, etc.) is displayed as well as a button to "Process" the application by redirecting the user to its respective processing page (either the past workflow or the new workflow). This also allows us to more easily search for applicants who are currently in the processing stage of their applications instead of searching across multiple pages.


tompollard:
Thanks Lucas, this seems to work well. Please could you make a few tweaks before we merge?:

1. Change "Processing" to something like "Pending" throughout. I think "Pending applications" more clearly communicates that the list contains applications that are pending a decision.
2. Change the "Past applications" in the menu to something like "All applications" (after this change, "past" applications doesn't seem quite right).
3. Perhaps also remove "past-" from the URL path for this page (for the same reason as point 2).
3. It might be worth removing "Ongoing applications" from the menu? I don't think it really serves a purpose after this change.
4. Minor thing (not directly related to this PR), but perhaps use this an an opportunity to remove "...Applications" from the tab headers at https://physionet.org/console/past-credential-applications/successful (i.e. just display "Successful", "Unsuccessful", "Pending"). It is clear that the page is referring to applications.

Lucas-Mc:
> Thanks Lucas, this seems to work well. Please could you make a few tweaks before we merge?:
> 
> 1. Change "Processing" to something like "Pending" throughout. I think "Pending applications" more clearly communicates that the list contains applications that are pending a decision.
> 2. Change the "Past applications" in the menu to something like "All applications" (after this change, "past" applications doesn't seem quite right).
> 3. Perhaps also remove "past-" from the URL path for this page (for the same reason as point 2).
> 4. It might be worth removing "Ongoing applications" from the menu? I don't think it really serves a purpose after this change.
> 5. Minor thing (not directly related to this PR), but perhaps use this an an opportunity to remove "...Applications" from the tab headers at https://physionet.org/console/past-credential-applications/successful (i.e. just display "Successful", "Unsuccessful", "Pending"). It is clear that the page is referring to applications.

Nice suggestions! They should be done here but it's a major change so make sure I didn't accidentally miss something 😄

physionet-django/console/templates/console/credential_applications.html
```diff
@@ -1,62 +1,91 @@
...
+        {% else %}
+          <p><i class="fas fa-check" style="color:green"></i> No applications to show.</p>
+        {% endif %}
+      </div>
       {% endif %}
     </div>
   </div>
 </div>
-{% endblock %}
+{% endblock %}
```
tompollard: quick nitpicks: please could you add back the newline here?

physionet-django/console/templates/console/credential_pending_user_list.html
```diff
@@ -0,0 +1,41 @@
...
+                    </button>
+                {% endif %}
+            </td>
+        {% endwith %}
+        </tr>
+        {% endfor %}
+    </tbody>
+</table>
+{% include "console/pagination.html" with pagination=p_applications %}
+</div>
```
tompollard: quick nitpicks: please could you add back the newline here?

physionet-django/console/views.py
```diff
@@ -1149,7 +1137,7 @@ def process_credential_application(request, application_slug):
         messages.error(request, """The application has already been
             processed. It may have been withdrawn by the applicant or
             handled by another administrator.""")
-        return redirect('credential_applications')
+        return redirect('credential_applications',status='pending')
```
tompollard: quick nitpicks: please could you add a space between the comma and "status"?

tompollard (commented):
Looks good, thanks Lucas. I've added a couple of comments about super minor formatting issues. Please could you take a look at these? I'll then merge!

PR #1270, opened 2021-02-18
Fix server error raised when anonymous users attempt to view project history

tompollard:
If an anonymous user (i.e. a user who is not logged in) attempts to view the history of a project, a server error is raised. For example, clicking one of the following links as an anonymous user will raise a server error:

http://localhost:8000/projects/published/demowave/1.0.0/submission-history/
http://localhost:8000/projects/rejected/demopsnm/submission-history/

This change redirects the user to the login page if they attempt to view the history. If the user does not have permission to view the page after logging in, they will receive a 404.


tompollard:
Looks great, thanks for sorting this out @elfeto!

PR #1273, opened 2021-02-23
Display ORCID for credentialing references, if available

tompollard:
When reviewing applications for credentialed access, it is helpful to know if the reference has an associated ORCID. This pull request:

1) Adds three methods to the `CredentialReview` model: (a) `ref_user_flag` which can be used to determine if the reference is a registered user (b) `ref_credentialed_flag` which can be used to determine if the reference is credentialed and (c) `get_reference_user` which can be used to retrieve the user object for the reference.

2) Displays an ORCID in the "Reference" section of credentialing applications if an ORCID is linked to the reference's user profile.

Screen Shot 2021-02-23 at 17 58 59

I'll follow up with another pull request with a minor clean up to the formatting for the chunk displayed in the screenshot above.

### DIFF
--- a/physionet-django/console/templates/console/application_display_table.html
+++ b/physionet-django/console/templates/console/application_display_table.html
@@ -66,12 +66,12 @@ Applied: {{ application.application_datetime|date }}</br>
   <li>Organization: <mark>{{ application.reference_organization }}</mark></li>
   <li>Position: <mark>{{ application.reference_title }}</mark></li>
   <li>Relationship: <mark>{{ application.get_reference_category_display }}</mark></li>
-  {% if known_active %}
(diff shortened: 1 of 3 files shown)


briangow:
This looks good!  If someone is using a reference who has a PhysioNet account in their application, it'd be good if they provided the email account associated with the references PhysioNet account.  We may consider adding something to this effect in the documentation.

tompollard:
> This looks good! If someone is using a reference who has a PhysioNet account in their application, it'd be good if they provided the email account associated with the references PhysioNet account. We may consider adding something to this effect in the documentation.

Good point, I'll try to look at adding a note somewhere in the application form.

PR #1274, opened 2021-02-24
Formatting changes to the application information in the credentialing workflow

tompollard:
Follows up on https://github.com/MIT-LCP/physionet-build/pull/1273 with a few minor formatting changes to the application information that is displayed in the credentialing workflow:

1. Fixes the text highlighting for "Active User" and "Credentialed User", so that only the status is highlighted rather than the whole line.

Screen Shot 2021-02-23 at 19 57 43

2. Only displays reference information if a reference is provided (otherwise displays "No reference provided")

Screen Shot 2021-02-23 at 19 53 19

3. Moves the "[Search for name and affiliation]" link to the "Profile" section for consistency with the Reference section.

Screen Shot 2021-02-23 at 19 58 02

PR #1275, opened 2021-02-24
Credentialing application form: Recommendations for a timely review

tompollard:
Follows previous changes to display ORCID IDs for references as well as applicants (ref https://github.com/MIT-LCP/physionet-build/pull/1274) in the credentialing workflow. 

This pull request tweaks the text on the application form (http://localhost:8000/credential-application/), adding "Recommendations for a timely review". One of the points is that it is helpful if the reference is registered and they have linked their ORCID ID.

### DIFF
--- a/physionet-django/user/templates/user/credential_application.html
+++ b/physionet-django/user/templates/user/credential_application.html
@@ -11,16 +11,23 @@
 {% block content %}
 <div class="container">
   {% include "message_snippet.html" %}
-  <h1>PhysioNet Credentialing Application</h1>
+  <h1>Application for Credentialed Access</h1>
   <hr>
 
   <p>Please use the form below to apply for PhysioNet credentialing. In order to apply:</p>
   <ul>
     <li>Complete the CITI Program in "Data or Specimens Only Research", an online course that covers ethics of human research and patient privacy. Instructions are provided <a href="{% url 'citi_course' %}">here</a>.</li>
     <li>If you have an ORCID iD please link to it in your settings (<a href="{% url 'edit_orcid' %}">ORCID settings</a>) as this may help us expedite your application by making it easier to verify your identity.</li>
-    <li>Check the form carefully before submission. <strong>Incomplete or erroneous applications will be rejected.</strong></li>
   </ul>
 
+  <h2>Recommendations for a timely review</h2>
+  <hr>
+  <p>The following steps will help to ensure a timely review of your application:</p>
+  <ul>
+    <li>Listing a reference is always helpful. Remember to prompt your reference to expect an email about your application.</li>
+    <li>It is also helpful if your reference is a registered user and they have linked their ORCID iD to their profile. Please provide the email associated with their account.</li>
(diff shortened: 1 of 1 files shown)


briangow:
This is a great idea, I'm going to leave a couple of comments with specific lines of code.

physionet-django/user/templates/user/credential_application.html
```diff
@@ -11,16 +11,23 @@
...
 <div class="container">
   {% include "message_snippet.html" %}
-  <h1>PhysioNet Credentialing Application</h1>
+  <h1>Application for Credentialed Access</h1>
   <hr>
 
   <p>Please use the form below to apply for PhysioNet credentialing. In order to apply:</p>
   <ul>
     <li>Complete the CITI Program in "Data or Specimens Only Research", an online course that covers ethics of human research and patient privacy. Instructions are provided <a href="{% url 'citi_course' %}">here</a>.</li>
     <li>If you have an ORCID iD please link to it in your settings (<a href="{% url 'edit_orcid' %}">ORCID settings</a>) as this may help us expedite your application by making it easier to verify your identity.</li>
```
briangow, line 20: Perhaps we should move this line about the applicant adding their own ORCID iD to the new timely review section below.

physionet-django/user/templates/user/credential_application.html
```diff
@@ -11,16 +11,23 @@
...
-    <li>Check the form carefully before submission. <strong>Incomplete or erroneous applications will be rejected.</strong></li>
   </ul>
 
+  <h2>Recommendations for a timely review</h2>
+  <hr>
+  <p>The following steps will help to ensure a timely review of your application:</p>
+  <ul>
+    <li>Listing a reference is always helpful. Remember to prompt your reference to expect an email about your application.</li>
+    <li>It is also helpful if your reference is a registered user and they have linked their ORCID iD to their profile.</li>
```
briangow: Do you think we should be more direct about providing a references PhysioNet email (if they are a registered user)?  Perhaps something like:  "It is also helpful if your reference is a registered user (please provide the email associated with their PhysioNet account if they are registered) and they have linked their ORCID iD to their profile."

physionet-django/user/templates/user/credential_application.html
tompollard: I thought about moving, but ended up leaving it here because I kind of like it being a soft requirement. Lucas's [pull request](https://github.com/MIT-LCP/physionet-build/pull/1189) moves this whole section, so perhaps we can rethink when we review his PR?

physionet-django/user/templates/user/credential_application.html
tompollard: Good point, thanks. I've updated this chunk as suggested!

physionet-django/user/templates/user/credential_application.html
briangow: Sounds good!

physionet-django/user/templates/user/credential_application.html
briangow: Looks good to me, thanks!

Issue #1279, opened 2021-03-03
Auto-save project creation forms to avoid data being lost  


tompollard:
When contributors are creating projects, they are required to populate a series of forms. The data is only saved when the "Save" button is clicked. Some of these forms are long (e.g. https://physionet.org/projects/<SLUG>/content/). 

It can waste a lot of time and effort if something happens to someone (e.g. @alistairewj) if he or she user unintentionally leaves the page before saving.

We should address this issue, perhaps by doing one or more of the following things:

- saving drafts of the text
- warning the user before they leave the page if there is unsaved content
- ...any other ideas?

bemoody:
> It can waste a lot of time and effort if something happens to someone
> (e.g. @alistairewj) if he or she user unintentionally leaves the page
> before saving.

Sounds like there's a story there... sorry, Alistair!

If I recall correctly, CKEditor does have a way to save automatically
or to put a "save" button on the toolbar so you can save without
submitting the whole form / reloading the page.  Of course we would
need something to handle that on the backend.

I am slightly disinclined to save automatically since we don't keep any
history of edits.

Definitely using the browser's "warn before navigating away if there
are unsaved changes" would be a good idea.

alistairewj:
> Definitely using the browser's "warn before navigating away if there
are unsaved changes" would be a good idea.

Yeah, this would have prevented my loss of data and sounds like a good change. A misplaced finger resulted in a three-finger swipe becoming a two-finger swipe, which triggered the back button, and never has there been a tale of more woe... well, maybe there's one.

tompollard:
I've been playing around with the autosave package for ckeditor and it works well. It handles the versioning issue really nicely:

1. the page autosaves every ~20 seconds
2. if you navigate away from the page without saving, then when you return you are shown a message to say that the most recent text was not saved

![Screen Shot 2021-03-03 at 14 47 17](https://user-images.githubusercontent.com/822601/109866536-5ff4ff00-7c33-11eb-8895-a96d37dce546.png)

3. You can then check and accept/reject a diff of the changes.

![Screen Shot 2021-03-03 at 14 52 20](https://user-images.githubusercontent.com/822601/109866595-7307cf00-7c33-11eb-8d87-e932f728ef3d.png)

PR #1280, opened 2021-03-03
Auto save ckeditor forms to help prevent loss of user created content #1279

tompollard:
As discussed in #1279, form data is only saved when the user clicks the "save" button. Users can easily lose content that they have created if they navigate away from the page without saving.

This pull request adds "autosave" functionality to the ckeditor form fields. These fields tend to contain large passages of text, so are some of the most important fields to protect. 

The functionality is described below:

1. the page autosaves every ~20 seconds (and a small "autosaved" message is displayed).
2. when returning to an unsaved page, the following message is displayed:

Screen Shot 2021-03-03 at 14 47 17

3. the user can check and accept/reject a diff of the changes.

Screen Shot 2021-03-03 at 14 52 20


bemoody:
Interesting, so to be clear, we're talking about auto saving on the
*client* side; it gets put in localStorage or something.

I haven't tested it but it seems like this interface might be somewhat
confusing (especially if you closed the window, went off and edited
the project on another machine, then later came back to the first
machine.)  But I don't have a horse in this race and any improvement
here is probably better than nothing.

I assume it's smart enough to associate the autosaved content with a
particular project (form URL) and a particular form field?  And I
assume it's also smart enough to *delete* the autosaved content once
the form is successfully submitted?

tompollard:
> Interesting, so to be clear, we're talking about auto saving on the *client* side; it gets put in localStorage or something.

That's right, a file is saved in the client's local storage (screenshot below from Google Chrome inspector).

Screen Shot 2021-03-03 at 18 20 04

The key is based on the URL and the form ID (e.g.: `autosave_http://localhost:8000/projects/OTeYYnJlM6DCKMxl87QA/content/_id_abstract`) and the value contains the encoded text.

> I haven't tested it but it seems like this interface might be somewhat confusing (especially if you closed the window, went off and edited the project on another machine, then later came back to the first machine.)

I agree, it could be a little for confusing for someone working in multiple locations. There is a time out that can be set on the autosave (default is 24 hours) which may help. 

I tried simulating this use case by working in multiple browsers, and I don't think it is a huge problem. The interface does a fairly good job at explaining what is happening (i.e. you have a temporary version that doesn't match the current version. do you want to look at the differences?).

> I assume it's smart enough to associate the autosaved content with a
particular project (form URL) and a particular form field?  

Yes, each field is identified by URL + form ID.

> And I assume it's also smart enough to *delete* the autosaved content once
the form is successfully submitted?

Yes, the client-side files are deleted when the page is saved.

> But I don't have a horse in this race and any improvement here is probably better than nothing.

If you and/or @alistairewj would prefer to just warn users when leaving the page, then I'm happy to go for that instead. In this case, ideally the user would only be warned if they have unsaved changes (not sure from the top of my head how that would be done). 

One benefit of autosave plugin approach is that it is pretty straightforward to undo if people don't like it. To undo, we just need to remove the "autosave" item from the config (and ideally delete the plugin content).

alistairewj:
I think I prefer the auto-save over a "warning, you are going to leave" dialog box. It's more likely to be effective since everyone has been hard-wired to ignore pop-up dialog boxes warning you about stuff. Feature looks nice!

bemoody:
After testing this, I do have concerns about the UI.  The fact that it
(briefly) says "Auto Saved" will make people think that the page was,
in fact, *saved* and they don't need to click the submit button.

It appears this message can be suppressed by setting:

    'autosave': {'messageType': 'no'},

tompollard:
That's a good point. Would you be happy using the plugin if we suppress the message? (thanks for finding the config for this)

tompollard:
Another minor issue that I didn't notice previously is that the "An autosaved version of this content has been found" popup is displayed once for each form that has been saved, which is a little confusing. It would be good to work out how to display the pop up only once for the page (preferably in a later pull request!).

bemoody:
Yeah, doesn't look like there's an easy way to fix that.

We also need to include the license for the package
(https://github.com/w8tcha/CKEditor-AutoSave-Plugin/blob/master/LICENSE.md),
and while you're at it, please include the original .css and .js
source files that the .min.css and .min.js files were generated
from
(https://github.com/w8tcha/CKEditor-AutoSave-Plugin/tree/master/autosave/css,
https://github.com/w8tcha/CKEditor-AutoSave-Plugin/tree/master/autosave/js).

Other than that this looks okay; hopefully it will be helpful and it's
easy to remove later if we find problems.

tompollard:
Thanks Benjamin, the license and source (js/css) files have been added.

bemoody:
Thanks!

Issue #183, opened 2018-11-27
Add figures for projects


cx1111:
Best to be able to embed them within the description fields via ckeditor.

cx1111:
From the result of today's discussion, we realized that since all figures need to be public, regardless of whether or not the project is open or protected, figures should not be published into the project file root itself, rather in `/static/published-figures/slugname/version/` or something similar.

tompollard:
Is there a reason why we wouldn't have slug as the top of the folder (e.g. `/static/slugname/published-figures/version/`)? There may be other project level resources that we may want to add in future, so this seems more flexible.

elfeto:
If the content would be figures, and the only thing they have in common is that they are figures, is ideal and cleaner to group them by that, rather than the slugname.

bemoody:
I'll repost my comment here where it belongs:

> To properly support images in project pages:
> 
> - Authors need to be able to upload image files somewhere.  This might or might not be the same as the project *files* area, but:
>   - must be subject to an overall size limit (either per-project or per-author)
>   - must be removed when the project is deleted
>   - must only be accessible by authors and editors
> 
> - Once the project is published, images need to be publicly accessible.
>   - They must be publicly accessible even if the project is restricted.
>   - Authors must not be able to modify or delete them.
>   - This *presumably* requires rewriting "src" attributes at publication time.
> 
> - SafeHTMLField needs to reject external images.
> 
> - Need convenient integration with CKEditor.
> 
> My initial thought was that images *would* be stored in the project files area, but this clearly wouldn't work for restricted projects.
> 
> At the same time, we could be much more restrictive about images than about project files in general - we could have a limit on the number of images per project, the size of each file, and require them all to be stored in a single flat directory.

PR #1284, opened 2021-03-08
Allow img tags in CKEditor. Ref #183

tompollard:
This pull request relaxes the ckeditor config to allow the img tag to be used. This change is needed because the CinC challenge organisers have several projects that they would like update, and some of these projects include figures.

### Background

There are a few projects that contain figures (2 or 3 challenges, I believe). e.g.:

- https://physionet.org/content/challenge-2019/1.0.0/ 

CKEditor (the project content editor) does not currently allow img tags to be embedded in content, so adding the images was a bit hacky (and involved editing the HTML directly in the database).

We now need to allow the contributors to update the projects. This is an issue because CKEditor will strip the existing img tags in the new version.

A simple fix is to enable to img tags in CKEditor. This allows the user to upload figure images as files and then link to the images using an img tag. There are some limitations with this approach (see #183) but I think it is a reasonable temporary solution.

As discussed in #183, it would be good to come up with a clean way of handling figures in project content. This might involve installing the ckeditor plugin. e.g.:

- https://ckeditor.com/docs/ckeditor4/latest/examples/fileupload.html
- https://ckeditor.com/cke4/addon/uploadimage

### DIFF
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -352,6 +352,7 @@ CKEDITOR_CONFIGS = {
             'h4': True,
             'h5': True,
             'h6': True,
+            'img': {'attributes': ['src', 'alt', 'width', 'height']},
             '*': {'attributes': _generic_attributes,
                   'classes': _math_classes},
         },


briangow:
This looks good to me.

bemoody:
This doesn't look like it should have any effect.  img is already part
of _inline_tags (though maybe it shouldn't be...)

Do you have an example of something that is rejected before and accepted afterwards?

PR #1285, opened 2021-03-09
Fix internal project links

bemoody:
This makes a few improvements to the handling of HTML text, particularly the text of project descriptions.

The major changes for project authors are:

- When a project moves from active to published, or published to active, internal links in the project description are updated accordingly.  If you write `<a href="/projects/SHuKI1APLrwWCqxSQnSk/preview/RECORDS">`, then when the project is published it will be replaced by `<a href="/content/mitdb/1.0.0/RECORDS">`.

- "src" attributes (currently permitted only for "img") cannot refer to an external server.  If you write `` it will be replaced by ``.  If you write `` it will be replaced by ``.

The less visible changes are:

- For links pointing to the same server, the scheme and hostname are stripped.  This is needed since the server is accessible at multiple hostnames and projects can even be transplanted from `physionet.org` to `staging.physionet.org`... clicking what appears to be an internal link shouldn't send you to a different hostname.

- *Relative* links will be expanded so they are always relative to the server root.  For example, `<a href="RECORDS">` becomes `<a href="/content/mitdb/1.0.0/RECORDS">`.  This is needed since the same HTML text is shown for every subdirectory of the project (and abstracts are shown in lots of other places, though as a matter of policy we don't want links or images in abstracts anyway.)


bemoody:
Sorry, made a couple of small thinkos.

path_subs was constructed assuming it would match only the path, rather than the full URL.  For example, prefix_map={'/foo': '/bar'} should remap "/foo?x=1" to "/bar?x=1":

```
--- a/physionet-django/project/utility.py
+++ b/physionet-django/project/utility.py
@@ -578,8 +578,9 @@ def __init__(self, base_url=None, my_hostnames=None, prefix_map=None):
                 oldpath = oldpath.rstrip('/')
                 newpath = newpath.rstrip('/')
                 # Match either oldpath exactly, or a string starting
-                # with oldpath followed by a slash
-                oldpath_pattern = '^' + re.escape(oldpath) + '(?![^/])'
+                # with oldpath followed by a slash, question mark, or
+                # number sign
+                oldpath_pattern = '^' + re.escape(oldpath) + '(?![^/?#])'
                 oldpath_re = re.compile(oldpath_pattern)
                 self.path_subs.append((oldpath_re, newpath))
 
```

The *base* URL should be "/content/mitbih/1.0.0/", not "/content/mitbih/1.0.0" (better not to use os.path for URLs anyway):

```
--- a/physionet-django/project/models.py
+++ b/physionet-django/project/models.py
@@ -759,14 +759,18 @@ def file_base_url(self):
         Return the base URL for downloading files from this project.
         """
         # file_url requires a non-empty path
-        return os.path.dirname(self.file_url('', 'dummy'))
+        url = self.file_url('', 'x')
+        assert url.endswith('/x')
+        return url[:-1]
 
     def file_display_base_url(self):
         """
         Return the base URL for displaying files in this project.
         """
         # file_display_url requires a non-empty path
-        return os.path.dirname(self.file_display_url('', 'dummy'))
+        url = self.file_display_url('', 'x')
+        assert url.endswith('/x')
+        return url[:-1]
 
     def edit_log_history(self):
         """
```

tompollard:
Thanks Benjamin, this should help to avoid broken links where URLs are included in the project content. 

Side note, but I think it would be good for us to avoid hyperlinks in the project content altogether where possible, because they are difficult to maintain and a little fragile. e.g.:

- when we "deprecate" (or whatever we want to call it) a project, links to files will break.
- if content is ever mirrored/replicated in a new archive (e.g. if MIT Libraries took responsibility for project descriptions because PhysioNet was unable to continue), internal links would likely break.

PR #1286, opened 2021-03-10
Fixes large CSV viewing

Lucas-Mc:
Currently, large CSVs are rendered to cover the entire page even expanding past its parent's div's CSS width and height leading to very wide and long pages. This change fixes that by setting overflow behavior to scroll instead of expanding past the page.

### DIFF
--- a/physionet-django/project/templates/project/file_view.html
+++ b/physionet-django/project/templates/project/file_view.html
@@ -39,7 +39,7 @@
         </li>
       </ul>
     </div>
-    <div class="card-body">
+    <div class="card-body data-card">
       {% block file_content %}
       <div class="text-center">
         This file cannot be viewed in the browser.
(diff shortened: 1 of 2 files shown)


bemoody:
This is not an improvement.  When I load a page, I want to be able to
(a) scroll immediately by pressing the arrow keys, and (b) see as much
of the text as possible on the screen at a time.

Putting the content into an inner scrolled element makes it more
difficult to navigate (how many times do I need to press tab? where
*is* that damn horizontal scrollbar anyway?) and wastes my precious
screen space (88 horizontal pixels and 177 vertical pixels, in this
case, consumed by white borders and distracting junk.)

And nested scrollbars are just *ugly*!

Suffice it to say this sort of thing is something I find intensely
aggravating about many websites and not something to emulate.

Lucas-Mc:
Thanks, I'll leave it how it is. I suppose it's not an issue.

bemoody:
Hey Lucas, I'm sorry.  I think my previous message came out sounding a
lot harsher than I meant it.

The existing style could certainly do with improvement, and I would
prefer to see something to improve the visual appearance without
sacrificing usability; but it's not that big a deal.

Issue #1292, opened 2021-03-12
Alternative credentialing workflow for classes and large groups


kepaik:
Something to work on.

Low bandwidth: simple flag to identify it's for a class. That way we can know that it should be dealt with differently.

Long-term: More automated workflow, potentially involving the course directors themselves managing approval lists.

kepaik:
Is there a way to add images to comments?

Otherwise, if we could add 1 or two web pages at the top of the credentialing application, this could be an easy way to filter course applications.
heading: 09.14 - Credentialing UX
in: https://docs.google.com/document/d/1ejHjAvw85q53mY_Bu2gymivz30Mo9za-4QJI8QYW2L0/edit

PR #1293, opened 2021-03-12
Adds course list for credentialing processing #1292

Lucas-Mc:
This change allows admin users to add references into the system who are operating classes. The user is instructed to submit the name of the reference, their email, their organization, and the name of the class. A list of classes will be kept here: http://127.0.0.1:8000/console/class-list/. There, admin users can also remove the class after a given time if needed (maybe when the class ends so processing doesn't get confusing).

On the credential application processing page (i.e. http://127.0.0.1:8000/console/credential-applications/qgRhw3V6PGBH7m5Qp5on/process/), the admin user will see if the provided reference has an active class in the system and list the class name if available, else it will say "No class provided".

On the backend, all submitted classes are saved, just either in an "active" or "inactive" state denoted by the `BooleanField` in the model called `active`. This allows us to search for previous classes in the database for whatever reason we see fit. A creation and deactivation datetime are also stored.

Fixes (the easy part of) #1292.


tompollard:
(An updated version of this is in the works...please hold off from reviews!)

tompollard:
Superseded by https://github.com/MIT-LCP/physionet-build/pull/1777

PR #1295, opened 2021-03-16
Update how to use ORCID info in .env.example to refer to the instructions under physionet-build on Github.

briangow:
This points the user from .env.example to https://github.com/MIT-LCP/physionet-build/blob/dev/deploy/README.md .  Therefore, we only need to maintain the detailed instructions for using ORCID in one place.


tompollard:
i'm thinking about the generic version of the platform that we want to build, and how we want to avoid references to "physionet".

instead of referencing a physionet specific URL, could we refer the user to the readme (which should be part of the package)?

briangow:
Thanks @tompollard , I've updated this to refer the user to the README without using a hard-coded URL.

Lucas-Mc:
Is this finished @tompollard ? It still refers to PhysioNet?

tompollard:
Yeah, just discussed with Brian. I think it's okay to merge. I'd prefer not to have the reference to PhysioNet in there, but there are a whole bunch of places that this will need to be addressed, so let's not worry for now!

PR #1296, opened 2021-03-18
Change account purge date to 3 days.

elfeto:
As requested from Tom, I am changing the account purge timeframe from 7 days to 3 days.


tompollard:
Thanks Felipe, this looks good to me!

When an activation link is followed after the account has been deleted, the following message is displayed:

```
Invalid Activation Link

There was an error with your link. Please try again or contact the site administrator..
```

It would be good us to provide a more informative message (e.g. explaining the possible cause and suggest a fix like re-registering) + remove the stray `.` at the end.

PR #1299, opened 2021-03-19
Allows users to create a course and add students #1292

Lucas-Mc:
This change allows registered users to create a course and add students to that course. In order to achieve this instructor status, someone with admin privileges must add the registered user to a list of Known Instructors (http://localhost:8000/console/known-instructors/) by entering the primary email address from their PhysioNet account.

Once added, the instructor user may now add courses from their Settings/Courses page (http://localhost:8000/settings/courses/). There, the instructor will submit the name of their course and can either "Manage" or "Remove" the course. Managing the course allows the instructor to submit the emails of their students. This can either be one at a time (`email1@mit.edu`) or in a comma separated fashion (`email1@mit.edu, email2@mit.edu, ...`). The students must be registered users as well prior to addition to the course list. Once added, students can be removed if needed.

Admin privileged users will, at any time, be able to view all the courses (active and inactive) of all the Known Instructors. For example, if the user `george` is added, all of their courses can be seen here: http://localhost:8000/console/known-instructors/george/courses/. 

If this gets merged, I plan on finding out a way to best address these courses in the credential application processing workflow.

Fixes (the hard part of) #1292.


danamouk:
I just rebased the `ref_course` branch on `dev`.

alistairewj:
Does this need to be specific to courses? I can think of other situations where more than 5+ users in a group may want approval.

tompollard:
@alistairewj good point, we should make sure that whatever approach we use is generalizable. I liked your suggestion of supporting temporary credentialing for workshops.

danamouk:
Thanks @Lucas-Mc this looks good. I tested it, and found some of the following points that we could look into. @tompollard and @bemoody, there might be somethings we might want to factor in to make the process more streamlined with the new credentialing system.
- Once we login as an instructor to add students, if 5 students are registered and one student is not registered, then all of the students are not added to the course.
- Once we login as an administrator, the page changes the username and it is unclear why.

tompollard:
Superseded by https://github.com/MIT-LCP/physionet-build/pull/1777

PR #1300, opened 2021-03-19
Tweaks reference response page to sort by verification status

Lucas-Mc:
This change tweaks the reference response step to sort by reference response verification status. Simply, applications where the reference has responded will be placed at the top. Currently, the applications are sorted by submission date, which is nice for the other stages, but we won't be passing on the applications in this stage without hearing from a reference so this kind of sort is nice.

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -1386,7 +1386,8 @@ def credential_processing(request):
         credential_review__status=40).order_by('application_datetime')
     # Awaiting reference response
     response_applications = applications.filter(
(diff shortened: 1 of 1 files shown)


tompollard:
I agree that it would be good to display applications that have received responses at the top. After this change, how are the remaining applications sorted (i.e. those that have not received a response)?

Lucas-Mc:
Nice catch, I just fixed it to sort first by reference verification then by application date which is better.

tompollard:
thanks lucas

PR #1301, opened 2021-03-19
Allows credential application reviewers to reset reviews #1248

### LINKED ISSUE #1248: Undo credentialing process queue
Would be useful to be able to unwind a stage or undo any review for a candidate.
Especially at this moment, since we have parallel workflows with kpierce, but also long term, in case someone was mistakenly advanced to a later stage without adequate review.  Would be fine to just kick to the initial review, rather than a specific stage.

Lucas-Mc:
This small change allows credential application reviewers to "reset" their review of the application. Essentially, in the backend, this just deletes its connected credential review which sends it back to the Initial Review stage. Once reset, it will show back up on KP's management page.

One thing to note is that if we contact an applicant's reference then reset the application, the application will still show up on KP's management page but it will say the reference has not been contacted. This may lead us to double contact the reference but I don't think this will happen very often.

Fixes #1248.

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -1412,6 +1412,7 @@ def credential_processing(request):
             application.credential_review.delete()
             new_review = CredentialReview.objects.create(application=application)
             application.save()
+            messages.success(request, 'The application has been reset')
 
     return render(request, 'console/credential_processing.html',
         {'applications': applications,
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -1403,6 +1403,16 @@ def credential_processing(request):
     final_applications = applications.filter(
         credential_review__status=60).order_by('application_datetime')
 
+    if request.method == 'POST':
+        if 'reset_application' in request.POST:
+            try:
+                application = CredentialApplication.objects.get(slug=request.POST['reset_application'])
+            except CredentialApplication.DoesNotExist:
+                raise Http404()
+            application.credential_review.delete()
+            application.reference_contact_datetime = None
+            application.save()
+
     return render(request, 'console/credential_processing.html',
         {'applications': applications,
         'initial_applications': initial_applications,
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -1410,7 +1410,7 @@ def credential_processing(request):
             except CredentialApplication.DoesNotExist:
                 raise Http404()
             application.credential_review.delete()
-            application.reference_contact_datetime = None
+            new_review = CredentialReview.objects.create(application=application)
             application.save()
 
     return render(request, 'console/credential_processing.html',
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -1410,8 +1410,6 @@ def credential_processing(request):
             except CredentialApplication.DoesNotExist:
                 raise Http404()
             application.credential_review.delete()
-            new_review = CredentialReview.objects.create(application=application)
-            application.save()
             messages.success(request, 'The application has been reset')
 
     return render(request, 'console/credential_processing.html',
--- a/physionet-django/console/templates/console/credential_processing.html
+++ b/physionet-django/console/templates/console/credential_processing.html
@@ -63,7 +63,11 @@
                 {% else %}
                   <td>{{ application.time_elapsed }} days</td>
                 {% endif %}
-                <td><a href="{% url 'process_credential_application' application.slug %}">Process</a></td>
+                <td>
+                  <form action="{% url 'process_credential_application' application.slug %}">
+                    <input class="btn btn-success" type="submit" value="Process"/>
+                  </form>
+                </td>
               {% endwith %}
               </tr>
             {% endfor %}
@@ -113,7 +117,11 @@
                     <button class="btn btn-danger" name="reset_application" value="{{ application.slug }}" type="submit">Reset</button>
                   </form>
                 </td>
-                <td><a href="{% url 'process_credential_application' application.slug %}">Process</a></td>
+                <td>
+                  <form action="{% url 'process_credential_application' application.slug %}">
+                    <input class="btn btn-success" type="submit" value="Process"/>
+                  </form>
+                </td>
               {% endwith %}
               </tr>
             {% endfor %}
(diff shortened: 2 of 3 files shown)


tompollard:
Deleting the review seems a little destructive! Any reason not to just reset the status?

Lucas-Mc:
What do you mean reset the status? We want KP to see it too and why does it matter if there's a review already if everything is `null` when it will just be populated later just the same as this simple approach.

tompollard:
> What do you mean reset the status? We want KP to see it too and why does it matter if there's a review already if everything is null when it will just be populated later just the same as this simple approach.

The CredentialReview object has a `status` attribute. It looks like setting `status = 0` (edit: or 10?) would achieve what you would like to do, so I'm just trying to understand why we delete the object instead.

```
class CredentialReview(models.Model):
    """
    Reviews for the CredentialApplications.
    """
    REVIEW_STATUS_LABELS = (
        ('', '-----------'),
        (0,  'Not in review'),
        (10, 'Initial review'),
        (20, 'Training'),
        (30, 'ID check'),
        (40, 'Reference'),
        (50, 'Reference response'),
        (60, 'Final review')
    )

    application = models.OneToOneField('user.CredentialApplication',
                                       related_name='credential_review',
                                       on_delete=models.CASCADE)

    status = models.PositiveSmallIntegerField(default=10,
        choices=REVIEW_STATUS_LABELS)
```

Lucas-Mc:
Essentially my point is, why should there be a `CredentialReview` object if we are resetting the application review? I don't want to salvage anything from the previous review when I do that. The application will still show up in Initial Review along with the other applications without (and with) a `CredentialReview` object. I thought the object only gets generated when someone clicks on process then doesn't do anything so why would we want that to be the default? Why not just like it's a fresh application?

Lucas-Mc:
Also, won't all of the responses still be there from whatever the previous reviewer submitted? Such as `fields_complete`, `appears_correct`, etc.

tompollard:
> Why not just like it's a fresh application?

For me, essentially because I don't think it is a fresh application. It's an application that has been looked at, and then moved back to the start of the workflow (probably for many different reasons). Keeping this in mind may help us to address issues like the one you mention ("This may lead us to double contact the reference but I don't think this will happen very often.")

tompollard:
> Also, won't all of the responses still be there from whatever the previous reviewer submitted? Such as fields_complete, appears_correct, etc.

Yes, they will be there unless you decide that they should be reset too.

Lucas-Mc:
Even your solution will lead to KP possibly double contacting unless we want to treat that specific type of application differently to make it not show up on his page (`old_app` flag? new status?). I can delete the review then give it a fresh review with `status=10` if that's better.

tompollard:
> I can delete the review then give it a fresh review with status=10 if that's better.

I don't think deleting objects is something we should treat lightly, and I'm unclear why it is necessary. It's easy to end up with downstream errors when an object is deleted and we lose useful information (like the fact that the reference was contacted).

> Even your solution will lead to KP possibly double contacting unless we want to treat that specific type of application differently to make it not show up on his page

If we do not delete the application, we retain the information that the reference is contacted. This means that we can address the issue you highlighted, rather than ignore it. e.g. if the reference has been contacted already, you could flag this for Ken.

Lucas-Mc:
Okay, sure, maybe I'll do that and reset everything except the reference contact information. But, what are we supposed to do with that information? If I see an application in Initial Review where the reference has already been contacted I'll be very confused. I don't think this will happen because we shouldn't be resetting if the reference has been contacted (maybe I can remove the button for those apps). This will make everything 100x easier.

tompollard:
The potential for references to be double contacted could lead to a significant amount of confusion (for us and the referees) so I agree that we need a way to prevent this. I don't have a good suggestion without looking into the details, but a quick fix is to do what you suggest (only display the button if the reference has not been contacted). It might be easier to resolve this issue once there is only a single workflow.

bemoody:
Certainly I would agree that if the reference has been contacted, that
information ought to be kept.  But that's not part of the
CredentialReview object.

So, okay, you could write a function to reset 25 or 26 fields
individually, or try to do it systematically, but either way seems
fragile.  Deleting the entire object is simpler, and it doesn't look
like there would be any problems with doing so.

These models are not well documented, though.  The documentation should
explain things for someone who is not familiar with the codebase and
has only a high-level understanding of the functions of the site.  Why
does the object exist?  When and how is it created and updated?  When
and how is it deleted?

If the purpose of the object is well documented, then it's easier to
determine the implications of deleting it.  If the documentation says
"this object can be deleted by the reviewer clicking the 'Reset
Application' button", then we will think twice about putting
information there that we might want to keep.

Whether it should be *possible* to reset the review after the reference
has been contacted is a question of procedure, which I will leave to
Kenneth and others to decide. :)  But I don't see a technical reason
you can't contact the reference, then later decide to redo the
training/identity checks, and skip over the "contact reference" step
when you get to it.

tompollard:
Thanks @bemoody. @Lucas-Mc, let's go with this then. Please could you make a couple of tweaks before we merge?:

- Add a note to the docstring to make it clear that the object may be deleted (as suggested by Benjamin above).
- Add a "`messages.success(request, 'The application has been reset')`" to confirm the action.
- Make the "Process" link a button too (green?) for consistency (optional).
- Add an "are you sure?" step after the reset button is clicked (optional).

physionet-django/console/views.py
```diff
@@ -1403,6 +1403,17 @@ def credential_processing(request):
     final_applications = applications.filter(
         credential_review__status=60).order_by('application_datetime')
 
+    if request.method == 'POST':
+        if 'reset_application' in request.POST:
+            try:
+                application = CredentialApplication.objects.get(slug=request.POST['reset_application'])
+            except CredentialApplication.DoesNotExist:
+                raise Http404()
+            application.credential_review.delete()
```
tompollard, line 1412: Perhaps put these database actions (delete and create) into a `with transaction.atomic():` clause so that if one step fails, both fail?

physionet-django/console/views.py
```diff
@@ -1403,6 +1403,17 @@ def credential_processing(request):
...
+            new_review = CredentialReview.objects.create(application=application)
+            application.save()
```
tompollard: I don't think this `application.save()` is needed because the object is not being modified. I was initially worried that the step might re-save the original review object (leading to duplicate reviews) but I don't think this is the case.

tompollard (commented):
Thanks Lucas, I added a couple of quick thoughts. Once you've looked at these points, we should be good to merge.

tompollard:
thanks lucas!

PR #1302, opened 2021-03-19
Adds reference email to credential processing workflow #1237

### LINKED ISSUE #1237: View/Sort by Reference Email on processing page
Currently only shows whether reference has been contacted, but that is not useful on the first tab.
This would be useful for batch processing of emails, but also Ken Pierce's request to sometimes have priority on certain applications.

Would be really useful to be able to sort on selected fields, such as reference email, but also maybe Institution

Lucas-Mc:
This change does multiple things for the credential application processing workflow:
1. Adds the submitted reference email to the table
2. Fixes the layout of the table to stay within the limits of the page (helps when doing half screen)
3. Removes reference contact and verification information prior to the Reference Response stage since it's irrelevant

Fixes #1237.

### DIFF
--- a/physionet-django/console/static/console/css/console.css
+++ b/physionet-django/console/static/console/css/console.css
@@ -526,3 +526,20 @@ footer.sticky-footer {
   list-style-type: none;
   margin: 1rem;
 }
+
+.table-cred {
+  table-layout: fixed;
+  width: 100%;
+}
+.table-index {
+  width: 5%;
+}
+.table-user {
+  width: 10%;
+}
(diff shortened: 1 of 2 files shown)


tompollard:
Looks good, thanks Lucas.

> Removes reference contact and verification information prior to the Reference Response stage since it's irrelevant

These columns have been useful in the past because they helped us to identify issues where applications appeared in both Ken and Sophie's workflows. I think these issues are now resolved, so we should be safe to remove the columns.

PR #1306, opened 2021-03-22
Adds option to download multiple files at once for published projects #1175

### LINKED ISSUE #1175: Please provide a 'select all' option for files in a Physionet project
It would be helpful if there was a 'select all' option under the files section of a Physionet project.  This would allow the user to quickly select and move or delete a large number of files.  This may be necessary if they are changing the organization of their project or creating an updated version of the project which requires a new structure.  When the project has many files it is tedious and time consuming to select each file individually.

Lucas-Mc:
This change gives the user the option to download multiple files at once by using checkboxes next to the file name in the files panel. Currently, the user is required to manually go through each file and press the download button which is more tedious and, on my current setup, the files open when they are finished downloading which interrupts me trying to download multiple files at once. Further, the download button is far from the file name and sometimes it's difficult to tell if I'm downloading the correct file.

The user simply checks off which files they wish to download and then clicks the "Download Selected Items" button which automatically opens up the download URL for each of the selected files. 

Fixes #1175.

### DIFF
--- a/physionet-django/project/templates/project/files_panel.html
+++ b/physionet-django/project/templates/project/files_panel.html
@@ -1,19 +1,24 @@
 <div class="card-header">
-  Folder Navigation:
-  {% spaceless %}
-    <span class="dir-breadcrumbs">
-      {% for breadcrumb in dir_breadcrumbs %}
-        {% if forloop.counter == dir_breadcrumbs|length %}
-          <span class="dir-breadcrumb-self">{{ breadcrumb.name }}</span>
-        {% else %}
-          <a href="{{ breadcrumb.rel_path }}#files-panel"
-             onclick="return navigateDir('{{ breadcrumb.full_subdir }}')"
-             class="dir-breadcrumb-up">{{ breadcrumb.name }}</a>
-          <span class="dir-breadcrumb-sep">/</span>
-        {% endif %}
-      {% endfor %}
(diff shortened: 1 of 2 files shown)


elfeto:
> It would be helpful if there was a 'select all' option under the files section of a Physionet project. 

It would still be helpful to have the "select all" if we are going to approve this PR since the issue it fix started with that statement.

briangow:
@Lucas-Mc this update is much welcomed! When clicking Download All Items it was only downloading the first file for me.  Also, is it possible to allow downloading individual folders (and their included content)?

Lucas-Mc:
Hey @briangow, downloading individual folders and its nested contents may be much harder since an AJAX request happens in the background to load a folder's contents when it's clicked on so let's keep that as a feature request.

Also, as mentioned in our conversation offline, it appears like Chrome blocks multiple downloads at first due to security risks but after the user allows it for the hostname then it should work subsequently.

briangow:
Hey @Lucas-Mc , I'm able to download multiple files at once after giving site permissions to the Chrome pop-up blocker.  Do you think we should put a note about allowing pop-ups somewhere (perhaps below the box that shows all the files)?

Lucas-Mc:
Nice! I think @tompollard should take a look first to see since he's better at judging the appearance of things

PR #1307, opened 2021-03-22
Displays debug toolbar when in development mode #1156

### LINKED ISSUE #1156: Display profiler console when in the development environment

Lucas-Mc:
This change displays the debug toolbar when in development mode. Metrics displayed here include things such as SQL queries, number of static files, CPU time, etc.

Fixes #1156.

### DIFF
--- a/physionet-django/physionet/settings/development/base.py
+++ b/physionet-django/physionet/settings/development/base.py
@@ -12,6 +12,10 @@ INSTALLED_APPS += [
     'debug_toolbar',
 ]
 
+INTERNAL_IPS = [
+    '127.0.0.1',
+]
+
 MIDDLEWARE += ['debug_toolbar.middleware.DebugToolbarMiddleware', ]
 
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
--- a/physionet-django/physionet/urls.py
+++ b/physionet-django/physionet/urls.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.conf.urls import include
 from django.contrib import admin
 from django.urls import path
@@ -66,3 +67,8 @@ urlpatterns = [
     path('robots.txt', lambda x: HttpResponse("User-Agent: *\Allow: /", 
         content_type="text/plain"), name="robots_file"),
 ]
+
+if settings.DEBUG:
+    import debug_toolbar
+    # debug toolbar
+    urlpatterns.append(path('__debug__/', include(debug_toolbar.urls)))


bemoody:
Thanks, Lucas.

I am a little concerned that the staging configuration
(physionet/settings/staging.py) enables the debug flag.  Probably that
should be turned off for safety, *even if* debug features are only
accessible to localhost (which I haven't checked.)

But having this enabled for development will be really useful!

Lucas-Mc:
Thanks Benjamin! That's true, what do we currently use the `DEBUG` flag for on staging? If we do turn off that flag, we should be sure that it doesn't hinder anything like detailed stack traces for errors.

tompollard:
We should also avoid the import and URL pattern unless needed. If we only want this on systems with DEBUG=TRUE, then see: https://django-debug-toolbar.readthedocs.io/en/1.0/installation.html

```
if settings.DEBUG:
    import debug_toolbar
    urlpatterns += patterns('',
        url(r'^__debug__/', include(debug_toolbar.urls)),
    )
```

Lucas-Mc:
Thanks @tompollard, I moved the URL import into a conditional based on the debug settings.

tompollard:
Thanks Lucas, this looks good to me. 

Do you want to set the debug flag to false on staging too, as suggested by @bemoody? https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/settings/staging.py

It's easy enough to temporarily switch the flag to true on staging when necessary for debugging. In most cases, I think we only really need debug=true when on dev.

Lucas-Mc:
Thanks for the clarification Tom! I was just waiting to see if changing the `DEBUG` flag would affect anything on staging but since not I can switch now.

elfeto:
I was reading a bit on this (debug toolbar)
There are several things needed for this to work, as stated here:
https://django-debug-toolbar.readthedocs.io/en/latest/installation.html

>  MIDDLEWARE += ['debug_toolbar.middleware.DebugToolbarMiddleware', ]

>import debug_toolbar
urlpatterns = [
    ...
    path('__debug__/', include(debug_toolbar.urls)),
]

>INTERNAL_IPS = [
    # ...
    '127.0.0.1',
    # ...
]

and obviously, `DEBUG=True`
Without these things, it wont work... 

The only place the `DebugToolbarMiddleware` and `INTERNAL_IPS` are set in  `physionet/settings/development/base.py` not in staging.
Even if the URL is added it wont work without that.

I would say its safe to assume that the DEBUG can stay in staging, BUT we might want to make staging as close to prod as possible removing the DEBUG option.

tompollard:
Thanks Felipe -

> ... we might want to make staging as close to prod as possible removing the DEBUG option.

Nice point, this makes a good case for switching off DEBUG on staging.

bemoody:
> Nice point, this makes a good case for switching off DEBUG on staging.

Yeah, I think that's probably a good idea, but could be done
independently of this pull request.  I'm happy either way.

Issue #1312, opened 2021-03-29
Autosave not cleared when form is saved


bemoody:
The autosave plugin will try to automatically save the text you've entered in localStorage.  However, it doesn't delete the auto-saved text when the form is submitted.

Examples of how this can be a problem:

- Go to /projects/create/, enter an abstract, and submit the form; the new project is created.  Go to /projects/create/ to create another project.  Now the form field is blank, so it gives a warning.

- Go to /projects/SHuKI1APLrwWCqxSQnSk/content/ and click the table button to create a table.  Drag the handle at the bottom of the table to make it taller.  Submit the form.  In the autosaved text (and even in the HTTP response), the table has a "height" attribute, but ckeditor strips this attribute out so it doesn't match, and it gives you a warning.

- For various reasons it is desirable to make automated changes to the HTML content which will cause the text to differ from what was submitted.

One possibility would be to set the "saveDetectionSelectors" config property, e.g. if you set that property to "button[name='edit_description']" then the plugin will clear its local storage when a button with that name is clicked.  This isn't ideal, since clicking the button doesn't mean the data was received or saved by the server.  Better would be to somehow clear the storage in the response *after* the form is saved.

tompollard:
@bemoody thanks for picking this up. Odd, I did some testing where I thought I saw the autosaved data in the browser being deleted when I clicked submit. I'll try to look into this.

PR #1314, opened 2021-03-30
Clear autosave data when a form is submitted

bemoody:
The autosave plugin will automatically save rich text fields to your browser's local storage.  (Well, most of the time it will.  Sometimes it seems to just not do anything and I'm not sure why.)  If you close the window, and reopen the form later, it will ask if you want to restore the autosaved text.

However, when the form is actually *saved* to the server, we want to remove the old autosave data, because the text on a subsequent page load may be different (either because the text was modified by bleach or other server-side filters, or because it's modified by ckeditor, or because you edited the same project in another browser, etc.)

So we add an additional script, which should be loaded as part of the *next* page (the page that's loaded when you submit the form - not necessarily the same page containing the form itself), that deletes the autosaved content.  We tell it which autosaved content to delete by setting a cookie in the response to the POST request.

This handles the following forms:

 - creating a project (/projects/create/)
 - editing a project (/projects/*/content/)
 - creating a news item (/console/news/add/)
 - editing a news item (/console/news/edit/*/)
 - making a data acccess request (/request-access/*/*/)
 - copyediting a project (/console/submitted-projects/*/copyedit/)

This does *not* handle any of the Django admin forms.  If you have access to the admin forms you are expected to know what you're doing.

One thing I am not entirely sure about is whether this script will correctly delete the cookie if the cookie is marked as "secure".  It might be that you need to do something like:

    document.cookie = 'saved_cke_fields=;path=/;SameSite=strict;' +
        'expires=Thu, 01 Jan 1970 00:00:00 GMT;secure';

if the existing cookie is secure, and:

    document.cookie = 'saved_cke_fields=;path=/;SameSite=strict;' +
        'expires=Thu, 01 Jan 1970 00:00:00 GMT';

if it isn't.  I haven't yet tested that.

This should fix issue #1312.

### DIFF
--- a/physionet-django/console/templates/console/copyedit_submission.html
+++ b/physionet-django/console/templates/console/copyedit_submission.html
@@ -10,6 +10,7 @@
 
 {% block local_js_top %}
 <script src="{% static 'custom/js/cookie.js' %}"></script>
+<script src="{% static 'custom/js/clear-autosave.js' %}"></script>
 <script src="{% static 'project/js/dynamic-formset.js' %}"></script>
 <script src="{% static 'custom/js/copy-to-clipboard.js' %}"></script>
 {% endblock %}
--- a/physionet-django/console/templates/console/news_console.html
+++ b/physionet-django/console/templates/console/news_console.html
@@ -4,6 +4,11 @@
 
 {% block title %}News Items{% endblock %}
 
+{% block local_js_top %}
+<script src="{% static 'custom/js/cookie.js' %}"></script>
+<script src="{% static 'custom/js/clear-autosave.js' %}"></script>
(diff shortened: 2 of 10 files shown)


bemoody:
> One thing I am not entirely sure about is whether this script will correctly delete the cookie if the cookie is marked as "secure".

The script seems to work fine as-is, though it'd still be worth checking if the browser behavior is different for *cookies sent from an actually secure server* vs. cookies sent from a local server with a self-signed certificate.

tompollard:
Nice work, thanks for fixing this. Works well on Staging, though I haven't checked the secure cookie issue. We can address this later if necessary.

PR #1315, opened 2021-04-05
Fixes editor assignment for projects with single quotes

Lucas-Mc:
This change fixes the Editor Assignment button for projects where the title includes a single quotation (e.g. `Lucas's Project`). Currently, the single quote is being rendered as `&#39;` in the HTML which causes problems so I added the `escapejs` filter which turned it to `\u0027` which fixes the issue.


bemoody:
Good catch.

This shows why it's dangerous to embed data in inline scripts (either
in script tags or in "onclick" attributes, it's much the same).  A much
better way to write this would be to store the project title as a data
attribute:

   <a data-project-title="{{ project }}" onclick="...">

and have the event handler function ('set_project_id') retrieve that
attribute's value.

PR #1316, opened 2021-04-05
Reload nginx when new certificates are installed

bemoody:
When certbot installs a new certificate, nginx needs to be told to "reload" in order to start using it.  When using the "webroot" aka "don't-mess-with-my-server-configuration" renewal mechanism, this reloading doesn't happen automatically.

Of course, it's not easy to test something that only happens every two months.  This script has been installed on the staging server for a few weeks, and appears to have done its job.


tompollard:
thanks benjamin

PR #1327, opened 2021-04-22
Add a custom location for all logs in the background tasks

elfeto:
At the moment all logs for the Django background tasks are redirected stdout and stderr to the logging system which are kinda lost.

Here I add a custom capture to get all those logs onto the default Django logging location.

### DIFF
--- a/deploy/common/etc/rsyslog.d/background_tasks.conf
+++ b/deploy/common/etc/rsyslog.d/background_tasks.conf
@@ -1,0 +1,3 @@
+# Filter to get all background tasks and send them to a custom location
+:msg, contains, "background_task.tasks" -/data/log/background_tasks/django_background_tasks.log
+& ~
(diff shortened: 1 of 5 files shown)


bemoody:
Filtering on message text seems very fragile.  If you want to use
syslog then I think a better way would be to specify "SyslogIdentifier"
in the unit file and filter based on that.

Another option would be for the unit file to write to the log directly
(StandardError=append:/data/log/background_tasks/django_background_tasks.log
or something like that).

bemoody:
> Filtering on message text seems very fragile.  If you want to use
> syslog then I think a better way would be to specify
> "SyslogIdentifier" in the unit file and filter based on that.

Something like this should work
(/etc/systemd/system/django-background-tasks.service):

[Service]
...
StandardError=syslog
SyslogIdentifier=django-background-tasks
...

and (/etc/rsyslog.d/background_tasks.conf):

:programname, isequal, "django-background-tasks" -/data/log/background_tasks/django_background_tasks.log

bemoody:
"StandardError=syslog" is already there, it doesn't need to be there
twice. :)

bemoody:
Thanks!

Issue #1328, opened 2021-04-22
GitHub action not running test cases


bemoody:
For some reason, the GitHub action (see pull #1320) is not actually running the test cases.

Looking at the most recent test for dev (run 775028130), the log shows:

```
2021-04-22T17:10:19.5797807Z   Applying user.0032_orcid... OK
2021-04-22T17:10:19.5798504Z   Applying user.0033_credentialapplication_reference_organization... OK
2021-04-22T17:10:21.8898721Z Installed 29 object(s) from 1 fixture(s)
2021-04-22T17:10:21.8899393Z Installed 4 object(s) from 1 fixture(s)
2021-04-22T17:10:21.8899948Z Installed 4 object(s) from 1 fixture(s)
2021-04-22T17:10:21.8900509Z Installed 894 object(s) from 3 fixture(s)
2021-04-22T17:10:21.8901070Z Copied demo media and static files.
2021-04-22T17:10:22.8453864Z 
2021-04-22T17:10:22.8454663Z Skipping setup of unused database(s): default.
2021-04-22T17:10:22.8456044Z ----------------------------------------------------------------------
2021-04-22T17:10:22.8456522Z Ran 0 tests in 0.000s
2021-04-22T17:10:22.8457052Z 
2021-04-22T17:10:22.8457341Z OK
2021-04-22T17:10:22.8457767Z System check identified no issues (0 silenced).
2021-04-22T17:10:24.1753968Z 
2021-04-22T17:10:24.1754810Z System check identified no issues (0 silenced).
2021-04-22T17:10:24.1756699Z ----------------------------------------------------------------------
2021-04-22T17:10:24.1757443Z Ran 0 tests in 0.000s
2021-04-22T17:10:24.1757840Z 
2021-04-22T17:10:24.1758282Z OK
2021-04-22T17:10:24.9554225Z Name                                                            Stmts   Miss  Cover   Missing
2021-04-22T17:10:24.9555820Z ---------------------------------------------------------------------------------------------
2021-04-22T17:10:24.9556875Z physionet-django/console/__init__.py                                0      0   100%
2021-04-22T17:10:24.9557839Z physionet-django/console/admin.py                                   1      0   100%
```

I haven't ever seen the message "Skipping setup of unused databases" before so I have no idea what that means.

But clearly "Ran 0 tests" is not what we want. :/

bemoody:
One issue is that DJANGO_SETTINGS_MODULE is not being set.

[...]

PR #1330, opened 2021-04-22
Fix for tests not running in the Github action

briangow:
Before calling `manage.py` do `cd physionet-django` since calling it from the directory above was not running the tests.  Add `DJANGO_SETTINGS_MODULE` ENV variable.

### DIFF
--- a/.github/workflows/physionet-build-test.yml
+++ b/.github/workflows/physionet-build-test.yml
@@ -46,9 +46,12 @@ jobs:
           sudo -u postgres psql -c "create database physionet;" -U postgres
       - name: install and setup wfdb
         run: |
-          wget https://github.com/bemoody/wfdb/archive/master.tar.gz -O wfdb.tar.gz
+          wget https://github.com/bemoody/wfdb/archive/10.6.2.tar.gz -O wfdb.tar.gz
           tar -xf wfdb.tar.gz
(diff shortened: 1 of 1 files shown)


.github/workflows/physionet-build-test.yml
```diff
@@ -46,12 +46,9 @@ jobs:
           sudo -u postgres psql -c "create database physionet;" -U postgres
       - name: install and setup wfdb
         run: |
-          wget https://github.com/bemoody/wfdb/archive/10.6.2.tar.gz -O wfdb.tar.gz
+          wget https://github.com/bemoody/wfdb/archive/master.tar.gz -O wfdb.tar.gz
```
tompollard: it would be good to pin a version if possible, so that we can keep the tests in sync with whatever version we are running on our live servers.

briangow: Thanks for the suggestion, I have brought this up in #1328 .

bemoody:
> it would be good to pin a version if possible, so that we can keep
> the tests in sync with whatever version we are running on our live
> servers.

Yeah.  I think 10.6.2 is what is installed on the servers.  It doesn't
make a huge difference (for now) what exact version we use but it is
better to pin it to a particular version.

briangow:
I've switched back to wfdb v10.6.2.

Issue #1258, opened 2021-02-08
Shippable Ending May 3, 2021


Lucas-Mc:
It looks like Shippable is ending their services effective March 3rd, 2021.... How should we handle the transition?

Here's some more info:
https://blog.shippable.com/the-next-step-in-the-evolution-of-shippable-jfrog-pipelines

tompollard:
I think GitHub Actions does what we need. 

https://lab.github.com/githubtraining/github-actions:-continuous-integration

https://docs.github.com/en/actions/learn-github-actions

It looks like there is a Django Continuous Integration template that might be helpful for getting started: https://github.com/actions/starter-workflows/blob/c59b62dee0eae1f9f368b7011cf05c2fc42cf084/ci/django.yml

PR #1335, opened 2021-05-04
Remove Shippable references from README

Lucas-Mc:
This change removes the Shippable references from the README now that it is deprecated (see #1258). Maybe we should think about finding a way to reproduce these with Github Actions?

This may help: https://dev.to/thejaredwilcurt/coverage-badge-with-github-actions-finally-59fa


tompollard:
Thanks Lucas (think the badges are more effort than they are worth for now!)

PR #1337, opened 2021-05-05
Adjusts credential reject email to clarify comments

Lucas-Mc:
This change adjusts the reject email which is sent when credential applications are rejected. Currently, the comments written by the reviewer are hidden beneath the standard list of suggestions which makes it difficult to understand exactly went wrong with the application since most of the things on the list would have been competed correctly.

This change only displays that list for cases where comments are disabled (KP's workflow in which comments are for internal use only). For the new workflow, since the reviewer is required to submit comments in order to reject an application, these comments will be the only things added to the email, not the list as it is currently.

It also keeps this line present for both emails (which was only present in the new workflow previously):
```
If you are able to address these issues, please open a new
credentialing application at
https://physionet.org/credential-application/.
```

### DIFF
--- a/physionet-django/notification/templates/notification/email/process_credential_complete.html
+++ b/physionet-django/notification/templates/notification/email/process_credential_complete.html
@@ -1,7 +1,7 @@
 {% load i18n %}{% autoescape off %}{% filter wordwrap:70 %}
 Dear {{ applicant_name }},
 
-{% if application.status == 1 %}Thank you for applying for credentialed access to PhysioNet. Your application was not approved. This may be for one of the following reasons:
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks Lucas. The opening and closing text ("This may be for one of the following reasons:..."; "If you are able to address these issues...") expects multiple issues, but the reviewer comment may refer to a single issue or multiple issues. Could you tweak the wording to deal with this?

Lucas-Mc:
Thanks @tompollard for the nice find! I tweaked the wording a bit to make it more general, but not sure if it's the best way to tweak it

sophie53:
thanks Lucas, it is more clear having the comments shown on the top. when would that be changed? I mean to the applicants.

tompollard:
thanks! let's give this a try.

PR #1339, opened 2021-05-12
Quickfix: Change the source for downloading zipfiles from Google Cloud to local servers

tompollard:
Users can choose to download a zipfile contained the contents of a project. If the project has been uploaded to Google Cloud, then a cloud bucket is the default source for download. 

Screen Shot 2021-05-12 at 11 38 56

Due to budget constraints, this is a temporary change that switches the default download server to local storage.

### DIFF
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -364,7 +364,12 @@
         <ul>
           {% if project.gcp %}
             {% if project.gcp.sent_zip %}
-              <li><a href="https://storage.googleapis.com/{{ project.gcp.bucket_name }}/{{ project.zip_name }}">Download the ZIP file</a> ({{ compressed_size }}).</a>
+            <!-- Temporary fix: Change default location to local servers -->
+            <!-- 
+              <li><a href="https://storage.googleapis.com/{{ project.gcp.bucket_name }}/{{ project.zip_name }}">
+                  Download the ZIP file</a> ({{ compressed_size }}).</a></li>
+             -->
+              <li><a href="{% static project.zip_url %}">Download the ZIP file</a> ({{ compressed_size }})</li>
             {% elif project.compressed_storage_size %}
               <li><a href="{% static project.zip_url %}">Download the ZIP file</a> ({{ compressed_size }})</li>
             {% endif %}


elfeto:
As conversed this will stay until we get more credits, and then should be reversed.

Lucas-Mc:
I may be wrong, but I thought the standard HTML comments (`<!--` and `-->`) didn't work for Django templates? Doesn't it have to be `{% comment %}` and `{% endcomment %}`?

tompollard:
@Lucas-Mc well spotted and now you mention it I also would expect the HTML comments not to work around the template tags. Oddly it looks like they do in this case (the change is live):

e.g. see: view-source:https://physionet.org/content/eicu-crd/2.0/

```
<h5>Access the files</h5>
<ul>
<li><a href="/content/eicu-crd/get-zip/2.0/">Download the ZIP file</a> (5.1 GB)</li>
<li><a href="/projects/eicu-crd/2.0/request_access/3">Request access</a> to the files using the <a href="https://console.cloud.google.com/storage/browser/eicu-crd-2.0.physionet.org/">Google Cloud Storage Browser</a>. Login with a Google account is required.</li>
<li><a href="/projects/eicu-crd/2.0/request_access/4">Request access</a> using Google BigQuery.</li>
<li>Download the files using your terminal: <pre class="shell-command">wget -r -N -c -np --user tompollard --ask-password https://physionet.org/files/eicu-crd/2.0/</pre></li>
</ul>
```

Hopefully we can revert the change shortly, so I'll leave this for now unless anyone has concerns.

bemoody:
> I may be wrong, but I thought the standard HTML comments (`<!--` and
> `-->`) didn't work for Django templates? Doesn't it have to be `{%
> comment %}` and `{% endcomment %}`?

Django doesn't do anything special with HTML comments, so you can use
HTML comments and it will copy them into the output.  If you actually
want an HTML comment, that's fine (it's useful sometimes.)

However, it is dangerous to put any template commands inside an HTML
comment (because, obviously, Django doesn't understand this is a
comment.)

In general you should avoid "commenting out" - if you want to remove
something, then remove it; keeping history is what git is for.

tompollard:
> In general you should avoid "commenting out" - if you want to remove
something, then remove it; keeping history is what git is for.

Yep, fair point!

PR #1340, opened 2021-05-12
Change gsutil for requesterpays

elfeto:
Here we specify in the gsutil command the project that will be charged for the GCP transactions.

### DIFF
--- a/physionet-django/notification/templates/notification/email/notify_gcp_access_request.html
+++ b/physionet-django/notification/templates/notification/email/notify_gcp_access_request.html
@@ -7,7 +7,7 @@ The URL for the storage bucket in GCP is:
 https://console.cloud.google.com/storage/browser/{{ project.gcp.bucket_name }}
 
 You can use the following command to copy the content of the bucket to your computer:
-gsutil -m cp -r gs://{{ project.gcp.bucket_name }} DESTINATION
+gsutil -m -u YOUR_PROJECT_ID cp -r gs://{{ project.gcp.bucket_name }} DESTINATION
 
 {% else %}
 You have requested access to use {{ project }} in GCP BigQuery.
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -375,7 +375,7 @@
             {% endif %}
             {% if project.gcp.sent_files %}
               <li>Access the files using the Google Cloud Storage Browser <a href="https://console.cloud.google.com/storage/browser/{{ project.gcp.bucket_name }}/">here</a>. Login with a Google account is required.
-              <li>Access the data using Google Cloud "gsutil": <pre class="shell-command">gsutil -m cp -r gs://{{ project.gcp.bucket_name }} DESTINATION</pre></li>
+              <li>Access the data using Google Cloud "gsutil": <pre class="shell-command">gsutil -m -u YOUR_PROJECT_ID cp -r gs://{{ project.gcp.bucket_name }} DESTINATION</pre></li>
             {% endif %}
           {% elif project.compressed_storage_size %}
             <li><a href="{% static project.zip_url %}">Download the ZIP file</a> ({{ compressed_size }})</li>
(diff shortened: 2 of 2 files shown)


tompollard:
Thanks Felipe. Before this change, I get a requestor pays message when trying to download:

```
gsutil -m cp -r gs://eicu-crd-2.0.physionet.org .

BadRequestException: 400 Bucket is requester pays bucket but no user project provided.
CommandException: 1 file/object could not be transferred.
```

After this change I get a "requires authentication" message.

```
gsutil -m -u my-billing-project cp -r gs://eicu-crd-2.0.physionet.org .

ServiceException: 401 Requester pays bucket access requires authentication.
CommandException: 1 file/object could not be transferred.
```

Is there anything we can do to clarify how the user should authenticate?

bemoody:
eicu-crd-2.0 *should* require authentication, no?  That first command
never should have worked.

elfeto:
So we are talking about different things here.

> Is there anything we can do to clarify how the user should authenticate?

I would say read the gsutil documentation and how to configure it...
There exist a lot of tutorials and documentations in GCP about how to use this. 

> eicu-crd-2.0 *should* require authentication, no?  That first command
never should have worked.

Most projects are open to the public, but in the case for the eicu-crd-2.0 and MIMIC its not.

For public buckets one only needs the command with a GCP project and it will work.
For the non public buckets one needs to be part of an access list in which there is the credentialed process ect...
Once, a user is credentialed, and he request access to the GCP MIMIC and/or eICU, then the email from that user is added to the access list granting him access to the files.

tompollard:
Good point, and I guess eicu-crd wasn't the best example. https://physionet.org/content/ptb-xl/1.0.1/ is an example of a public dataset.

`gsutil -m -u physionet-data cp -r gs://ptbdb-1.0.0.physionet.org .`

It sounds like adding the authentication flag (+ instructions) is a good idea for two reasons: (1) requestor pays (2) access controlled datasets.

tompollard:
> I would say read the gsutil documentation and how to configure it...

Could we link to the documentation to help people out with this? e.g. Something like:

```HTML
Access the data using Google Cloud (please refer to 
the <a href="https://cloud.google.com/storage/docs/gsutil/">gsutil</a> documentation for guidance): 
```

bemoody:
> I would say read the gsutil documentation and how to configure it...
> There exist a lot of tutorials and documentations in GCP about how to
> use this. 

I took a very quick look at the documentation and didn't see anything
obvious, so I'm not at all surprised that people would find this
confusing.  Can you give a 30-second version?

elfeto:
If you follow the steps 1-6 in the section of `Authenticate Google Cloud SDK` the following link, it will show how to initialize the `gsutil` command and how to select a default project.
https://cloud.google.com/storage/docs/gsutil_install

That link also helps for the installation of the command.

If you guys want, we can add some instructions for this later because we would need to think where to put them...
I thought of a news item, but that in time will be lost among news...

This is the point a FAQ or Tutorial section comes to mind :D

tompollard:
> If you follow the steps 1-6 in the section of Authenticate Google Cloud SDK the following link, it will show how to initialize the gsutil command and how to select a default project.
https://cloud.google.com/storage/docs/gsutil_install

The gsutil_install link looks like a good one to me (especially the "Setting Up Credentials to Access Protected Data" bit).

> This is the point a FAQ or Tutorial section comes to mind :D

yeah, we need to add the tutorial project type (and merge https://github.com/MIT-LCP/physionet-build/pulls :/)

bemoody:
> If you follow the steps 1-6 in the section of `Authenticate Google
> Cloud SDK` the following link, it will show how to initialize the
> `gsutil` command and how to select a default project.
> https://cloud.google.com/storage/docs/gsutil_install

Perfect.  Could we add something like this in the email message as well:

    To access this bucket from the command line, you will need to set
    up Google Cloud credentials as described here:
      https://cloud.google.com/storage/docs/gsutil_install#authenticate

tompollard:
I would like to merge this because it is an improvement on what we have now. I've raised an issue at https://github.com/MIT-LCP/physionet-build/issues/1344 for the suggested change to the email text.

PR #1345, opened 2021-05-18
Adds sorting to credential processing tables

Lucas-Mc:
This change allows admin users to sort the credentialing application processing table by: User, Full Name, Email, Reference Email, and Time Elapsed. Since the IDs are no longer relevant, I removed them from the table. I further added sorting icons from Font Awesome to let the user know that a column is sortable and what its status is (neutral, sorting up, sorting down).


tompollard:
Looks good, thanks Lucas (and sorry for the delay!)

Issue #1341, opened 2021-05-12
Unbranding: Remove "physionet" from docstrings


tompollard:
@shulim as a small first step towards creating our unbranded version of the repo, please could you remove all instances of "physionet" from the docstrings? If it isn't possible to remove any of these instances, please make a note (e.g. listing them in this issue).

shulim:
Here are the files where I decided to keep mentions of "physionet" because they are in docstring examples or commented code:

- https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/console/utility.py
- https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/models.py
- https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/test_browser.py
- https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/test_views.py
- https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/notification/utility.py
- https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/user/management/commands/sendnewemailtokens.py

Additionally, I kept the mention of "physionet" in https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/test_utility.py because it mentions running "physionet.utility.zip_dir," so this would be better changed later when the physionet module is renamed.

Lucas-Mc:
What about the maintenance messages in `.env.example`?
```
#SYSTEM_MAINTENANCE_MESSAGE=PhysioNet is undergoing maintenance, and projects cannot be edited. The site will be back online at 16:00 GMT.

PAUSE_CREDENTIALING_MESSAGE=PhysioNet will not be taking new applications for credentialed access until 4 January 2021. We apologize for the inconvenience.
```

tompollard:
Shulammite has compiled a list of tasks and there is a fair amount of work to do (including the points that you mention). This issue focused on docstrings. I asked Shulammite to make a note of places where she decided to leave the docstrings as they were (which is list you see above).

PR #1347, opened 2021-05-21
Removes 'physionet' from docstrings

shulim:
This addresses #1341 by removing mentions of "physionet" in docstrings. I decided to keep mentions of "physionet" as an example in some docstrings.


tompollard:
thanks!

PR #1348, opened 2021-05-27
Updated grammar

Hsaeedx:
Replaced "Awaiting Publish" tab title on admin console with "Awaiting Publication".


tompollard:
Thanks for the contribution Hasan. Looks good!

Issue #1354, opened 2021-06-02
Unbranding: Moving toward an "unbranded" platform


shulim:
There have been a number of issues raised with "Unbranding" in the name, so in this issue I will aim to tie together the existing issues, which could serve as a springboard for future unbranding-related work. In addition, after taking a look at "physionet" mentions in the code, I have come across a number of files with which I am not sure about how to deal, so I'll also include a section about them.

With the aim of moving toward a platform that can support not only PhysioNet, but other data repositories, it is important to remove explicit mentions of "physionet" in the code that may interfere with other platforms' branding. To this end, here is a list (which will likely grow) of action steps:

1. #1342: Create a config folder and move PhysioNet settings-specific files into this folder
2. #1351: Rename physionet package and update references
3. #1352: Replace email and form mentions of "PhysioNet"
4. #1353: Replace "PhysioNet" mentions in HTML templates

Note: the "demo" and "fixture" files likely do not need to be touched at this stage, since they are PhysioNet-specific examples.

Here are files that contain mentions of "physionet" about which I'm unclear, as well as some thoughts:

- https://github.com/MIT-LCP/physionet-build/blob/8f4a5548e1b54266b2c37fa4923e74e08a9bd2fe/physionet-django/user/management/commands/resetdb.py: we probably don't want to call the database "physionet" in the future, so I think we just need to supply a different database name
- https://github.com/MIT-LCP/physionet-build/blob/2f7ff6ce22a14b5b5145f8e31317a2fdc7414f32/physionet-django/project/models.py: need to somehow reformat the citations to be generalizable
- https://github.com/MIT-LCP/physionet-build/blob/8f4a5548e1b54266b2c37fa4923e74e08a9bd2fe/physionet-django/user/validators.py: I'm not clear on what this validator does, but perhaps bad_words can be made configurable, if necessary
- https://github.com/MIT-LCP/physionet-build/blob/8f4a5548e1b54266b2c37fa4923e74e08a9bd2fe/physionet-django/project/pnmigration.py: includes multiple physionet URLs
- https://github.com/MIT-LCP/physionet-build/blob/8f4a5548e1b54266b2c37fa4923e74e08a9bd2fe/physionet-django/user/test_forms.py: includes physionet URL
- https://github.com/MIT-LCP/physionet-build/blob/8f4a5548e1b54266b2c37fa4923e74e08a9bd2fe/physionet-django/lightwave/views.py: includes physionet URL
[...]

PR #1355, opened 2021-06-02
Remove unnecessary "physionet" mentions

shulim:
These small changes remove mentions of "physionet" that are not needed for code functionality or clarity, which is a small step toward addressing #1354.


tompollard:
thanks!

PR #1360, opened 2021-06-18
Added Submissions page to admin console to track when projects reached milestones

Hsaeedx:
Page can be accessed under Admin Console > Usage Stats > Submissions

For all ActiveProjects and PublishedProjects, the dates that they were created, submitted, resubmitted (if applicable)
and published (if applicable) are tracked and counted up for each month over the last 18 months.

Currently does not consider projects that have never been submitted


tompollard:
Nice, thanks for the contribution Hasan. We'll take a look!

tompollard:
Nice work, thanks @Hsaeedx! Let's work out next steps at our catch up on Monday.

PR #1361, opened 2021-06-28
Dockerize PhysioNet

grzegorz-wcislo:
## What's included in the PR?

1. Poetry config (https://python-poetry.org/) - For improved dependency version management. Currently, it is only used by the Dockerfile to build the PhysioNet image, using it to run on the MIT deployment will also require some changes in your config files.
1. settings files unification - We wanted to move the config to environment variables and keep as few specific settings files as possible (to stick to the 12-factor methodology, https://12factor.net/).
1. Dockerfile - for creating the PhysioNet image. It is reasonably optimized in terms of layers and size.
1. .dockerignore - for excluding files from the docker image (it contains most of the things from the .gitignore file and some other files that are not needed in the image). The syntax of this file is slightly different than the .gitignore file (https://zzz.buzz/2018/05/23/differences-of-rules-between-gitignore-and-dockerignore/)
1. docker-compose - used to create a development environment that uses docker containers for PhysioNet and Postgresql. It's missing some quality-of-life features for loading fixtures and running tests.

## What changes are needed at your end

The `test-update.sh` file might need some work because it uses a sqlite database.

Due to the changes to setting files you will need to update the staging and production `.env` files to include some more keys (for example `MEDIA_X_ACCEL_ALIAS='/protected'`).


briangow:
I tried following the README to setup the docker container with physionet-build.  I'm using a Macbook Pro with the M1 chip.  I get the following error when running `docker-compose build`:

```
Error: pg_config executable not found.
#12 6.953    
#12 6.953    pg_config is required to build psycopg2 from source. Please add the directory
#12 6.953    containing pg_config to the $PATH or specify the full executable path with the
#12 6.953    option:
#12 6.953    
#12 6.953      python setup.py build_ext --pg-config /path/to/pg_config build ...
#12 6.953    
#12 6.953    or with the pg_config option in 'setup.cfg'.
```

My postgres was installed locally via homebrew.  This is what I get outside the container at a command line (though I don't think Docker is depending on this anyhow):

```
which pg_config
/opt/homebrew/bin/pg_config

brew link postgresql
Warning: Already linked: /opt/homebrew/Cellar/postgresql/13.3
To relink, run:
  brew unlink postgresql && brew link postgresql
```

Does the Dockerfile need to be updated as suggested here ( https://stackoverflow.com/questions/46711990/error-pg-config-executable-not-found-when-installing-psycopg2-on-alpine-in-dock ) ?

tompollard:
@briangow have you tried the steps at: https://stackoverflow.com/questions/66888087/cannot-install-psycopg2-with-pip3-on-m1-mac ?

It looks like there are a couple of dependencies you may need to install:

```
brew install libpq --build-from-source
brew install openssl

export LDFLAGS="-L/opt/homebrew/opt/openssl@1.1/lib -L/opt/homebrew/opt/libpq/lib"
export CPPFLAGS="-I/opt/homebrew/opt/openssl@1.1/include -I/opt/homebrew/opt/libpq/include"

pip3 install psycopg2
```

grzegorz-wcislo:
@briangow Turns out the `psycopg2-binary` package was problematic (it used a precompiled binary on amd64 but tried to compile from source on arm64 architecture), I changed it to `psycopg2` which always compiles from source and installed the required compilation dependencies in the Dockerfile.

You shouldn't have to install postgresql locally for the docker-compose commands to work.

briangow:
Thanks @tompollard , @grzegorz-wcislo !  After pulling the update, I am able to run through the rest of the setup in the README.  However, when I run:


```
(base) ➜  physionet-build git:(upsidelab-dockerize) ✗ docker-compose exec dev /bin/bash
root@54c0bce026e5:/code# cd *django
root@54c0bce026e5:/code/physionet-django# python manage.py resetdb
```

I observe the following error message:


```
dropdb: database removal failed: ERROR:  database "physionet_dev" is being accessed by other users
DETAIL:  There is 1 other session using the database.
createdb: database creation failed: ERROR:  database "physionet_dev" already exists
Removed all media files and targeted static files.
Operations to perform:
  Apply all migrations: admin, auth, background_task, contenttypes, notification, project, sessions, sites, user
Running migrations:
  No migrations to apply.
```

I don't see any errors in the database shell when I save new content from http://localhost:8000/ though.  Is this error because the `docker-compose exec dev /bin/bash` is connected to the setup from `docker-compose up` ?

Also, I do not get the error message above when running:


```
(base) ➜  physionet-build git:(upsidelab-dockerize) ✗ docker-compose exec test /bin/bash
root@ed03cd256496:/code# cd *django
root@ed03cd256496:/code/physionet-django# python manage.py resetdb
Removed all media files and targeted static files.
Operations to perform:
  Apply all migrations: admin, auth, background_task, contenttypes, notification, project, sessions, sites, user
Running migrations:
  Applying user.0001_initial... OK
.
.
.
```

grzegorz-wcislo:
I'm happy to hear that the changes worked!

When running the `docker-compose up` command the `dev` container is running a Django web server. Because of that, it's impossible to delete the database (because the server is connected to it). To reset the development database you have to stop the containers and run a temporary container using `docker-compose run --rm dev /bin/bash` and reset the database this way. After you reset the development database you can exit the container and run docker-compose up again.

The test container does not run a web server and because of that resetting, the database is simpler (the `dev `and `test` containers use separate databases).

Keep in mind you start with an empty and migrated database and you don't have to reset it right away.

briangow:
Ok, makes sense, thanks for the explanation @grzegorz-wcislo !

Hsaeedx:
I was able to get this working on my Windows machine, and it's a significant improvement over the virtual environment. Running a virtual environment using Ubuntu made it such that loading pages took around 7 seconds, whereas page loading in the dockerized version is near instantaneous.

Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM python:3.9-slim
```
tompollard: It looks like the development environment is using 3.9. @bemoody has asked whether we can match this to the version used in tests (3.7)?

.github/workflows/physionet-build-test.yml
```diff
@@ -26,16 +44,17 @@ jobs:
           apt-get update --yes
       - name: install and configure needed software
```
tompollard, line 31: It's kind of nice to have  `ln -sT .env.example .env` step that takes the `.env.example` file and uses this for the test build. Is there a reason why we don't move all/most of the environment variables above to the .env.example?

I could see it may make sense to include the PhysioNet specific variables in this`xxx-build-test.yml` workflow, with the idea that there would be additional workflows for other implementations (e.g. `tcairem-build-test.yml`). If this is the idea, then should we have a separate test workflow that builds from the .env.example defaults?

physionet-django/physionet/settings/production.py
```diff
@@ -1,68 +0,0 @@
-import os
```
tompollard: Overall we agree that moving to a single `settings.py` module makes sense. 

It would be helpful to keep the original staging.py and production.py modules temporarily to give us time to switch the servers to the new configuration. Is this possible?

docker-compose.yml
```diff
@@ -0,0 +1,86 @@
...
+      - ./physionet-django:/code/physionet-django:rw
+      - test_data:/data:rw
+    networks:
+      - test
+  db:
+    image: postgres:13-alpine
+    environment:
+      POSTGRES_USER: physionet
+      POSTGRES_PASSWORD: password
+      DEV_DB: physionet_dev
```
tompollard, line 65: Please could you briefly explain the purpose of the two separate databases here? (`DEV_DB` and `TEST_DB`)

tompollard:
@grzegorz-wcislo thanks, this is looking good. @bemoody, @briangow and I briefly ran though things together. I've posted a few comments/questions above that came up during the discussion. Most of the points are pretty minor. 

The only real worry was that without staging.py and production.py,  our servers will break. Could we keep these files in place temporarily to give time to update our build process on the live servers?

Dockerfile
grzegorz-wcislo: Sure, it's a good idea to bump the python version in the future, though

docker-compose.yml
grzegorz-wcislo: The reason for using 2 databases is pretty simple: the test suit requires a very specific database contents (from `loaddemo`). Using a separate database for development makes it possible to freely change the development database contents (adding/deleting users, projects etc.) while ensuring that the tests will not see those changes. Because of that there is no need to run subsequent `resetdb` or `loaddemo` on the test database after making changes via `localhost:8000`.

.github/workflows/physionet-build-test.yml
grzegorz-wcislo: That's a good point, I'll try to extract the common configuration from here and the `docker-compose` file to `.env.example`. There will be slight differences between CI and running locally.

physionet-django/physionet/settings/production.py
grzegorz-wcislo: Yes, I'll bring them back

grzegorz-wcislo:
@tompollard @bemoody @briangow I pushed the changes we discussed earlier, I'll squash them into the previous commits after your re-review :)

.env.example
```diff
@@ -1,7 +1,27 @@
...
+# Database
+DB_USER=physionet
+DB_PASSWORD=password
+DB_HOST=localhost
+DB_NAME=physionet
+
+# Emails
+EMAIL_HOST=localhost
+DEFAULT_FROM_EMAIL=localhost@localhost
+CONTACT_EMAIL=PhysioNet Contact <contact@dev.physionet.org>
```
tompollard: not sure whether this changes anything functionally, but perhaps good for use of quotes to be consistent with the line below.

tompollard:
Thanks @grzegorz-wcislo, this addresses the concerns that we raised yesterday. @bemoody, please say by the end of today if you have any additional requests before this is merged!

.env.example
grzegorz-wcislo: Fixed!

bemoody:
> Thanks @grzegorz-wcislo, this addresses the concerns that we raised
> yesterday.

Agreed, I appreciate your quick work on this!

> @bemoody, please say by the end of today if you are not
> okay for this to be merged.

Unfortunately this isn't something we can rush. ;)  Even ignoring all
the docker stuff, there are non-trivial changes that need to be
considered and tested.

It's important to continue running tests using the same packages used
in production.  That means,
in .github/workflows/physionet-build-test.yml, we want to use 'pip3
install -r requirements.txt'.  It's still not clear to me what exactly
"poetry" does or why we would want to use it; but that change, if it's
desirable, is something that should be addressed separately.

Removing all of the physionet-django/physionet/settings/development/
also breaks a number of things.  It's important to be able to run a
demo server using an unprivileged user account (no containers or VMs
required.)  It's useful to be able to set up experimental servers using
'git clone', back them up using 'tar', and delete them using 'rm' (no
screwing around with postgresql required.)  Obviously such
functionality could be re-added later, but I'd rather not remove it in
the first place.

(Again, I am in favor of moving everything to a single universal
settings module, and removing these old modules once their functionality
is integrated into the universal one.  I think that keeping the old
modules is the easiest way forward for now.)

It looks like DATACITE_PREFIX, DATACITE_USER, DATACITE_PASS,
ORCID_CLIENT_ID, ORCID_CLIENT_SECRET, and ORCID_SCOPE are defined in
physionet.settings.production but not in physionet.settings.staging (or
physionet.settings.development.*).  This configuration is ugly (and
right now I don't even care about datacite/orcid stuff actually working
in development/staging); maybe the best thing would be to move the
definitions from physionet/settings/settings.py to
physionet/settings/base.py ?  That is to say, just forget about
"_TEST_" and use the same variable names everywhere.

grzegorz-wcislo:
@bemoody I think I addressed your concerns with environment variables and configuration files. I also adjusted the github actions to test against both requirements.txt and poetry.

bemoody:
From some quick tests this looks good.

It would probably be wise to add the new variables to .env (and update
the datacite/orcid stuff) on staging and production before pushing,
though I don't expect it to cause any problems either way.  I can take
care of that tomorrow unless somebody else wants to do it. :)

tompollard:
@bemoody happy to merge? we can do some more testing on staging before it goes to the live server.

tompollard:
@grzegorz-wcislo this is good to go! do you want to rebase/squash before we merge?

grzegorz-wcislo:
@tompollard I squashed the changes, everything should be good to go :)

tompollard:
thanks!

PR #1365, opened 2021-07-07
Fixed Key error where months outside of 18 month range were accessed

Hsaeedx:
If projects were created, submitted, or published within the same year before 18 months
previous to the current month, the program would attempt to index into that month despite
being out of range (i.e. if the current month is July of 2021, then 18 months ago would be
February 2020. If a project was submitted in January 2020, the program would attempt to
access that month, even though it's outside the 18 month period, causing a Key Error).


tompollard:
thanks hasan!

PR #1369, opened 2021-07-13
Refactoring project models - Part 1

cx1111:
First few PRs will be mainly splitting up content into multiple modules without any functional changes. Then there will be proper logic refactoring.

To ensure nothing changed in the DB, I tested with `./manage.py makemigrations --dry-run` after every commit.

If everyone agrees with the approach shown in this PR, I will follow up with the rest of the project models.


tompollard:
Please could you add a little more detail about the bigger plan? I don't have any major objections to this restructuring, but it's difficult to decide whether it is helpful without more context.

cx1111:
There's going to be a lot of work on adding support for different filesystem backends like S3 and GCS for projects. Currently, the logic is very difficult to follow, extend, and test, especially for new contributors such as the Upside team. I don't blame them since the `models.py` file has about 2500 lines.

I want to refactor a lot of the projects code (without changing the functionality) as follows:

1. Split code into different modules by functionality.
2. Refactor code dealing with project-specific files into isolated modules. I've seen a couple of view functions which should be converted to instance methods, and even a couple of file path validation bugs which need to be fixed.
3. Where there is a large amount of similar logic for file processing between different project types, I want to use abstract classes and methods to enforce that each class has its own overrides.

Basically I want to split up the code, clean up project file logic, fix file processing bugs, and add scaffolding to enable the codebase to be more efficiently extended.

physionet-django/project/forms.py
```diff
@@ -772,8 +770,8 @@ def __init__(self, editable=True, **kwargs):
         licenses = License.objects.filter(
             resource_types__icontains=str(self.instance.resource_type.id))
         # Set allowed access policies based on license policies
-        available_policies = [a for a in range(len(Metadata.ACCESS_POLICIES)) if licenses.filter(access_policy=a)]
-        self.fields['access_policy'].choices = tuple(Metadata.ACCESS_POLICIES[p] for p in available_policies)
+        available_policies = [a for a in range(len(ACCESS_POLICIES)) if licenses.filter(access_policy=a)]
```
JerzySpendel: This will fail if somebody decides to change the definition of `ACCESS_POLICIES` to something like:
```
    ACCESS_POLICIES = (
        (2, 'Open'),
        (4, 'Restricted'),
        (6, 'Credentialed'),
    )
```
I know there's no reason to, but it would be safer it this line of code doesn't assume that the first element is also an index of that element in the collection. So I would propose something like that:
```suggestion
        available_policies = [policy[0] for policy in ACCESS_POLICIES if licenses.filter(access_policy=policy[0]).exists()]
```
Additionally .exists() might be a bit more efficient than evaluation to bool in the if statement.

cx1111: Updated. The old code was indeed very strange.

physionet-django/project/models.py
```diff
@@ -4,40 +4,36 @@
...
+from django.db import models, transaction
 from django.forms.utils import ErrorList
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.html import format_html, strip_tags
 from django.utils.text import slugify
 from background_task import background
-from django.utils.crypto import get_random_string
 
+from project.modelcomponents.authors import (AuthorInvitation, Affiliation, PublishedAffiliation,
```
tompollard: Not a big deal but slightly bothersome that any new classes in the modelcomponents modules will need to be explicitly imported here (this may not be immediately obvious to people when updating the modules) but I guess (1) there's not much we can do about that other than maybe making a note somewhere and (2) those modules should be relatively stable once the refactoring is done anyway.

tompollard:
i'm okay with this if it makes the refactoring easier. @bemoody please could you add your thoughts?

tompollard:
@cx1111 just had a chat with Benjamin and he is good to merge too. Will do the merge after https://github.com/MIT-LCP/physionet-build/pull/1361.

bemoody:
I have no problem with moving code around if it makes things easier to
maintain, but I'm not sure about the specifics here.  In particular it's
not clear what "generic.py" is for, or why it's called that.

As far as SafeHTMLField is concerned, I think most Django packages use
"fields.py" for custom fields so perhaps that's what we should do.
Though, I always get confused between "model field" and "form field"
classes so I may be wrong about this.

As for BaseInvitation, wouldn't it make more sense to put that in
authors.py?

physionet-django/project/models.py
cx1111: We could also `import *` for these particular instances. Might be more maintainable. What do you think?

cx1111:
> I have no problem with moving code around if it makes things easier to maintain, but I'm not sure about the specifics here. In particular it's not clear what "generic.py" is for, or why it's called that. As far as SafeHTMLField is concerned, I think most Django packages use "fields.py" for custom fields so perhaps that's what we should do. Though, I always get confused between "model field" and "form field" classes so I may be wrong about this. As for BaseInvitation, wouldn't it make more sense to put that in authors.py?

I've moved `SafeHTMLField` to `fields.py`. Good idea.

I'll leave `BaseInvitation` in `generic.py` for now since `StorageRequest` will inherit from it. If it makes sense to restructure, we can do so in follow-up PRs.

Issue #1362, opened 2021-06-29
Clarify purpose of the "Usage notes" section in the author guidelines


tompollard:
In our [author guidelines](https://physionet.org/about/publish/), we currently say:

> Usage notes: This section should provide the reader with information relevant to reuse. This might include technical instructions or ideas for projects that could make use of the resource. Please describe the following: file structure and organization; file naming conventions; issues users should be aware of (e.g. known issues and limitations) in using your resource. You can also provide specific examples of file structure and organization.

This doesn't really align with our current thinking about what the section should include. I'd suggest changing the text to something like:

> Usage notes: This section should provide the reader with information relevant to reuse. In particular we suggest discussing: (1) how the data has already been used (citing relevant papers); (2) the reuse potential of the dataset; (3) known limitations that users should be aware of when using the resource; and (4) any complementary code or datasets that might be of interest to the user community.

tompollard:
@UBDM In case it is helpful, I've added some brief notes below. Give me a shout if you have any questions!

1) Take a look at the contribution guidelines in the [project readme](https://github.com/MIT-LCP/physionet-build#readme), including the linked resources (PEP8; tips for pull requests)
2) Make sure that you're able to run PhysioNet on your local machine.
3) Open a shell and `git pull` to make sure you have the latest version of the code.
4) Create a new branch for your feature (`git switch -c fix_usage_notes`)
5) Find the relevant template and update the text. Check the new page looks okay locally.
6) `git status` will allow you to review the files that you've changed.
7) `git diff FILENAME` will give you a diff of the changes on FILENAME.
8) Run `python manage.py test` to check that the tests are still passing (note: you'll need to make sure you have a fresh version of the database before running the tests).
9) `git add <filename>` to prepare the changes for commit.
10) `git commit -m "Your commit message goes here"` to commit the changes. It is good to reference the issue in your commit message too, with a `#` followed by the issue number. e.g. `Clarify usage notes section. Fixes #1363`.
11) `git push --set-upstream origin fix_usage_notes` to push the change to GitHub
12) Go to https://github.com/MIT-LCP/physionet-build/pulls and open a new pull request. The clearer the title and description in your pull request, the more likely you'll get a smooth review! At minimum, please briefly (1) explain why the change is necessary (2) summarize the change. 
13) Switch back to the development branch so that you're ready for your next issue (`git switch dev`).
[...]

PR #1374, opened 2021-07-20
Updates Usage notes section in the author guidelines. ref #1362

danamouk:
This is a change to the Usage notes supplementary text (physionet-django->templates->about->author guidelines.html) to reflect the updated text. The updated text includes information on "(1) how the data has already been used (citing relevant papers); (2) the reuse potential of the dataset; (3) known limitations that users should be aware of when using the resource; and (4) any complementary code or datasets that might be of interest to the user community". ref #1362.


tompollard:
Looks good to me, thanks Dana!

PR #1379, opened 2021-08-06
GCS Object Storage

grzegorz-wcislo:
In this PR we add an alternative way to host all the static, media and project files using Google Cloud Storage.

All feedback is greatly appreciated!


bemoody:
I'm not going to get into the details at this point; the following is just a high-level review of your approach.

Question zero is why not use gcs-fuse (https://cloud.google.com/storage/docs/gcs-fuse)?  I have no experience with it so perhaps it has serious flaws that make it unworkable in this context.  If it does work, though, it could save a *huge* amount of effort.

Putting "if" statements all over the place is not a good idea.  For one thing it means the logic is scattered: mixing different levels of logic makes it difficult to modify or debug either the low-level logic (e.g. adding a new storage backend) or the high-level logic (e.g. a new mechanism for uploading/editing files.)  Secondly, having everything keyed on a single global settings variable means you cannot have *some* projects using one storage backend while *others* use another, which means migrating from one backend to another is impossible.

What you should do instead is to define an abstract class that encapsulates all project file operations, and define multiple implementations (derived classes), one that uses a local filesytem backend and one that uses GCP.

A basic version might look like this:
```
    class ProjectFiles:
        def create_file(self, path, content):
            raise NotImplementedError

        def delete_file(self, path):
            raise NotImplementedError

        def rename_file(self, old_path, new_path):
            raise NotImplementedError

    class LocalProjectFiles(ProjectFiles):
        def __init__(self, project_root):
            self.project_root = project_root

        def create_file(self, path, content):
            path = os.path.join(self.project_root, path)
            utility.write_uploaded_file(file=content,
                write_file_path=path, overwrite=False)

        def delete_file(self, path):
            path = os.path.join(self.project_root, path)
            utility.remove_items([path], ignore_missing=False)

        def rename_file(self, old_path, new_path):
            old_path = os.path.join(self.project_root, old_path)
            new_path = os.path.join(self.project_root, new_path)
            utility.rename_file(old_path, new_path)

    class GCPProjectFiles(ProjectFiles):
        def __init__(self, bucket_id):
            self.bucket_id = bucket_id

        def create_file(self, path, content):
            ...
```
(obviously this needs a lot more methods and documentation.)

Then, the PublishedProject and UnpublishedProject classes should each have a method (or property) that returns an appropriate ProjectFiles instance for that project.

Methods like PublishedProject.file_root() will be deprecated; everything that formerly used file_root() will need to be changed to use one of the ProjectFiles methods instead.

We will need to have some way of testing the code.  Encapsulation like the above is a good start: if we can see that LocalProjectFiles and GCPProjectFiles have the same API, and we're confident that they work the same way, then we don't need to test every site feature using both backends.  At the same time, we do want to actually test the functionality of each individual backend, which means we will want some way of mocking GCP for testing.  Perhaps this could be done with 'requests_mock' as we do for other remote services.

bemoody:
Installing this version of `cryptography`, using the Debian 10 version of `pip`, apparently requires compiling it from source.  I think this will be unnecessary after upgrading to Debian 11; for now, please add "rustc" and "libssl-dev" to the `apt-get install` command in `deploy/test-server/install-pn-test-server`.

docker-compose.yml
```diff
@@ -70,3 +71,14 @@ services:
     networks:
       - dev
       - test
+    ports:
+      - 5432:5432
+
+  gcs:
```
grzegorz-wcislo, line 79: You could add this container to the `depends_on` in the test container

physionet-django/physionet/test_gcs.py
```diff
@@ -0,0 +1,207 @@
...
+from google.api_core.client_options import ClientOptions
+from google.auth.credentials import AnonymousCredentials
+from google.cloud.exceptions import NotFound
+from google.cloud.storage import Client, Bucket, Blob
+
+from physionet.gcs import GCSObject, GCSObjectException
+from physionet.settings.base import StorageTypes
+
+
+SKIP__GCS_INTEGRATION = os.environ.get('TEST_GCS_INTEGRATION') == 'false'
```
grzegorz-wcislo: I'd use decouple here like so:

```py
from decouple import config

SKIP__GCS_INTEGRATION = not config('TEST_GCS_INTEGRATION', default=False, cast=bool)
```

By the way, is there a reason for the double underscore here?

physionet-django/project/modelcomponents/activeproject.py
```diff
@@ -80,7 +84,11 @@ class ActiveProject(Metadata, UnpublishedProject, SubmissionInfo):
     MAX_SUBMITTING_PROJECTS = 10
     INDIVIDUAL_FILE_SIZE_LIMIT = 10 * 1024**3
     # Where all the active project files are kept
-    FILE_ROOT = os.path.join(settings.MEDIA_ROOT, 'active-projects')
+
+    FILE_ROOT = os.path.join(
+        (settings.MEDIA_ROOT if settings.STORAGE_TYPE == StorageTypes.LOCAL else settings.GCP_STORAGE_BUCKET_NAME),
```
grzegorz-wcislo: Maybe we can add a property like `FILE_ROOT` to the ProjectFile classes to get rid of this

requirements.txt
```diff
@@ -1,198 +1,384 @@
-#### Requirements for running the server ####
-
-bleach==3.3.0 \
- --hash=sha256:6123ddc1052673e52bab52cdc955bcb57a015264a1c57d37bea2f6b817af0125 \
- --hash=sha256:98b3170739e5e83dd9dc19633f074727ad848cbedb6026708c8ac2d3b697a433
+bleach==3.3.1; (python_version >= "2.7" and python_full_version < "3.0.0") or (python_full_version >= "3.5.0") \
```
grzegorz-wcislo: Most of those changes seem unnecessary. I think we should only modify the 2 changed libraries according to `poetry.lock`

grzegorz-wcislo (commented):
Some small comments

bemoody:
Some package hashes appear to be missing/incorrect:
```
THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    google-auth-httplib2==0.1.0 from https://files.pythonhosted.org/packages/ba/db/721e2f3f32339080153995d16e46edc3a7657251f167ddcb9327e632783b/google_auth_httplib2-0.1.0-py2.py3-none-any.whl#sha256=31e49c36c6b5643b57e82617cb3e021e3e1d2df9da63af67252c02fa9c1f4a10 (from -r ../requirements.txt (line 188)):
        Expected sha256 a07c39fd632becacd3f07718dfd6021bf396978f03ad3ce4321d060015cc30ac
             Got        31e49c36c6b5643b57e82617cb3e021e3e1d2df9da63af67252c02fa9c1f4a10

    google-crc32c==1.1.2 from https://files.pythonhosted.org/packages/95/88/edeaf3260740bad1e56e3a6f4f08a5f4a797ca64dcdc97da5596dc2e9350/google_crc32c-1.1.2-cp37-cp37m-manylinux1_x86_64.whl#sha256=34a97937f164147aefa53c3277364fd3bfa7fd244cbebbd5a976fa8325fb496b (from -r ../requirements.txt (line 195)):
        Expected sha256 e6458c41236d37cb982120b070ebcc115687c852bee24cdd18792da2640cf44d
             Got        34a97937f164147aefa53c3277364fd3bfa7fd244cbebbd5a976fa8325fb496b
```

bemoody:
I haven't finished looking over the whole diff, but it looks reasonable so far.  Here are a few small issues and suggestions:

## physionet-django/console/forms.py
```
+        if settings.STORAGE_TYPE == StorageTypes.GCP:
+            del self.fields['make_zip']
```
Don't use `if settings.STORAGE_TYPE == ...`.  Instead, define a method to check whether ProjectFiles supports `make_zip`.

Also, instead of deleting the field, it would probably be better to make it hidden and set a default value:
```
        if not ProjectFiles().can_make_zip():
            self.fields['make_zip'].widget = forms.HiddenInput()
            self.fields['make_zip'].initial = 0
```

## physionet-django/console/views.py
```
-            published_project = project.publish(slug=slug,
-                make_zip=int(publish_form.cleaned_data['make_zip']))
+            
+            make_zip = publish_form.cleaned_data.get('make_zip')
+            make_zip = int(make_zip) if make_zip else False
+
+            published_project = project.publish(slug=slug, make_zip=make_zip)
```
This should be handled by making the form field hidden (see above.)

```
-        elif 'make_checksum_file' in request.POST:
+        elif 'make_checksum_file' in request.POST and settings.STORAGE_TYPE == StorageTypes.LOCAL:
```
Again, don't use `if settings.STORAGE_TYPE == ...`.  In this case the submission button(s) should be hidden if the relevant features are unsupported.  There is no need for the view function to check storage type here.

```
-         'legacy_author_form': legacy_author_form})
+         'legacy_author_form': legacy_author_form,
+         'storage_type': settings.STORAGE_TYPE})
```
Instead of passing `storage_type`, pass flags for specific capabilities:
```
         'can_make_zip': ProjectFiles().can_make_zip(),
         'can_make_checksum_file': ProjectFiles().can_make_checksum_file(),
```

## physionet-django/physionet/urls.py

I don't think it's wise to conditionally add URL patterns - rather, URLs that are unsupported by the configuration should return a 404 error.  Note also that you *haven't* removed all of the lightwave URLs here.

## physionet-django/physionet/views.py

As in `manage_published_project` above, 'storage_type' should not be passed as a template parameter.  I think the only reason this is being done here is to suppress the LightWAVE links on front-page projects - is that correct?

The better way to do that would be to define a method of the PublishedProject object (or maybe a utility function in the lightwave module), to determine whether the project supports LightWAVE.

grzegorz-wcislo:
Reply from @fiedosiukr :

## physionet-django/console/forms.py
If we decided to hide the field instead of removing it, there still would be a possibility to edit the field via the inspect element or send a request manually. In this case it doesn't change anything, because zipping the file is not implemented. The form is located in the admin console, so probably a random user will not have access to the form, but in my opinion it is wiser to keep it removed.

The `can_make_zip()` method makes a lot of sense and I will add it.

## physionet-django/console/views.py
```
-           published_project = project.publish(slug=slug,
-                make_zip=int(publish_form.cleaned_data['make_zip']))
+            
+            make_zip = publish_form.cleaned_data.get('make_zip')
+            make_zip = int(make_zip) if make_zip else False
+
+            published_project = project.publish(slug=slug, make_zip=make_zip)
```

If we keep the `make_zip` field deleted this part probably shouldn't be changed. The field will not be present all the time, so we need to handle the situation when it is not present.
```
-        elif 'make_checksum_file' in request.POST:
+        elif 'make_checksum_file' in request.POST and settings.STORAGE_TYPE == StorageTypes.LOCAL:
```
You are right. The second part of the statement doesn't make sense at all. The `make_checksum_file()` and `make_zip()` methods do nothing when GCP storage type is used.

## physionet-django/physionet/urls.py
If a url is not added to the `urlpatterns` , it will simply return an 404 error. There is no need to modify the view itself. The debug toolbar is also added conditionally.

## physionet-django/physionet/views.py
In some cases we will need to pass the `storage_type` to the template in order to disable some template parts. Currently we use it to hide lightwave and wget stuff. Probably adding a custom method like `is_lightwave_supported()` to the `ProjectFiles` would be a better idea. The other option would be to add a custom context processor and the `storage_type` would be added to all templates.

bemoody:
> If we decided to hide the field instead of removing it, there still
> would be a possibility to edit the field via the inspect element or
> send a request manually. In this case it doesn't change anything,
> because zipping the file is not implemented. The form is located in
> the admin console, so probably a random user will not have access to
> the form, but in my opinion it is wiser to keep it removed.

If you want to guard against clients submitting an unsupported value,
then you can set the field disabled (`self.fields['make_zip'].disabled
= True`).

I didn't suggest that because if someone decides to tamper with an
admin form and it results in an error, that sounds like a perfectly
fine outcome to me. :)

But the point is to keep the logic for generating and parsing the form
within the form class.  Having the *view function* silently ignore the
absence of what is supposed to be a mandatory field doesn't seem like a
good idea to me.

> If a url is not added to the `urlpatterns` , it will simply return an
> 404 error. There is no need to modify the view itself. The debug
> toolbar is also added conditionally.

Maybe the entire "app" should be disabled?  I don't know.  I really
don't think the decision of whether a view is incidentally unsupported
belongs in an unrelated file.

The debug toolbar is messy, but the debug toolbar also *isn't a
user-facing API*.

> In some cases we will need to pass the `storage_type` to the template
> in order to disable some template parts.

That's the problem, though.  Checking the value of "storage_type",
either in Python logic or in template logic, is almost never the right
thing to do.

bemoody:
The `ProjectFiles.open` method returns a tuple of a file object and
the size of the file.  This seems messy, I think a function called
`open` ought to return a file object and nothing else (files have lots
of attributes; why is size special?)

There are then two possibilities for how to determine the file size
when needed:

1. `ProjectFiles.open` could store the size as an attribute of the
   file object:

```
class LocalProjectFiles(BaseProjectFiles):
    def open(self, path, mode='rb'):
        infile = open(path, mode)
        infile.size = os.stat(infile.fileno()).st_size
        return infile

...

class FileView:
    def __init__(self, project, path, file):
        self.project = project
        self.path = path
        self.file = file
        self._basename = os.path.basename(path)
        self._size = file.size
        self._url = project.file_url('', path)
```

2. `FileView.__init__` could determine the size by seeking:

```
class LocalProjectFiles(BaseProjectFiles):
    def open(self, path, mode='rb'):
        return open(path, mode)

...

class FileView:
    def __init__(self, project, path, file):
        self.project = project
        self.path = path
        self.file = file
        self._basename = os.path.basename(path)
        self._size = file.seek(0, os.SEEK_END)
        file.seek(0)
        self._url = project.file_url('', path)
```

fiedosiukr:
> If you want to guard against clients submitting an unsupported value, then you can set the field disabled (`self.fields['make_zip'].disabled = True`). I didn't suggest that because if someone decides to tamper with an admin form and it results in an error, that sounds like a perfectly fine outcome to me. :) But the point is to keep the logic for generating and parsing the form within the form class. Having the *view function* silently ignore the absence of what is supposed to be a mandatory field doesn't seem like a good idea to me.

That makes sense. I disabled the field instead of deleting it.

> Maybe the entire "app" should be disabled? I don't know. I really don't think the decision of whether a view is incidentally unsupported belongs in an unrelated file.

Disabling and enabling the application would still depend on the storage type or on an additional environment variable. Conditional appending LightWave URLs to the `urlpatterns` would still be necessary.

> That's the problem, though. Checking the value of "storage_type", either in Python logic or in template logic, is almost never the right thing to do.

It seems like it is going to be necessary in some of the cases.

> The ProjectFiles.open method returns a tuple of a file object and
the size of the file. This seems messy, I think a function called
open ought to return a file object and nothing else (files have lots
of attributes; why is size special?)

This part wasn't implemented by me, but you are completely right. There is nothing special about the file size. I made the changes, I went with the first possibility.

bemoody:
```
THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    google-auth-httplib2==0.1.0 from https://files.pythonhosted.org/packages/ba/db/721e2f3f32339080153995d16e46edc3a7657251f167ddcb9327e632783b/google_auth_httplib2-0.1.0-py2.py3-none-any.whl#sha256=31e49c36c6b5643b57e82617cb3e021e3e1d2df9da63af67252c02fa9c1f4a10 (from -r ../requirements.txt (line 245)):
        Expected sha256 a07c39fd632becacd3f07718dfd6021bf396978f03ad3ce4321d060015cc30ac
             Got        31e49c36c6b5643b57e82617cb3e021e3e1d2df9da63af67252c02fa9c1f4a10

```

fiedosiukr:
> ```
> THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
>     google-auth-httplib2==0.1.0 from https://files.pythonhosted.org/packages/ba/db/721e2f3f32339080153995d16e46edc3a7657251f167ddcb9327e632783b/google_auth_httplib2-0.1.0-py2.py3-none-any.whl#sha256=31e49c36c6b5643b57e82617cb3e021e3e1d2df9da63af67252c02fa9c1f4a10 (from -r ../requirements.txt (line 245)):
>         Expected sha256 a07c39fd632becacd3f07718dfd6021bf396978f03ad3ce4321d060015cc30ac
>              Got        31e49c36c6b5643b57e82617cb3e021e3e1d2df9da63af67252c02fa9c1f4a10
> ```

It's strange, because the hash is valid. I added one additional hash. You can view the hashes [here](https://pypi.org/project/google-auth-httplib2/#files).

alistairewj:
Maybe a cached corrupted install package?

tompollard:
> Maybe a cached corrupted install package?

That's what someone who'd secretly modified the package and was trying to hide their tracks would say.

alistairewj:
@bemoody @tompollard this PR has been updated with all the style checks required by #1421 and now passes - it's ready for final checks!

fiedosiukr:
@tompollard @bemoody I think we are ready to finally merge this pull request.

I checked the user application and the profile picture uploads work properly. I decided not to create a separate pull request regarding the flake8 configuration, because we would need to merge it first and then rebase this branch once again. I will not include unrelated changes in the next pull requests.

bemoody:
Thanks, @fiedosiukr.  Unfortunately, there are a couple of things in your last commit that still don't look quite right to me.

The first issue is about how we move/copy the files at publication time.  In physionet-django/project/modelcomponents/activeproject.py, you now have:
```
        ProjectFiles().publish(self, published_project)

        try:
            with transaction.atomic():
                ...
                # Remove the ActiveProject
                self.delete()

                return published_project

        except Exception:
            # Move the files to the active project directory
            if settings.STORAGE_TYPE == StorageTypes.LOCAL:
                os.rename(published_project.file_root(), self.file_root())
            elif settings.STORAGE_TYPE == StorageTypes.GCP:
                ProjectFiles().rm_dir(published_project.file_root())
            raise

        if settings.STORAGE_TYPE == StorageTypes.GCP:
            ProjectFiles().rm_dir(self.file_root())
```

First, don't change "except" to "except Exception" here; the original code was correct.  If for any reason the transaction is not completed, then the active project directory should be rolled back to its original state.

Second, in the GCP case you want to delete the active project files, after the transaction succeeds but *before* returning from the function.  So "return published_project" should be moved down to the very bottom.

What would be better, I think, is to implement this using *three* functions in ProjectFiles:

```
class LocalProjectFiles(BaseProjectFiles):
    ...
    def publish_initial(self, active_project, published_project):
        # Move project files from active to published.
        if not os.path.isdir(published_project.project_file_root()):
            os.mkdir(published_project.project_file_root())
        os.rename(active_project.file_root(), published_project.file_root())

    def publish_complete(self, active_project, published_project):
        # Publication finished - nothing more to do.
        pass

    def publish_rollback(self, active_project, published_project):
        # Publication failed - move files from published back to active.
        os.rename(published_project.file_root(), active_project.file_root())
    ...

class GCSProjectFiles(BaseProjectFiles):
    ...
    def publish_initial(self, active_project, published_project):
        # Copy project files from active to published.
        bucket_name = published_project.project_file_root()
        try:
            create_bucket(bucket_name)
        except Conflict:
            # FIXME: I don't think this exception should be ignored!
            # If the bucket already exists, for whatever weird reason,
            # then publication should fail.  Otherwise you might end
            # up with content in the published project that wasn't
            # what the author/editor approved.
            pass

        active_project_path = self._dir_path(active_project.file_root())
        published_project_path = self._dir_path(published_project.file_root())

        GCSObject(active_project_path).cp_dir_content(GCSObject(published_project_path), None)

    def publish_complete(self, active_project, published_project):
        # Publication finished - delete the active project files.
        self.rm_dir(active_project.file_root())

    def publish_rollback(self, active_project, published_project):
        # Publication failed - delete the published project files.
        self.rm_dir(published_project.file_root())
        # FIXME: this should also delete the bucket, which was created
        # by publish_initial above.
    ...
```

Then the above code in publish() becomes:
```
        ProjectFiles().publish_initial(self, published_project)

        try:
            with transaction.atomic():
                ...
                # Remove the ActiveProject
                self.delete()

        except:
            ProjectFiles().publish_rollback(self, published_project)
            raise

        ProjectFiles().publish_complete(self, published_project)
        return published_project
```

bemoody:
The second issue is about `active_project_storage_used`, `published_project_storage_used`, and `_storage_used`.  You haven't documented what exactly these functions are expected to return, and it's not consistent:

- `GCSProjectFiles.active_project_storage_used` returns a single integer.

- `GCSProjectFiles.published_project_storage_used` returns a single integer.

- `GCSProjectFiles._storage_used` returns a tuple of two integers, the second of which isn't used.

- `LocalProjectFiles.active_project_storage_used` returns a single integer.

- `LocalProjectFiles.published_project_storage_used` returns a tuple of two integers.

The existing function `PublishedProject.storage_used` is expected to return two integers (total size of the raw files, and size of the zip file if any), but that function is not a great example to emulate. :)

fiedosiukr:
> First, don't change "except" to "except Exception" here; the original code was correct. If for any reason the transaction is not completed, then the active project directory should be rolled back to its original state.

Leaving a bare except is an anti-pattern, even flake8 complains about that. The functionality doesn't get changed by adding Exception. All of the exceptions thrown by the inner block of transaction.atomic() derive from Exception class.

> class GCSProjectFiles(BaseProjectFiles):
>     ...
>     def publish_initial(self, active_project, published_project):
>         # Copy project files from active to published.
>         bucket_name = published_project.project_file_root()
>         try:
>             create_bucket(bucket_name)
>         except Conflict:
>             # FIXME: I don't think this exception should be ignored!
>             # If the bucket already exists, for whatever weird reason,
>             # then publication should fail.  Otherwise you might end
>             # up with content in the published project that wasn't
>             # what the author/editor approved.
>             pass

The buckets are currently created in the following way: `f"physionet-{project_slug}"`. Each version of a project is stored in the exactly the same bucket. If the exception wasn't ignored, we wouldn't be able to publish a new project version.

> The second issue is about active_project_storage_used, published_project_storage_used, and _storage_used. You haven't documented what exactly these functions are expected to return, and it's not consistent.

Each method returns a tuple now.

bemoody:
> Leaving a bare except is an anti-pattern, even flake8 complains
> about that. The functionality doesn't get changed by adding
> Exception. All of the exceptions thrown by the inner block of
> transaction.atomic() derive from Exception class.

Not so.  KeyboardInterrupt can theoretically occur at any time.  If a
KeyboardInterrupt occurred here, we would want to roll back the
filesystem state.

Why is "except:" scary?  Because it's never correct to handle all
exceptions in the same way.  "except Exception:" is almost always wrong
for the same reason.

But here we are not *handling* exceptions at all, we are only doing
cleanup in case of a nonlocal exit.  Here, we very specifically want to
do that cleanup in every possible case *unless* the function completed
normally.  If Python had a syntax for "finally-if-exiting-nonlocally"
then we would use that; it doesn't, so instead we write "except: ...
raise".

If you prefer, you could write "except BaseException:" instead of
"except:", though I disagree that the extra verbosity is helpful.

bemoody:
> The buckets are currently created in the following way:
> `f"physionet-{project_slug}"`. Each version of a project is stored in
> the exactly the same bucket. If the exception wasn't ignored, we
> wouldn't be able to publish a new project version.

Understood.  This is something to address in pull #1411, but I thought
it worth pointing out since it jumped out at me.

fiedosiukr:
> If you prefer, you could write "except BaseException:" instead of
"except:", though I disagree that the extra verbosity is helpful.

I'm fine with both solutions, but when using a bare except,  we have to ignore another flake8 rule - [E722](https://www.flake8rules.com/rules/E722.html).

bemoody:
OK with me either way.  @tompollard, do you think it is better to
ignore these errors in flake8, or to change "except" to "except
BaseException"?

tompollard:
@bemoody My preference is to name the exception to avoid having to deal with the Flake8 rule, but either is fine with me! Easy to change later :)

tompollard:
Thinking about it a little more, it's definitely good to have Flake8 looking out for bare excepts so my preference is to keep the rule unless strong reason not to.

bemoody:
> > The second issue is about active_project_storage_used,
> > published_project_storage_used, and _storage_used. You haven't
> > documented what exactly these functions are expected to return, and
> > it's not consistent.
> 
> Each method returns a tuple now.

Unfortunately, this is still unclear and inconsistent.  I don't
see why any of these methods should be returning a tuple.

If you think there is a compelling reason for doing so, please document
(in the doc strings of BaseProjectFiles.active_project_storage_used and
BaseProjectFiles.published_project_storage_used) exactly what values
these methods are expected to return.

fiedosiukr:
> Unfortunately, this is still unclear and inconsistent. I don't see why any of these methods should be returning a tuple. If you think there is a compelling reason for doing so, please document (in the doc strings of BaseProjectFiles.active_project_storage_used and BaseProjectFiles.published_project_storage_used) exactly what values these methods are expected to return.

Fixed.

fiedosiukr:
@tompollard @bemoody Are we ready to merge it?

tompollard:
@fiedosiukr yes, I believe so! @bemoody would you like to merge?

bemoody:
@tompollard, please go ahead and merge if you like.  However, unless
something urgent comes up, I would avoid pushing this to the production
server until next week.

tompollard:
@bemoody Thanks, I'll merge later today, after a final test. 

> I would avoid pushing this to the production server until next week.

Got it, I'll hold back as suggested.

alistairewj:
> @tompollard, please go ahead and merge if you like. However, unless something urgent comes up, I would avoid pushing this to the production server until next week.

we have a deployment on GCP right now so we'll let you know if anything comes up!

PR #1380, opened 2021-08-16
Made simple change towards unbranding

SichengH:
This is my first commit to Physionet-building. I would like to start with something small before jumping onto the real issues. 

This change is very simple, just got rid of the unnecessary words "to Physionet" in the sentence. The intuition is this may help with unbranding, and makes the platform generalizable.


tompollard:
Looks good, thanks!

PR #1382, opened 2021-08-25
Updates text for PhysioNet unbranding. ref #1380

danamouk:
Updated the text on the Share page under "Sharing on Physionet" heading from "We invite you to share your resources with the PhysioNet community to "We invite you to share your resources with the research community."  This is to help with the unbranding of PhysioNet, similar to #1380.


tompollard:
Looks good, thanks Dana!

Issue #1155, opened 2020-11-08
Add tool to the console for activating accounts


tompollard:
#1154 adds a page to the console for managing individual users. It would be good to add functionality to this page that allows us to activate (and perhaps deactivate) user accounts via the console. 

If implementing a tool for deactivating accounts, we would need to be careful that deactivation didn't lead to the account being deleted by a cron task (which could be bad, if the account was associated with projects).

PR #1385, opened 2021-08-30
Adds user activation status on user profile page. ref #1155

danamouk:
As noted in #1155, this update is the first step towards adding the activation status for a user on this page (http://127.0.0.1:8000/console/user/manage/adamfinn/) for example.

The page for a specific user can be accessed from this page (http://127.0.0.1:8000/console/users/all/) from the Users>All Users tab by clicking on a specific username under the Username column.

The update uses the is_active attribute that returns True or False for a specific user object under class User (AbstractBaseUser).

`python manage.py shell` is used to interface with the database to access the User class from physionet-build/physionet-django/user/models.py path, run methods/attributes on an object, run django queries and save any changes to a user's activation status.

The Bootstrap's CSS framework is used to make the change on the physionet-django>console>templates>console>
templates>console>user_management.html page:

```
<div class="row mb-1">
    <div class="col-md-3">
      Activation status:
    </div>
    <div class="col-md-9">
      {{ subject.is_active }}
    </div>
</div>
```

### DIFF
--- a/physionet-django/console/templates/console/user_management.html
+++ b/physionet-django/console/templates/console/user_management.html
@@ -18,6 +18,14 @@
       {{ subject.get_full_name }}
     </div>
   </div>
+  <div class="row mb-1">
+    <div class="col-md-3">
+      Activation status:
+    </div>
+    <div class="col-md-9">
+      {{ subject.is_active }}
+    </div>
+  </div>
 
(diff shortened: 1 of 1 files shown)


tompollard:
Looks good, thanks!

PR #1386, opened 2021-08-31
bump max filename length from 50 to 60 characters

tompollard:
Currently project files and folders must be a maximum of 50 characters long. This arbitrary limit is mostly fine, but I am reviewing a project with a large number of folders and files, many of which exceed the limit. To avoid having to ask the contributors to reorganise their files, I would like to bump the limit from 50 to 60.


briangow:
Looks good!

Issue #1178, opened 2020-12-30
"Deprecating" a project using the console raises a typeerror


tompollard:
It was necessary to "deprecate" a project (https://physionet.org/content/semeval2015/1.0/). I did this using the "manage" page for the published project in the console. Clicking the deprecate button raised a typeerror in the following chunk:

https://github.com/MIT-LCP/physionet-build/blob/c8f0d0ca210ac3b305d9d3a9e8592160e80bb8ad/physionet-django/project/utility.py#L162:184

```
Traceback:

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/handlers/exception.py" in inner
 34.             response = get_response(request)

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
 115.                 response = self.process_exception_by_middleware(e, request)

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
 113.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/contrib/auth/decorators.py" in _wrapped_view
 21.                 return view_func(request, *args, **kwargs)

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/contrib/auth/decorators.py" in _wrapped_view
 21.                 return view_func(request, *args, **kwargs)

File "./console/views.py" in manage_published_project
 766.                     delete_files=int(deprecate_form.cleaned_data['delete_files']))

File "./project/models.py" in deprecate_files
 1938.             self.remove_files()

File "./project/models.py" in remove_files
 1926.         clear_directory(self.file_root())

File "./project/utility.py" in clear_directory
 190.     remove_items(os.path.join(directory, i) for i in os.listdir(directory))

File "./project/utility.py" in remove_items
 184.         raise TypeError
```

PR #1391, opened 2021-09-02
Fixed error when deprecating published projects and deleting files

SichengH:
This commit fixed the problem mentioned in issue #1178. 

The problem with that issue is: a sequence of error messages will pop up when deprecating a published project and chose "Yes" in deleting the files. However, the deprecating process will be completed despite the error. 

With Tom's help, we found that the "item" that is associated with the files that are to be deleted have the types  "Generatortype" but the code only checks for "tuple" or "list". 

Adding "Generatortype" fixed the problem.

### DIFF
--- a/physionet-django/project/utility.py
+++ b/physionet-django/project/utility.py
@@ -189,7 +189,8 @@ def clear_directory(directory):
     """
     Delete all files and folders in a directory.
     """
-    remove_items(os.path.join(directory, i) for i in os.listdir(directory))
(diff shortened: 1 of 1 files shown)


bemoody:
This code is messy already, do we need to make it messier?  Why not get
rid of the inline generator expression and just say:

def clear_directory(directory):
    for i in os.listdir(directory):
        remove_items([os.path.join(directory, i)])

tompollard:
thanks, looks good to me!

PR #1393, opened 2021-09-10
Add ping route and uwsgi.ini settings for k8s deployment

grzegorz-wcislo:
* Add `/ping/` route as a kubernetes readiness probe
* Add a `uwsgi.ini` configuration for the kubernetes deployment with JSON logging (solution described here: https://www.velebit.ai/blog/tech-blog-uwsgi-json-logging/)
* Add `single-interpreter` uWSGI option to PhysioNet deployments (https://uwsgi-docs.readthedocs.io/en/latest/FAQ.html#what-are-the-most-important-options-for-performance-and-robustness-in-the-uwsgi-environment)
* Add an option to enable JSON logging in Django (with an environment variable `JSON_LOGGING`)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,7 +39,7 @@ PAUSE_CREDENTIALING_MESSAGE='PhysioNet will not be taking new applications for c
 # The Secret, can be changed under the:
 #  - GCP console -> API & Services -> Credentials
 # Changing this incorrectly will cause that nothing will be sent to GCP
-GOOGLE_APPLICATION_CREDENTIALS=json
+# GOOGLE_APPLICATION_CREDENTIALS=json
 GCP_DELEGATION_EMAIL=email
 
 # AWS
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM python:3.7-slim-buster
 
 RUN apt-get update -y \
     && apt-get upgrade -y \
-    && apt-get install build-essential libseccomp-dev libpq-dev postgresql-client wget zip -y --no-install-recommends \
+    && apt-get install build-essential libseccomp-dev libpq-dev libpcre3-dev postgresql-client wget zip -y --no-install-recommends \
     && rm -rf /var/lib/apt/lists/*
 
 RUN wget https://github.com/bemoody/wfdb/archive/10.6.2.tar.gz -O wfdb.tar.gz \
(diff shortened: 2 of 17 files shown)


bemoody:
Please split this into multiple commits that only do one thing each.

grzegorz-wcislo:
@bemoody Done!

bemoody:
91d027a7164b1111ad9b4f3f814b8d23d905c1c7

Looks good to me.

551845e1413d7a955d4f721a4a0f460321fa0c4f

This (in base.py) is already confusing to follow and you're making it more so.  It might make more sense to define an entirely separate configuration, or it might make sense to define one or more additional 'handlers' that can be switched on or off.

Anyway, if you're making major changes to the logging configuration then I'd like to see a detailed comparison of what types of messages are getting logged, where, in what formats.  In particular, the "verbose" logs are important for debugging, I don't think you want to ignore them.

4c4585393d109240a09066332b8a9de281608a82

Not yet tested.

01f37736aa97f1e5782f55bd25c0870dfb58f72e

Looks good to me.

969aab8fb11f45c41c153da133a630e32e690db1

On live and staging, .env contains:
```
GOOGLE_APPLICATION_CREDENTIALS=PhysioNet-Data-credentials.json
```
and the actual data file is in `/physionet/physionet-build/physionet-django/PhysioNet-Data-credentials.json` (also there's a `PhysioNet-Data-credentials.p12`; is that needed?)

I would assume that the environment variable ought to be set to the absolute path, and I don't particularly want to have a hardcoded default here, so I think base.py should read:
```
if config('GOOGLE_APPLICATION_CREDENTIALS', default=None):
    os.environ['GOOGLE_APPLICATION_CREDENTIALS'] = os.path.join(
        BASE_DIR, config('GOOGLE_APPLICATION_CREDENTIALS'))
```
(in other words, if this variable is already defined with an absolute path, use that; if it's defined with a relative path then resolve it relative to BASE_DIR; otherwise don't set the environment variable at all.)

grzegorz-wcislo:
@bemoody Do you use the same service account for the JSON and p12 keys? It's possible to just use the google.auth.default() credentials (which will use the JSON file specified by the GOOGLE_APPLICATION_CREDENTIALS). The usage of [ImpersondatedCredentials](https://google-auth.readthedocs.io/en/master/_modules/google/auth/impersonated_credentials.html) is required instead of `create_delegated`).

After this change, the deprecated oauth2client library will no longer be needed. Also, we want to bump the versions of the other Google-related libraries in the Object Storage PR.

grzegorz-wcislo:
I agree with the environment variable change, makes sense to me

grzegorz-wcislo:
@bemoody I pushed the GOOGLE_APPLICATION_CREDENTIALS and logging changes. I'll remove the p12 key in a separate PR.

tompollard:
@grzegorz-wcislo sorry, please could you rebase on dev?

@bemoody happy for this to be merged?

grzegorz-wcislo:
@tompollard Done!

tompollard:
Thanks @grzegorz-wcislo, looks good to me.

Issue #1364, opened 2021-07-02
Displaying "Pending project approval" in the outstanding tasks panel at /projects/


tompollard:
If a project is pending approval by an author, it is appears in the "Pending project approval" panel on the /projects page:

![Screen Shot 2021-07-02 at 10 30 17](https://user-images.githubusercontent.com/822601/124291347-7bf5a800-db22-11eb-84ab-95fa3064a21e.png)

Submitting authors have the ability to approve the project on behalf of all authors. However, if they approve the project for themselves (but not their co-authors), the project drops out of the pending-approval panel.

I think it would make sense to update our logic so that projects that are pending approval are displayed in the submitting authors panel until all authors have approved. This gives the submitting author the opportunity to approve on their colleagues behalf if necessary.

SichengH:
Looks like this issue is solved.

PR #1397, opened 2021-09-15
Maintaining pending approval panel for submitting authors when not all author approves the publication

SichengH:
When the submitting author is viewing the project page. If not all authors approve the publication, the submitting author will see and have access to the approval for publication page for other authors. 

Related issue #1364

### DIFF
--- a/physionet-django/project/views.py
+++ b/physionet-django/project/views.py
@@ -213,8 +213,11 @@ def project_home(request):
     pending_revisions = []
     for p in projects:
         if (p.submission_status == 50
-                and not p.authors.get(user=user).approval_datetime):
-            pending_author_approvals.append(p)
+                and not p.all_authors_approved()):
+            if p.authors.get(user=user).is_submitting:
+                pending_author_approvals.append(p)
+            elif not p.authors.get(user=user).approval_datetime:
+                pending_author_approvals.append(p)
         if (p.submission_status == 30
                 and p.authors.get(user=user).is_submitting):
             pending_revisions.append(p)


tompollard:
Thanks @SichengH, it would be helpful if you could give a slightly more detailed description in the pull request, so we can understand what you are aiming for. 

After making your changes, try testing as a submitting author and non-submitting author to check that the update does what you want it to do. Are you intentionally removing the "outstanding task" for non-submitting authors?

SichengH:
Updated branch, fixed a bug in the previous commit. 

Old logic:

When a project is at the pending approval stage. All authors will have the Outstanding tasks panel displayed on the project page. The action they are able to take is to click "approve the project" and then go to the submission page to approve the project. On the submission page, the submitting author has the option to approve for all the other authors; however, when the submitting author approve for themselves and then the Outstanding tasks panel disappears. 

New logic:

On whether to display the Outstanding tasks panel, if the current author is the submitting author. The code loop through all the authors on the project, and create a new variable "num_not_approval" to record how many authors didn't approve the project. If the count is not 0, the Outstanding tasks panel is displayed for the submitting author and the author has a choice to go into the submission page and approve the project on other authors' behalf. 

Test performed:

1. Reset db and load demo
2. Log in as "rgmark", submit the project "MIMIC-III Clinical Demo". Several changes are made so the project is able to submit.
3. Approve the project as "admin" and finish copy edit. 
4. Log in as "aewj" the non-submitting author on that project, see Outstanding tasks panel appears. Approve project and Outstanding task is gone. 
5. Log in as  "rgmark" see Outstanding tasks panel appears, going and approve for self. Outstanding tasks panel still appears. 
6. Log in as "george" see Outstanding tasks panel
7. Log in as "rgmark" on and approve project on behalf of "george", Outstanding tasks panel gone for both users.

bemoody:
Please write commit messages that explain the reason for each change.  Don't use `git commit -m`.

Don't put parentheses around the expression in a single-line "if" statement: for example, write `if p.submission_status == 50:` rather than `if (p.submission_status == 50):`

Always put a space after "if": `if not u.approval_datetime:`, not `if(not u.approval_datetime):`

In this case, you should use the existing method `p.all_authors_approved()` to check whether all authors have approved publication, rather than re-implementing the logic here.

physionet-django/project/views.py
```diff
@@ -212,9 +212,13 @@ def project_home(request):
     missing_affiliations = []
     pending_revisions = []
     for p in projects:
-        if (p.submission_status == 50
-                and not p.authors.get(user=user).approval_datetime):
-            pending_author_approvals.append(p)
+        if p.submission_status == 50:
```
tompollard: Removing some of the nesting might make this more readable? e.g.:

```
if p.submission_status == 50 and not p.all_authors_approved():
    if p.authors.get(user=user).is_submitting:
        pending_author_approvals.append(p)
    elif not p.authors.get(user=user).approval_datetime:
        pending_author_approvals.append(p)
```

SichengH: > Removing some of the nesting might make this more readable? e.g.:
> 
> ```
> if p.submission_status == 50 and not p.all_authors_approved():
>     if p.authors.get(user=user).is_submitting:
>         pending_author_approvals.append(p)
>     elif not p.authors.get(user=user).approval_datetime:
>         pending_author_approvals.append(p)
> ```

Thanks, Tom. This logic is much clear.

tompollard:
looks good to me, thanks sicheng!

Issue #1160, opened 2020-11-13
Notification emails should distinguish between "New project" and "New version"


tompollard:
Various emails are sent out between the submission to publication. For example "A new project has been submitted"; "A new project has been published".

These emails should be tweaked to distinguish between new projects and new versions. So for example, instead of "A new project has been submitted" perhaps we could say "A new version has been submitted".

PR #1398, opened 2021-09-21
changed 'new project' to 'new version' in the notification email

SichengH:
If a project has been previously submitted. The notification email will have the title" A new project has been updated" instead of "A new project has been submitted"

Solves issue #1160

### DIFF
--- a/physionet-django/notification/utility.py
+++ b/physionet-django/notification/utility.py
@@ -185,7 +185,10 @@ def submit_notify(project):
                   [email], fail_silently=False)
 
     # notify editorial team
-    subject = 'A new project has been submitted: {0}'.format(project.title)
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks Sicheng, I have added a few quick thoughts below:

- How about using `exists()` which will return True/False and may be slightly faster than `all()`?: 
https://docs.djangoproject.com/en/3.2/ref/models/querysets/#django.db.models.query.QuerySet.exists

- Minor thing, but perhaps switch the `if` clause to begin with the true statement for readability?:

```
if True:
    do x
else:
    do y
```

SichengH:
Thanks for the feedback. The change has been applied and tested on different scenarios.

tompollard:
Looks good, thanks!

Issue #1073, opened 2020-06-16
Ordering issue with "Approve on behalf of co-authors" step


tompollard:
We received following note from a submitting author of a project that was at the "Awaiting author approval" step:

> As I was accepting on behalf of two of my collaborators, there was a small glitch in the programming of the approval buttons. I started to approve from bottom to top. But when clicking the lowest button it approved the upper one (using Google Chrome). 

This sounds like a similar issue to https://github.com/MIT-LCP/physionet-build/issues/590.

bemoody:
I don't see anything obviously wrong with the Python.

But I do notice that in awaiting_authors_table.html, it looks like
it's creating a lot of buttons and divs with the *same* id, meaning
that when you click the "approve" button, it might not activate the
correct "dialog".

(We might be able to catch bugs like this by running an HTML validator
as part of the test suite.)

It might in any case be worth clarifying the text to say "If you are
satisfied with the final state of the project, you may approve the
publication on behalf of {{author.name}}".

SichengH:
I have recreated the bug, after many tries. Here is the tricky part, whenever you have two or more, authors that are not submitting authors are waiting for approval.  No matter which button you click, the pop-up window will show up and ask you to approve the first author. This works (1), when there is only one author needed to be approved (2) when you are approving authors in an orderly manner.

PR #1402, opened 2021-09-23
fixed approve on behalf order problem 

SichengH:
Fixed the problem described in issue #1073.

The bug happens at the author approval stage of the project submission. When a submitting author trying to approve the project on behalf of other authors and there is more than one other authors. (Picture below)

Screen Shot 2021-09-21 at 9 54 27 AM

For the two approve on behalf buttons displayed, whichever got clicked, the pop-up modal will have the author displayed on top displayed. 

The reason for this to happen is because in the template page, the code loop through all the authors' names but the 'approve-on-behalf-modal' doesn't have a unique id, therefore, every time the modal will display the first author's name and pass on the same id. 

Fixed this by adding 'author.id' as part of the 'data-target' object in the approve-on-behalf-modal.

Tested on the demo projects with three authors.

### DIFF
--- a/physionet-django/project/templates/project/awaiting_authors_table.html
+++ b/physionet-django/project/templates/project/awaiting_authors_table.html
@@ -27,13 +27,13 @@
       </td>
      {% if is_submitting and not author.approval_datetime and author.user != user %}
       <td>
-        <button id="approve-on-behalf-publication-modal-button" type="button" class="btn btn-success" data-toggle="modal" data-target="#approve-on-behalf-modal">Approve
+        <button id="approve-on-behalf-publication-modal-button" type="button" class="btn btn-success" data-toggle="modal" data-target="#approve-on-behalf-modal-{{ author.id }}">Approve
         </button>
-        <div class="modal fade" id="approve-on-behalf-modal" tabindex="-1" role="dialog" aria-labelledby="approve-on-behalf-modal" aria-hidden="true">
+        <div class="modal fade" id="approve-on-behalf-modal-{{ author.id }}" tabindex="-1" role="dialog" aria-labelledby="approve-on-behalf-label-{{ author.id }}" aria-hidden="true">
           <div class="modal-dialog" role="document">
             <div class="modal-content">
               <div class="modal-header">
-                <h5 class="modal-title">Approve on behalf of {{ author.name }}</h5>
+                <h5 class="modal-title" id="approve-on-behalf-label-{{ author.id }}">Approve on behalf of {{ author.name }}</h5>
                 <button type="button" class="close" data-dismiss="modal" aria-label="Close">
                   <span aria-hidden="true">&times;</span>
                 </button>
@@ -44,7 +44,7 @@
                     <p>If you are satisfied with the final state of the project, you may approve the publication.</p>
                   </div>
                   <div class="modal-footer">
-                    <button id="approve-publication-button" class="btn btn-success" type="submit" value="{{ author.id }}" name="approve_publication">Approve Publication</button>
+                    <button id="approve-publication-button-{{ author.id }}" class="btn btn-success" type="submit" value="{{ author.id }}" name="approve_publication">Approve Publication</button>
                     <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
                   </div>
                 </form>


tompollard:
Thanks Sicheng, looking at the example in: https://getbootstrap.com/docs/4.0/components/modal/ it sounds like fixing the `data-target` field on the button and the `id` on the modal are the necessary bits.

From reading https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute, I'm not sure that the change to `aria-labelledby=` is quite right.

For `aria-labelledby=`, \*I think\* we want to:
- change `aria-labelledby=` to something like `approve-on-behalf-label-{{ author.id }}`
- add the ID to `<h5 class="modal-title">` (i.e. set this to `<h5 class="modal-title" id="approve-on-behalf-label-{{ author.id }}">`
- (see the live demo at: https://getbootstrap.com/docs/4.0/components/modal/ for reference)

SichengH:
> Thanks Sicheng, looking at the example in: https://getbootstrap.com/docs/4.0/components/modal/ it sounds like fixing the `data-target` field on the button and the `id` on the modal are the necessary bits.
> 
> From reading https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute, I'm not sure that the change to `aria-labelledby=` is quite right.
> 
> For `aria-labelledby=`, *I think* we want to:
> 
> * change `aria-labelledby=` to something like `approve-on-behalf-label-{{ author.id }}`
> * add the ID to `<h5 class="modal-title">` (i.e. set this to `<h5 class="modal-title" id="approve-on-behalf-label-{{ author.id }}">`
> * (see the live demo at: https://getbootstrap.com/docs/4.0/components/modal/ for reference)

Aria feature changed and tested ok. 

Changed the modal title seems break the code. I am wondering why would we want to change it. The current code:

```<h5 class="modal-title">Approve on behalf of {{ author.name }}</h5>```

works fine from my perspective.

**Comment by tompollard** [comment]
> Changed the modal title seems break the code. I am wondering why would we want to change it. The current code:

What is meant by "breaks the code"? For me, the "Approve" buttons work as expected after the change.

Running the page content through the W3C validator by copying and pasting the HTML into https://validator.w3.org/nu/#textarea highlights some other issues that relate to this bug.  The "approve-publication-button" is also duplicated, so may need the same treatment?

**Comment by SichengH** [comment]
> > Changed the modal title seems break the code. I am wondering why would we want to change it. The current code:
> 
> What is meant by "breaks the code"? For me, the "Approve" buttons work as expected after the change.
> 
> Running the page content through the W3C validator by copying and pasting the HTML into https://validator.w3.org/nu/#textarea highlights some other issues that relate to this bug. The "approve-publication-button" is also duplicated, so may need the same treatment?

For "break the code" I am talking about the line with <h5...

**Comment by tompollard** [comment]
> For "break the code" I am talking about the line with <h5...

Thanks Sicheng. I'm just trying to understand what "broken" looks like. Do you get an error message, does the page fail to behave in the way you expect, does the ARIA text not render as it is supposed to, etc.

**Comment by SichengH** [comment]
Title for the pop up modal didn't show up. No error message.

**Comment by tompollard** [comment]
> Title for the pop up modal didn't show up. No error message.

Please could you try making the change again (those changes listed below), and then post the HTML for the chunk that you've edited?

> * change `aria-labelledby=` to something like `approve-on-behalf-label-{{ author.id }}`
> * add the ID to `<h5 class="modal-title">` (i.e. set this to `<h5 class="modal-title" id="approve-on-behalf-label-{{ author.id }}">`
> * (see the live demo at: https://getbootstrap.com/docs/4.0/components/modal/ for reference)

**Comment by SichengH** [comment]
```
<button id="approve-on-behalf-publication-modal-button" type="button" class="btn btn-success" data-toggle="modal" data-target="#approve-on-behalf-modal-{{ author.id }}">Approve
        </button>
        <div class="modal fade" id="approve-on-behalf-modal-{{ author.id }}" tabindex="-1" role="dialog" **aria-labelledby="approve-on-behalf-label-{{ author.id }}"** aria-hidden="true">
          <div class="modal-dialog" role="document">
            <div class="modal-content">
              <div class="modal-header">
                **<h5 class="modal-title">Approve on behalf of {{ author.name }}</h5>**
                <button type="button" class="close" data-dismiss="modal" aria-label="Close">
                  <span aria-hidden="true">&times;</span>
                </button>
```

Here is the code. 

First, I already changed the "aria-labelledby" part. Also tested using a chrome extension, it is able to read the author. 

Second, the h5 title already has "author.name", I don't understand why would we want to change that. Plus, after I changed it per request, the code breaks, where the whole modal title didn't show.

**Comment by tompollard** [comment]
To be a little clearer, I have copied the example from https://getbootstrap.com/docs/4.0/components/modal/ below. Note the `exampleModalLabel` field in this example:

```html
<!-- Button trigger modal -->
<button type="button" class="btn btn-primary" data-toggle="modal" data-target="#exampleModal">
  Launch demo modal
</button>

<!-- Modal -->
<div class="modal fade" id="exampleModal" tabindex="-1" role="dialog" aria-labelledby="exampleModalLabel" aria-hidden="true">
  <div class="modal-dialog" role="document">
    <div class="modal-content">
      <div class="modal-header">
        <h5 class="modal-title" id="exampleModalLabel">Modal title</h5>
        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
          <span aria-hidden="true">&times;</span>
        </button>
      </div>
      <div class="modal-body">
        ...
      </div>
      <div class="modal-footer">
        <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
        <button type="button" class="btn btn-primary">Save changes</button>
      </div>
    </div>
  </div>
</div>
```

In your post above, I can see that you have made first suggested change:

> change aria-labelledby= to something like approve-on-behalf-label-{{ author.id }}

The second change doesn't seem to have been made.

> add the ID to `<h5 class="modal-title">` (i.e. set this to `<h5 class="modal-title" id="approve-on-behalf-label-{{ author.id }}">`

SichengH:
New update:

On top of fixing the ordering problem when submitting authors approve projects on behalf of other authors. (Details can be found in the previous discussions)

The newest version of the PS also made sure the ARIA function is active. ARIA is short for accessible rich internet applications, where the screen reader was able to read necessary information to people with visions problems. 

In the newest update, "author.id" has been added to both the title and the text of the pop-up modal. So when a modal pops up, the user has the option to hear what is being written on the pop-up window, in this particular case, the message about "approve the project on behalf of" and also the author name.  Tested using a Chrome extension called "Screen Reader".

tompollard:
Thanks Sicheng, looks good to me! The w3c validator (https://validator.w3.org/) highlights a few warnings/errors for the page, but that's probably the case for most of our pages and we can deal with them another time.

Issue #1396, opened 2021-09-15
Decommission email address: webmaster@physionet.org


SichengH:
Since we are no longer using webmaster@physionet.org and there are quite a few spam emails that passed through and forwarded to us every day. Why don't we decommission that email address?

tompollard:
Decommissioning the email address sounds like a good idea. The email address is still occasionally used by people, so we will need to work out how to set up a bounce/autoreply that provides people with the new contact details.

PR #1403, opened 2021-09-23
changed email address webmaster to contact

SichengH:
This pull request is related to issue #1396 

We want to decommission webmaster@physionet.org and replace it with contact@physionet.org instead. Thanks to @elfeto for setting up an auto-reply to direct users to the new email address. 

This pull request replaces webmaster@physionet.org in several places in the repo with contact@physionet.org. 

Now, we can stop forwarding emails send to webmaster@physionet.org to our team members.


tompollard:
I think this can be closed. Most of the places where the email address appears are now handled in the database.

PR #1404, opened 2021-09-27
Debug Toolbar - Ignore INTERNAL_IPS in development

grzegorz-wcislo:
Currently, the debug toolbar is invisible when developing using Docker. I changed it so that in the `development` environment `INTERNAL_IPS` are not taken into account and the toolbar is always visible when `DEBUG = True`. In other environments (`staging`, `production`) `INTERNAL_IPS` has to be set to prevent unauthorized access to the debug info.

### DIFF
--- a/physionet-django/physionet/settings/settings.py
+++ b/physionet-django/physionet/settings/settings.py
@@ -27,6 +27,11 @@ if DEBUG:
     INSTALLED_APPS += ['debug_toolbar']
     MIDDLEWARE += ['debug_toolbar.middleware.DebugToolbarMiddleware']
 
+    if ENVIRONMENT == 'development' and not RUNNING_TEST_SUITE:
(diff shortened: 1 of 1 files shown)


bemoody:
Hm, for some reason I thought INTERNAL_IPS could have wildcards.

I don't think the toolbar should ever be enabled if DEBUG is false.
I'm not sure why it should be any more complicated than that.

grzegorz-wcislo:
I thought so too but sadly it does not work...

This PR changes the logic so that:
- if debug is false -> no toolbar
- if debug is true and environment is development -> toolbar
- if debug is true and other environment -> toolbar depends on INTERNAL_IPS

Maybe nesting the if calls will make this more clear (the added if only changes toolbar configuration, but if debug is false the toolbar is never added as an application or middleware)

tompollard:
thanks @grzegorz-wcislo!

PR #1407, opened 2021-10-01
Add and configure Sentry

grzegorz-wcislo:
Setup sentry according to this guide: https://docs.sentry.io/platforms/python/guides/django/.

To use the integration one has to set the `SENTRY_DSN` environment variable (the value is available after creating a project in Sentry). To control how many events are sent to Sentry set the `SENTRY_SAMPLE_RATE` to a smaller fraction (eg. 0.1 would send 1 out of 10 errors to Sentry). Setting the uWSGI option `enable-threads` is also recommended.

### DIFF
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -439,6 +439,17 @@ logging.config.dictConfig({
     },
 })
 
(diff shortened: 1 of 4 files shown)


tompollard:
Thanks @grzegorz-wcislo, looks good to me.

PR #1408, opened 2021-10-04
Updates url links for users from public to private profiles on the credentialing processing page & adds credentialing history to the user's private profile page. 

danamouk:
In this pull request two changes are made to improve the credentialing system:

1. Updating user's public profiles to private profiles that can be accessed by reviewers (admin users).
>On the credentialing processing page, http://localhost:8000/console/credential_processing/, the links attached to the username under the `user` column are changed to link to private user profiles instead of public user profiles across all tabs including `Initial review`, `Training check`, `ID check`,  `Reference check`, `Reference response`, and `Final Review`. 

2. Providing an intra-reviewer feed to help other reviewers track the log of previous reviewers.
>The credentialing history is added to the private profile page, e.g. http://localhost:8000/console/user/manage/rgmark/.
Under the `Credentialing History` heading, all the user's submitted credentialing applications are listed from the latest to the oldest. Information on the date the credentialing application was submitted, the status of the application, the reviewer and the reviewer comments are also included to guide reviewers during the credentialing process.

### DIFF
--- a/physionet-django/console/templates/console/user_management.html
+++ b/physionet-django/console/templates/console/user_management.html
@@ -76,8 +76,9 @@
   {% endif %}
 
   <br />
+  <h3>Projects</h3>
   {% for status, group in projects.items %}
-  <h3>{{ status }} projects</h3>
+  <h4>{{ status }} projects</h4>
     <ul>
       {% for project in group %}
         {% if status == "Unsubmitted" %}
@@ -99,5 +100,23 @@
     </ul>
   {% endfor %}
 
+  <h3>Credentialing history</h3>
+  <ul>
+  {% for cred_app in credentialing_app %}
+    <li><strong> Application {{ forloop.counter }}</strong><br />
+    Date: {{ cred_app.application_datetime }}<br />
+    {% if cred_app.status == 0 %} 
+      Status: Pending<br />
+    {% else %}
+      Status: {{ cred_app.get_status_display }}<br />
+    {% endif %}
+    Reviewer: {{ cred_app.responder }}<br />
+    Reviewer comments: {{ cred_app.responder_comments }}
+    </li>
+  {% empty %}
+    <li>No applications found.</li>
+  {% endfor %}
+  </ul>
+  
 </div>
 {% endblock %}
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -938,12 +938,13 @@ def user_management(request, username):
     projects['Archived'] = ArchivedProject.objects.filter(authors__user=user).order_by('-archive_datetime')
     projects['Published'] = PublishedProject.objects.filter(authors__user=user).order_by('-publish_datetime')
 
-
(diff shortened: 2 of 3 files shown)


physionet-django/console/views.py
```diff
@@ -938,12 +938,13 @@ def user_management(request, username):
     projects['Archived'] = ArchivedProject.objects.filter(authors__user=user).order_by('-archive_datetime')
     projects['Published'] = PublishedProject.objects.filter(authors__user=user).order_by('-publish_datetime')
 
-
+    credentialing_app = CredentialApplication.objects.filter(user=user, status=1).order_by("-application_datetime")
```
tompollard: Ideally the "Credentialing history" section should display all applications that have been submitted by the user. I don't think we want the "`status=1`" here, which limits to rejected applications only.

physionet-django/console/templates/console/user_management.html
```diff
@@ -99,5 +99,15 @@ <h3>{{ status }} projects</h3>
...
 
+  <h3> Credentialing history </h3> 
+  {% for cred_app in credentialing_app %}
+  <p><strong> Application {{ forloop.counter }}</strong> <br />
+  Date: {{ cred_app.application_datetime }} <br />
+  Status : {{ cred_app.get_status_display }} <br />
+  Reviewer: {{ cred_app.responder }} <br />
+  Reviewer comments: {{ cred_app.responder_comments }} <br />
+  </p>
+  {% endfor %}
```
tompollard, line 118: If there is no history, I think we want to say something like "No applications found." rather than leaving the section blank. For example of a blank section in the demo data, see http://localhost:8000/console/user/manage/ethelknox/

Before you reimplemented the changes, we had used the `for … empty` pattern (https://docs.djangoproject.com/en/3.2/ref/templates/builtins/) to do this.

physionet-django/console/templates/console/user_management.html
```diff
@@ -99,5 +99,15 @@ <h3>{{ status }} projects</h3>
     </ul>
   {% endfor %}
 
+  <h3> Credentialing history </h3>
```
tompollard: Previously we had updated the formatting of this page slightly, so that the header structure was consistent between the "Projects" section and the "Credentialing history" section.

I don't remember exactly how we had done this, but I think the general idea was that we created a new h3 called "Projects" and then the looped headers ("Unsubmitted", "Submitted" etc) were h4. e.g.

```
<h3>Projects</h3>
{% for status, group in projects.items %}
<h4>{{ status }}</h4>
    <ul>
      {% for project in group %}
        {% if status == "Unsubmitted" %}
```

We can then follow a similar pattern for the Credentialing history.

tompollard (commented):
Thanks @danamouk, I think some of your original changes are missing in this pull request. Please could you take another look?

danamouk:
thanks @tompollard for the comments, I think I have resolved the points raised.

tompollard:
Looks good, thanks Dana!

PR #1411, opened 2021-10-12
Separate buckets for different project versions

fiedosiukr:
This PR add functionality to create a separate bucket for each version of a published project. Bucket names are constructed in the following way: `<slug>-<version>.<domain>`. This naming convention allows to easily manage bucket permissions. Dots in the versions are replaced by underscores.

This branch will be rebased when #1379 is merged.


fiedosiukr:
@tompollard @bemoody This one is ready for your reviews.

bemoody:
In physionet-django/project/projectfiles/gcs.py:

    def get_project_file_root(self, slug, version, access_policy, klass):
        name = f'{slug}-{version.replace(".", "_")}'[:63]
        return f'{name}.{settings.GCP_DOMAIN}'

I think we want to be consistent with the existing GCP bucket naming scheme (for future compatibility, plus I think this looks neater): "mitdb-1.0.0.physionet.org" rather than "mitdb-1_0_0.physionet.org".  This is implemented by bucket_info() in console/utility.py.  Note that this also allows adding a prefix to the bucket name (GCP_BUCKET_PREFIX), which is useful for testing.

Also, we don't want to truncate the name to 63 characters: it should be impossible for a project slug plus version to exceed 63 characters, but if that ever did happen, it'd be better to generate an invalid name (which should result in an error from GCP) rather than silently truncating (which could result in different project versions having the same bucket name.)

Regarding the error handling at publication time:

    def publish_initial(self, active_project, published_project):
        bucket_name = published_project.project_file_root()
        try:
            create_bucket(bucket_name)
        except Conflict:
            pass

        active_project_path = self._dir_path(active_project.file_root())
        published_project_path = self._dir_path(published_project.file_root())

        GCSObject(active_project_path).cp_dir_content(GCSObject(published_project_path), None)

We want to ensure that the published bucket is an exact copy of the active project.  One possibility is to raise an exception if the bucket already exists, and delete the newly-created bucket in case of an error, i.e. something like this:

    def publish_initial(self, active_project, published_project):
        bucket_name = published_project.project_file_root()
        create_bucket(bucket_name)
        try:
            active_project_path = self._dir_path(active_project.file_root())
            published_project_path = self._dir_path(published_project.file_root())
            GCSObject(active_project_path).cp_dir_content(GCSObject(published_project_path), None)
        except BaseException:
            delete_bucket(bucket_name)
            raise

    def publish_rollback(self, active_project, published_project):
        delete_bucket(published_project.project_file_root())

Another possibility would be to reuse the existing bucket, but raise an exception if the existing bucket contains any files.  A third possibility would be to reuse the existing bucket but silently delete any existing files; this seems risky.

bemoody:
Just a very brief note on the issue I mentioned earlier: if ProjectFiles.get_project_file_root returns a project-version-specific path then it shouldn't be used for "core project" files (PublishedProject.project_file_root).

In the GCP model there won't be a "core project directory" (at least, I don't think it's needed or really makes sense.)  So I think that PublishedProject.project_file_root should be deprecated and PublishedProject.file_root should map to ProjectFiles.get_project_file_root.

If that's not totally clear, I'll try to follow up with more specific code suggestions later. :)

PR #1412, opened 2021-10-12
Feature/gcs direct files downloads

fiedosiukr:
This PR adds functionality to choose whether the files should be accessible after the publication of the project (`allow_file_downloads` field). This feature is only enabled when `ENABLE_FILE_DOWNLOADS_OPTION` environment variable is set to True.  There is also an additional environment variable `(COPY_FILES_TO_NEW_VERSION`) which defines whether to copy the files to a new project version or not.

image


fiedosiukr:
@tompollard @bemoody Ready for your reviews.

tompollard:
@fiedosiukr moving conversation from Slack:

> ...there is a problem with upgrade test workflow. This PR includes one migration file which simply adds a column to every project table. The tests are run on the old code base with the new migrations applied and 4 of them fail due to this for loop:

```
for attr in [f.name for f in Metadata._meta.fields] + [f.name for f in SubmissionInfo._meta.fields]:
    setattr(published_project, attr, getattr(self, attr))
```

> It iterates over all of Metadata's fields, but the new field (allow_file_downloads ) is not present yet and when calling save()  method, the ORM tries to insert null value into the database and an exception is raised. All of the tests pass, only the workflow fails. In order to fix that, we could make the publish()  method use PublishedProject.objects.create()  instead. How does it sound?

tompollard:
@fiedosiukr has pointed out that some of the tests are failing because the code is referencing the `allow_file_downloads` column before it has been added to the database. 

https://github.com/MIT-LCP/physionet-build/blob/76d18ba172ecaa85b353ed4afe142a572b9ad1d6/physionet-django/project/modelcomponents/activeproject.py#L224-L226

@bemoody Any thoughts on how to deal with this? The tests are at: https://github.com/MIT-LCP/physionet-build/runs/4703324880?check_suite_focus=true

fiedosiukr:
@tompollard the code is not aware of the column in the database and the ORM tries to insert null in its place. The simplest fix would be to allow the null column, but it doesn't seem to be a proper solution. 

The problem is also connected with this loop:
https://github.com/MIT-LCP/physionet-build/blob/76d18ba172ecaa85b353ed4afe142a572b9ad1d6/physionet-django/project/modelcomponents/activeproject.py#L225-L227

tompollard:
would adding the [`MIGRATE_AFTER_INSTALL = True`](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/user/management/commands/getmigrationtargets.py) to the migration fix the issue?

fiedosiukr:
> would adding the [`MIGRATE_AFTER_INSTALL = True`](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/user/management/commands/getmigrationtargets.py) to the migration fix the issue?

I have already tried that and then the code is referencing the column that is not present.

tompollard:
Hmm, let's hang on for @bemoody's thoughts then. My understanding is that these tests are helping us to avoid messing things up users with existing sessions when we run the migration, so i guess the tests are doing their job here.

@fiedosiukr please could you say more about the following suggestion? i.e. what chunk of code would you be changing to allow the null column?

> The simplest fix would be to allow the null column, but it doesn't seem to be a proper solution.

What about cleaning up the following chunk so that we only copy fields that exist on the project model. e.g. call `get_fields()` on the `ArchivedProject` and then only populating these fields?

https://github.com/MIT-LCP/physionet-build/blob/76d18ba172ecaa85b353ed4afe142a572b9ad1d6/physionet-django/project/modelcomponents/activeproject.py#L219-L228

Side issue but iI'm unclear why we create the object and then copy the fields instead of doing this all with `create()`

bemoody:
Note that this pull request isn't doing what it claims: you're not actually performing any access control based on 'allow_file_downloads'.

Setting aside the question of what this feature actually does or ought to do, I'll just address the issue of the migrations.

Recall that when the application code is installed on the server, there will be a period of time when some processes (uwsgi workers and/or django-background-tasks) are running the old application code, and some processes are running the new code.  That means that the database schema must somehow be palatable to both the old and the new application code.  If the new code expects to find a particular column, and the old code doesn't, that means the column must *exist* in the database (so the new code can query it) and it must be *nullable* (so the old code can disregard it for insertions.)  This means that (with limited exceptions) whenever you *add* a new column it must be nullable afterwards, and whenever you *remove* a column it must be nullable beforehand.

(The reason this is necessary is https://code.djangoproject.com/ticket/470 .)

This doesn't mean it's impossible to add or remove a non-nullable column: it simply needs to be done in two separate migrations.  The first (pre-install) migration creates the column; the second (post-install) migration changes it to be non-nullable.

In this case, it'd look like this:
```
# physionet-django/project/migrations/0047_allow_file_downloads_1.py

from django.db import migrations, models
    dependencies = [
        ('project', '0046_auto_20210504_1504'),
    ]
    operations = [
        migrations.AddField(
            model_name='activeproject',
            name='allow_file_downloads',
            field=models.BooleanField(default=True, null=True),
        ),
        migrations.AddField(
            model_name='archivedproject',
            name='allow_file_downloads',
            field=models.BooleanField(default=True, null=True),
        ),
        migrations.AddField(
            model_name='publishedproject',
            name='allow_file_downloads',
            field=models.BooleanField(default=True, null=True),
        ),
    ]

# physionet-django/project/migrations/0048_allow_file_downloads_2.py

from django.db import migrations, models
    MIGRATE_AFTER_INSTALL = True
    dependencies = [
        ('project', '0047_allow_file_downloads_1'),
    ]
    operations = [
        migrations.AlterField(
            model_name='activeproject',
            name='allow_file_downloads',
            field=models.BooleanField(default=True),
        ),
        migrations.AlterField(
            model_name='archivedproject',
            name='allow_file_downloads',
            field=models.BooleanField(default=True),
        ),
        migrations.AlterField(
            model_name='publishedproject',
            name='allow_file_downloads',
            field=models.BooleanField(default=True),
        ),
    ]
```

physionet-django/project/modelcomponents/publishedproject.py
```diff
@@ -201,6 +201,9 @@ def has_access(self, user):
         if self.deprecated_files:
             return False
 
+        if not self.allow_file_downloads:
+            return False
```
tompollard, line 205: To me this if statement feels like it is misinterpreting the original intent of the `has_access` method. I think it could be confusing in future. 

The method is asking the question "Does the user have access to the files". If `allow_file_downloads` is False, the answer might still be "yes, they do have access to the files but only in the cloud environment".

Perhaps we could have an argument that specifies the type of access? e.g. `access_type="download"`? Or a separate "can_download" method?

physionet-django/user/management/commands/loaddemo.py
```diff
@@ -128,6 +128,7 @@ def copy_demo_static():
     # the time of publication
     ppdir = os.path.join(effective_static_root, 'published-projects')
     for dirpath, subdirs, files in os.walk(ppdir):
+        print(dirpath, subdirs, files)
```
tompollard, line 131: Is this intentional?

physionet-django/project/templates/project/published_project.html
```diff
@@ -312,8 +312,8 @@ <h5 class="card-header">Versions</h5>
     </div>
     <!-- /.sidebar -->
   </div>
-
   <h2 id="files">Files</h2>
+  {% if project.allow_file_downloads %}
```
tompollard, line 316: Minor thing but easier to navigate these nested `if` statements if they are indented a couple of spaces.

physionet-django/project/modelcomponents/publishedproject.py
tompollard: @bemoody has convinced me that this isn't an issue for now, so I think we can ignore my comment.

physionet-django/project/forms.py
```diff
@@ -750,9 +751,15 @@ class AccessMetadataForm(forms.ModelForm):
...
-        help_texts = {'access_policy': '* Access policy for files.',
-                      'license': "* License for usage. <a href='/about/publish/#licenses' target='_blank'>View available.</a>"}
+        fields = ('access_policy', 'license', 'allow_file_downloads')
+        help_texts = {
+            'access_policy': '* Access policy for files.',
+            'license': "* License for usage. <a href='/about/publish/#licenses' target='_blank'>View available.</a>",
+            'allow_file_downloads': (
+                '* This option allows to enable/disable direct files downloads from the '
+                'platform. It cannot be changed after the publication of the project!'
+            ),
```
tompollard, line 761: @bemoody and I think this comment is not especially clear and it may lead to confused authors, but we are happy to leave it as it is for now and perhaps fix later.

tompollard:
Looks good to me! 

In another pull request, we might want to:
- Remove the [print statement](https://github.com/MIT-LCP/physionet-build/pull/1412/files/e7cb0c8c390b33a171eb0ace26c4ed14eb84e1d3#diff-9731e4e10d477d1a627768dc9cff11c761141c772e65aba814c86618b090df13)
- Disable downloads on the [project preview template](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/templates/project/project_preview.html) too (@alistairewj  ?)

PR #1414, opened 2021-10-15
The green page

fiedosiukr:
This PR repairs the green page and the related ones. There were over 100 database queries for the demo data, the amount has been reduced to 5.


tompollard:
thanks @fiedosiukr, this page is temporary and should be removed before long. The ordering of applications is unusual and it can be difficult to test! We'll take a look shortly.

Issue #1418, opened 2021-10-26
Style Check for New Code


rguo123:
## Description
To support good coding practice, we would like to have a code style check for incoming pull requests. This check should automatically run as part of the CI/CD pipeline and only be applied to new code changes.

## Acceptance Criteria
Step in Github Actions pipeline to check code style for incoming code changes.

tompollard:
This was implemented at some point in the past.

PR #1421, opened 2021-10-27
[1418] PEP8 Style Check on Modified Code Only

rguo123:
# Issue
#1418 

# Description
This PR adds the [flake8](https://flake8.pycqa.org/en/latest/) to project dev dependencies and introduces a code style check for *new lines of code* either being PR'ed or merged into dev.

# Modified Code
Minor changes had to be made to the existing physionet-build-test workflow:
1. `git` needs to be installed and updated. Otherwise, the `actions/checkout@v2` has different behavior and the `git` command will not be found in the container.
2. `actions/checkout@v2` must be moved to after `git` is updated. 
3. We add in a `fetch_depth: 0` for the checkout action. This will fetch the entire git history during the clone step as opposed to a shallow copy. We need this in order to compare modified code between branches and older commits. (Note: the entire history may not be necessary as described [here](https://stackoverflow.com/questions/65944700/how-to-run-git-diff-in-github-actions). I kept it as 0 to keep it more robust but if this runs into notable performance issues, we can change it to a smaller depth).

# Solution
The solution is straightforward. If the workflow event trigger is a pull request, we fetch the PR base branch from the original repository. This is necessary if the incoming PR is from a forked repo. We then simply have to run `git diff` to get the modified lines and pipe the result into `flake8`. 

I added a similar step to run on non PRs events to just run flake8 on modified lines relative to the same repository's `dev` branch. This is useful in case developers are triggering the physionet-build-test workflow on other branches for testing before PR. 

# Steps to Validate Solution:
0. Fork physionet-build repo
1. Copy this branch to get workflow changes
2. Introduce PEP8 error
3. Create PR
4. Workflow should fail on `Run code style check on changed files relative to PR base branch` step

To test the second change,
5. [Add manual trigger to physionet-build-test workflow](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow)
6. Manually trigger physionet-build-test workflow on branch
7. Workflow should fail on `Run code style check relative to local dev branch` step

### DIFF
--- a/.github/workflows/physionet-build-test.yml
+++ b/.github/workflows/physionet-build-test.yml
@@ -25,15 +25,22 @@ jobs:
       matrix:
         pip3: ['poetry', 'requirements.txt']
     steps:
-      - name: Checkout physionet-build repo
-        uses: actions/checkout@v2
 
       - name: Update packages
         run: apt-get update --yes
 
       - name: Install and configure needed software
         run: |
-          apt-get install sudo python3-dev python3-pip build-essential libpq-dev postgresql zip wget python-virtualenv --yes
+          apt-get install sudo git python3-dev python3-pip build-essential libpq-dev postgresql zip wget flake8 python-virtualenv --yes
+
+      # Note: run checkout after updating git
+      - name: Checkout physionet-build repo
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Create .env file
+        run: |
           ln -sT .env.example .env
 
       - name: Install repo dependencies
@@ -83,3 +90,15 @@ jobs:
           python manage.py loaddemo
           coverage run --source='.' manage.py test --verbosity=3 --keepdb
           coverage report -m
+
+      - name: Run code style check on changed files relative to PR base branch
+        if: github.event_name == 'pull_request'
+        run: |
+          git rev-parse --verify ${{ github.event.pull_request.base.sha }}
+          git diff ${{ github.event.pull_request.base.sha }} HEAD | flake8 --diff
+
+      - name: Run code style check relative to local dev branch
+        if: github.event_name != 'pull_request'
+        run: |
+          git fetch origin dev
+          git diff origin/dev HEAD | flake8 --diff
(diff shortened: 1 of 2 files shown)


bemoody:
Sounds great!

I would recommend not to pull in flake8 as a run-time dependency (in requirements.txt), and instead just to install the package with `apt-get install flake8`.

The following looks complicated, somewhat fragile, and probably unnecessary:
```
          git remote add upstream https://github.com/${{ github.event.pull_request.base.repo.full_name }}.git
          git fetch upstream ${{ github.event.pull_request.base.ref }}
          git diff upstream/${{ github.event.pull_request.base.ref }} HEAD | flake8 --diff
```

In a pull request, `HEAD` should be the automerge commit (`refs/pull/X/merge`) which means that, AFAIK, `HEAD^1` should always be the base onto which we are merging.  Indeed, trying to fetch the base branch by name might fetch the *wrong* commit if that branch is updated while this test is running.

(Moreover, the commands above presumably wouldn't work in a private repository!)

So I think that (along with `fetch_depth`) all you should need is this:
```
git diff HEAD^1 HEAD | flake8 --diff
```

But it might be more robust to instead write:
```
git diff ${{ github.event.pull_request.base.sha }} HEAD | flake8 --diff
```

When your changes are ready, please squash them into a single commit, or a couple of commits that do one thing each, with clear commit messages explaining the reason for each change.  Please do not use `git commit -m`.

bemoody:
I don't know whether github runs these commands in `bash` with `set -o pipefail` (or something similar), though they probably do, but to be safe it might be even more robust to write:
```
git rev-parse --verify ${{ github.event.pull_request.base.sha }}
git diff ${{ github.event.pull_request.base.sha }} HEAD | flake8 --diff
```

rguo123:
Hi @bemoody thanks for taking a look! Found the git suggestions very helpful! 

One question I had, I thought it would be better to leave flake8 as part of requirements/poetry so developers would be able to use it in case they want to run it locally or mimic the build workflow. Is there a reason we want to only include flake8 as a workflow dependency?

bemoody:
Simply put, we have too many pip dependencies already and are in over
our heads.  There may be nothing wrong with this package in particular,
but the size of the list is worrying.  Ultimately the problem is that
PyPI has no end-to-end authentication and no QA.

On the other hand, I don't see any reason we couldn't add the Debian
package to the Docker and/or VM test environments.

rguo123:
@bemoody Makes sense! I reset the poetry and requirements.txt files and only added flake8 as part of workflow. Also updated readme to specify that `flake8` needs to be installed if developers want to run it locally.

tompollard:
Looks great, thanks everyone. Let's give this a try.

@rguo123 super clear PR message...lesson for us all I think!

PR #1424, opened 2021-11-01
Silence disallowed host log messages

bemoody:
Partially revert aea362a1a7264b43c4b66842c8ac46e17f453df7:

When a client requests a URL from an unknown hostname (one that is not
listed in "ALLOWED_HOSTS"), django generates a log message at level
ERROR.  Since this is not a server bug, it should not result in an
email message to the administrators.  Therefore, raise the log level
for this logger to CRITICAL.


tompollard:
thanks @bemoody and https://github.com/MIT-LCP/physionet-build/pull/1421

PR #1426, opened 2021-11-03
Fix test suite on development system with lightwave

bemoody:
Running the test suite (in development config, on a system with WFDB installed) was broken because:

- loaddemo creates "physionet-django/static/wfdbcal", and then TestMixin.setUp tries to overwrite that same file and chokes.

- When running the test suite, lightwave shouldn't be accessing "physionet-django/static" anyway; it should be using "physionet-django/static/test" instead.

- Putting DBCAL_FILE in the right place causes TestMixin.tearDown to choke because it's trying to chmod a file we don't own.

- TestMixin.setUp chokes if the test wasn't previously torn down properly.

So clearly this has been broken for a while, I don't know how long.  But these changes should make things more robust.

### DIFF
--- a/physionet-django/lightwave/views.py
+++ b/physionet-django/lightwave/views.py
@@ -13,13 +13,11 @@ from project.views import project_auth
 
 
 # PUBLIC_ROOT: chroot directory for public databases
-if settings.STATIC_ROOT:
-    PUBLIC_ROOT = settings.STATIC_ROOT
-else:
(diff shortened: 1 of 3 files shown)


bemoody:
Actually, it's weird that lightwave.views.PUBLIC_DBPATH is "published-projects" and not "/published-projects", but that's how it's currently done on live and seems to work.

tompollard:
looks good to me, thanks benjamin

PR #1429, opened 2021-11-04
Debranding

fiedosiukr:
This PR allows admins to modify About and Share pages sections. There is an additional menu option in the admin console where users are able to edit the sections. The platform name and the footer data was moved to the environment variables.


bemoody:
Just some quick thoughts!

I don't think this belongs in the "project" app.  There's too much in there already!  It could go in the "physionet" app (where the views are now) or, maybe better, it could be moved to its own app ("about"? "docs"?)

I don't like "PLATFORM_NAME" either, I know some people like to call PhysioNet a "platform" but that's not what the word means in a technical context.  "SITE_NAME" would be more understandable.

I'm guessing the changes to lightwave/static/ were unintentional and should be undone.

I think that before trying to implement this we need to discuss what the result should look like.  I *don't* like having parts of the document structure defined by template files while other parts are defined by the database.  I would much rather see a flexible model that basically makes the *entire page* admin-editable with some sort of template system to handle dynamic content.

fiedosiukr:
@tompollard @bemoody Read for your reviews.

physionet-django/console/templates/console/console_navbar.html
```diff
@@ -194,6 +194,14 @@
         </ul>
       </li>
 
+      <!-- static pages -->
+      <li class="nav-item {% if static_pages_nav %}active{% endif %}" data-toggle="tooltip" data-placement="right">
+        <a id="nav_console_news" class="nav-link" href="{% url 'static_pages' %}">
+          <i class="far fa-window-maximize"></i>
+          <span class="nav-link-text">Static Pages</span>
```
tompollard, line 201: Nitpick but looking at the other menu items, there is a space between the icon and the text! Please could you add this?

tompollard:
Thanks @fiedosiukr, looks good to me. Reminder for self that before this goes to our live server we would need to: (1) update the environment file and (2) migrate about page content to the database.

physionet-django/console/templates/console/console_navbar.html
fiedosiukr: Fixed.

tompollard:
A side benefit of this change is that it makes it easier to distinguish between development ("PhysioNet-Dev"), testing ("PhysioNet-Test") and production ("PhysioNet") sites. Many times in the past I almost made changes on the live site unintentionally, because it looked so similar to the testing site.

Issue #1423, opened 2021-10-29
Use a distinct hostname for bulk downloads


bemoody:
Since the expiration of the DST Root CA X3, some clients are not able to connect.  In particular, both `wget` and `curl` on Debian 9 (for example) will fail to connect, even though the ISRG Root X1 is trusted, because these clients first try to follow the certificate chain provided by the server, and give up when that leads to an expired root.

(Which is silly on numerous levels.)

To make things easier it might be helpful to create another hostname (e.g. `dl.physionet.org`) which would be accessible to older wget versions, while the main `physionet.org`/`www.physionet.org` continues to use the current certificate chain for compatibility with older web browsers.

https://www.openssl.org/blog/blog/2021/09/13/LetsEncryptRootCertExpire/

PR #1432, opened 2021-11-04
Configurable hostname for bulk downloads

bemoody:
I think it's probably a good idea to use a different hostname in the example wget commands (and potentially other places) for bulk downloads.  See issue #1423.

(I haven't defined what that hostname might be, but I was thinking 'dl.physionet.org' for brevity, or maybe 'data.physionet.org' or 'files.physionet.org' or 'download.physionet.org'.)

It might be nice, at the same time, to stick an '-nH' and '--cut-dirs' into the wget command.

This here is a simple kludge and we may want to think about making it more general/flexible, but it should hopefully be a quick fix for the TLS issue.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,9 @@ SECRET_KEY=secret
 ALLOWED_HOSTS=localhost,127.0.0.1
 SITE_ID=4
 
+# Alternate hostname to be used in example download commands
+#BULK_DOWNLOAD_HOSTNAME=downloads.example.org
+
 # Database
 DB_USER=physionet
 DB_PASSWORD=password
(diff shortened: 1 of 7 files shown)


tompollard:
@bemoody looks good to me. please could you rebase?

tompollard:
@bemoody in an effort to try to clear the backlog of PRs, I rebased this myself. In hindsight, I don't think this was a great idea, but it is done now. Are you happy to merge?

bemoody:
Sure, looks good.

tompollard:
Thanks! Do we have a `BULK_DOWNLOAD_HOSTNAME` to add, or shall I leave this blank for now?

PR #1433, opened 2021-11-05
Flake8 configuration

fiedosiukr:
This PR adds the flake8 configuration file. The maximum line length has been set to 119, the value is recommended by Django (79 worsens the readability). It also ignores [F401](https://www.flake8rules.com/rules/F401.html) rule, because all of the models are imported into a single file and flake8 complains about that.

We decided not to force people to use any specific code formatter, so we didn't add `darker` to the dependencies. If someone wants to use `darker`, he needs to install it manually. In order to format the code, simply run `make format`.


tompollard:
Thanks @fiedosiukr looks good to me.

PR #1436, opened 2021-11-09
User login using SSO via Shibboleth

grzegorz-wcislo:


physionet-django/sso/auth.py
```diff
@@ -0,0 +1,26 @@
...
+
+UserModel = get_user_model()
+
+
+class RemoteUserBackend:
+    def authenticate(self, request, remote_user=None):
+        if not remote_user:
+            return
+
+        user = self._get_user_or_none(shibboleth_id=remote_user)
```
grzegorz-wcislo: Rename shibboleth_id to something more generic

physionet-django/user/models.py
```diff
@@ -331,6 +330,7 @@ class User(AbstractBaseUser):
         validators=[validators.UsernameValidator()],
         error_messages={
             'unique': "A user with that username already exists."})
+    sso_id = models.CharField(max_length=50, unique=True, null=True, blank=False)
```
grzegorz-wcislo: Check maximum length in the shibboleth docs

kshalot:
Figured I'll leave a few notes to make it easier to review (I can't edit the PR, so adding it as a comment).

#### The /sso routes
All the Django`/sso` routes are not directly accessed through the Internet. The point is that all traffic going to `https://domain/sso/*` is proxied to a SAML Service Provider that then authenticates the user, assigning him a token which is sent as the `SSO_REMOTE_USER_HEADER` header to Django. In short, what the Service Provider does:
1. Receives traffic going to `/sso/<path>`
2. Authenticates the user
3. If authentication was successful, proxies to Django's `/sso/<path>` internally.

So Django's `/sso` routes can be thought of as protected resources that can only be accessed after being authenticated through SAML.

SAML_SSO

#### Login
After authenticating at the SP's `/sso/login`, the user is internally proxied to Django's `/sso/login` with the SSO header set. So the Django starts handling the user only after they are already authenticated.

#### Register
If the user is not in the database, they are redirected from `/sso/login` to `/sso/register` which, again, passes through the Service Provider that then redirects the user to Django's `/sso/register` with the appropriate header set. The form is the same as for a regular user, with the addition of an invisible `sso_id` which is the user's persistent identifier assigned by the Identity Provider.

#### Activation
The request is proxied through the Service Provider so the header is set. There is no password setup required after successfully activating.

kshalot:
### Navbar with SSO enabled
Screenshot 2022-03-02 at 16 30 57

### Login view with SSO enabled
Screenshot 2022-03-02 at 16 31 05

tompollard:
Looks good, thanks!

tompollard:
@bemoody @briangow @danamouk this pull request added some settings to the environment file (`.env`), so you may want to add them:

```
# Single Sign-On
ENABLE_SSO="False"
SSO_REMOTE_USER_HEADER="HTTP_REMOTE_USER"
SSO_LOGIN_BUTTON_TEXT="Single Sign-On"
```

PR #1438, opened 2021-11-23
Remove arguments.

fiedosiukr:
This fix removes positional arguments that were not supposed to be there.


fiedosiukr:
@bemoody @tompollard For some reason I forgot to amend the last commit and the methods are broken. It would be nice if we could merge it today.

PR #1439, opened 2021-11-24
Ignore F403 rule.

fiedosiukr:
We need to ignore the [F403](https://www.flake8rules.com/rules/F403.html), because all of the models are imported into a single file and flake8 complains about that.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,3 @@
 [flake8]
 max-line-length = 119
-ignore = F401, W503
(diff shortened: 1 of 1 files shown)


fiedosiukr:
@tompollard @bemoody

tompollard:
thanks, merging! maybe we should explicitly call out the imports instead of `import *`, but ignoring the rule seems like the right thing for now.

PR #1440, opened 2021-11-24
dev-entrypoint.sh: only wait for GCS_HOST if set

grzegorz-wcislo:
Right now you can't run the development docker-compose configuration without using GCS.

This PR ensures that the startup scripts waits for the GCS_HOST to be available only if the GCS_HOST environment variable. is present


tompollard:
thanks, looks good!

PR #1441, opened 2021-11-24
Change unix socket to uwsgi tcp socket in docker

grzegorz-wcislo:
It's more convenient to use a TCP socket for the uwsgi protocol instead of a UNIX socket when using containers. This allows for easy communication with a separate container with Nginx to communicate with the uWSGI server easily using `uwsgi_pass`. I chose port 8001 for no particular reason.


tompollard:
thanks Grzegorz, looks good to me

PR #1442, opened 2021-11-25
Add new access policy - contributor review.

fiedosiukr:
This PR adds a new access policy called Contributor Review. In order to access the project files, a user has to send a data access request and then it has to be accepted by one of the reviewers.


tompollard:
🟥

tompollard:
@fiedosiukr I think you mentioned that you wanted to update the migration for this, so I'll review the PR when once the new migration is pushed.

fiedosiukr:
@bemoody It seems like some test cases related to one of the templates fail. They only fail after reverting the late migrations. Do you have any ideas how to fix that? It is probably connected with the static files.

bemoody:
Not relevant to the migration issue (which I'll comment about in a bit) but this looks suspicious:

```
--- a/physionet-django/project/views.py
+++ b/physionet-django/project/views.py
@@ -1836,8 +1836,7 @@ def sign_dua(request, project_slug, version):
     else:
         raise Http404()
 
-    if project.deprecated_files or not project.access_policy or project.has_access(user) \
-        or project.is_self_managed_access:
+    if project.deprecated_files or not project.access_policy or project.has_access(user):
         return redirect('published_project',
                         project_slug=project_slug, version=version)
 
```

For self-managed-access projects, we do not currently create or use a DUASignature object, and we shouldn't allow users to create a "self-service" DUASignature object (even if currently that object would not have any effect.)

I think this should instead be:
```
    if (project.access_policy not in {AccessPolicy.RESTRICTED, AccessPolicy.CREDENTIALED}
            or project.deprecated_files or project.has_access(user)):
```

bemoody:
The issue here is that after applying the first migration, an existing "self-managed" published project is not considered "contributor-reviewed".

If we installed this code, applying the early migration but not the late migration, any credentialed user would then be able to access existing self-managed projects.  So the test case is doing its job there!

I think what we want is that after applying the early migration, or after unapplying the late migration, these projects should have "access_policy = CONTRIBUTOR_REVIEW", and also "is_self_managed_access = True".  So what I think we should do is:

- Take the RunPython from the current migration 0052 and add it to the end of 0051.

- Add the following RunPython to the beginning of 0053:
```
def migrate_forward(apps, schema_editor):
    pass

def migrate_backward(apps, schema_editor):
    ActiveProject = apps.get_model('project', 'ActiveProject')
    ArchivedProject = apps.get_model('project', 'ArchivedProject')
    PublishedProject = apps.get_model('project', 'PublishedProject')

    ActiveProject.objects.filter(access_policy=AccessPolicy.CONTRIBUTOR_REVIEW).update(
        is_self_managed_access=True
    )
    ArchivedProject.objects.filter(access_policy=AccessPolicy.CONTRIBUTOR_REVIEW).update(
        is_self_managed_access=True
    )
    PublishedProject.objects.filter(access_policy=AccessPolicy.CONTRIBUTOR_REVIEW).update(
        is_self_managed_access=True
    )
```
- Delete the old 0052 and rename 0053 to 0052; there's no reason we need three migration scripts here.

Here's what we should expect to see with those migrations:
```
$ ./manage.py resetdb && ./manage.py loaddemo && ./manage.py migrate project 0050
$ sqlite3 db.sqlite3 'select pp.slug, pp.is_self_managed_access, pp.access_policy, li.slug from project_publishedproject as pp inner join project_license as li on pp.license_id = li.id'
demoeicu|0|2|physionet-credentialed-health-data-license-150
demoecg|0|0|creative-commons-attribution-40-international-public-license
demobsn|0|0|mit-license
demoselfmanaged|1|2|physionet-credentialed-health-data-license-150
demopsn|0|0|gnu-general-public-license-gpl-version-3
demowave|0|0|open-data-commons-attribution-license-v10

$ ./manage.py migrate project 0051
$ sqlite3 db.sqlite3 'select pp.slug, pp.is_self_managed_access, pp.access_policy, li.slug from project_publishedproject as pp inner join project_license as li on pp.license_id = li.id'
demoeicu|0|2|physionet-credentialed-health-data-license-150
demoecg|0|0|creative-commons-attribution-40-international-public-license
demobsn|0|0|mit-license
demoselfmanaged|1|3|physionet-contributor-review-health-data-license-150
demopsn|0|0|gnu-general-public-license-gpl-version-3
demowave|0|0|open-data-commons-attribution-license-v10

$ ./manage.py migrate project 0052
$ sqlite3 db.sqlite3 'select pp.slug, pp.access_policy, li.slug from project_publishedproject as pp inner join project_license as li on pp.license_id = li.id'
demoeicu|2|physionet-credentialed-health-data-license-150
demoecg|0|creative-commons-attribution-40-international-public-license
demobsn|0|mit-license
demoselfmanaged|3|physionet-contributor-review-health-data-license-150
demopsn|0|gnu-general-public-license-gpl-version-3
demowave|0|open-data-commons-attribution-license-v10

$ ./manage.py migrate project 0051
$ sqlite3 db.sqlite3 'select pp.slug, pp.is_self_managed_access, pp.access_policy, li.slug from project_publishedproject as pp inner join project_license as li on pp.license_id = li.id'
demoeicu|0|2|physionet-credentialed-health-data-license-150
demoecg|0|0|creative-commons-attribution-40-international-public-license
demobsn|0|0|mit-license
demoselfmanaged|1|3|physionet-contributor-review-health-data-license-150
demopsn|0|0|gnu-general-public-license-gpl-version-3
demowave|0|0|open-data-commons-attribution-license-v10
```

**However**, this will fail the upgrade test because the *production* code makes the assumption that the access policy is one of the three known values.  That should be fixed by pull #1468.

tompollard:
Thanks for getting to the bottom of this @bemoody. #1468 is now merged.

fiedosiukr:
@bemoody thanks for the comprehensive explanation.

physionet-django/project/fixtures/demo-project.json
```diff
@@ -1873,6 +1834,22 @@
     "dua_html_content": ""
   }
 },
+{
+  "model": "project.license",
+  "pk": 10,
```
tompollard: Shouldn't this be `"pk": 11,`? Surprised it doesn't raise an error somewhere.

physionet-django/project/modelcomponents/publishedproject.py
```diff
@@ -203,22 +203,24 @@ def has_access(self, user):
...
+            return True
+        elif self.access_policy == AccessPolicy.RESTRICTED:
+            return user.is_authenticated and DUASignature.objects.filter(project=self, user=user).exists()
+        elif self.access_policy == AccessPolicy.CREDENTIALED:
+            return (
+                user.is_authenticated
+                and user.is_credentialed
+                and DUASignature.objects.filter(project=self, user=user).exists()
+            )
+        elif self.access_policy == AccessPolicy.CONTRIBUTOR_REVIEW:
```
tompollard, line 216: There seems to be an issue with this `elif` chunk. If a user's initial request for access is rejected, it seems they can never gain access. To reproduce:

1. As "Admin", create a contributor reviewed project.
2. As a new user, request access to the data
3. As Admin, reject the request.
4. As the new user, request access to the data.
5. As Admin, approve the request.
6. The new user is unable to access the data.

tompollard (commented):
Thanks @fiedosiukr! There are a couple of minor issues that I came across when testing. Please could you take a look? Briefly the issues are:
1. Duplicate key in fixture?
2. If the first access request is rejected, it seems the user can never gain access.

physionet-django/project/modelcomponents/publishedproject.py
fiedosiukr: The issue was in the template. Probably the new user isn't credentialed.
image

physionet-django/project/fixtures/demo-project.json
fiedosiukr: Fixed! The license was loaded twice, because it was defined in two separate files. Currently only the `contributor-review-license.json` is used.

tompollard:
Thanks @fiedosiukr, looks good to me!

PR #1444, opened 2021-12-01
Run test-upgrade.sh as part of test suite

bemoody:
This adds another GitHub workflow script to run `test-upgrade.sh`.

(The purpose of this script is to test that the server can be "live-upgraded", i.e. that it will work when installing to the live server using `git push`.  For example, it tests that migrations don't break the running server.)

This assumes there is a branch called `origin/production` corresponding to the current version from which we are upgrading.

This workflow duplicates some of the functionality of `physionet-build-test.yml`, but it doesn't *entirely* duplicate it.  Notable differences:

- `physionet-build-test.yml` tests the postgresql backend
- `physionet-upgrade-test.yml` tests the sqlite backend
- `physionet-upgrade-test.yml` tests random query ordering
- `physionet-build-test.yml` tests setup using poetry (as well as using pip)
- `physionet-build-test.yml` runs coverage tests
- `physionet-build-test.yml` runs code style tests

Moreover, this workflow has the significant limitation that it will not be automatically re-run when the `production` branch is pushed.

So there's a lot that could be improved here, but I still think adding this workflow would be beneficial.

### DIFF
--- a/.github/workflows/physionet-upgrade-test.yml
+++ b/.github/workflows/physionet-upgrade-test.yml
@@ -1,0 +1,60 @@
+name: Debian / Upgrade Test
+
+on:
+  push:
+    branches:
+      - dev
+  pull_request:
+    branches:
+      - dev
+
+jobs:
+  testupgrade:
+    name: Upgrade Test
+    runs-on: ubuntu-latest
+    container: debian:10
+    steps:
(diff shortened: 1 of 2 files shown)


tompollard:
looks good to me, thanks!

PR #1446, opened 2021-12-07
Fix internal links used by the second level navigation menu on the Publish page

tompollard:
In https://github.com/MIT-LCP/physionet-build/pull/1429 we moved content from the [About page](https://physionet.org/about/) and the [Sharing page](https://physionet.org/about/publish/) to the database. 

The menu items on the [About page](https://physionet.org/about/) correctly refer the user to relevant content within the page using internal hyperlink links, but the chunk of code that generates these links was mistakenly omitted from the [Sharing page](https://physionet.org/about/publish/). 

This small pull request corrects the omission. Slightly unrelated, but there was an unnecessary `<div>` ... `</div>` on the About page that I removed at the same time.

### DIFF
--- a/physionet-django/templates/about/about.html
+++ b/physionet-django/templates/about/about.html
@@ -27,18 +27,17 @@ About
           </ul>
         </div>
       </div>
-    </div><div class="main-content">
-      <div>
-        {% for section in sections %}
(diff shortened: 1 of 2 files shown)


briangow:
Looks good to me!

PR #1448, opened 2021-12-10
Debranding - footer data and overlapping logo

fiedosiukr:
I forgot to push the last commit and this PR contains changes that were not included:
- moved footer data into environment variables
- fixed logo overlapping the menu items


physionet-django/physionet/context_processors.py
```diff
@@ -9,3 +9,11 @@ def access_policy(request):
 
 def platform_name(request):
     return {'SITE_NAME': settings.SITE_NAME}
+
+
+def footer_data(request):
```
tompollard: Do we need a separate context processor for each variable? Is there a reason not to have a single `platform_config(request)` processor with SITE_NAME, FOOTER_MANAGED_BY, etc?

fiedosiukr: I combined them.

tompollard:
Looks good, thanks!

PR #1452, opened 2021-12-20
Allow the platform STRAPLINE to be configured alongside SITENAME

tompollard:
The strapline on our homepage ("The Research Resource for Complex Physiologic Signals") is specific to PhysioNet. Collaborators reusing the platform, like Toronto, will want to customise the strapline.

This pull request adds a STRAPLINE setting to the environment file, so it can be configured alongside SITENAME. After merging, .env files will need to be updated with `STRAPLINE=my strapline`


fiedosiukr (approved):

.env.example
```diff
@@ -85,8 +85,9 @@ GCP_STATIC_BUCKET_NAME=
 GCP_PROJECT_ID=
 GCP_BUCKET_LOCATION=
 
-SITE_NAME=
-EMAIL_SIGNATURE=
-FOOTER_MANAGED_BY=
-FOOTER_SUPPORTED_BY=
+SITE_NAME="DataShare"
```
briangow, line 89: @tompollard , perhaps a comment should be added above the new variables as I don't think `# Variables required to be set when storage type is set to GCP` applies to these new variables.

tompollard: Good idea, thanks for the suggestion. I added `# Site-specific content` as the header (best I could come up with for now, but hopefully it does the job and we can tweak later!)

briangow:
@tompollard, this looks good to me!

PR #1459, opened 2022-01-05
Bump requirements file to match recent dependabot updates to Django and sqlparse

tompollard:
Dependabot recently opened pull requests to update Python versions of Django (#1454) and sqlparse (#1453). For reasons that are unclear to me, Dependabot didn't update requirements.txt to match. 

This pull request bumps the versions of the two packages in requirements.txt to match those in poetry.lock.

Note that there was already a mismatch in versions between poetry.lock and requirements.txt for both packages. I haven't checked, but I assume this is the case for many other packages too.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -100,8 +100,9 @@ pyparsing==2.4.7 \
  --hash=sha256:ef9d7589ef3c200abe66653d3f1ab1033c3c419ae9b9bdb1240a85b024efc88b
 
 # required by Django 2.2.3
-sqlparse==0.2.4 \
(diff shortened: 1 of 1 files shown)


bemoody:
I don't see an obvious way to check what is different between what poetry installs and what pip installs.  Running the commands in 'physionet-build-test.yml' and then running 'pip freeze' gives a bunch of junk like

bleach @ file:///home/user/.cache/pypoetry/artifacts/0c/19/da/291df1a8b71e9bd208e3e6f801afd6567b9ef47036b167c11e25f5c96c/bleach-3.3.0-py2.py3-none-any.whl

instead of the expected

bleach==3.3.0

so, yeah.  But at least, this appears to work.

PR #1463, opened 2022-01-20
Update the order of citations in the grey box for PhysioNet/CinC challenges

briangow:
For the PhysioNet/CinC challenges this changes the citation order in the grey box at the top of the project so that the original publication citation is first, the PhysioNet project citation is next, and the standard PhysioNet citation is last. 

This does not change the order for other "community" challenges or non-challenge projects. 

This has been tested on staging for the 2019-2021 challenges but since publication data for the legacy challenges didn't exist until I added it recently (for 2015-2018) that data isn't on staging. Therefore, it hasn't been tested for legacy projects.


briangow:
I was able to do a test of a legacy project (for year 2015) on staging after figuring out how to create a PublishedPublication object from the command line.

Citations for the recent and legacy projects are working as expected.

bemoody:
There's a lot of duplicated template code here.  Note that in dev,
lines 98-142 of project_preview.html are nearly identical to lines
82-129 of published_project.html.

So before making any changes to that section of the template, it would
be better to pull that code out into a separate html file.

briangow:
@bemoody thanks for pointing out the duplicated code between project_preview.html and published_project.html! I'll look into breaking that code out into a separate html file and then make my updates there.

briangow:
@bemoody , I have made the requested updates:
- I moved the duplicate code from published_preview.html and published_project.html into citation_box.html
- I applied my updates to change the citation order for PhysioNet/CinC challenges
- I updated the code to allow for unpublished and published projects. This required burying the `project.is_legacy` statements under a  `{% if project.is_published %}`. This results in even more `if` `else` statements 
- I checked that this is working on staging for published and unpublished projects

briangow:
I am pushing an updated commit which fixes the indentation level of `{% if publication %}` in the case when the project is published but isn't a PhysioNet/CinC challenge.

bemoody:
Thanks!  We can simplify this by defining all unpublished projects as
non-legacy:

    class UnpublishedProject(models.Model):
        ...
        @property
        def is_legacy(self):
            return False

Then the "if project.is_published" can go away.

briangow:
Ok, yeah, defining `is_legacy` for unpublished projects is another option. If it is worth doing that to simply this block of code, then I'll work on implementing that. Thanks @bemoody !

briangow:
hi @bemoody , I updated this PR with your suggestion. It appears to be working properly on the staging server.

PR #1468, opened 2022-02-02
Handle unknown access policies in template tags

bemoody:
The access_badge and access_description template tags should not crash if a project access policy is set to an unknown value.  We want to allow for the possibility of adding new access policy values, so ensure the site won't crash if we do so.

This is a prerequisite for making the upgrade tests pass in pull #1442 (i.e., we should merge this change to production first.)

Of course, there may be other places that make the implicit assumption that access policy is 0, 1, or 2.  In particular, PublishedProject.has_access is quite fragile.  However, I don't think this is an immediate concern for pull #1442, and that pull rewrites has_access to be more robust for the future.


tompollard:
Looks good, thanks Benjamin!

tompollard:
(now on the live server)

PR #1470, opened 2022-02-04
Remove the green credentialing workflow

tompollard:
We currently have two parallel workflows for credentialing. One based on the old platform (used by KP) and one developed for the new platform. Maintaining two workflows makes development more complicated than it needs to be.

This pull request removes the old workflow, which is no longer used. The main changes are:
- Remove the HTML templates that rendered the old workflow
- Redirect the URLs for the old workflow to the relevant page in the new workflow
- Remove email templates that were used only in the old workflow
- Remove some awkward logic that prevented applications from appearing in both workflows at the same time. Now all applications should appear in the new workflow.

It is possible that the new workflow will now contain some applications that were being handled by KP (e.g. applications where KP had contacted the referee and not received a response for several months). I don't think this should be an issue. Worst case is that the referee gets contacted again.

@fiedosiukr is in the process of making some changes to the credentialing workflow. If anything, this pull request should simplify that work. @alistairewj @kepaik


briangow:
@tompollard , looks good to me!

PR #1471, opened 2022-02-10
Debranding - error pages.

fiedosiukr:
This PR removes hard-coded emails from the error templates.


tompollard:
Looks good to me, thanks!

PR #1473, opened 2022-02-10
Reduce query number.

fiedosiukr:
This PR reduces the database queries amount in some of the admin console views.

### DIFF
--- a/physionet-django/console/templates/console/users_admin.html
+++ b/physionet-django/console/templates/console/users_admin.html
@@ -29,7 +29,7 @@
             <td>{{ user.email }}</td>
             <td>{{ user.join_date }}</td>
             <td>{{ user.last_login|date }}</td>
-            <td>{{ user.login_time.values_list|length }} </td>
+            <td>{{ user.login_time_count }} </td>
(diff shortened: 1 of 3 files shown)


tompollard:
It looks like this changes the information displayed on the user pages. Not necessarily a bad thing, but possibly not intended? Previously the pages aggregated the number of logins, but now each login appears to be listed on a separate row.

e.g.

1. http://localhost:8000/console/users/admin/

Screen Shot 2022-02-10 at 11 26 39

2. http://localhost:8000/console/users/all/

Screen Shot 2022-02-10 at 11 43 02

fiedosiukr:
The number of logins is still aggregated.

1. http://localhost:8000/console/users/all/
image
2. http://localhost:8000/console/users/admin/
image

tompollard:
@fiedosiukr odd, that's not what I saw when testing. I'll take another look!

tompollard:
I was testing in my local environment, rather than Docker, so I assume this is a Postgres vs SQLlite issue. @bemoody Do we care that behaviour is different with SQLlite?

PR #1464, opened 2022-01-25
Separate training and credentialing

fiedosiukr:
This PR removes training part from the credentialing process and makes it a separate workflow. There is a new section in the user settings - training. Each training type has its own `valid_duration` which can be either `None` (the training will never expire) or `timedelta`. In order to access the project with the credentialed access policy, a user has to finish the required training. The required training is set during project creation - access section.

PR #1474, opened 2022-02-10
ProjectFiles - add cp_file method

fiedosiukr:
This method is required to make the upgrade workflow pass in the #1464.


tompollard:
Thanks, looks good to me.

PR #1479, opened 2022-02-15
Access and GCP logs

fiedosiukr:
This PR adds two new log types. Logs are created every time an authenticated user views a published project. If the user accesses the same project within the given time period (`LOG_TIMEDELTA` environment variable, 10 minutes by default), the previously created log is updated. The logs are also created for signed URLs (only enabled for GCP storage type).

image
Generated logs may be viewed in the admin console.


fiedosiukr:
For some reason installing packages inside `virtualenv` using poetry stopped working. This PR doesn't include any new packages, so it should work as before. The problem also occurs in #1464.

fiedosiukr:
@bemoody @tompollard @alistairewj The issue with the poetry workflow disappeared 🎉 .

tompollard:
Thanks @fiedosiukr. I think the logging is a great addition but I need to think a bit more about how the GCP side of things is handled here. It feels slightly odd to have GCP functionality that doesn't relate to our GCP implementation. 

I'm wondering whether we can either (1) modify things so that the GCP logging works for us too or (2) make the distinction between Toronto GCP and PhysioNet GCP clearer.

kshalot:
@tompollard Regarding the conversation we had during our last HDN meeting - everything seems to work as intended.

**Reminder on what it was about**: You noted that project accesses are not being logged - log count was not being incremented.

I think I know why you got that impression - logs are being stored in "buckets", with `LOG_TIMEDELTA` denoting the "length" of the "bucket". So when a user accesses the project **earlier that LOG_TIMEDELTA after his previous access**, a separate log entry will not be created but the count will be increased in the previous log (bucket).

**Log overview: count = the total count of log rows (buckets/sessions).
Not incremented with each access, but with each access that was after last_access + LOG_TIMEDELTA**
project_logs

**Project log overview: count = the number of accesses within that row/bucket/session.
Count of the last bucket is incremented each time the user accesses the project.**
log_details

You must have been looking at the overview of logs for all projects - on that page it seems like the count is not incrementing. If that behaviour is unintuitive then alternatively we could sum the count of each bucket and display that instead of the bucket count.

tompollard:
thanks for the explanation @kshalot, this makes sense! 

i had thought we were capturing anonymous users as well as registered users, but i guess this is not the case. 

It would be good to discuss how we might capture anonymous users too, perhaps at next catch up.

tompollard:
@bemoody @briangow @danamouk 

This change added another new setting to the .env file:

```
# Logging period in minutes
# i.e. time that must elapse before a view is considered "new"
LOG_TIMEDELTA=10
```

PR #1481, opened 2022-02-15
Remove account migration notification

fiedosiukr:
This PR removes the account migration notification which was displayed on the login page.


tompollard:
thanks!

PR #1482, opened 2022-02-21
Change license for self managed project.

fiedosiukr:
The license should be set to 11 instead of 10 for the self managed project.


tompollard:
thanks!

Issue #1469, opened 2022-02-04
Display the status of credentialing applications to users?


tompollard:
We receive many emails from people asking about the status of their credentialing application. This creates extra work for us and our users. Now that we follow a single credentialing workflow, we have the ability to display a fairly granular status level. Should we display a "application status" message at: https://physionet.org/settings/credentialing/ ? 

My vote is yes. Okay @bemoody @alistairewj? 

Note, in the past we displayed a "reference contacted" status message in the credentialing system to applicants. See: https://github.com/MIT-LCP/physionet-build/pull/1042/commits/35a2e0dabcc794a88cc0295f0bfcfcca72ecfdf8. This was removed at the request of KP.

alistairewj:
Seems very appropriate!

tompollard:
Should also tag @kepaik

PR #1485, opened 2022-02-24
Display status of credentialing application to user. Ref #1469

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/1469, we would like to display the status of credentialing applications to users. This pull request displays the status to users on the application page (http://localhost:8000/settings/credentialing/). 

e.g.:

Screen Shot 2022-02-23 at 18 44 39

The `get_review_status` logic is slightly hacky because the `REVIEW_STATUS_LABELS` are not well suited to being displayed publicly:

```
    REVIEW_STATUS_LABELS = (
        ('', '-----------'),
        (0,  'Not in review'),
        (10, 'Initial review'),
        (20, 'Training'),
        (30, 'ID check'),
        (40, 'Reference'),
        (50, 'Reference response'),
        (60, 'Final review')
    )
```

Instead of modifying the labels - which would require a migration - I have left them as they are for now. We can implement something cleaner once the forthcoming updates to the credentialing system are merged.

### DIFF
--- a/physionet-django/user/models.py
+++ b/physionet-django/user/models.py
@@ -941,13 +941,14 @@ class CredentialApplication(models.Model):
         Get the current review status of a credentialing application. Hacky.
         Could be simplified to return self.credential_review.status later.
         """
-        if hasattr(self, 'credential_review') and self.credential_review.status >= 10:
-            title_dict = {a: k for a, k in CredentialReview.REVIEW_STATUS_LABELS}
-            status = title_dict[self.credential_review.status]
-            if "review" not in status and "check" not in status:
-                status = status + " review"
-        else:
+        if not hasattr(self, 'credential_review'):
             status = 'Awaiting review'
+        elif self.credential_review.status <= 40:
+            status = 'Awaiting review'
+        elif self.credential_review.status == 50:
+            status = 'Awaiting a response from your reference'
+        elif self.credential_review.status >= 60:
+            status = 'Awaiting final approval'
 
         return status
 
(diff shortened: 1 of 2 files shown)


briangow:
Hi @tompollard , this looks good but I wonder if it is meaningful to the user to show a status of "Training review" or "ID check review" as they probably don't know what we are referring to. I think these three categories might make sense:

credential_review.status <= 40 -> "Awaiting review"
credential_review.status = 50 -> "Awaiting a response from your reference"
credential_review.status = 60 -> "Awaiting final approval"

tompollard:
@briangow Thanks for looking. I almost took the approach that you're suggesting, but ended up getting scared off by https://github.com/MIT-LCP/physionet-build/pull/1464/files which changes the labels.

Once that PR is merged, the statuses become:

credential_review.status <= 30 -> "Awaiting review"
credential_review.status = 40 -> "Awaiting a response from your reference"
credential_review.status = 50 -> "Awaiting final approval"

I'd thought it was better to have slightly awkward language that didn't break when we merged the other changes, but on reflection I think you're right that we might just confuse people.

Let's hold off on this for now. I'll update the PR to follow your approach after https://github.com/MIT-LCP/physionet-build/pull/1464/ is merged.

*Not that it really matters, but you shouldn't have seen "ID check review". "Review" is only added to fields that don't already contain "check" or "review"

briangow:
Ok, sounds good. If keeping it the way you have it would allow merging it and it wouldn't break after #1464 is merged then that makes sense to me.

tompollard:
@briangow I've temporarily updated the `get_review_status` method to follow your suggestion. I've also noted that the method will need updating once #1464. Please could you take a look and merge if you are happy? Thanks!

briangow:
@tompollard , this looks good to me and works as expected.

PR #1489, opened 2022-02-28
Add about static page content to demo database

briangow:
This adds content from the /about/ and /about/publish static pages to loaddemo.


bemoody:
Thanks!

I would rename 'static-pages.json' to 'demo-physionet.json'.  Existing "X/fixtures/demo-X.json" files are loaded automatically (see find_demo_fixtures()).  Better to avoid creating more special cases and make the purpose of the data clear.

tompollard:
Looks good to me, thanks!

briangow:
@bemoody , thanks for the review and the suggestion. I've updated this.

PR #1493, opened 2022-02-28
Add a StaticPage model so about pages don't need to be hard-coded in enums

briangow:
This PR adds a new StaticPage model which will be used in the existing Section model (as seen in physionet/models.py) in place of the hard-coded pages from physionet/enums.py. This is a necessary step towards creating a generalizable view and template for the about pages as discussed in #1489. The previous approach required adding hard-coded page names in enums while this approach attempts to resolve any page from the StaticPage object in the database. 

Thanks to @tompollard for creating the code for the StaticPage model!

Thanks to @bemoody for outlining the steps needed to create the migration files!


bemoody:
```
diff --git a/physionet-django/physionet/fixtures/demo-physionet.json b/physionet
-django/physionet/fixtures/demo-physionet.json
index daee1621..c6ff6d39 100644
--- a/physionet-django/physionet/fixtures/demo-physionet.json
+++ b/physionet-django/physionet/fixtures/demo-physionet.json
@@ -4,7 +4,7 @@
   "pk": 1,
   "fields": {
     "title": "About",
-    "url": "about"
+    "url": "/about/"
   }
 },
 {
@@ -12,7 +12,7 @@
   "pk": 2,
   "fields": {
     "title": "Share",
-    "url": "share"
+    "url": "/about/publish/"
   }
 },
 {
diff --git a/physionet-django/physionet/migrations/0002_auto_20220228_1217.py b/physionet-django/physionet/migrations/0002_auto_20220228_1217.py
index f8621c80..ee63a0fe 100644
--- a/physionet-django/physionet/migrations/0002_auto_20220228_1217.py
+++ b/physionet-django/physionet/migrations/0002_auto_20220228_1217.py
@@ -14,17 +14,28 @@ def migrate_forward(apps, schema_editor):
             url='/about/', title='About')
 
         (about_pub_page, _) = StaticPage.objects.get_or_create(
-            url='/about/share', title='Share')
+            url='/about/publish/', title='Share')
 
         for section in Section.objects.all():
             if section.page == 'Page.ABOUT':
                 section.static_page = about_page
-            if section.page == 'Page.SHARE':
+            elif section.page == 'Page.SHARE':
                 section.static_page = about_pub_page
+            else:
+                raise ValueError(section.page)
             section.save()
 
     def migrate_backward(apps, schema_editor):
-        pass
+        Section = apps.get_model('physionet', 'Section')
+
+        for section in Section.objects.all():
+            if section.static_page.url == '/about/':
+                section.page = 'Page.ABOUT'
+            elif section.static_page.url == '/about/publish/':
+                section.page = 'Page.SHARE'
+            else:
+                raise ValueError(section.static_page.url)
+            section.save()
 
     dependencies = [
         ('physionet', '0001_initial'),
@@ -44,5 +55,10 @@ def migrate_backward(apps, schema_editor):
             name='static_page',
             field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.CASCADE, to='physionet.StaticPage'),
         ),
+        migrations.AlterField(
+            model_name='section',
+            name='page',
+            field=models.CharField(blank=True, null=True, max_length=16),
+        ),
         migrations.RunPython(migrate_forward, migrate_backward),
     ]
```

bemoody:
I'm not sure if we want to allow arbitrary URLs for static pages.  I think we probably do want to use the actual URL (like "/about/") as the key for the StaticPage object.  Anyway we do need to be consistent here ;)

briangow:
Thanks @bemoody for the suggested updates, I will make those. 

Regarding the url paths, we will go with your suggestion. We will use the convention of specifying `/about/` for `static_url`'s in the database.

bemoody:
I think you need to switch the order of the AlterField and RunPython in 0002_auto_20220228_1217.py.

You're also missing the trailing slash in `/about/publish/` in line 17.

PR #1498, opened 2022-03-03
Fix mismatch in django-storages version between requirements.txt and poetry

tompollard:
Currently there is a mismatch in versions of django-storages. This fixes that by bumping both version to 1.12.3. Currently the versions are 1.12.2 (poetry) and 1.11.1 (requirements.txt). Small issue but fixing here will help to make forthcoming version updates a little cleaner.


briangow:
Looks good to me, thanks!

PR #1501, opened 2022-03-03
Bump django_background_tasks from v1.2.0 to v1.2.5

tompollard:
Bump [django_background_tasks](https://pypi.org/project/django-background-tasks/) from v1.2.0 to v1.2.5. This change is required to move to Django 3 (see: https://github.com/MIT-LCP/physionet-build/issues/1095).

A couple of changes were needed on the imports (thanks @kshalot!):
- `from background_task.models_completed import CompletedTask` => `from background_task.models import CompletedTask`
- `from django.contrib.staticfiles.templatetags.staticfiles import static` => `from django.templatetags.static import static`

PR #1502, opened 2022-03-03
add alternative imports as preparation for upgrading django_background_tasks

tompollard:
In #1501, tests fail because the code is run against the existing production branch, which is using v1.2.0 of django_background_tasks. This pull request adds support for both import statements (old v1.2.0 vs new v1.2.5).

Once this pull request is merged, we should be able to bump the version of django_background_tasks to v1.2.5.

### DIFF
--- a/physionet-django/project/admin.py
+++ b/physionet-django/project/admin.py
@@ -1,5 +1,8 @@
 from background_task.models import Task
-from background_task.models_completed import CompletedTask
+try:
+    from background_task.models_completed import CompletedTask
+except ImportError:
+    from background_task.models import CompletedTask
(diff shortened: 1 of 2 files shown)


bemoody:
```
+try:
+    from django.contrib.staticfiles.templatetags.staticfiles import staticgit
+except ImportError:
+    from django.templatetags.static import static
```
Something seems fishy

briangow:
Looks good, thanks!

tompollard:
> +try:
> +    from django.contrib.staticfiles.templatetags.staticfiles import staticgit
> +except ImportError:
> +    from django.templatetags.static import static

:facepalm:

Issue #1491, opened 2022-02-28
Settings object has no attribute ERROR_EMAIL


bemoody:
Commit 8d6ae6d819b0578f97123eccca1d5444d2cd893f introduced a variable 'settings.ERROR_EMAIL', but this is only defined in physionet.settings.development.base and physionet.settings.settings (not in physionet.settings.base.)

Consequently all django-generated error pages are 502ing.

PR #1505, opened 2022-03-03
Define ERROR_EMAIL

bemoody:
Define the ERROR_EMAIL variable so that we can generate appropriate 404 / 403 / etc. error pages.  Fix issue #1491.

The way this stuff is configured is not the best; it'd be good to clean it up so we only have to define the hostname in one place.  (I don't think we can/should rely on Site objects when defining variables here in the settings module.)

Of course what we really want to be aiming for is to unify staging / production / settings so we can use physionet.settings.settings everywhere...


tompollard:
Looks good, thanks!

PR #1506, opened 2022-03-03
Tab completion and history in test-server

bemoody:
When installing a test server using install-pn-test-server (or ssh-install-pn-test-server), create a Python startup script to work around the default brokenness of virtualenv.  Not a serious issue by any means, just a minor convenience.


tompollard:
thanks!

PR #1507, opened 2022-03-04
Fix static pages console views url params

kshalot:
I found some issues with the new static pages changes in the admin console. #1436 introduces a new static page so I had to fix it anyway to test that PR's changes.

### `str` url params
Static pages are accessed by their url in the admin console:
https://github.com/MIT-LCP/physionet-build/blob/83b29d9124ab80253d1678feeffaf05afa9b5a57/physionet-django/console/views.py#L1991-L1995

The first problem is that the url expects an `str` param which won't match slashes (and static page urls contain slashes)
https://github.com/MIT-LCP/physionet-build/blob/83b29d9124ab80253d1678feeffaf05afa9b5a57/physionet-django/console/urls.py#L97

Screenshot 2022-03-04 at 11 46 05

Using `<path:xxx>` params won't fix the issue since then urls like https://github.com/MIT-LCP/physionet-build/blob/83b29d9124ab80253d1678feeffaf05afa9b5a57/physionet-django/console/urls.py#L103 won't be matched correctly.

### Mixing `page.title` with `page.url`
The other issue is the fact that sometimes the pages are linked by the `title` while the views expect the `url` which creates errors as well. Example:

#### View expects url
https://github.com/MIT-LCP/physionet-build/blob/83b29d9124ab80253d1678feeffaf05afa9b5a57/physionet-django/console/views.py#L2042-L2043

#### Title being sent
https://github.com/MIT-LCP/physionet-build/blob/83b29d9124ab80253d1678feeffaf05afa9b5a57/physionet-django/console/views.py#L2053-L2057

https://github.com/MIT-LCP/physionet-build/blob/83b29d9124ab80253d1678feeffaf05afa9b5a57/physionet-django/console/templates/console/static_page_sections_edit.html#L13

### Solution
I just used the PK in the url param instead of the url. Alternatively title could be used but I think it doesn't fit semantically for the title to unambiguously identify a page (e.g. in case of two pages with the same title)

### DIFF
--- a/physionet-django/console/templates/console/static_page_sections.html
+++ b/physionet-django/console/templates/console/static_page_sections.html
@@ -36,10 +36,10 @@
                         </form>
                     </td>
                     <td>
-                        <a href="{% url 'static_page_sections_edit' page section.id %}" class="btn btn-sm btn-primary" role="button">Edit</a>
+                        <a href="{% url 'static_page_sections_edit' page.pk section.id %}" class="btn btn-sm btn-primary" role="button">Edit</a>
                     </td>
                     <td>
-                        <form method="post" action="{% url 'static_page_sections_delete' page section.id %}">
+                        <form method="post" action="{% url 'static_page_sections_delete' page.pk section.id %}">
                             {% csrf_token %}
                             <button type="submit" class="btn btn-sm btn-danger" role="button">Remove</button>
                         </form>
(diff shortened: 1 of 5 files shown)


tompollard:
@briangow was working on fixing this, sorry @kshalot we should have said! Brian please could you take a look at this pull request and see whether it does what you need?

kshalot:
> @briangow was working on fixing this, sorry @kshalot we should have said! Brian please could you take a look at this pull request and see whether it does what you need?

No problem, it wasn't time consuming anyway. @briangow if you have a better solution ready already then just close this PR.
I amended the commit here to fix a linter issue 👎

briangow:
@kshalot , this looks good to me! Thanks for cleaning up the issues I caused, sorry for the extra work.

PR #1510, opened 2022-03-11
Create generalized url pattern, view, template and nav bar for use with `StaticPage`

briangow:
This PR uses the `StaticPage` model and generalizes the associated url pattern, view, and template for the `/about/` and `/about/publish` pages. 

Now instead of a hardcoded path in `/physionet/urls.py` the `/about/publish` page is caught by using `about/<path:static_url>/`where `static_url` comes from the `StaticPage` object. Similarly, a new view has been created, `static_view`, for the `physionet` app which takes the `static_url` and attempts to render a page with content from an associated `Section` object . The template that gets called from `static_view` renders the page based these objects. 

The nav bar at the top of each page on the site has also been updated to allow it to be built based upon new fields that have been added to the `StaticPage` model. The new fields, `nav_bar` and `nav_order` determine whether a link to a given page appears in the nav bar and in what order those links appear.  The `demo-physionet` fixture has been updated to include these new fields for the `/about/` and `about/publish` pages.


physionet-django/physionet/models.py
```diff
@@ -4,14 +4,23 @@
...
 
 class StaticPage(models.Model):
+    """
+    Allows pages on the site to be created via the admin tool. Controls whether a given page appears in the navigation
+    bar.
+    """
     title = models.CharField(max_length=64)
     url = models.CharField(max_length=64, unique=True)
+    nav_bar = models.BooleanField(default=False)
+    nav_order = models.IntegerField(unique=True, null=True)
```
kshalot: 1. Since this column defines the order, maybe it makes more sense to support only positive integers?
2. Setting just `null=True` will only prevent the db from setting the `NOT NULL` constraint but all the Django forms and logic will still treat this field as required - e.g. now it's impossible to create a new static page that does not have this field set through the admin console. Adding `blank=True` allows this field to not be set through a form (I assume that was the expectation)
```suggestion
    nav_order = models.PositiveSmallIntegerField(unique=True, null=True, blank=True)
```

physionet-django/templates/navbar_content.html
```diff
@@ -7,17 +11,15 @@
...
-
-  <li class="nav-item">
-    <a id="nav_about" class="nav-link" href="{% url 'about' %}">
-        About
-    </a>
-  </li>
+  {% for page in static_pg %}
+    {% if page.nav_bar %}
+      <li class="nav-item">
+        <a id="nav_{{ page.title | lower }}" class="nav-link" href="{{ page.url }}" >
```
kshalot: Aren't you missing the `underscore` filter here? All the other ids in this PR use it, now multi-word page titles will have spaces in their ID.
```suggestion
        <a id="nav_{{ page.title | underscore | lower }}" class="nav-link" href="{{ page.url }}" >
```

Since the `underscore` filter is only used in this combination then maybe it would be a good idea to just combine the functionality, like this:
```python
@register.filter
def snake_case(string):
    return string.replace(' ', '_').lower()
```
and use that in the templates instead of chaining`underscore | lower` each time

kshalot (approved):
Left two small things to consider, but LGTM 👍

physionet-django/physionet/models.py
briangow: Thanks @kshalot , these are good suggestions! I've implemented them.

physionet-django/templates/navbar_content.html
briangow: Thanks @kshalot , I've added the `underscore` filter to the navbar template. We decided to keep it as its own filter so it could be used independent of `lower` if desired in the future.

tompollard:
Looks good, thanks!

Issue #1514, opened 2022-03-19
Training page in console is very slow to load


tompollard:
The new page in the console that lists training applications is extremely slow (>30 seconds) to load when there are a large number of applications. See https://localhost:8000/console/training/ on the live site. We should take a look to see what we can do to speed things up.

tompollard:
Fixed in https://github.com/MIT-LCP/physionet-build/pull/1517

PR #1518, opened 2022-03-22
Add search bar to the training pages in the console

tompollard:
In https://github.com/MIT-LCP/physionet-build/pull/1517 we added pagination to the training pages in the console. This fixed the issue of slow loading pages (https://github.com/MIT-LCP/physionet-build/issues/1514) but it is now difficult to find applications because they are distributed across multiple pages.

This pull request: 
1. adds a search bar to the training pages in the console (e.g. http://localhost:8000/console/training/review). Applications can be searched by username, firstname, lastname, and email.
2. lists applications from oldest to newest (rather than newest to oldest) at the request of Sophie.

allow applications to be searched. also order applications from old to new, instead of new to old.

### DIFF
--- a/physionet-django/console/templates/console/training_list.html
+++ b/physionet-django/console/templates/console/training_list.html
@@ -45,17 +45,14 @@
     <div class="card-body">
       <div class="tab-content">
         {% if trainings %}
+          <input type="search" oninput="search('{% url "training_list" status=status %}', value);" placeholder="Search..." >
           {% if status == 'review' %}
-            <input type="search" oninput="search('{% url "training_list" status="review" %}', value);" placeholder="Search...">
             {% include 'console/review_training_table.html' %}
           {% elif status == 'valid' %}
-            <input type="search" oninput="search('{% url "training_list" status="valid" %}', value);" placeholder="Search...">
             {% include 'console/valid_training_table.html' %}
           {% elif status == 'expired' %}
-            <input type="search" oninput="search('{% url "training_list" status="expired" %}', value);" placeholder="Search...">
             {% include 'console/expired_training_table.html' %}
           {% elif status == 'rejected' %}
-            <input type="search" oninput="search('{% url "training_list" status="rejected" %}', value);" placeholder="Search...">
             {% include 'console/rejected_training_table.html' %}
           {% endif %}
         {% else %}
--- a/physionet-django/console/templates/console/training_list.html
+++ b/physionet-django/console/templates/console/training_list.html
@@ -6,6 +6,12 @@
 
 {% load console_templatetags %}
 
+{% block local_js_top %}
+<script src="{% static 'custom/js/cookie.js' %}"></script>
+<script src="{% static 'project/js/dynamic-formset.js' %}"></script>
+<script src="{% static 'console/js/search-console.js' %}"></script>
+{% endblock %}
+
 {% block local_css %}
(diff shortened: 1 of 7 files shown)


physionet-django/console/templates/console/training_list.html
```diff
@@ -40,12 +46,16 @@
       <div class="tab-content">
         {% if trainings %}
           {% if status == 'review' %}
+            <input type="search" oninput="search('{% url "training_list" status="review" %}', value);" placeholder="Search...">
```
kshalot: I'd render this input only once, under `if trainings` but above all the status conditions. Current status is available in the template so this would work:
```html
<input
  type="search"
  oninput="search('{% url "training_list" status=status %}', value);"
  placeholder="Search..."
>
```

physionet-django/console/views.py
```diff
@@ -1550,6 +1550,22 @@ def training_list(request, status):
 
     display_training = training_by_status[status]
 
+    if request.method == 'POST':
+        if "search" in request.POST:
+            display_training = search_training_applications(request, display_training)
+            if status == 'review':
```
kshalot: It's a matter of subjective preference but to me this code would look better like this:
```python
template_by_status = {
  'review': 'console/review_training_table.html',
  'valid': 'console/valid_training_table.html',
  'expired': 'console/expired_training_table.html',
  'rejected': 'console/rejected_training_table.html',
}
return render(request, template_by_status[status], {'trainings': display_training, 'status': status})
```

No branching and it's immediately clear that the template is the only thing that's different for different statuses. It's also possible to go as far as to do `f"console/{status}_training_table.html"` but coupling the file name doesn't feel right.

tompollard: Much cleaner, thanks for the suggestion!

physionet-django/console/templates/console/training_list.html
tompollard: Good point, thanks!

tompollard:
@kshalot Great suggestions, thanks. I have incorporated your changes into the PR.

kshalot:
LGTM!

@tompollard one last thing I noticed (not related to you changes at all) - I did not notice when adding the pagination that I duplicated the urls:
https://github.com/MIT-LCP/physionet-build/blob/d3b73f1e1bbb0a01ec7d3b8aebad0e282a850d06/physionet-django/console/urls.py#L75-L76
Navigation through anchors works (because I missed the `/` at the end of the first url), but if you try going to e.g. `/console/training/1` manually (by typing in the browser bar), it's going to blow up, as it will match the first url that filters by status. I'd suggest adding the forward slash to the first path `training/<status>/` and renaming the second path to e.g. `training/view/<int:pk>` to avoid undefined behaviour.

I'm mentioning it now because you can change it immediately in this PR. If not, I can also add it later myself.

kshalot (approved):

kshalot:
[6e316d8](https://github.com/MIT-LCP/physionet-build/pull/1518/commits/6e316d8a5da15e504d081b4f66005583234e10ff) - LGTM 👍

tompollard:
@bemoody Non-standard for to be merging my own PR but @kshalot has reviewed and he is happy, so I'm going to go ahead and merge which I hope is okay with you. 

Side note, but @cx1111 made a good argument people merging their own PRs after approval, so we could consider making it new policy in future.

PR #1487, opened 2022-02-28
Permissions system

fiedosiukr:
## Notes
The Django admin console is now only accessible for superusers (`is_superuser` field). A superuser has all of the existing permissions.
## Roles

1. Admin
    - assigning project editor
    - reviewing assigned projects
    - credentialing
    - storage requests
    - viewing project access
    - viewing users
    - featured content
    - news
    - usage stats
    - static pages

2. Managing Editor
    - assigning project editor
    - reviewing assigned projects
    - viewing project access
    - storage requests
    - usage stats

3. Handling Editor
    - reviewing assigned projects
    - viewing project access

4. Reviewer
    - reviewing assigned projects

5. Writer
    - featured content
    - news
    - static pages

6. Credentialer
    - credentialing
    - viewing users

PR #1519, opened 2022-03-22
Split DUA and License

kshalot:
# Todo
- (?) create a data use agreement for every license
- create a sample DUA for unit testing purposes
- once #1487 is merged, it needs to be integrated with the permissions system

## Notes
- ~~in order to apply all migrations on a fresh database instance, one of the migration was modified. Probably needs a discussion.~~ I created a separate migration with the changes instead - testing whether it won't break anything (@kshalot) 

## Changes
- creates a separate table containing data use agreements
- makes license and DUA objects independent
- DUA may be only selected for credentialed and contributor review projects
- adds view to edit licenses and DUAs in the admin console


tompollard:
I need to investigate a little more, but it seems there is a bug in the migration of existing agreements. To reproduce the issue:

### 1. Run the following steps on the dev branch (i.e. prior to the changes in the PR)

a. Go to http://localhost:8000/login/ and login as `rgmark`.
b. Sign the DUA for the [demo eICU project](http://localhost:8000/content/demoeicu/2.0.0/).
c. Go to the Agreements page: http://localhost:8000/settings/agreements/. This page shows:

> Screen Shot 2022-03-22 at 14 54 00

d. Click the "View" link to view the signed agreement (direct link should be http://localhost:8000/settings/agreements/2/). This page displays:

> Screen Shot 2022-03-22 at 14 51 43

### 2. Now switch to the PR branch

a. Migrate the data with `python manage.py migrate`.
b. Now view the previous two pages. The [Agreements page](http://localhost:8000/settings/agreements/) now shows:

> Screen Shot 2022-03-22 at 14 56 09

c. The signed agreement (probably at http://localhost:8000/settings/agreements/2/) now displays:

> Screen Shot 2022-03-22 at 14 56 57

kshalot:
@tompollard I looked into it and found a few more issues on my own:

# Easy fixes

## [6094701](https://github.com/MIT-LCP/physionet-build/pull/1519/commits/6094701f6ba3d80db7a595ab178a32756e7a941f) & [3fd78e0](https://github.com/MIT-LCP/physionet-build/pull/1519/commits/3fd78e006eda9530d5526d833081245909746180)
This is the issue that you found. The migration now correctly sets the new DUA field for existing projects + the DUA is accesses correctly in the template.
Screenshot 2022-03-23 at 10 26 08

## [f6ad519](https://github.com/MIT-LCP/physionet-build/pull/1519/commits/f6ad5190c4d89f21637055888c843804a5e926c5)
Exactly the same problem as above - `license.dua_html_content` -> `dua.html_content`

## [fbcbd86](https://github.com/MIT-LCP/physionet-build/pull/1519/commits/fbcbd864ec0addd6a42860e551f836be9f5dfafc)
Simply wrong parameter name.

## [a437661](https://github.com/MIT-LCP/physionet-build/pull/1519/commits/a4376616e552e456f35ce64d79e90831e77fc891)
DUA slug was set to be unique which made it impossible to create new versions of DUAs.

# Tricky

These two issues are related to hard-coded licenses in the code - I'm not 100% sure which DUA we need to replace them with. Also hard coding this might not be a great idea - how about introducing some settings that would control these values? E.g:
```
CREDENTIAL_APPLICATION_DUA_SLUG=''
CREDENTIAL_APPLICATION_DUA_VERSION=''
ANONYMOUS_ACCESS_DUA_SLUG=''
ANONYMOUS_ACCESS_DUA_VERSION=''
```

## [47e993b](https://github.com/MIT-LCP/physionet-build/pull/1519/commits/47e993b6dd297a80921c2e0daff17aecfc18cd8f)
This is something that I won't be able to figure out on my own - right now the anonymous access DUA is being accessed through the license with slug `physionet-restricted-health-data-license-150` - I changed it to `physionet-restricted-health-data-dua` temporarily - we need to figure out the value we want there.

## [a95de83](https://github.com/MIT-LCP/physionet-build/pull/1519/commits/a95de834a91bfcef54534e897702b8a6a616e27d)
This one is the most confusing to me - I have no idea which License has id=6, so I don't know which DUA we need there.

kshalot:
@tompollard Got the workflows working - renaming a ManyToMany column was problematic (`required_training` -> `required_trainings`) but looks fine now. Encountered some other errors along the way and fixed those as well.

kshalot:
I'm also wondering about the reversibility of the [0059_license_dua_2](https://github.com/MIT-LCP/physionet-build/pull/1519/files#diff-115d969898605d54b4c7b0c7822f350fa2e4b2cf1edc7b3ed945d2f80a0046c8) migration - it removes the DUA fields from the License model and there is no good way to bring those fields back (as in - copy data from the DUA table into dua fields in License when reversing this migration) since there is no way to tell which DUA belonged to which license. I don't know whether this is relevant for you (I assume you won't reverse this migration) but figured I'll just put it out there.

tompollard:
thanks @kshalot, i'll take a look shortly. glad to hear you managed to get the tests passing and hope it wasn't too painful! i'm not too worried about the reversability of [0059_license_dua_2](https://github.com/MIT-LCP/physionet-build/pull/1519/files#diff-115d969898605d54b4c7b0c7822f350fa2e4b2cf1edc7b3ed945d2f80a0046c8).

physionet-django/console/forms.py
```diff
@@ -934,6 +941,43 @@ def clean(self):
...
+            'access_template',
+            'access_policy',
+            'project_types',
+        )
+        labels = {'html_content': 'Content'}
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields['access_policy'].choices = AccessPolicy.choices(gte_value=AccessPolicy.CREDENTIALED)
```
tompollard: Restricted datasets require a DUA, so I think this should be `gte_value=AccessPolicy.RESTRICTED`

tompollard (commented):
Thanks, looks good to me! There are couple of issues that I think we need to follow up on later:

1. It should not be possible to delete an DUA that someone has signed. Right now, when I delete a DUA, I destroy all of the records of anyone signing it (e.g. the agreements at https://physionet.org/settings/agreements/ now appear blank)
2. The tools used to create/edit the license and DUA need a bit more explanation. e.g. on http://localhost:8000/console/duas/, what is the meaning of "is active", "access template", etc.

physionet-django/console/forms.py
tompollard: Updated in https://github.com/MIT-LCP/physionet-build/pull/1519/commits/7a12b0d12d36b86607efb9748db5cc9608089c10

PR #1521, opened 2022-03-22
Handle lightwave subdirectories in active projects

bemoody:
LightWAVE allows you to break up a large database into "sub-databases" where each one has its own RECORDS file, and the top-level RECORDS file just lists the subdirectories (indicated by a trailing slash.)  This feature isn't widely used and was a bit of a kludge to begin with, but it's used in the MIMIC-III Waveform Database and is useful to make it easier to navigate other large databases.

This feature was broken for *active* (as opposed to published) projects because lightwave_project_server would unconditionally override the client-specified database path, always setting it to the root of the project.


tompollard:
Thanks Benjamin!

PR #1522, opened 2022-03-22
Install LightWAVE in test server

bemoody:
Download and install WFDB and LightWAVE as part of `install-pn-test-server`, to allow easy testing.


tompollard:
thanks!

PR #1524, opened 2022-03-23
Tests on Debian 11

bemoody:
Run workflow tests on Debian 11 to see if it will work.


bemoody:
It seems to be compiling cffi from source now and wasn't before.  After upgrading OS, try updating python packages and figure out which ones (still) need to be compiled from source.

tompollard:
Looks good, thanks!

PR #1527, opened 2022-03-24
Add Code of Conduct management

kshalot:


physionet-django/console/forms.py
```diff
@@ -975,7 +973,14 @@ class Meta:
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        self.fields['access_policy'].choices = AccessPolicy.choices(gte_value=AccessPolicy.RESTRICTED)
+        self.fields['access_policy'].choices = AccessPolicy.choices(gte_value=AccessPolicy.CREDENTIALED)
```
tompollard: This looks like it is undoing https://github.com/MIT-LCP/physionet-build/pull/1519/commits/7a12b0d12d36b86607efb9748db5cc9608089c10.

It should be possible to assign a DUA to restricted access projects, so should this stay as `gte_value=AccessPolicy.RESTRICTED`?

physionet-django/user/forms.py
```diff
@@ -668,8 +668,6 @@ def __init__(self, user, *args, **kwargs):
         self.fields['training_type'].initial = self.training_type
 
         if self.training_type is not None:
-            self.fields['training_type'].help_text = self.training_type.description
```
tompollard: Does this line need to be removed? 

In https://github.com/MIT-LCP/physionet-build/pull/1529/commits/2938c10470a414ab3dfd6bd20b91170d01e27323#diff-1616a969a01973c697c08fe3fe6011689bdaa0d7684d5746f130651bc7572667 it looks like @adrianryt is fixing a problem with the rendering of the field in the template. Are these two changes addressing the same issue in different ways?

physionet-django/console/forms.py
kshalot: Yeah, good catch. I don't know which commit reverted this change - I'll make it correct again and let you know.

physionet-django/user/forms.py
kshalot: @tompollard Seems Rafał removed it in one of his commits. I restored it to the way it's done on dev.

physionet-django/console/forms.py
kshalot: Brought `AccessPolicy.RESTRICTED` back.

tompollard:
Looks great, thanks! In a future PR, we may want to:

- Allow the user to view a signed code of conduct at: http://localhost:8000/settings/agreements/ (at the moment, there is no link to the signed code of conduct, as far as I can see).
- Display a copy of the signed code of conduct in the "DataShare credentialing application notification" email to the user.

PR #1528, opened 2022-03-24
DUA/License - minor bugfixes

kshalot:
## 6d393a0ec633cdf71c4a504e024b074e4bb9ac20 & 0520f45329558d592ba775e65555a9f9bd30cbc9
Each button referenced the same ID and each modal had the same ID assigned - this means that regardless of which delete button was clicked, the first modal from the top was always opened and the first DUA/License from the top was deleted.

## f567be00ffddb59801d6a846e0f6a96d1e5d4462
Slugs should clearly be unique because of these views:
https://github.com/MIT-LCP/physionet-build/blob/f7b6d06a4f7cc61a882d412745e033c14ddc7d92/physionet-django/physionet/urls.py#L41-L42

Both expect to get at most one DUA/License by the slug:
https://github.com/MIT-LCP/physionet-build/blob/f7b6d06a4f7cc61a882d412745e033c14ddc7d92/physionet-django/physionet/views.py#L68-L83


tompollard:
Thanks Karol!

PR #1537, opened 2022-03-29
Fix bug at credential_application 

adrianryt:
When invalid form data was provided at Credential Application, variable 'code_of_conduct' wasn't initialized before use.


tompollard:
thanks for picking this up

PR #1538, opened 2022-03-29
WIP: Add cloud-native environments app host logic

kshalot:
## Research Environments
This is the host portion of the research environment integration. The application's code will be hosted under https://github.com/T-CAIREM/hdn-research-environment - it will soon become public.

Some of the logic here might change today/tomorrow, hence the WIP tag.

### What's changed
- `environment` urls added conditionally
- New `gcp-research-environment` project DataAccess that provides the platform with the IAM group that grants access to data in GCP. This may change to be automatically generated, but the API is currently very limiting when it comes to naming so it might not be possible to base this value on the slug/version combination. It has to be set when the project is published and can be later changed through the DataAccess form (only when `ENABLE_RESEARCH_ENVIRONMENTS=True`)

### What's missing
- [ ]  Add the Django app to poetry and requirements.


physionet-django/project/managers/project.py
```diff
@@ -0,0 +1,45 @@
+from django.db.models import Manager, Q
+from project.models import AccessPolicy, DUASignature, DataAccessRequest
+from user.models import Training, TrainingType
+
+
+class PublishedProjectManager(Manager):
+    def accessible_by(self, user):
```
kshalot, line 7: We need this logic in the environments app. Figured it may come useful somewhere else as well.

PR #1536, opened 2022-03-29
Add SOURCE_CODE_REPOSITORY_LINK variable to the .env.example file

adrianryt:
In this PR I add SOURCE_CODE_REPOSITORY_LINK variable to the .env.example file.

PR #1539, opened 2022-03-29
Remove example project link from edit_cloud template

adrianryt:
In this PR I create EXAMPLE_PROJECT_FILES_LINK variable with default value: None. It can be configured with env file.
In the .env.example file I put the previous used url. 

Note: Please let's first merge [#1536](https://github.com/MIT-LCP/physionet-build/pull/1536). Then I will resolve conflicts in this PR.

### DIFF
--- a/physionet-django/project/views.py
+++ b/physionet-django/project/views.py
@@ -719,7 +719,7 @@ def project_access(request, project_slug, **kwargs):
         access_form = forms.AccessMetadataForm(data=request.POST, instance=project, editable=editable)
         if access_form.is_valid():
             access_form.save()
-            messages.success(request, 'Your access metdata has been updated.')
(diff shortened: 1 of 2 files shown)


tompollard:
Thanks @adrianryt. Instead of adding this template variable, I wonder whether we should just remove `(for example, https://physionet.org/content/eicu-crd-demo/#files)` from the template? The example doesn't really seem necessary.

bemoody:
Yeah, I would just get rid of it.  Also the name
"EXAMPLE_PROJECT_FILES_LINK" doesn't convey that this link needs to
point to some existing published project that is known to be integrated
with (unspecified?) cloud services.

adrianryt:
@tompollard good point, I removed example link.

PR #1541, opened 2022-03-30
Fix bug that shows total uncompressed file size is 0B 

adrianryt:
Note: Please, do not merge yet(need more investigation)

After creating project with files, Project view shows: "Total uncompressed file size: 0B ".
My solution is to call function that calculates project's file size after publishing.

### DIFF
--- a/physionet-django/physionet/gcs.py
+++ b/physionet-django/physionet/gcs.py
@@ -95,7 +95,8 @@ class GCSObject:
     def size(self):
         """Size of the object/all objects in the dictionary, in bytes."""
         if self.is_dir():
-            return sum(obj.size for obj in self.bucket.list_blobs(prefix=self.name))
+            prefix = '' if self.name == '/' else self.name
+            return sum(obj.size for obj in self.bucket.list_blobs(prefix=prefix))
 
         file = self.bucket.get_blob(self.blob.name)
         if not file:


bemoody:
Calculating storage size is potentially a long-running process.  We don't want to do that in publish().

In fact what we *should* be doing is running all of the long-running stuff asynchronously before publish() happens.

- We should generate SHA256SUMS.txt (and LICENSE.txt) after the editor clicks "finish copyedit" and before notifying the authors that the project is ready for approval.

- We should set file permissions read-only after all authors approve the project and before notifying the editor that the project can be published.

- We should calculate the total size, and scan for invalid file names, and maybe generate SHA256SUMS.txt as well, every time the author clicks "submit" and before notifying the editor that the project is ready for review.

bemoody:
Consider, as a short-term solution, calculating the size in console.forms.CopyeditForm.save.  (Which really ought to be a method of ActiveProject, not buried in console.)

Anyway, why does this matter?  Because publishing a project is a "big switch" and we really do not want this to freeze and leave the editor wondering if it worked or not.

adrianryt:
Thanks, @bemoody for the comment, after some investigations it turned out that the problem lies elsewhere.
Function:
```
sum(obj.size for obj in self.bucket.list_blobs(prefix=self.name))
```
in gcp.py was returning 0. It was because 
```
self.bucket.list_blobs(prefix=self.name) 
```
with prefix='/' returns an empty list - blobs in the bucket don't have a root prefix.
A quick fix for the problem(used also in other places) is to change the prefix from '/' to ''.
I amend committed the previous changes.

tompollard:
@bemoody happy to merge this?

bemoody:
I haven't tested it, but seems fine!

PR #1543, opened 2022-03-30
Rate limits on /register/

bemoody:
Put some limits on /register/ to reduce abuse.

The first speed bump hopefully should prevent a lot of unintelligent bots from registering, simply by requiring that you load the /register/ page and wait at least 15 seconds before submitting it.

Secondly, disallow someone from creating a huge number of accounts at once from the same address.  This requires recording the address from which each account was registered.


tompollard:
Thanks for looking at this. I just timed how long it takes me to fill in a form like this and it is ~5 seconds, because the fields are autofilled by my browser. Could we reduce to say 10 seconds? I really think we should consider captcha in addition to this.

bemoody:
Agreed it's an annoyance in development.  If you're a real person
registering for a real account then surely you're not auto-filling and
15 seconds doesn't seem overly burdensome.  Maybe it is, I don't know.

tompollard:
For me, any form like this is auto-populated by my browser (Chrome), so I think the 5 seconds is fairly realistic. I click the email field, my browser displays a list of profiles to choose from (comprising email, name, etc). I pick one and then the form is complete. 

I'm okay to merge this as it stands, but I think a significant proportion of people will get the "Please wait and try again message". It's not clear to me how problematic this will be. Probably just a minor annoyance, but it does require people to read the error message, which isn't a given!

tompollard:
Side note, but I'd like to try to merge https://github.com/MIT-LCP/physionet-build/pull/1487 first, so we might need to regenerate the migration before merging. Regenerating the migrations for https://github.com/MIT-LCP/physionet-build/pull/1487 is a little less straightforward.

bemoody:
> Regenerating the migrations for
> https://github.com/MIT-LCP/physionet-build/pull/1487 is a little less
> straightforward.

It really shouldn't be, I volunteer to do the sed-work if necessary.

I'd like to get the change to record registration IP deployed ASAP.

PR #1545, opened 2022-03-31
Add email notifications for (1) submission of training reports and (2) "immediate acceptance" of credentialing applications

adrianryt:
Add email notifications after approving the credential application and after submitting training.

Email is now sent when:
-user submits a credentialing application
-user submits a training application
-credentialing application is approved/rejected
-training application is approved/rejected

### DIFF
--- a/physionet-django/notification/utility.py
+++ b/physionet-django/notification/utility.py
@@ -693,6 +693,23 @@ def process_training_complete(request, training, include_comments=True):
     message.send(fail_silently=False)
 
 
+def training_application_request(request, training):
+    """ Notify user of training submit """
+
+    subject = f'{settings.SITE_NAME} training application notification'
+    body = loader.render_to_string(
+        'notification/email/notify_training_request.html', {
+            'training': training,
+            'applicant_name': training.user.get_full_name(),
+            'domain': get_current_site(request),
+            'url_prefix': get_url_prefix(request),
+            'signature': settings.EMAIL_SIGNATURE,
+            'footer': email_footer(), 'SITE_NAME': settings.SITE_NAME
+        })
+    send_mail(subject, body, settings.DEFAULT_FROM_EMAIL,
+              [training.user.email], fail_silently=False)
+
+
 def credential_application_request(request, application):
     """
     Notify user of credentialing decision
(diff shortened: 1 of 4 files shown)


tompollard:
Thanks @adrianryt. It looks like this mainly relates to training notifications, rather than credentialing (which are now two separate processes). Please could you add a more descriptive title for the issue, and perhaps update the description to outline the main changes? 

As far as I remember we already send notifications when a user applies for credentialing and when a credentialing decision is made, so I'm surprised that these emails need to be added (note looked at code yet, so perhaps I've misunderstood).

adrianryt:
Hi @tompollard, I think there was a bug and we didn't send an email after 'immediate accept' (approving the credential application case) also I added an email notification after submitting training by the user (since we send email notification after submitting credentialing)

physionet-django/notification/utility.py
```diff
@@ -693,6 +693,25 @@ def process_training_complete(request, training, include_comments=True):
...
+    body = loader.render_to_string(
+        'notification/email/notify_training_request.html', {
+            'training': training,
+            'applicant_name': training.user.get_full_name(),
+            'domain': get_current_site(request),
+            'url_prefix': get_url_prefix(request),
+            'signature': settings.EMAIL_SIGNATURE,
+            'footer': email_footer(), 'SITE_NAME': settings.SITE_NAME
+        })
+    print("DUPA", subject)
```
tompollard: I just learnt my first word in Polish!

kshalot:
@tompollard I removed all the _bad_ lines. I'm looking at this commit though and verifying it. The reported issue was that:
1. There is no mail sent to the user after their credentialing application is approved.
2. There is no mail sent to the user after their training is submitted.

Seems that at least issue 1. is not really an issue - I just verified that the email is in fact being sent. Must have been a one time fluke for the person reporting the issue.

tompollard:
Thanks @kshalot, I haven't yet tested but from the code everything now looks good to me. Unless I'm missing something, it looks like Adrian is right about the notification email...I added an "Immediate Accept" button that appears on the very first stage of the credentialing process, but I think I accidentally omitted the "Send an email" step.

kshalot:
> Thanks @kshalot, I haven't yet tested but from the code everything now looks good to me. Unless I'm missing something, it looks like Adrian is right about the notification email...I added an "Immediate Accept" button that appears on the very first stage of the credentialing process, but I think I accidentally omitted the "Send an email" step.

Yeah, I just verified that:
1. The credentialing accept email is not sent only when using the "Immediate accept" button
2. No email is sent when submitting a training.

tompollard:
Looks good to me, thanks @adrianryt and @kshalot

Issue #1533, opened 2022-03-25
Update location of training files


tompollard:
Training files are saved in two different locations on our servers - "training" and "trainings" - as a result of a minor mix-up (https://github.com/MIT-LCP/physionet-build/pull/1464/commits/b66f59d2aec5f7229b4e7ee3a4bd2596fb345709 and https://github.com/MIT-LCP/physionet-build/pull/1519/files). @kshalot outlined the solution below:

> Yeah, the initial “split training and credentialing” PR used singular and the follow-up DUA PR changed it to plural. Since the DUA PR came later I kind of assumed that it’s supposed to be the more up-to-date in terms of naming so I chose plural when resolving conflicts, but I agree - training refers to some kind of process so uncountable is probably the way to go here.

> So the problem is that for a couple of days before the DUA PR got merged, the training certificates were saved under training/ and now after the DUA PR is merged it’s saved under trainings/ so I assume you just got a couple of PDFs that are in the wrong place now. I created a PR that saves the certificates under training/ again - https://github.com/MIT-LCP/physionet-build/pull/1532/files.
> - When it comes to cleanup then you’ll have to:
> - Copy the files from trainings/ to training/
> - Change the out of date urls in the database, e.g. with this script:

> ```from django.db.models import F, Func, Value
> 
> Training.objects.update(
>     completion_report=Func(
>             F('completion_report'),
>             Value('trainings'), Value('training'),
>             function='replace',
>     )
> )
> ```
>  3. Remove the trainings/ directory
> That should do the trick.

> Also I think that the pdfs were saved under credential-applications earlier, right? If you want then you could move them to training/ as well, then everything would be in the same place.

tompollard:
I have migrated all training reports in the `trainings` and `credential-applications` directories to the "`/training/`" directory. I also updated the database to point to the new location following Karol's suggestion above.

### Reports in `/trainings/`

The path was updated in the database with:

```
from user.models import Training
Training.objects.update(
    completion_report=Func(
            F('completion_report'),
            Value('trainings'), Value('training'),
            function='replace',
    )
)
```

### Reports in `/credential-applications/`

The path was updated in the database with:

```
from user.models import Training
from django.db.models import F, Func, Value


Training.objects.update(
    completion_report=Func(
            F('completion_report'),
            Value('credential-applications'), Value('training'),
            function='replace',
    )
)
```

### Testing

[...]

PR #1550, opened 2022-04-05
Remove unused function for getting the training_report_path

tompollard:
While consolidating the paths for training reports on our server, I noticed a redundant function:

```
def training_report_path(instance, filename):
    """
    Storage path of CITI training report
    """
    return 'credential-applications/{}/{}'.format(instance.slug, 'training-report.pdf')
```

As far as I can see, this function isn't used. Instead, we now use:

```
def get_training_path(instance, filename):
    """
    Storage path for training reports relative to media root.
    """
    return f'training/{instance.slug}/training-report.pdf'
```

Side note, is that `get_training_path` asks for filename, but doesn't use it anywhere. This might be a bug? Shouldn't `get_training_path` specify the type of training in the filename? Otherwise I think training reports will be overwritten if there are multiple types @alistairewj @kshalot.


tompollard:
hmm, do we have to keep it because it appears in the migrations? @bemoody

kshalot:
### Unused `training_report_path` function
My hunch is that this function was simply renamed to `get_training_path` but the old name was kept for the sake of old migrations (which you mentioned). Honestly it looks like you could just remove that function and change it's invocations to `get_training_path` in the migrations.

### Unused `filename` parameter
The parameter is actually necessary because it's part of the FileField "contract". You can see it described here - https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.FileField.upload_to. The function's arity has to be 2.

This actually shouldn't introduce any conflict since the report is created under a directory denoted by the training's slug - the slug is unique for each training (training == training completion by a user, not a specific training type).

tompollard:
thanks for your helpful comments @kshalot

tompollard:
@bemoody please could you take a look at this when you have a chance?

bemoody:
It looks reasonable and tests are passing.  I haven't really followed
this transition, other than I remember some things were failing when
Karol or somebody tried to write a migration to move the files
automatically.

At this point, have all the files been moved
into '/data/pn-media/training/' on the live server?  Presumably we want
to do the same thing in 'demo-files/media'.

bemoody:
> Side note, is that get_training_path asks for filename, but doesn't use it anywhere. This might be a bug?

I could be completely wrong, but I would guess that the filename argument here is the filename provided by the web browser.  We probably want to ignore that.

tompollard:
> At this point, have all the files been moved into '/data/pn-media/training/' on the live server?

Yes, all the files should be in the new location (see: https://github.com/MIT-LCP/physionet-build/issues/1533).

> Presumably we want to do the same thing in 'demo-files/media'.

I hadn't thought about this, but sounds like a good idea. Can we raise this as an issue and do it later or would you like it addressed in this PR?

tompollard:
> I could be completely wrong, but I would guess that the filename argument here is the filename provided by the web browser. We probably want to ignore that.

Sorry, missed this. See Karol's response to this point above (copied below):

> Unused filename parameter
The parameter is actually necessary because it's part of the FileField "contract". You can see it described here - https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.FileField.upload_to. The function's arity has to be 2.

> This actually shouldn't introduce any conflict since the report is created under a directory denoted by the training's slug - the slug is unique for each training (training == training completion by a user, not a specific training type).

bemoody:
> I hadn't thought about this, but sounds like a good idea. Can we raise this as an issue and do it later or would you like it addressed in this PR?
Let's do it separately (see: issue #1561).

tompollard:
(@bemoody confirmed that he's happy to merge)

bemoody:
It looks reasonable and tests are passing.  I haven't really followed
this transition, other than I remember some things were failing when
Karol or somebody tried to write a migration to move the files
automatically.

At this point, have all the files been moved
into '/data/pn-media/training/' on the live server?  Presumably we want
to do the same thing in 'demo-files/media'.

PR #1552, opened 2022-04-05
Tidy up nginx configuration

bemoody:
These changes shuffle the nginx configuration around, removing most of the duplicated stuff (between http and https, and between production and staging) to be hopefully more maintainable.

These changes *should not* have any observable effects.


tompollard:
Thanks!

PR #1554, opened 2022-04-05
Make .env readable only by www-data group

bemoody:
The file .env contains extremely sensitive information and should not be world-readable.  It'd be better to have a special group just for accessing this file (and maybe the rest of /physionet/), but for now, the simplest thing is to make this file readable only by members of the www-data group.  (All of the admin users are members of www-data.)

This is now done on the staging/production/live servers; do the same in install-pn-test-server for consistency.


tompollard:
thanks, looks good to me.

PR #1555, opened 2022-04-06
Fix credentialing application status displayed to users. Ref #1485.

tompollard:
In https://github.com/MIT-LCP/physionet-build/pull/1485 we added the status of credentialing applications to the user settings page: http://localhost:8000/settings/credentialing/

In https://github.com/MIT-LCP/physionet-build/pull/1464 we removed one of the statuses from the credentialing workflow. As a result, we are now displaying the wrong status to users (specifically, we display "awaiting review" when applications are in fact waiting for response from the referee).

This pull request updates the logic to display the correct status to users.


briangow:
Thanks @tompollard !

PR #1556, opened 2022-04-06
Remove the 'initial review' stage from the credentialing process.

tompollard:
The initial review stage in the credentialing process displays the following questions:

Screen Shot 2022-04-06 at 12 54 40

Based on feedback from people doing the credentialing, this step is not necessary. It just adds an extra couple of clicks to the credentialing process.

In this pull request, I:
1. Remove the initial review stage of the credentialing workflow
2. Ensure that the statuses of existing credentialing applications are updated (by subtracting 10 when appropriate).

PR #1558, opened 2022-04-12
Display 'ID check' tab by default on credentialing page

tompollard:
In https://github.com/MIT-LCP/physionet-build/pull/1556, we removed the "initial review" stage from the credentialing page of the console (http://localhost:8000/console/credential_processing/).

Before the change, visiting http://localhost:8000/console/credential_processing/ would display a list of applications at the "initial review" stage by default. 

We did not set a new default list of applications, so now the page is empty when loaded (and the user must select a tab to view).  This change displays the "ID check" list by default.


briangow:
Thanks @tompollard , looks good to me!

PR #1563, opened 2022-04-20
Display 'contact reference' form on correct page

tompollard:
In an earlier pull request (https://github.com/MIT-LCP/physionet-build/pull/1556), we removed the "initial review" stage of the credentialing workflow. 

When we made this change, we failed to move the "Contact Reference" form to the earlier stage of the credentialing workflow. 

As a result, the contact reference form is incorrectly displayed on the "Final review" page. This change moves the contact reference form to the correct stage of the credentialing workflow.

Screen Shot 2022-04-20 at 17 22 08

### DIFF
--- a/physionet-django/console/templates/console/process_credential_application.html
+++ b/physionet-django/console/templates/console/process_credential_application.html
@@ -28,7 +28,7 @@
   <div class="col-sm-5">
     <div class="row">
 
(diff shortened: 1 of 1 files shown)


briangow:
Looks good @tompollard , thanks!

PR #1564, opened 2022-04-21
Link to pending training and credentialing applications in the user console. Fix "skip reference check" bug

tompollard:
This pull request makes a couple of minor updates to the credentialing system:

1. Adds link to pending training and credentialing applications from the user console (e.g. see: http://localhost:8000/console/user/manage/george/)
2. Fixes a bug with the "skip reference check" button. Currently the button moves the application to "status=50". This is incorrect because the final review staging has a status of 40.


briangow:
@tompollard , this looks good!

PR #1568, opened 2022-05-03
Remove radio buttons from the ID check stage of the credentialing workflow

tompollard:
Currently credentialing involves a lot of button clicking! This is slow and tiresome for credentialing admin. An example of the current questions is shown below:

Screen Shot 2022-05-03 at 10 52 24

This change removes the radio buttons from the ID check stage of the credentialing workflow. The points are instead displayed in a "Guidelines" box. The guidelines should be improved later.

One of the nice things about moving the ID check points from the database to a free text section is that it makes the system more flexible (i.e. it is much easier to add/change the guidelines now that they are not tracked in the database).

An example of a new ID check page is shown below (this can be accessed at: http://localhost:8000/console/credential-applications/qMNskHKOcrsnpNHJwGKx/process/?).

Screen Shot 2022-05-03 at 10 55 58

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -464,37 +464,19 @@ class PersonalCredentialForm(forms.ModelForm):
     """
     Form to respond to a credential application in the ID check stage
     """
-
     decision = forms.ChoiceField(choices=REVIEW_RESPONSE_CHOICES,
                                  widget=forms.RadioSelect)
 
     class Meta:
         model = CredentialReview
-        fields = ('user_searchable', 'user_has_papers',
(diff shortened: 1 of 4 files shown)


briangow:
Thanks @tompollard , this looks good to me. We should probably create an issue about removing the unused fields from the database. Also, a future improvement could save a click by converting the remaining radio button into separate "Pass to next stage" and "Reject" buttons.

tompollard:
> We should probably create an issue to remove the unused fields from the database.

I'll do this as a final step.

> Also, a future improvement could save a click by converting the remaining radio button into separate "Pass to next stage" and "Reject" buttons.

Good point! @kepaik may be able to implement this tomorrow.

PR #1569, opened 2022-05-03
Remove radio buttons from the reference check stage of the credentialing workflow 

tompollard:
This pull request does the same as https://github.com/MIT-LCP/physionet-build/pull/1568 but for the "reference check" stage. The goal is to remove time-consuming button clicks.

This change removes the radio buttons from the reference check stage of the credentialing workflow. The points are instead displayed in a "Guidelines" box. The guidelines should be improved later.


tompollard:
(rebased after #1568 was merged)

briangow:
@tompollard , looks good, thanks!

PR #1570, opened 2022-05-03
Remove radio buttons from the reference response stage of the credentialing workflow

tompollard:
This pull request does the same as https://github.com/MIT-LCP/physionet-build/pull/1568 and https://github.com/MIT-LCP/physionet-build/pull/1569 but for the "reference response" stage. The goal is to remove time-consuming button clicks.

This change removes the radio buttons from the reference response stage of the credentialing workflow. The points are instead displayed in a "Guidelines" box. The guidelines should be improved later.

Side note, it feels like the credentialing workflow is missing a step when a response from a referee has been received. I may add this step in a later PR.

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -563,13 +563,9 @@ class ResponseCredentialForm(forms.ModelForm):
 
     class Meta:
         model = CredentialReview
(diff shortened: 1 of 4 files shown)


briangow:
@tompollard , looks good to me, thanks!

Issue #1571, opened 2022-05-03
Consider removing unused credentialing fields from user models 


tompollard:
In PRs https://github.com/MIT-LCP/physionet-build/pull/1570, https://github.com/MIT-LCP/physionet-build/pull/1569, and https://github.com/MIT-LCP/physionet-build/pull/1568 we removed several radio buttons from the credentialing workflow.

These radio buttons populated fields in the `CredentialReview` model in `user/models.py`. Now that the fields are no longer check during credentialing, we could consider removing the fields from the database.

By removing the fields we would be losing some information on past applications, but I don't think this is a problem. The fields that can be removed are indicated with a "Consider removing" comment, as shown below:

```python
class CredentialReview(models.Model):
    """
    Reviews for the CredentialApplications.

    NOTES
    -----
    This relational model will be deleted in the case that a credential
    reviewer decides to "reset" the application, meaning reset it back to the
    "Not in review" stage.

    """
    REVIEW_STATUS_LABELS = (
        ('', '-----------'),
        (0,  'Not in review'),
        (10, 'ID check'),
        (20, 'Reference'),
        (30, 'Reference response'),
        (40, 'Final review'),
    )

    application = models.OneToOneField('user.CredentialApplication',
                                       related_name='credential_review',
                                       on_delete=models.CASCADE)

    status = models.PositiveSmallIntegerField(default=10,
        choices=REVIEW_STATUS_LABELS)

    # Initial review questions
    # No longer checked. Consider removing these.
    fields_complete = models.NullBooleanField(null=True)
    appears_correct = models.NullBooleanField(null=True)
    lang_understandable = models.NullBooleanField(null=True)

    # ID check questions
    # No longer checked. Consider removing these.
    user_searchable = models.NullBooleanField(null=True)
    user_has_papers = models.NullBooleanField(null=True)
    research_summary_clear = models.NullBooleanField(null=True)
    course_name_provided = models.NullBooleanField(null=True)
    user_understands_privacy = models.NullBooleanField(null=True)
    user_org_known = models.NullBooleanField(null=True)
    user_details_consistent = models.NullBooleanField(null=True)

    # Reference check questions
    # No longer checked. Consider removing these.
    ref_appropriate = models.NullBooleanField(null=True)
    ref_searchable = models.NullBooleanField(null=True)
    ref_has_papers = models.NullBooleanField(null=True)
    ref_is_supervisor = models.NullBooleanField(null=True)
[...]

Issue #1095, opened 2020-07-06
Upgrading to Django 3


tompollard:
Django 3 was released at the end of 2019 (https://www.djangoproject.com/weblog/2019/dec/02/django-3-released/). 

> "With the release of Django 3.0, Django 2.2 has reached the end of mainstream support. The final minor bug fix release (which is also a security release), 2.2.8, was issued today. Django 2.2 is an LTS release and will receive security and data loss fixes until April 2022. All users are encouraged to upgrade before then to continue receiving fixes for security issues."

We should look into upgrading at some point this year.

kshalot:
We did a very small spike to see what the upgrade from 2.2 to 3.2 would break in physionet. Here are the findings:

#### General
- [x] `django_background_tasks` needs to be bumped. [Their documentation](https://django-background-tasks.readthedocs.io/en/latest/#supported-versions-and-compatibility) mentions 2.2 as the highest supported version. [The newer tags](https://github.com/arteria/django-background-tasks/commit/97b0a98823ef560837c6cab8e50d1530c2bcf572) in GH (published to pypi as well) mentioned added support for Django 3.0 though. Might be okay, but Django doesn't always stick to semantic versioning so 3.1/3.2 might have contained some breaking changes.
- [x] `django-debug-toolbar` needs to be bumped, `^2.0.0` should suffice
- [ ] 4 unit tests fail on `reverse` for urls that contain uidb64.
- [ ] Some easy deprecation warnings, e.g. `NullBooleanField` => `BooleanField(null=True)`

#### Minor
- [x] `from background_task.models_completed import CompletedTask` was removed in newer versions of `django_background_tasks` - change to `from background_task.models import CompletedTask`
`from django.contrib.staticfiles.templatetags.staticfiles import static` => `from django.templatetags.static import static`

tompollard:
Now running Django 3!

PR #1574, opened 2022-05-04
Replace `NullBooleanField` with `BooleanField(null=True)`

tompollard:
`NullBooleanField` is deprecated in version 3.0 of Django (ref https://github.com/MIT-LCP/physionet-build/issues/1095). See: https://code.djangoproject.com/ticket/31369. In preparation for this deprecation, `BooleanField` was updated to support a `null=True` argument. 

This pull request switches to the new syntax (i.e. from `NullBooleanField` with `BooleanField(null=True)`). Note that most (or all?) of the fields that are being updated could potentially be removed altogether (see: https://github.com/MIT-LCP/physionet-build/issues/1571).


tompollard:
(I didn't intend to include 4ff240aee607f550024c6691b2b7b70e1a5efec9 but it wouldn't hurt to merge that too!).

briangow:
@tompollard , in addition to the old migration files I see a couple of other files which still use `NullBooleanField`:

physionet-django/project/modelcomponents/submission.py
physionet-django/project/modelcomponents/generic.py

tompollard:
@briangow thanks for picking that up. looks like i broke something in the update...looking into it!

tompollard:
Not investigated, but I \*think\* the failing tests are a result of the (hacky) way some of the review response choices are coded:

https://github.com/MIT-LCP/physionet-build/blob/dceccf0e44fbb0b527800dc00130929b2b5bc6bd/physionet-django/console/forms.py#L48-L65

```
FAIL: test_edit (console.test_views.TestState)
Edit a project. Request resubmission, then accept.
  File "/__w/physionet-build/physionet-build/physionet-django/console/test_views.py", line 116, in test_edit
    self.assertTrue(project.author_editable())
AssertionError: None is not true

FAIL: test_edit_reject (console.test_views.TestState)
Edit a project, rejecting it.
  File "/__w/physionet-build/physionet-build/physionet-django/console/test_views.py", line 96, in test_edit_reject
    self.assertTrue(ArchivedProject.objects.filter(slug=project.slug))
AssertionError: <QuerySet []> is not true
```

tompollard:
I think this is now good to merge. The issue is that the `YES_NO_UNDETERMINED` choice requires the field to differentiate between `""` and `None`. For now, I've added the `blank=True` flag to the fields using these choices. 

It would be best to update the fields to properly capture the different choices using an enum. I would like to do this in a later step (and will raise an issue for the task if this PR is merged).

PR #1575, opened 2022-05-04
Ensure files are closed after use.

tompollard:
While running tests in preparation for upgrading to Django 3, I noticed that files are opened but not closed in a couple of places, raising a ResourceWarning. To display these warnings, run:

```python -Wa manage.py test```

(the deprecation warnings are silenced by default. -Wa enables them).

Example warning:

```
...../projects/physionet-build/physionet-django/project/modelcomponents/activeproject.py:60: ResourceWarning: unclosed file <_io.BufferedReader name='...../projects/physionet-build/physionet-django/static/test/published-projects/mitbih/1.0.0/RECORDS'>
  with open(os.path.join(root, f), 'rb').read(2) as fline:
```

This pull request uses the `with open(path) as f:` syntax to fix the issue in two places:
1. In the `move_files_as_readonly` function.
2. In `project/test_views.py`

### DIFF
--- a/physionet-django/project/modelcomponents/activeproject.py
+++ b/physionet-django/project/modelcomponents/activeproject.py
@@ -59,10 +59,10 @@ def move_files_as_readonly(pid, dir_from, dir_to, make_zip):
             for f in files:
                 with open(os.path.join(root, f), 'rb') as file:
                     fline = file.read(2)
-                    if fline[:2] == b'#!':
-                        os.chmod(os.path.join(root, f), 0o555)
+                    if fline == b'#!':
+                        os.chmod(file.fileno(), 0o555)
                     else:
-                        os.chmod(os.path.join(root, f), 0o444)
+                        os.chmod(file.fileno(), 0o444)
 
             for d in dirs:
                 os.chmod(os.path.join(root, d), 0o555)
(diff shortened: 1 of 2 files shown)


bemoody:
Oof, yeah, that could have been messy.  In move_files_as_readonly, the
files are presumably being GCed immediately, but if they weren't then that
could DOS itself with file descriptors.

This could perhaps be changed to:

                with open(os.path.join(root, f), 'rb') as file:
                    fline = file.read(2)
                    if fline[:2] == b'#!':
                        os.chmod(file.fileno(), 0o555)
                    else:
                        os.chmod(file.fileno(), 0o444)

Probably doesn't make much difference, but avoids some minor redundancy.

bemoody:
Actually, the "[:2]" is unnecessary too.

tompollard:
Thanks @bemoody, I made the improvements that you suggested.

bemoody:
Works great, thanks!                                                    

After pushing to production: `sudo systemctl restart django-background-tasks.service`

tompollard:
Thanks!

> After pushing to production: sudo systemctl restart django-background-tasks.service

Got it, will do this.

PR #1577, opened 2022-05-04
force_text() is deprecated in Django 3. replace with force_str()

tompollard:
In Django 3, force_text() has been deprecated in favor of force_str(). force_text() is an alias of force_str(). https://django.fun/docs/django/en/4.0/releases/3.0/#django-utils-encoding-force-text-and-smart-text

This pull request changes all instances of force_text() to force_str()


briangow:
@tompollard , looks good to me, thanks!

PR #1582, opened 2022-05-07
Bump django-ckeditor package to 5.9.0. ref #1095

tompollard:
The django-ckeditor package provides the rich text editor that we use on various forms (e.g. the forms used by data contributors to generate the project description. This pull request bumps the package to version V5.9.0 (which added support for Django 3). 

See: https://pypi.org/project/django-ckeditor/#id6 for the django-ckeditor changelog. Note to self: these are the upgrade commands that I'm running:

```bash
pip install django-ckeditor==5.9.0
hashin django-ckeditor==5.9.0
poetry add django-ckeditor=^5.9.0
```

PR #1583, opened 2022-05-07
Bump django-js-asset to 1.2.2 in requirements.txt to match poetry. Ref #1582

tompollard:
The version of django-js-asset is not matched between requirements.txt and poetry. This pull request bumps django-js-asset in requirements.txt to v1.2.2 to match poetry.

This change is needed for https://github.com/MIT-LCP/physionet-build/pull/1582 to [pass tests](https://github.com/MIT-LCP/physionet-build/runs/6336093254?check_suite_focus=true):

```
ERROR: Cannot install -r requirements.txt (line 14) and django-js-asset==0.1.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested django-js-asset==0.1.1
    django-ckeditor 5.9.0 depends on django-js-asset>=1.2.2

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -186,9 +186,9 @@ pyparsing==2.4.7 \
 sqlparse==0.4.2 \
     --hash=sha256:0c00730c74263a94e5a9919ade150dfc3b19c574389985446148402998287dae \
     --hash=sha256:48719e356bb8b42991bdbb1e8b83223757b93789c00910a616a071910ca4a64d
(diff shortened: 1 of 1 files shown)


briangow:
@tompollard , this looks good to me but I'm not able to test it.

PR #1584, opened 2022-05-10
Replace dropdown menus in credentialing review process with accept/reject buttons

tompollard:
We are in the process of implementing several changes to the credentialing workflow with the goal of improving usability and reducing the number of button clicks (e.g. https://github.com/MIT-LCP/physionet-build/pull/1570). 

The workflow can be found at: http://127.0.0.1:8000/console/credential_processing/. This change replaces the dropdown accept/reject menus with separate accept and reject buttons.

e.g. http://127.0.0.1:8000/console/credential-applications/5U5sjoUd3b0QmHMGZm5C/process/? displays the following buttons: 

Screen Shot 2022-05-10 at 17 18 22

A few notes on the changes:
- Please forgive the (auto applied) formatting changes in [physionet-django/console/views.py](https://github.com/MIT-LCP/physionet-build/compare/tp/cred_radio_1?expand=1#diff-f6a42ec9e0dd1026ae0ec1016b6549d4bc69526a31f5dcfb9ea14eff72dcf9d7). I'd really prefer to avoid having to spend time cleaning this up. If the changes result in conflicts with existing PRs, I will deal with those conflicts.
- I have not updated the buttons on the "Contact Reference" stage of the workflow. I plan to modify this stage in a later pull request so have left it unchanged for now.
- The use of `intermediate_credential_form` is pretty confusing and it can also be removed after I have implemented the change discussed above.

@kepaik this is the change we discussed. Please could you do some testing on your local machine? Let me know if you need help setting this up.

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -784,6 +784,17 @@ class TrainingQuestionFormSet(forms.BaseModelFormSet):
                 )
 
 
+class CredentialReviewForm(forms.Form):
+    reviewer_comments = forms.CharField(widget=forms.Textarea(attrs={'rows': 5}), required=False)
+
+    def clean(self):
+        if self.errors:
+            return
+
+        if not self.cleaned_data['reviewer_comments']:
+            raise forms.ValidationError('If you reject, you must explain why.')
+
+
 class TrainingReviewForm(forms.Form):
     reviewer_comments = forms.CharField(widget=forms.Textarea(attrs={'rows': 5}), required=False)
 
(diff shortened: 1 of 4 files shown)


kepaik:
1) Small, but significant problem: Once you click a button to process, it boots you to the root credentialing page, instead of the next tab for the same application.

2) Might help for clarity to rename "Accept" to "Accept ID", and similar on subsequent tabs.

3) It seems like the "reviewer comment" box on the "Final Review" tab does not get passed to the user in the auto response. It might be helpful to include that like we do for rejections to supplement with clarity and reinforcing messages about the DUA for borderline cases.

Haven't had a chance to test the reference process.

tompollard:
@kepaik thanks for the quick review. 

> Small, but significant problem: Once you click a button to process, it boots you to the root credentialing page, instead of the next tab for the same application.

I'd thought this change was desirable so it was a new feature! I've switched back to the old behaviour.

> Might help for clarity to rename "Accept" to "Accept ID", and similar on subsequent tabs.

Done.

> It seems like the "reviewer comment" box on the "Final Review" tab does not get passed to the user in the auto response. It might be helpful to include that like we do for rejections to supplement with clarity and reinforcing messages about the DUA for borderline cases.

I haven't touched this aspect of the form, so what you see is the current behaviour. The comments functionality definitely needs (a lot of) improvement, but I would like to deal with this in a future pull request.

PR #1587, opened 2022-05-12
Credential email

lamawmouk:
I wanted to make changes to the code base. This pull request makes two changes to be used on the credentialing pages:

- changing the "Edit Emails" page to prompt users to use an institutional/academic if applicable in the Primary Email section.
- changing the "Application for Credentialed Access" section to prompt users to add an institutional/academic email.

### DIFF
--- a/physionet-django/user/templates/user/credential_application.html
+++ b/physionet-django/user/templates/user/credential_application.html
@@ -16,7 +16,7 @@
 
   <p>Please use the form below to apply for {{ SITE_NAME }} credentialing. In order to apply:</p>
   <ul>
-    <li>If you have an institutional/academic please add it in your settings (<a href="{% url 'edit_emails' %}">Emails settings</a>).</li>
+    <li>If you have an institutional/academic email address please add it in your settings (<a href="{% url 'edit_emails' %}">Emails settings</a>).</li>
     <li>If you have an ORCID iD please link to it in your settings (<a href="{% url 'edit_orcid' %}">ORCID settings</a>) as this may help us expedite your application by making it easier to verify your identity.</li>
   </ul>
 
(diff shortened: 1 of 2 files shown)


physionet-django/user/templates/user/credential_application.html
```diff
@@ -16,6 +16,7 @@ <h1>Application for Credentialed Access</h1>
 
   <p>Please use the form below to apply for {{ SITE_NAME }} credentialing. In order to apply:</p>
   <ul>
+    <li>If you have an institutional/academic please add it in your settings (<a href="{% url 'edit_emails' %}">Emails settings</a>).</li>
```
tompollard: Looks like a word is missing here? "If you have an institutional/academic XXX please add it...".

physionet-django/user/templates/user/edit_emails.html
```diff
@@ -62,7 +62,7 @@ <h5 class="modal-title">Remove Email</h5>
 
 <br>
 <h4>Primary Email</h4>
-<p>The email address used for authentication, and email notifications.</p>
+<p>Please use an institutional/academic email address. This will be used for authentication, and email notifications.</p>
```
tompollard: We should probably acknowledge that not everyone will have an institutional/academic email address. We could say something like: 

"The email address used for authentication, and email notifications. We recommend using an institutional/academic email address where possible."

tompollard (commented):
Thanks Lama, it's a good idea to update this text. I've added a couple of suggestions.

physionet-django/user/templates/user/credential_application.html
lamawmouk: thanks @tompollard, added the email address to the text.

physionet-django/user/templates/user/edit_emails.html
lamawmouk: sounds good, added the sentence as part of the edit emails section.

tompollard:
Looks good to me, thanks Lama!

PR #1592, opened 2022-05-20
Swap Ethics and Release Notes

bemoody:
This is a stylistic change to swap the order of the "Ethics" and "Release Notes" sections in published projects.

My reasoning for doing this is:

1. "Usage Notes" (or "Evaluation") and "Release Notes" are both what I would consider technically heavy sections.  "Release Notes", if present, should provide a detailed technical explanation of what has changed, which pertains directly to "Usage Notes", and generally *doesn't* pertain to "Ethics" (or to "Acknowledgements" or "Conflicts of Interest").

2. "Ethics" is typically a very brief statement that looks strange when placed between two (potentially) long and detailed sections.

I like putting "Ethics" above "Acknowledgements" rather than below but that's mostly a gut feeling and I could be persuaded otherwise.

I've reviewed the projects that have been published since the "Ethics" section was added (from March 10 to May 16), and I don't think that reordering these sections has a substantial impact on the flow of the document.  I haven't reviewed the projects currently in the pipeline.

### DIFF
--- a/physionet-django/project/templates/project/challenge_content.html
+++ b/physionet-django/project/templates/project/challenge_content.html
@@ -18,15 +18,15 @@
 {{ project.usage_notes|safe }}
 <hr>
 
-{% if project.ethics_statement %}
-  <h2 id="ethics">Ethics</h2>
-    {{ project.ethics_statement|safe }}
-  <hr>
-{% endif %}
-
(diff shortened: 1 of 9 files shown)


tompollard:
Thanks, looks good to me.

PR #1593, opened 2022-05-25
Remove unneeded packages from install-pn-test-server

bemoody:
rustc and libssl-dev were needed to install the cryptography package on Debian 10.  They are not needed now, so remove them.


tompollard:
Looks good to me, thanks Benjamin

PR #1594, opened 2022-05-26
Create date_added field for the `Orcid` user model

briangow:
This adds a derived `date_added` field to the `Orcid` model in the user app. The `date_added` is derived from the `token_expiration` for consistency with the objects that are already in the database. The token expires after 20 years so it can be used to calculate the date the object was created, taken as 20 years prior to expiration. 

Even though it would be more accurate, I did not use `relativedelta` from the `python-dateutil` package to generate the 20 year offset because it is not already installed in our docker environment (but it is in requirements.txt) and doesn't seem necessary for this application. Instead I used `timedelta(days=20*365.2422)` which isn't aware of when it is a leap year but should be sufficient for the `date_added` field. 

I have tested this PR on my local machine but not on staging. I have done the following locally:

- confirmed that old `Orcid` records in the database which don't have a `date_added` field are properly populated with `date_added` once the migration is run
- confirmed that new `Orcid` records generated by connecting to the ORCID API continue to work as expected and generate the correct `date_added` field, as derived from the `token_expiration`. 
- confirmed that I can roll back to a prior migration (i.e. from 0044 to 0043). In `migrate_backward` I simply use `pass` as @tompollard didn't think it was necessary to do a `.delete` of the `date_added` field in this case.

### DIFF
--- a/physionet-django/user/models.py
+++ b/physionet-django/user/models.py
@@ -2,6 +2,7 @@ import logging
 import os
 import uuid
 from datetime import timedelta
+from datetime import datetime
 
 from django.conf import settings
 from django.contrib import messages
@@ -657,6 +658,7 @@ class Orcid(models.Model):
     token_type = models.CharField(max_length=50, default='', blank=True)
     token_scope = models.CharField(max_length=50, default='', blank=True)
     token_expiration = models.DecimalField(max_digits=50, decimal_places=40, default=0)
+    datetime_added = models.DateTimeField(auto_now_add=True)
 
     class Meta:
         default_permissions = ()
--- a/physionet-django/user/fixtures/demo-user.json
+++ b/physionet-django/user/fixtures/demo-user.json
@@ -9772,7 +9772,8 @@
     "refresh_token": "",
     "token_type": "",
     "token_scope": "",
-    "token_expiration": "0E-40"
+    "token_expiration": "0E-40",
+    "datetime_added": "2021-04-10T15:51:15.446Z"
   }
 },
 {
--- a/physionet-django/user/migrations/0044_orcid_datetime_added.py
+++ b/physionet-django/user/migrations/0044_orcid_datetime_added.py
@@ -1,0 +1,31 @@
+from django.db import migrations, models
+import django.utils.timezone
+from datetime import datetime, timedelta
+
+
+def migrate_forward(apps, schema_editor):
+    orcid_model = apps.get_model('user', 'Orcid')
+    for row in orcid_model.objects.all():
+        row.datetime_added = datetime.fromtimestamp(row.token_expiration) - timedelta(days=20*365.2422)
+        row.save()
+
+
+def migrate_backward(apps, schema_editor):
+    pass
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('user', '0043_auto_20220406_1229'),
+    ]
+
+    operations = [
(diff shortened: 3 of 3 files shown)


bemoody:
> Even though it would be more accurate, I did not use `relativedelta`
> from the `python-dateutil` package to generate the 20 year offset
> because it is not already installed in our docker environment (but it
> is in requirements.txt) and doesn&#39;t seem necessary for this
> application. Instead I used `timedelta(days=20*365.2422)` which
> isn&#39;t aware of when it is a leap year but should be sufficient
> for the `date_added` field. 

You know better than I do, but I'd guess that the expiration is
actually calculated as some fixed number of seconds in the future (so
what you're doing is perfectly correct.)  If it really is calculated in
terms of calendar years then you could do:

t = datetime.datetime.fromtimestamp(...)
t = t.replace(year=(t.year - 20))

but, y'know, as long as it more-or-less works for the existing data,
don't worry about it.

I think it might be wiser, though, to use a DateTimeField instead of a
DateField.  Even if the time isn't exactly right for historical data.

physionet-django/user/models.py
```diff
@@ -657,6 +658,7 @@ class Orcid(models.Model):
     token_type = models.CharField(max_length=50, default='', blank=True)
     token_scope = models.CharField(max_length=50, default='', blank=True)
     token_expiration = models.DecimalField(max_digits=50, decimal_places=40, default=0)
+    date_added = models.DateField(default=datetime(2200, 1, 1))
```
tompollard: In a previous version, I think you had something like `models.DateField(auto_now_add=True)`. Thought that seemed like a good idea. What was the reason to change?

physionet-django/user/views.py
```diff
@@ -448,6 +448,7 @@ def auth_orcid(request):
         orcid_profile.token_type = token.get('token_type')
         orcid_profile.token_scope = token.get('scope')
         orcid_profile.token_expiration = token.get('expires_at')
+        orcid_profile.date_added = datetime.fromtimestamp(token.get('expires_at')) - timedelta(days=20*365.2422)
```
tompollard: We used the expiry date to guess the `date_added` when it wasn't known. As we are now capturing the date directly, can't this just be date_added = now? I think if you use `auto_now_add=True` in the model, it may not be necessary to modify the view at all.

physionet-django/user/models.py
briangow: @tompollard , thanks for the comments. I decided to implement it like I did for consistency between the old objects which didn't have a `date_added` field. I choose this default to make it clear that it isn't a real date in the event that it doesn't get overwritten by the view as expected.

physionet-django/user/views.py
briangow: Please see my previous comment which also addresses this. I'm open to changing this if desired though.

physionet-django/user/models.py
briangow: @bemoody , I think we are going to change this back to what I had previously: 
`date_added = models.DateField(auto_now_add=True)`
This means that new objects will be created based on the current timestamp whereas old objects will be derived from the token_expiration.

briangow:
> I think it might be wiser, though, to use a DateTimeField instead of a
DateField.

Thanks @bemoody , I updated the field to a `DateTimeField`.

tompollard:
thanks brian, looks good to me!

Issue #1467, opened 2022-02-01
Consider adding "date_added" field to the Orcid model


tompollard:
There is some benefit in knowing when ORCIDs were added to user profiles (for example, we are currently interested in exploring what proportion of new users add ORCIDs to their profiles, and how this is changing over time).

The `user.Orcid` model does not currently capture the date when the ORCID was added. There is however a `token_expiration` date which appears to be added date + 40 [update: 20] years in unix timestamps. We may want to add a "date_added" field to the Orcid model and perhaps populate this with `token_expiration` - 40 [update: 20] years.

The following chunk generates a list of ORCIDs added over time from the `token_expiration` dates.

```python
import pandas as pd

df = pd.read_csv("orcid_dates.csv")

def convert_time(ts, string=False):
    if string:
        return datetime.utcfromtimestamp(ts).strftime('%Y-%m-%d %H:%M:%S')
    else:
        return datetime.utcfromtimestamp(ts)

df['date'] = df['unix_date'].apply(convert_time)
# df['date_fixed'] = df['date'] - pd.Timedelta(days=40*365)
# fixed offset below
df['date_fixed'] = df['date'] - pd.Timedelta(days=20*365) 

period = df.date_fixed.dt.to_period("M")
grouped = df.groupby(period)
grouped.count()
```

```
#             unix_date
# 2001-02            11
# 2001-03           129
# 2001-04           172
# 2001-05           165
# 2001-06           169
# 2001-07           178
# 2001-08           185
# 2001-09           186
# 2001-10           222
# 2001-11           188
# 2001-12           181
# 2002-01           195
# 2002-02            93

```

[update: @briangow pointed out that the offset is 20 years, not 40!]

tompollard:
The field was added in https://github.com/MIT-LCP/physionet-build/pull/1594

PR #1598, opened 2022-05-29
Fix: Record the time when Orcids are linked

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/1467, we would like to record the time when users add an Orcid to their profile.

https://github.com/MIT-LCP/physionet-build/pull/1594 added a new field to capture the datettime. This approach worked fine on dev and staging servers, but errored on production.

```
remote:     value = timezone.make_aware(value, default_timezone)
remote:   File "/home/physionet/python-env/physionet/lib/python3.9/site-packages/django/utils/timezone.py", line 270, in make_aware
remote:     return timezone.localize(value, is_dst=is_dst)
remote:   File "/home/physionet/python-env/physionet/lib/python3.9/site-packages/pytz/tzinfo.py", line 327, in localize
remote:     raise NonExistentTimeError(dt)
remote: pytz.exceptions.NonExistentTimeError: 2021-03-14 02:37:39.400000
```

The fix for this error is to make sure that datetimes are accompanied by a timezone. The easiest way to manage this is to use Django's timeaware function.

### DIFF
--- a/physionet-django/user/fixtures/demo-user.json
+++ b/physionet-django/user/fixtures/demo-user.json
@@ -9772,7 +9772,8 @@
     "refresh_token": "",
     "token_type": "",
     "token_scope": "",
-    "token_expiration": "0E-40"
+    "token_expiration": "0E-40",
+    "datetime_added": "2021-04-10T15:51:15.446Z"
   }
 },
 {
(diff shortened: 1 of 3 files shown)


tompollard:
Sorry for the thrash. The goal was to clean up a failed migration in https://github.com/MIT-LCP/physionet-build/pull/1594, but I ended up making things more complicated than they needed to be.

Issue #1557, opened 2022-04-11
Reference reordering triggers a 500 error


alistairewj:
It's possible to get a 500 Server Error by moving a reference in the list of references on the content page. Here's how I reproduced:

1. Had a project with 7 references.
2. Add a new blank reference.
3. Move the text of reference 7 to one of the other spots (I moved it to spot 1).
4. Remove the (blank) 8th reference.
5. Click save content

This consistently resulted in a 500 error (and I did it thrice on production just now, ~4:54pm). It looks like a duplicate key error:

> IntegrityError at /projects/yk9LlDASyilE4qnZ7Il8/content/
> duplicate key value violates unique constraint "project_reference_description_content_type_8b53e7b5_uniq"
DETAIL:  Key (description, content_type_id, object_id)=(Makam AN, Lanham HJ, Batchelor K, Samal L, Moran B, Howell-Stampley T, Kirk L, Cherukuri M, Santini N, Leykum LK, Halm EA. Use and satisfaction with key functions of a common commercial electronic health record: a survey of primary care providers. BMC medical informatics and decision making. 2013 Dec;13(1):1-7., 11, 881) already exists.

I don't think it's super unusual to re-order references so this is probably more common of an issue than it sounds.

tompollard:
yeah, this is a major bug that has been sitting around for too long! it's the same issue described in https://github.com/MIT-LCP/physionet-build/issues/1105. basically the reference object doesn't include an `order` attribute, and it should do!

tompollard:
@amitupreti iif you have time (and Alistair is okay with this) please could you take a look at this issue to see if you can come up with a fix?

I made a start on the initial part of the work, which is migrating from using primary key for sorting to an explicit order: https://github.com/MIT-LCP/physionet-build/pull/1602

Now we need to use the order in the reference tool. @Rutvikrj26 made a start, so could point you in the right direction.

superryeti:
@tompollard  Sounds great. Thank you. 

let me add this to my task list.

tompollard:
This should now be fixed (thanks @amitupreti!). References now have an `order` and this order is used to sort references.

Issue #1105, opened 2020-08-01
Unable to change references during copy-edit stage - reference deletion error and server 500 error


li-lcp:
During copy edit stage, the following errors occurred:

(1) deleting 1 reference by pressing on the "x" sign, surprisingly caused two references to be deleted.   (The particular sequence of actions that led to this error: adding 3 references at the bottom, and then attempting to delete the first reference on top, caused two references to disappear.)
(2) Changing/swapping the content of the references, and then press "Save Content" apparently lead to Server 500 errors persistently.

(My original intention was simply to  change the order of the references to match the order in which they appear in the main text.)

tompollard:
Thanks for raising this @li-lcp. I have added the traceback at the end of this post.

We need to look into point 1. Not sure what happened with that. I have added a note on point 2 below.

> (2) Changing/swapping the content of the references, and then press "Save Content" apparently lead to Server 500 errors persistently.

It looks like your changes are raising a `duplicate key` error, and I think this is an issue with the way the system is setup. I've not explored, but I think the problem is:

1) the database contains a list of references. each reference description must be unique within the project.
2) you are viewing the references in the form, then editing them (e.g. reordering).
3) when you click "save/submit" on the form, the system attempts to save the changes in the database
4) an integrity error is raised because the references that you are saving already exist in the database.

For now, you should be able to bypass the issue by making sure that new references are not an identical match to references that already in the system (e.g. add some temporary text when you save, then remove this temporary text in a later save step if necessary). In the longer term, the system for managing references should be improved.

```
Internal Server Error: /console/submitted-projects/M1X2WSSvkIOzXJd78Oho/copyedit/

IntegrityError at /console/submitted-projects/M1X2WSSvkIOzXJd78Oho/copyedit/
duplicate key value violates unique constraint "project_reference_description_content_type_8b53e7b5_uniq"
DETAIL:  Key (description, content_type_id, object_id)=(Epilepsy Foundations, Last accessed on July 17th, 2020. https://www.epilepsy.com/, 11, 51) already exists.


Request Method: POST
Request URL: https://www.physionet.org/console/submitted-projects/M1X2WSSvkIOzXJd78Oho/copyedit/

Traceback:

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/db/backends/utils.py" in _execute
 84.                 return self.cursor.execute(sql, params)

[...]

PR #1602, opened 2022-06-02
[WIP] Add order field to reference objects

tompollard:
# Background

Project descriptions include a "References" section that lists relevant papers, websites, etc. An example of a references section for a demo project is shown below (screenshot from http://localhost:8000/content/demoecg/10.5.24/):

Screen Shot 2022-06-02 at 14 04 47

# Problem

As discussed in https://github.com/MIT-LCP/physionet-build/issues/1557 and https://github.com/MIT-LCP/physionet-build/issues/1105, the way we manage these references is extremely buggy.  The problem is that there is no "order" field associated with the references. Instead the references are ordered by their primary key!

For both `ActiveProject` and `ArchivedProject`, the references are stored using the `Reference` class:

https://github.com/MIT-LCP/physionet-build/blob/1dc09c6ce67ad6d77c76a1ebe16a2dcdee65c883/physionet-django/project/modelcomponents/metadata.py#L575-L587

For `PublishedProject`, the references are copied to the `PublishedReference` class:

https://github.com/MIT-LCP/physionet-build/blob/1dc09c6ce67ad6d77c76a1ebe16a2dcdee65c883/physionet-django/project/modelcomponents/metadata.py#L601-L610

# Fixing the problem

This pull request:
- adds an `order` attribute to the `Reference` and `PublishedReference` classes.
- adds a migration that (a) adds the order field (2) migrates the order of existing references to the order field
- updates the view/templates to use the order field to display references.

What's needed before merging:
- We need to update the forms displayed in the following template so that newly created projects record `order` (instead of the ID): https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/templates/project/project_content.html

The current formset is:

https://github.com/MIT-LCP/physionet-build/blob/1dc09c6ce67ad6d77c76a1ebe16a2dcdee65c883/physionet-django/project/forms.py#L658-L691

The formset is rendered by a generic "item_list" template:
https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/templates/project/item_list.html

### DIFF
--- a/physionet-django/project/modelcomponents/metadata.py
+++ b/physionet-django/project/modelcomponents/metadata.py
@@ -603,14 +603,16 @@ class Reference(models.Model):
 
 class PublishedReference(models.Model):
     """
+    Reference field for PublishedProject
     """
     description = models.CharField(max_length=1000)
     project = models.ForeignKey('project.PublishedProject',
         related_name='references', on_delete=models.CASCADE)
+    order = models.PositiveIntegerField(null=True)
 
     class Meta:
         default_permissions = ()
-        unique_together = (('description', 'project'))
+        unique_together = (('description', 'project', 'order'))
 
 
 class Contact(models.Model):
--- a/physionet-django/project/modelcomponents/metadata.py
+++ b/physionet-django/project/modelcomponents/metadata.py
@@ -581,12 +581,13 @@ class Reference(models.Model):
     content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
     object_id = models.PositiveIntegerField()
     project = GenericForeignKey('content_type', 'object_id')
+    order = models.PositiveIntegerField(null=True)
 
     description = models.CharField(max_length=1000)
 
     class Meta:
         default_permissions = ()
-        unique_together = (('description', 'content_type', 'object_id'),)
+        unique_together = (('description', 'content_type', 'order'),)
 
     def __str__(self):
         return self.description
--- a/physionet-django/project/migrations/0066_migrate_references_to_use_order.py
+++ b/physionet-django/project/migrations/0066_migrate_references_to_use_order.py
@@ -1,0 +1,63 @@
+from django.db import migrations, models
+from project.models import PublishedProject, ActiveProject, ArchivedProject, LegacyProject
+
+
+def migrate_forward(apps, schema_editor):
+
+    for project in PublishedProject.objects.all():
+        refs = project.references.all().order_by('id')
+        for n, ref in enumerate(refs):
+            ref.order = n + 1
+            ref.save()
+
+    for project in ActiveProject.objects.all():
+        refs = project.references.all().order_by('id')
+        for n, ref in enumerate(refs):
+            ref.order = n + 1
+            ref.save()
+
+    for project in ArchivedProject.objects.all():
+        refs = project.references.all().order_by('id')
+        for n, ref in enumerate(refs):
+            ref.order = n + 1
+            ref.save()
+
+    for project in LegacyProject.objects.all():
+        refs = project.references.all().order_by('id')
+        for n, ref in enumerate(refs):
+            ref.order = n + 1
+            ref.save()
+
+
+def migrate_backward(apps, schema_editor):
+    pass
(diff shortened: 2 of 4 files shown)


tompollard:
@alistairewj is someone on your side able to help with the following point?:

> We need to update the form displayed in the following template so that newly created projects record "`order` (instead of the ID): https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/templates/project/project_content.html

physionet-django/project/modelcomponents/metadata.py
```diff
@@ -581,12 +581,13 @@ class Reference(models.Model):
...
     object_id = models.PositiveIntegerField()
     project = GenericForeignKey('content_type', 'object_id')
+    order = models.PositiveIntegerField(null=True)
 
     description = models.CharField(max_length=1000)
 
     class Meta:
         default_permissions = ()
-        unique_together = (('description', 'content_type', 'object_id'),)
+        unique_together = (('description', 'content_type', 'order'),)
```
superryeti, line 590: @tompollard  I am not sure if the unique is currently setup in the best possible way. Our current unique setup will cause false duplicate error in case where the reference and order of reference is same in two different projects(of same project type).

For example

Project A ( contentType=database)

| Order | reference | uniqueKey      |
|-------|-----------|----------------|
| 1     | 'a'       | (27, 'a', 1)   |
| 2     | 'b'       | (27, 'b', 2)   |
| 3     | 'c'       | (27, 'c', 3)   |
| 4     | 'd'       | (27, 'd', 4)   |


Project B ( contentType=database)

| Order | reference | uniqueKey      |
|-------|-----------|----------------|
| 1     | 'aa'       | (27, 'aa', 1)  |
| 2     | 'b'       | (27, 'b', 2)   |
| 3     | 'cc'      | (27, 'cc', 3)  |
| 4     | 'dd'      | (27, 'dd', 4)  |


as you can see we will have a conflict on the 2nd reference.


I think project might be better instead of content_type.

superryeti: maybe not `project` exactly, as the generic foreign key throws error when trying to use in `unique_together`

superryeti: not sure if adding object_id to unique_together is a good idea as that would most likely allow is to insert duplicates

for example
| Order | reference | uniqueKey                | content_id |
|-------|-----------|--------------------------|------------|
| 1     | 'a'       | (27, 'a', 1,11)    | 11   |
| 2     | 'b'       | (27, 'b', 2,12)    | 12   |
| 3     | 'c'       | (27, 'c', 3,13)    | 13   |
| 4     | 'd'       | (27, 'd', 4,14)    | 14   |
| 5     | 'a'       | (27, 'a', 5,15)    | 15   |


Although the front end will not send the duplicate reference, but updating the unique_together this way would make zero sense, as it would allow duplicate in database

tompollard: > Our current unique setup will cause false duplicate error in case where the reference and order of reference is same in two different projects(of same project type).

Yes, good catch, you're right! This doesn't make sense. 

> I think project might be better instead of content_type ... maybe not project exactly, as the generic foreign key throws error when trying to use in unique_together

I agree, `content_type` doesn't add anything useful here. I also agree that project would be a good alternative (and it's a shame that generic foreign keys make this difficult).

> not sure if adding object_id to unique_together is a good idea as that would most likely allow is to insert duplicates

Unless I'm missing something, I think it is fine to switch `object_id` with `content_type`. It is true that we may end up with duplicate reference strings in the database (slightly inefficient in terms of storage) but I don't think this is too much to worry about.

In summary, I think you should be good to change `content_type` to `object_id`.

superryeti: Alright, i think that makes sense `object_id` instead of `content_type` should work. The object ID is a integer that increments automatically. So it should help us prevent duplicate across different project.(and the order should help fix the reference error that we were trying to fix)

I will do more tests, and submit a PR in the monring

Thanks @tompollard

PR #1603, opened 2022-06-06
Credential email

lamawmouk:
I wanted to make some changes to the code base. This pull request makes two changes on the credentialing application:

- It shows a pop up with an alert message, if a non-institutional email address is added as a reference email on the "credentialing application" page by the user.
- It shows a pop up with an alert message, if the user inputs a non-institutional email address as the primary address on the "Edit emails" page.

### DIFF
--- a/physionet-django/user/templates/user/edit_emails.html
+++ b/physionet-django/user/templates/user/edit_emails.html
@@ -9,6 +9,7 @@
 <h1 class="form-signin-heading">Edit Emails</h1>
 
 <hr>
+
 <div class="card">
   <ul class="list-group list-group-flush">
     {% for email in associated_emails %}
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks @lamawmouk! Please could you remove the check from the settings page as discussed? We would like to give people freedom about which email address they use as their primary contact address.

In future work it would be great to run automated checks on the credentialing form in a similar way to the project submission, e.g.: https://github.com/MIT-LCP/physionet-build/blob/d5169d6f42d988545cb814c8861c7c7d0e66af09/physionet-django/project/templates/project/project_submission.html#L15-L49)

You could imagine a similar pop-up to the one below that instead prompts the user to add things like:

- an institutional email address,  
- an ORCID
- a reference
- etc

...along with buttons to "continue anyway" or "edit details".

Screen Shot 2022-06-21 at 17 32 54

cx1111:
I suggest breaking this down into smaller steps. We can add the ORCID stuff later.

This PR should check the provided reference email, and if it is detected to be one of those in the semi-banned list, shows a modal with the error message that allows them to go back, or submit the form anyway.

Steps:
1. Add HTML of the confirmation modal.
2. Create a Javascript function to be used as the onSubmit of the form's Submit button. https://www.w3schools.com/TAGS/ev_onsubmit.asp
   - First, call preventDefault to stop the form from being submitted. https://www.w3schools.com/jsref/event_preventdefault.asp
   - Validate the form field for the reference email.
   - If it is not a semi-banned email, then programmatically submit the form.
   - If it is a semi-banned email, show them a confirm diaglog (https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm) that warns them that the reference email is bad and provides a submit or cancel button.
      - If they click cancel, just close the modal.
      - If they click submit, then programmatically submit the form.
3. Register that function to the original submit button.

tompollard:
thanks!

PR #1604, opened 2022-06-08
Feature/theming

Rutvikrj26:
I have changed the theme.scss file and changed the base theme color to match the current Physionet Build. 

**This pull request should ideally not have any noticeable changes on the website** and is only supposed to replace the Bootstrap CSS files with the Scss setup.

The libraries required: Django-libsass & Django-compressor are added in the pyproject.toml file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -142,3 +142,6 @@ physionet-django/*credentials*
 
 # Pycharm / JetBrains / IntelliJ project specific settings folder
 .idea/
+
+# Cache CSS Files
+physionet-django/compressedfiles/CACHE
--- a/physionet-django/console/static/console/css/console.css
+++ b/physionet-django/console/static/console/css/console.css
@@ -255,10 +255,6 @@ body.fixed-nav {
   }
 }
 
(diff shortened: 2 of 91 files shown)


tompollard:
Thanks for looking at this @Rutvikrj26. I haven't had a chance to look at this properly, but I have a few quick questions before reviewing later:

The following files look like images for the Toronto platform. Should they be in this pull request?
- physionet-django/static/images/temerty-medicine.svg
- physionet-django/static/images/white-tcairem.png

Are the following private files needed?
- .uwsgi_plugins_builder/uwsgi.h
- .uwsgi_plugins_builder/uwsgiconfig.py

The cache folder also feels like it could be removed?
- physionet-django/CACHE/

Rutvikrj26:
Hello Tom,

During the initial work on the theme, I was not mindful of having the same deployment codebase as Physionet and added UofT banners, which were later removed by @alistairewj. They are not required anymore. 

The private files are for my visual studio setup and are not required.

The cache folder is generated by Django-libsass during deployment and is not required. 

Thank you for pointing this out, I will make sure the gitignore is updated and remove the extra files.
I'll push these changes right away.

PR #1619, opened 2022-06-24
Feature/theming

Rutvikrj26:
Replacing the current implementation of bootstrap with CSS to SCSS to adapt the DRY approach &  facilitate writing clean, easy, and less Code.

Some other benefits include variable-controlled CSS, nesting, and support mathematical operations.

Files Changed : static/bootstrap CSS files >> scss files
Added files: theme.scss: current theme same as the physionet build
Extra Libraries required: Django-compressor & Django-libsass | included in requirements.txt with hashes

This **_pull request should ideally not have any noticeable changes on the website_** and is only supposed to replace the Bootstrap CSS files with the Scss setup.


tompollard:
Thanks @Rutvikrj26, this looks good to me, but I'd like to understand more about why the django-compressor and django-libsass packages are needed (and avoid them if possible).

It looks like this pull request essentially adds (1) the bootstrap scss (2) custom theme files. My understanding is that scss compiles to css, so could we not do a single compilation step when the server is started?

Could the build process be hooked into the [./manage.py collectstatic](https://docs.djangoproject.com/en/4.0/howto/static-files/) step? https://pypi.org/project/django-sass/ might also be worth looking into as a more lightweight approach than django-compressor and django-libsass.

Rutvikrj26:
Thank you for your suggestions @tompollard. I read the documentation of Django-sass and was able to find a compile_sass function that can be used in the build script. I am currently trying to remove the Django-compressor and Django-libsass libraries replacing them with the Django-sass. Once I am able to set it up locally, I will try to make the bash script that can be run at the start of deployment before collectstatic command.

.gitignore
```diff
@@ -141,4 +141,7 @@ physionet-django/*credentials*
 *.iso
 
 # Pycharm / JetBrains / IntelliJ project specific settings folder
-.idea/
\ No newline at end of file
+.idea/
+
+# Cache CSS Files
+physionet-django/compressedfiles/CACHE
```
tompollard: I don't think we are using this folder now, so is change still necessary?

physionet-django/physionet/settings/base.py
```diff
@@ -529,3 +530,15 @@ class StorageTypes:
 
 SOURCE_CODE_REPOSITORY_LINK = config('SOURCE_CODE_REPOSITORY_LINK',
                                      default='https://github.com/MIT-LCP/physionet-build')
+
+#COMPRESS_PRECOMPILERS = (
```
tompollard: Can these comments be removed?

physionet-django/physionet/settings/base.py
```diff
@@ -529,3 +530,15 @@ class StorageTypes:
...
+#    ('text/x-scss', 'django_libsass.SassCompiler'),
+#)
+
+#COMPRESS_ROOT = os.path.join(BASE_DIR, 'compressedfiles')
+
+STATICFILES_FINDERS = [
+    'django.contrib.staticfiles.finders.FileSystemFinder',
+    'django.contrib.staticfiles.finders.AppDirectoriesFinder',
+#    'compressor.finders.CompressorFinder',
```
tompollard: Can this comment be removed?

pyproject.toml
```diff
@@ -30,6 +30,8 @@ python-json-logger = "^2.0.2"
 django-storages = { version = "^1.11.1", extras = ["google"] }
 "pdfminer.six" = "^20211012"
 requests = "2.27.1"
+django-compressor = "^3.1"
```
tompollard: are these packages required?

poetry.lock
```diff
@@ -151,6 +162,19 @@ python-versions = "*"
 [package.dependencies]
 six = ">=1.10.0"
 
+[[package]]
+name = "django-compressor"
```
tompollard: Are these packages needed?

physionet-django/templates/base_css.html
```diff
@@ -1,4 +1,7 @@
 {% load static %}
+<link rel="stylesheet" type="text/css" href="{% static 'bootstrap/css/theme.css' %}"/>
 <link rel="stylesheet" type="text/css" href="{% static 'bootstrap/css/bootstrap.css' %}"/>
-<link rel="stylesheet" type="text/css" href="{% static 'font-awesome/css/all.css' %}"/>
+<link rel="stylesheet" type="text/css" href="{% static 'bootstrap/css/bootstrap-grid.css' %}"/>
```
tompollard: Are all of these new stylesheets being used?

tompollard (commented):
Thanks @Rutvikrj26, looks good. I have added a few comments alongside the changes. Please could you take a look and update the pull request if needed?

physionet-django/templates/base_css.html
Rutvikrj26: These are all produced by django_sass after compiling the scss files. These are necessary for importing the style-sheets.

poetry.lock
Rutvikrj26: Removed from Poetry - had to rewrite poetry.lock

pyproject.toml
Rutvikrj26: Removed from Pyproject.toml

physionet-django/physionet/settings/base.py
Rutvikrj26: Removed all the comments

physionet-django/physionet/settings/base.py
Rutvikrj26: Removed the comments

Rutvikrj26:
Thank you for pointing out these issues @tompollard. I have resolved all of them. Please review and let me know if any changes need to be made from my end.

tompollard:
Thanks @Rutvikrj26, I think we're almost there. A couple of quick things:

- It looks like we are bumping the version for lot of packages here. Ideally we should only be changing versions where necessary. Let me know if it would be helpful to chat about this.
- Are you also adding the new dependency/ies to requirements.txt? If not, we should add them to this file too.
- It isn't clear to me that all of the new CSS imports in physionet-django/templates/base_css.html. We are now importing `bootstrap.css`, `bootstrap-grid.scss`, and `bootstrap-reboot.css`. Are all of these required? e.g. looking at this (old!) post on StackOverflow: https://stackoverflow.com/questions/40827280/differences-between-bootstrap-flex-bootstrap-grid-and-bootstrap-reboot, it sounds like `bootstrap-flex.scss` "Includes all the imports from the standard Bootstrap project, but enables the flexbox variable.". In this case, why would the `bootstrap.css` also be needed?

pyproject.toml
```diff
@@ -30,6 +30,7 @@ python-json-logger = "^2.0.2"
 django-storages = { version = "^1.11.1", extras = ["google"] }
 "pdfminer.six" = "^20211012"
 requests = "2.27.1"
+django-sass = "1.1.0"
```
tompollard: Adding quotes to the library name \*might\* fix the build error, e.g.:

`"django-sass" = "1.1.0"`

Rutvikrj26: Implementeed this.

PR #1621, opened 2022-06-28
Add TICKET_SYSTEM_URL variable to config and display in template

tompollard:
Currently we are dealing with support requests by email, which is becoming overwhelming. This pull request adds a `TICKET_SYSTEM_URL` to the config file. 

If `TICKET_SYSTEM_URL` is set, the following statement is added to the "edit training" and "edit credentialing" pages, where users submit their training and credentialing applications:

> To raise a support request, please [click here](url).

I'm sure we'll want to tweak the language etc later, but this is a start.


kepaik:
1) It would be good for link to open in a new window, since it's on a different website
`target="_blank"`

2) Also would be good to add styling for it to stand out a little bit with a tag around the text:
`<div class="alert alert-primary">`

Might want to change the status to a "card" instead of an "alert-warning", so it's not overwhelming.
`<div class="card"><div class="card-body">`

Otherwise, seems to work and looks good.

PR #1622, opened 2022-06-30
Updates reference column names on credentialing page.

lamawmouk:
This pull request renames the column headers in the tabs on the credentialing page. 
The `Reference Response` is changed to `Awaiting Reference Response` and the `Final Review` is changed to `Final Reference Review` to assist admin in understanding the meaning of each tab.

### DIFF
--- a/physionet-django/console/templates/console/credential_processing.html
+++ b/physionet-django/console/templates/console/credential_processing.html
@@ -22,7 +22,7 @@
         <a class="nav-link" id="reference-tab" data-toggle="tab" href="#reference" role="tab" aria-controls="reference" aria-selected="false">Reference Check {{ reference_applications|task_count_badge|safe }}</a>
       </li>
       <li class="nav-item">
-        <a class="nav-link" id="response-tab" data-toggle="tab" href="#response" role="tab" aria-controls="response" aria-selected="false">Reference Response {{ response_applications|task_count_badge|safe }}</a>
+        <a class="nav-link" id="response-tab" data-toggle="tab" href="#response" role="tab" aria-controls="response" aria-selected="false">Awaiting Reference Response {{ response_applications|task_count_badge|safe }}</a>
       </li>
       <li class="nav-item">
         <a class="nav-link" id="final-tab" data-toggle="tab" href="#final" role="tab" aria-controls="final" aria-selected="false">Final Review {{ final_applications|task_count_badge|safe }}</a>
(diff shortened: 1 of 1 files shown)


physionet-django/console/templates/console/credential_processing.html
```diff
@@ -22,10 +22,10 @@
         <a class="nav-link" id="reference-tab" data-toggle="tab" href="#reference" role="tab" aria-controls="reference" aria-selected="false">Reference Check {{ reference_applications|task_count_badge|safe }}</a>
       </li>
       <li class="nav-item">
-        <a class="nav-link" id="response-tab" data-toggle="tab" href="#response" role="tab" aria-controls="response" aria-selected="false">Reference Response {{ response_applications|task_count_badge|safe }}</a>
+        <a class="nav-link" id="response-tab" data-toggle="tab" href="#response" role="tab" aria-controls="response" aria-selected="false">Awaiting Reference Response {{ response_applications|task_count_badge|safe }}</a>
```
tompollard, line 25: Does this cause any issues with page width? If not, no problem, otherwise perhaps "Awaiting response" would be okay.

physionet-django/console/templates/console/credential_processing.html
```diff
@@ -22,10 +22,10 @@
...
       </li>
       <li class="nav-item">
-        <a class="nav-link" id="final-tab" data-toggle="tab" href="#final" role="tab" aria-controls="final" aria-selected="false">Final Review {{ final_applications|task_count_badge|safe }}</a>
+        <a class="nav-link" id="final-tab" data-toggle="tab" href="#final" role="tab" aria-controls="final" aria-selected="false">Final Reference Review {{ final_applications|task_count_badge|safe }}</a>
```
tompollard: This is the final review of the application, not just the reference check. I think leaving this as "Final review" may be better?

physionet-django/user/templates/user/credential_application.html
```diff
@@ -75,4 +75,17 @@ <h2>Research Area</h2>
 {% block local_js_bottom %}
 <script src="{% static 'custom/js/enable-tooltip.js' %}"></script>
 <script src="{% static 'user/js/credential-form-control.js' %}"></script>
+<script>
```
tompollard: This change seems unrelated to the original comment in the pull request. It looks like the changes for https://github.com/MIT-LCP/physionet-build/pull/1603 have been unintentionally included here?

physionet-django/console/templates/console/credential_processing.html
lamawmouk: Yes, no problem with page width. I think add some clarity for the user.

physionet-django/console/templates/console/credential_processing.html
lamawmouk: I think after approving the previous tab checks e.g. ID (identity of applicant), the 'Final Review' tab has the addition of the response from the reference before approval, therefore adding 'Final Reference Review'  adds clarity on what additions to look for when checking this tab before final approval of the applicate as other tab checks have been passed.

tompollard:
Looks good, thanks Lama

PR #1625, opened 2022-07-08
/settings/training/ 'completion_report'

tompollard:
We are currently seeing a fair amount of "KeyError at /settings/training/ 'completion_report'" errors, which I think are a result of people submitting the training form at: http://localhost:8000/settings/training/ without including a copy of their report (presumably because they somehow click submit before the report has had time to upload).

This pull request catches the server errors that are being raised in this case and prompts the user to upload their completion report.


briangow:
Looks good to me @tompollard !

PR #1629, opened 2022-07-20
Tp/ref order 1

Rutvikrj26:
Updates on the References Issue - Basic Working Version with duplicate order validation. Adding the order field and an additional validation check makes sure that INTEGRITY Error can't be replicated. 

This merge will update the MIT-LCP branch with the most basic working version. This version will show an extra field of order in the references part of the content. The reference order is tracked and used later on to sort the references for listing.

### DIFF
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -690,15 +690,6 @@ class ReferenceFormSet(BaseGenericInlineFormSet):
                     raise forms.ValidationError('References must be unique.')
                 descriptions.append(description)
 
-        # orders = []
-        # for form in self.forms:
-        #     # This is to allow empty unsaved form
-        #     if 'order' in form.cleaned_data:
-        #         order = form.cleaned_data['order']
-        #         if order in orders:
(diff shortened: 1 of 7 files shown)


tompollard:
Thanks @Rutvikrj26, this is getting there (and fixes the integrity errors that we currently see), but I find the functionality slightly confusing as a user. When adding three references for a new project, I see the following list:

Screen Shot 2022-09-22 at 3 29 14 PM

For this to be useable, I think we would want to tweak this to:

1.  replace the `->` arrows on the left with dropdown menus showing numbers/order (i.e. move the Order dropdown to the current position of the `->` arrows.
2. display the numbers in sequence. if someone changes the order and then clicks "save" the references should be reordered (in numerical order, 1,2,3,4, etc).

Note that the changes in this pull request should not affect other places that use similar ordered lists. e.g. the "Affiliations" section at the bottom of http://localhost:8000/projects/lwJxuN1OjJNocwAR9PmE/authors/ should not display the `->` arrows as they do currently.

Screen Shot 2022-09-22 at 1 40 38 PM

PR #1630, opened 2022-07-20
Adds event list page

danamouk:
This pull request adds an events list page as part of the user's app to help manage events such as classes, datathons, workshops, conferences, and symposiums that would eventually require data access on PhysioNet. 

The following additions were included:

- events submenu `console_navbar.html` with events list page `event.html` with a url path, `urls.py` to include information on host, event type, event created date, start date and end date.
- event model `models.py` to capture information on events, and participants model to capture information about participants in the event. The models were registered in `admin.py`.
- `demo-events.json` to add demo data for the events list page, and the applied `migrations`.
-  permissions `views.py` to enforce which users can view the event menu and event list page.
-  an enum `enums.py`  to account for the various choices as part of the event type variable in the event model `models.py`.


physionet-django/user/enums.py
```diff
@@ -19,3 +19,15 @@ class RequiredField(IntEnum):
     @classmethod
     def choices(cls):
         return tuple((option.value, option.name) for option in cls)
+
+
+class EventType(IntEnum):
+    COURSE = 0
+    DATATHON = 1
+    CONFERENCE = 2
```
tompollard, line 27: I don't think there is a need for all of these types now. Perhaps just limit to course and workshop?

physionet-django/user/migrations/0046_event_participants.py
```diff
@@ -0,0 +1,48 @@
...
+        ('user', '0045_auto_20220528_2248'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Event',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('name', models.TextField()),
+                ('event_type', models.PositiveSmallIntegerField(choices=[(0, 'COURSE'),
```
tompollard, line 21: This will need updating to match the updated enum (removing symposium etc).

physionet-django/user/admin.py
```diff
@@ -90,3 +90,7 @@ def render_change_form(self, request, context, *args, **kwargs):
 admin.site.register(models.Question)
 admin.site.register(models.TrainingQuestion)
 admin.site.register(models.TrainingRegex)
+
+# Register the Event and Participants on the site
```
tompollard, line 94: no need for this comment

tompollard (commented):
I have added a few thoughts.  @bemoody please could you take a look and let @danamouk know if any additional changes are needed?

physionet-django/user/models.py
```diff
@@ -1164,3 +1164,33 @@ class CodeOfConductSignature(models.Model):
...
+
+    class Meta:
+        unique_together = ('name', 'host')
+
+
+class Participants(models.Model):
+    """
+    Captures information about participants in an event.
+    """
+    participant = models.ForeignKey(User, on_delete=models.CASCADE)
```
tompollard, line 1189: update to match the model in your current branch (something like:
`user = models.ForeignKey(User, on_delete=models.CASCADE, related_name = "participants"`)

as we are considering a new app for events, you could hold off making updates to this pull request until we decide exactly what we want..

danamouk:
#1653 replaces this pull request.

PR #1631, opened 2022-07-28
Add 'resetdb --load-demo' command

bemoody:
I'm tired of typing `/home/benjamin/ve-pndjangoplus/bin/python3 manage.py resetdb && /home/benjamin/ve-pndjangoplus/bin/python3 manage.py loaddemo`.  Let's have a command-line option to do both things in one operation.


tompollard:
looks good, thanks!

PR #1637, opened 2022-08-02
bump django-autocomplete-light to v3.9.4. Ref #1095

tompollard:
This pull request bumps django-autocomplete-light (referenced in imports as `DAL`). 

The update is a requirement for moving to Django 3 because the current version of django-autocomplete-light makes calls to a deprecated django module, raising the following error:

```ImportError: cannot import name 'six' from 'django.utils'```

See: https://docs.djangoproject.com/en/3.2/releases/3.0/#removed-private-python-2-compatibility-apis for detail on this deprecation.


briangow:
Thanks @tompollard!

PR #1638, opened 2022-08-02
Loosen requirement in poetry to cffi ^1.15.1

tompollard:
I'd previously fixed the version of cffi to 1.15.1 in the poetry lock file. This pull request loosens the requirement to ^1.15.1 to avoid dependency issues when making updates in future.


briangow:
Thanks @tompollard!

PR #1639, opened 2022-08-02
In Django 3, Random() moves to a different module.

tompollard:
In https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/db/chaotic/compiler.py we import Random() to force random ordering for queries (to help spot when ordering is an issue).

The current code imports Random() with:

```from django.db.models.expressions import Random```

This is incompatible with Django 3, which requires:

```from django.db.models.functions import Random```

This pull request allows us to temporarily support both imports.


briangow:
Thanks @tompollard!

PR #1640, opened 2022-08-02
Bump Django from v2.2.28 to v3.0.14. Ref #1095

tompollard:
This pull request bumps Django to Django 3.0.14 (Ref #1095). Release notes: https://docs.djangoproject.com/en/4.0/releases/3.0/ 

To test the upgrade, upgrade to Django 3.0.14 and then run: `python -Wa manage.py test`. All tests should pass, with a single (repeated) warning:

>```/Users/tompollard/projects/physionet-build/env/lib/python3.9/site-packages/ckeditor/widgets.py:123: RemovedInDjango40Warning: force_text() is deprecated in favor of force_str(). 'value': conditional_escape(force_text(value)),```

To fix the warning, we should bump ckeditor. We replaced `force_text()` with `force_str()` throughout our codebase in https://github.com/MIT-LCP/physionet-build/pull/1577. 

Note that before bumping to later versions of Django 3 (i.e. >=3.2), we will need to make additional updates. These include:
- Setting the DEFAULT_AUTO_FIELD for auto indexes in settings.py: https://docs.djangoproject.com/en/3.2/ref/settings/#std-setting-DEFAULT_AUTO_FIELD
- Updating ckeditor

### DIFF
--- a/physionet-django/user/management/commands/getmigrationtargets.py
+++ b/physionet-django/user/management/commands/getmigrationtargets.py
@@ -130,7 +130,7 @@ class Command(BaseCommand):
 
         # Check if there are any migrations applied that we don't know
         # about.
-        ghosts = applied - set(desired)
+        ghosts = applied.keys() - set(desired)
         if ghosts:
             mlist = ', '.join('{}.{}'.format(app, name)
                               for app, name in sorted(ghosts))
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,4 +1,18 @@
 [[package]]
+name = "asgiref"
+version = "3.5.2"
+description = "ASGI specs, helper code, and adapters"
+category = "main"
+optional = false
+python-versions = ">=3.7"
+
+[package.dependencies]
+typing-extensions = {version = "*", markers = "python_version < \"3.8\""}
+
+[package.extras]
+tests = ["pytest", "pytest-asyncio", "mypy (>=0.800)"]
+
+[[package]]
 name = "bleach"
 version = "3.3.0"
 description = "An easy safelist-based HTML-sanitizing tool."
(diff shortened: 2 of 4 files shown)


tompollard:
@bemoody the following command is failing on the upgrade test:
`./manage.py getmigrationtargets --current`

The test fails at:
```ghosts = applied - set(desired)```

"`applied`" is a dictionary and "`desired`" is a list. 

It looks like this is a result of changes to django/db/migrations/executor.py a year or two ago, when `applied` switched from a set to a dict. See: https://github.com/django/django/commit/371ddade1e4e28827fd45e041c9410f8b4f01067

`applied` now looks like this:

```
{('user', '0024_associatedemail_verification_token'): <Migration: Migration 0024_associatedemail_verification_token for user>, ('user', '0002_auto_20181231_1623'): <Migration: Migration 0002_auto_20181231_1623 for user>, ('notification', '0003_news_project'): <Migration: Migration 0003_news_project for notification>, ('project', '0051_auto_20211129_0427'): <Migration: Migration 0051_auto_20211129_0427 for project>, ('user', '0033_credentialapplication_reference_organization'): <Migration: Migration 0033_credentialapplication_reference_organization for user>, ...
```

We're comparing this to `desired`, which looks like:

```
[('user', '0001_initial'), ('contenttypes', '0001_initial'), ('admin', '0001_initial'), ('admin', '0002_logentry_remove_auto_add'),
```

I think the solution is to call set() on `applied`, so the test becomes:
```ghosts = set(applied) - set(desired)```

or perhaps more explicitly:
```applied.keys() - set(desired)```

These compare the keys of `applied` to the `desired` set. Does one of these solutions sound good to you?

briangow:
Looks good @tompollard!

PR #1645, opened 2022-08-04
Feature/theming

Rutvikrj26:
Replacing the current implementation of bootstrap with CSS to SCSS to adapt the DRY approach & facilitate writing clean, easy, and less Code.

Some other benefits include variable-controlled CSS, nesting, and support mathematical operations.

Files added : static/bootstrap SCSS files
Added files: theme.scss: current theme same as the physionet build
Extra Libraries required: Django-sass & libsass(dependency) | included in requirements.txt with hashes

This pull request should ideally not have any noticeable changes on the website and is only supposed to replace the Bootstrap CSS files with the Scss setup.


tompollard:
Thanks @Rutvikrj26, this is getting there. 

- When we compile the CSS, we overwrite existing bootstrap css files. This will cause problems with deployment to the live server because when we push updates the remote files will have changed. 
   - I think the solution is to remove (`git rm`) any existing versions of the bootstrap files that are being written (e.g. those in https://github.com/MIT-LCP/physionet-build/tree/dev/physionet-django/static/bootstrap/css).
  - If we make the change above, we will also need to make sure that development servers include the CSS compilation step.
- `Write.sh` probably shouldn't be in the root of the project and it should have a more descriptive name. I'm not sure that it is needed at all though? If possible, I think we would prefer for style variables to go into a customisable theme file or the .env config file.

bemoody:
Sorry I haven't taken the time to review this.

Before we can merge this there are a few major issues (but hopefully not difficult to fix):

1. The system needs to have sensible default values if nothing is configured.

2. The deployment scripts must be tested and working.

3. The Github test suite must verify that the theme can be compiled before a pull request can be merged.

I don't think it's wise to write this logic as a shell script (for a lot of reasons).  Instead, add a management command (`physionet/management/commands/compilestatic.py` ?) that writes the `theme.scss` file and then invokes the `sass` command.

Rutvikrj26:
Excellent Suggestion @bemoody 
I created a management command : compilestatic, which covers both the theme file generation & SCSS compilation. Furthermore, I have added a run test in physionet-build-test.yml. The management command has a basic check to ensure the hex code is valid. It also has default values set in, which are replaced if an environment variable for the corresponding theme color is present. Finally, I have also modified the post hook script, which invokes the compilestatic command just before collectstatic.
@tompollard, @bemoody, please let me know if any issue comes up while testing the setup.

tompollard:
thanks @Rutvikrj26! please could you resolve the conflict with dev?

bemoody:
Thanks, this looks good!  A couple of small things I would improve:

* Please add the `compilestatic` command before `collectstatic` in `deploy/test-server/install-pn-test-server`.

* Setting configuration variables:
```
        # checking if DARK environemnt variable is set
        if os.environ.get('DARK'):
            # setting the the Dark Parameter to the environment variable
            dark = str(os.environ.get('DARK'))
            # checking if the Dark Parameter is a valid hex color
            if len(dark) != 7:
                raise CommandError('DARK environment variable is not a valid hex color')
            # announcing the change
            self.stdout.write(self.style.SUCCESS('DARK environment variable set to %s' % dark))
```
Variables in `.env` are not inserted into `os.environ`.  Instead, you can do this:
```
from decouple import config
# ...
        dark = config('DARK', default='#343A40')
        if len(dark) != 7:
            raise CommandError('DARK environment variable is not a valid hex color')
```

* Invoking the "sass" command:
```
            try:
                # compiling the static files
                os.system('python manage.py sass static/bootstrap/scss/ static/bootstrap/css/')
            except Exception as e:
                # printing the error message
                print(e)
                exit(1)            
```
The interpreter (`python`) might not be what you expect, and `os.system` doesn't handle all errors.  It's simpler to invoke the "sass" command this way:
```
from django.core.management import call_command

# ...
            call_command('sass', 'static/bootstrap/scss/',
                         'static/bootstrap/css/')
```
(You don't need to explicitly handle exceptions if all you're going to do is print an error message and exit, and it's often better *not* to do that since the default error message will show more information for debugging.)

Rutvikrj26:
The merge conflicts have been resolved @tompollard 
@bemoody I have implemented your suggestions. Please review and merge.

tompollard:
@Rutvikrj26 thanks! It looks like there are a couple of minor style issues that need fixing to pass tests:

```
physionet-django/user/management/commands/compilestatic.py:9:1: E302 expected 2 blank lines, found 1
physionet-django/user/management/commands/compilestatic.py:21:1: E302 expected 2 blank lines, found 1
physionet-django/user/management/commands/compilestatic.py:41:89: W292 no newline at end of file
```

bemoody:
Since theme.scss is dynamically generated, please remove it from the git repository (`git rm physionet-django/static/bootstrap/scss/theme.scss`).  Also, please add these files to `.gitignore`:
```
# Static content generated by "manage.py compilestatic"
physionet-django/static/bootstrap/css/theme.scss
physionet-django/static/bootstrap/css/*.css
```

The changes look good to me, though we will want to be a little bit careful & test deployment with the staging server.  Note that after installing the modified `post-receive`, it will break if you try to force push an older version.

@tompollard, can you check why tests aren't running?

Rutvikrj26:
@bemoody There are other css files not generated by scss compilation and added each newly generated file individually in the gitignore.

tompollard:
@bemoody it looks like external pull requests require someone to click the "Approve and run" button in github before the tests run. I guess we may need to check you permissions if you're not seeing this. We should probably come up with a more permissive rule for running tests.

tompollard:
This looks great to me. Hate to say there are a couple of minor bugs:

## Menu

In the top menu, something funky has happened with the down arrow that indicates a dropdown menu:

> Screen Shot 2022-09-15 at 10 55 53 AM

> Screen Shot 2022-09-15 at 11 00 45 AM

## Left alignment of content

There seems to be a left alignment issue for a fair amount of text:

1. e.g. on project pages (e.g. http://localhost:8000/content/demowave/1.0.0/), the "Files" section is shifted to the right

> Screen Shot 2022-09-15 at 11 05 44 AM

2. On many pages, non-header content is left of the headers. e.g.

http://localhost:8000/projects/QN5hLDij2mbCKUPgGaUL/access/

> Screen Shot 2022-09-15 at 11 09 17 AM

3. http://localhost:8000/console/user/manage/admin/

> Screen Shot 2022-09-15 at 11 08 25 AM

tompollard:
@Rutvikrj26 any chance that you could take a quick look at these issues today? If not, I suggest we merge now, test on staging, and then follow up with tweaks later.

Rutvikrj26:
All issues are now resolved: They were coming as I was importing both the theme file and the Boostrap file - but since the theme file itself imported the bootstrap.scss, only file needed to be generated and imported.
image
image
image
image

tompollard:
Thanks @Rutvikrj26, looks great!

PR #1652, opened 2022-08-24
update the max filename length to 100 characters

lamawmouk:
This change updates the code base to change the file length name from 60 to 100 characters long. This is done in order to avoid having to ask the contributors to reorganize their files.


tompollard:
Looks good to me, thanks Lama!

PR #1668, opened 2022-09-19
Adds a column that aggregates credentialing applications by event to the event list page.

danamouk:
This pull request, adds a column that adds a link that lists all outstanding credentialing applications to the event_list page. This is useful for admin to narrow down the participants to process that are relevant to a particular event.

- The next step could include adding a training column to view all outstanding training applications to the event_list page. 
- It also might be useful to add a flag to represent the number of users in a class, that are trained/credentialed to skip credentialing/training for courses that  already satisfy the credentialing/training requirement. 
- Right now after processing a credentialing application, the process is referred back to full list of users, a simple workflow could be devised to redirect admin to students in a batch, or back to the event_list if user's already satisfy the training/credentialing requirement.

### DIFF
--- a/physionet-django/console/templates/console/event.html
+++ b/physionet-django/console/templates/console/event.html
@@ -26,6 +26,7 @@
                   <th>Created</th>
                   <th>Started</th>
                   <th>Ended</th>
+                  <th>Credentialing</th>
                 </tr>
               </thead>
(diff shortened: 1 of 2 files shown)


tompollard:
Looks good, thanks Dana!

Issue #1664, opened 2022-09-13
Add a "past events" section to the events home page


tompollard:
Currently we display a list of events at https://localhost:8000/events/ under an "Active Events" title. We should add a new section called "Past Events". In the Past Events section we should 

- Display projects where the End Date of the event is in the past.
- Perhaps hide the sign up link for past events and prevent new users from signing up (e.g. if someone clicks a link for a past event, they might see a message saying "This event has now finished".)

Chrystinne:
@tompollard just to confirm if I get this straight: Is the link below the one we need to hide for past events?

![link to remove](https://user-images.githubusercontent.com/4949149/189998509-e3762a99-db90-4637-9f5f-04d92c02d50f.png)

If anyone has already received the link, we will replace the previous Event registration confirmation below with the "This event has now finished" message.

![enroll](https://user-images.githubusercontent.com/4949149/189998901-9194c316-eabc-4f9d-9b01-2de72fc8ec62.png)

tompollard:
Thanks Chrystinne, yes, that's what I was thinking (though we are making this up as we go along...please suggest alternative ideas if something better springs to mind).

PR #1669, opened 2022-09-19
Add Past Events section to the Events home page

Chrystinne:
It solves issue #1664:

- Add a "Past events" section to the events home page to display projects where the End Date of the event is in the past;

- Hide the sign-up link for past events to prevent new users from signing up;

- Display the message 'This event has now finished' if someone clicks a link for a past event.

image

### DIFF
--- a/physionet-django/physionet/views.py
+++ b/physionet-django/physionet/views.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 from os import path
 from re import fullmatch
 from urllib.parse import urljoin
+from datetime import datetime
 
 import notification.utility as notification
 from django.contrib import messages
@@ -237,8 +238,9 @@ def event_home(request):
     is_instructor = user.has_perm('user.add_event')
 
     # sqlite doesn't support the distinct() method
-    events = set(Event.objects.filter(Q(host=user) | Q(participants__user=user)))
-
+    events_previous = Event.objects.filter(Q(host=user) | Q(participants__user=user))
+    events = set(events_previous.filter(end_date__gte=datetime.now()))
+    past_events = set(events_previous.filter(end_date__lt=datetime.now()))
     event_form = AddEventForm(user=user)
 
     url_prefix = notification.get_url_prefix(request)
@@ -251,6 +253,7 @@ def event_home(request):
 
     return render(request, 'event_home.html',
                   {'events': events,
+                   'past_events': past_events,
                    'event_form': event_form,
                    'url_prefix': url_prefix,
                    'is_instructor': is_instructor
--- a/physionet-django/physionet/views.py
+++ b/physionet-django/physionet/views.py
@@ -271,7 +271,7 @@ def event_add_participant(request, event_slug):
 
     if event.end_date < datetime.now().date():
         messages.error(request, "This event has now finished")
-        return redirect(event_home)    
+        return redirect(event_home)
 
     if event.participants.filter(user=user).exists():
         messages.success(request, "You are already enrolled")
--- a/physionet-django/physionet/views.py
+++ b/physionet-django/physionet/views.py
@@ -238,9 +238,9 @@ def event_home(request):
     is_instructor = user.has_perm('user.add_event')
 
     # sqlite doesn't support the distinct() method
-    events_previous = Event.objects.filter(Q(host=user) | Q(participants__user=user))
-    events = set(events_previous.filter(end_date__gte=datetime.now()))
-    past_events = set(events_previous.filter(end_date__lt=datetime.now()))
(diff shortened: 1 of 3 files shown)


physionet-django/templates/event_home.html
```diff
@@ -115,6 +115,76 @@ <h5 class="modal-title">Participants</h5>
         </div>
     {% endfor %}
   {% endif %}
+
+<!-- Past events -->
+
+  <div class="card" >
+    <div class="card-header">
+      <h2>Past Events</h2>
+      <p class="card-text">Events hosted in the past</p>
```
tompollard: This can wait to a later pull request, but I think we should probably remove the "Events hosted in the past" and "Events to be hosted" subheaders. They essentially just repeat the header.

physionet-django/templates/event_home.html
```diff
@@ -115,6 +115,76 @@ <h5 class="modal-title">Participants</h5>
...
+    </div>
+    
+    <ul class="list-group list-group-flush">
+      {% if events %}
```
tompollard: Not sure it makes much of a difference, but you could consider dropping the `{% if events %} ... {% endif %}` here (and in the chunk below) and instead use the "For...Empty" syntax (https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#for-empty)

physionet-django/physionet/views.py
```diff
@@ -237,8 +238,9 @@ def event_home(request):
     is_instructor = user.has_perm('user.add_event')
 
     # sqlite doesn't support the distinct() method
-    events = set(Event.objects.filter(Q(host=user) | Q(participants__user=user)))
-
+    events_previous = Event.objects.filter(Q(host=user) | Q(participants__user=user))
```
tompollard: I'm not sure what "previous" is referring to in `events_previous`. `events_all` might be clearer? We could also consider renaming `events` to `active_events` for consistency with `past_events`?

physionet-django/templates/event_home.html
```diff
@@ -115,6 +115,76 @@ <h5 class="modal-title">Participants</h5>
...
+        <div class="modal-dialog modal-lg"  role="document">
+          <div class="modal-content">
+            <div class="modal-header">
+              <h5 class="modal-title">Participants</h5>
+              <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                <span aria-hidden="true">&times;</span>
+              </button>
+            </div>
 
+              <div class="table-responsive">
```
tompollard: Minor issue but this might be more readable if we moved the event entries into a separate template that is then imported (might also help to avoid duplicated code?). Not necessary to make the change though...we could refactor later.

tompollard (approved):
Looks good to me, thanks @Chrystinne. I added a few thoughts inline, but I don't think there's anything that _needs_ to be changed. Please could you take a quick look and see whether there is anything you'd like to update before we merge?

physionet-django/templates/event_home.html
Chrystinne: @ Yes, I agree. I added the subheader for the Past events because we had one for the Active events. But I also think we should remove them. I'll do it.

physionet-django/physionet/views.py
Chrystinne: I thought of renaming "events" to "active_events" before so I will change it now. Also, "events_all" makes more sense than "events_previsous".

alistairewj: * `active_events`
* `past_events`
* `all_events`

... seems more consistent

alistairewj: ... but from an IDE code completion point of view, `events_active`, `events_past`, and `events_all` will be easier to code with.

(sorry to bike-shed!).

tompollard: Anything is fine with me, just having consistency makes it more readable (yeah, wasn't thinking when I suggested events_all!).

Chrystinne: In this case, I've renamed the variables' names to events_active, events_past, and events_all.

physionet-django/templates/event_home.html
Chrystinne: Yes, the "For...Empty" syntax is better. I refactored the code to use it.

physionet-django/templates/event_home.html
Chrystinne: I've added the event entries into a separate template (event_entries.html) to avoid duplicate code.

Chrystinne:
@tompollard @alistairewj thanks for the comments. 

I believe I've finished all changes. I would appreciate it if you could have a look, @tompollard.

tompollard:
Looks great, thanks!

PR #1671, opened 2022-09-22
Adds a column that aggregates training applications by event to the event list page.

danamouk:
This pull request, adds a column that adds a link that lists all outstanding training applications to the event_list page. This is useful for admin to narrow down the participants to process that are relevant to a particular event.


tompollard:
Looks good to me, thanks Dana!

PR #1676, opened 2022-09-27
Adds description field to form on event home page.

danamouk:
This pull request adds a description field to form on event home page while hosts are creating events.


tompollard:
It looks like this pull request includes a bunch of unrelated formatting changes. Please could you remove these changes? I assume they were automatically applied by your code editor when you saved the file. In VSCode, you can switch this off with "preferences > settings > text editor > formatting"

danamouk:
thanks @tompollard the button under Editor: Format On Save was unclicked, and after `git reset --hard`, the changes in the local directory were removed, and the description field additions were added.

tompollard:
Thanks Dana! Are you happy with how the form looks? For me, the description field looks a little awkwardly placed:

Screen Shot 2022-09-28 at 9 29 01 AM

Ideally I think the box should sit below the title, like this:

Screen Shot 2022-09-28 at 9 55 02 AM

A quick fix would be to add a <br /> to the form_snippet.html before the field (`</label><br /> {{ field }}`) but probably better to work out how to deal with this in the css. We would need to be careful to avoid negatively affecting other forms that use `form_snippet.html`.

Issue #1596, opened 2022-05-27
"is_valid" and "is_expired" flags on training objects should be mutually exclusive. Ref #1595


tompollard:
There seems to be a bug in the way training reports are managed. I would expect the "is_valid" and "is_expired" flags for a Training object to be mutually exclusive:

https://github.com/MIT-LCP/physionet-build/blob/966e22f57aed1b8c06c4391f36b949c32e04d753/physionet-django/user/models.py#L1103-L1115

But it is possible for both flags to return True for a single training type:

```python
>>> a
<Training: Training object (x)>
>>> a.is_valid()
True
>>> a.is_expired()
True
```

We have users reporting that their training is shown as "active" but they are unable to access a dataset because they are required to complete training. They cannot upload new training because they already have active training, so they are stuck. See also #1595.

tompollard:
In the case of one of the users I am checking `a.training_type.valid_duration` is empty, so `is_expired` is returning True. We do not support expiry dates for training, so what should this field contain?

tompollard:
Also, if there is at least one user that returns "is_expired" for their training, why is the list at https://physionet.org/console/training/expired/ empty?

tompollard:
It looks like the expired list is generated with:

https://github.com/MIT-LCP/physionet-build/blob/966e22f57aed1b8c06c4391f36b949c32e04d753/physionet-django/user/managers.py#L22-L29

This needs a bit of investigation.

tompollard:
This template is suggesting that if a training type does not have a `training.valid_datetime` then the training is valid until the end of the 🌎

https://github.com/MIT-LCP/physionet-build/blob/7255ec760360b843abc2970853fc5fe9192fda96/physionet-django/console/templates/console/valid_training_table.html#L21-L27

tompollard:
I think the bug is in the following chunk:

```
 def is_expired(self): 
     if self.status == TrainingStatus.ACCEPTED: 
         if not self.training_type.valid_duration: 
             return True 
         else: 
             return self.process_datetime + self.training_type.valid_duration < timezone.now() 
```

I think whoever wrote this intended `if not self.training_type.valid_duration` to return False, not True (i.e. they intended for training types without a duration to never expire).

tompollard:
For now I have set our CITI training expiry date to 100 years in the future.

superryeti:
[...]

PR #1678, opened 2022-09-29
Correct is_expired and is_valid behaviour.

harryface:
Bug as raised here: https://github.com/MIT-LCP/physionet-build/issues/1596

There seems to be a bug in the way training reports are managed. The "is_valid" and "is_expired" flags for a Training object should be mutually exclusive, but it is possible for both flags to return True for a single training type.
If the valid duration does not exist, it should not be termed expired.


tompollard:
good catch, thanks @harryface!

Issue #1648, opened 2022-08-04
zxcvbn-python is deprecated. Replace with zxcvbn or remove.


tompollard:
We currently use the `zxcvbn-python`package to estimate the password strength.  The package is deprecated: https://pypi.org/project/zxcvbn-python/

We could replace the zxcvbn-python with zxcvbn (https://pypi.org/project/zxcvbn/) or instead remove the dependency altogether.

PR #1682, opened 2022-10-01
Replace the zxcvbn-python package with zxcvbn

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/1648, we currently use the zxcvbn-python package to estimate the password strength. The package is deprecated: https://pypi.org/project/zxcvbn-python/

This pull request replaces zxcvbn-python with zxcvbn (https://pypi.org/project/zxcvbn/). zxcvbn is a continuation of the original project.


briangow:
@tompollard looks good to me, thanks!

Issue #1655, opened 2022-09-08
Allow instructors to specify email domains (e.g. @mit.edu) for participants


tompollard:
In https://github.com/MIT-LCP/physionet-build/pull/1653/ we allow instructors to create classes and invite participants via a link. 

@alistairewj suggests that we allow instructors to specify an email domain for participants. e.g. in order to join the event, users must be associated with an @mit.edu email address.

This won't work for all events (e.g. perhaps not workshops) but I think it is a nice feature that could help to make the system less susceptible to misuse.

PR #1683, opened 2022-10-03
Adds allowed event domains to verify participants.

danamouk:
This pull request addresses issue #1655. It adds an `allowed_domains` field to the `Event` model along with the migrations, which allows instructors to specify allowed email domains (separated by commas) as part of the form on the `event_home` page. 

Currently, participants can register for an event through a shared event link from the instructor. This pull request, also alerts participants to register for an event using the allowed domains if the domain used does not match the domains listed by the instructor (participants are redirected to the `event_home` page).  If the participants domains match the domains specified by the instructor, the participants are directed to the `event_participant` page to confirm participation in the event.


tompollard:
Thanks @danamouk, this is almost there. I noticed a couple of bugs when testing:

1. If the domain list has spaces, these spaces seem to cause issues with the pattern matching. e.g. if you add "blah.com, mit.edu" when creating the event, someone with an @mit.edu address will not match (because it looks for " mit.edu"). You can fix this by stripping the white space (e.g. with `strip()`)
2. If the form validation fails when creating an event (e.g. because the domain list is not comma separated) then the error message isn't seen because the modal closes when the user tries to submit the form.

PR #1685, opened 2022-10-06
Sorting training list based on credentialing status

Chrystinne:
Sorting training list based on credentialing status. Now applicants who have been credentialed appear at top of the training list. 
It also adds the Credentialed column to the training list indicating if a user has already been credentialed.

### DIFF
--- a/physionet-django/console/templates/console/expired_training_table.html
+++ b/physionet-django/console/templates/console/expired_training_table.html
@@ -18,13 +18,7 @@
                     <td>{{ training.user.get_full_name }}</td>
                     <td>{{ training.user.email }}</td>
                     <td>{{ training.training_type.name }}</td>
-                    <td>
-                        {% if training.user.is_credentialed %}
-                            Yes
-                        {% else %}
-                            No
-                        {% endif %}
-                        </td>
+                    <td>{{ training.user.is_credentialed|yesno:"Yes,No" }}</td>
                     <td>{{ training.valid_datetime }}</td>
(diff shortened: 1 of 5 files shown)


physionet-django/console/templates/console/expired_training_table.html
```diff
@@ -17,6 +18,13 @@
                     <td>{{ training.user.get_full_name }}</td>
                     <td>{{ training.user.email }}</td>
                     <td>{{ training.training_type.name }}</td>
+                    <td>
+                        {% if training.user.is_credentialed %}
```
tompollard: This can be simplified with the Django "yesno" template tag (see: https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#yesno).

e.g. I think you should be able to replace the if clause with something like this: 
```
{{ training.user.is_credentialed|yesno:"Yes,No" }}
```

tompollard (commented):
Looks good, thanks Chrystinne! 

There is a "yesno" template tag provided by Django that could replace the if clauses: https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#yesno. 

e.g. `{{ training.user.is_credentialed|yesno:"Yes,No" }}`

physionet-django/console/templates/console/expired_training_table.html
Chrystinne: @tompollard I did the suggested changes, Tom. Thank you so much.

tompollard:
Thanks, looks good to me!

Issue #1679, opened 2022-09-30
Credentialing Console: Add column with training status


kepaik:
1. Active
2. Under Review
3. None

Chrystinne:
@kepaik I've created the pull request below to solve this issue:

https://github.com/MIT-LCP/physionet-build/pull/1689

Chrystinne:
@kepaik Have you thought about how the system should work (i.e., which information should be shown on the Training Status) in situations where:

1) A user has multiple CITI training applications (e.g., a "rejected" application and an application "under review");

2) A user has multiple types of training (e.g. CITI training, Datathon).

I've added the Training Status column, but I haven't considered these situations on my Pull Request (https://github.com/MIT-LCP/physionet-build/pull/1689). 

So, I came up with this possible solution: turning the two columns "Training Type" and "Training Status" into one (Training Status). So Training Status would list every training available for each user along with its status. (e.g. CITI program: Active, Datathon training: Under review). 

Would that make sense to you? Or do you have a better idea?

kepaik:
Right now, I don't have any other types of training, so it's not useful to me for credentialing, only whether it's active or not.  I don't know if Tom or Alistair have other plans.

To clarify, this request was for the Training reports, which you've already completed, not the Credentialing processing.  For me, it's not useful there as it's not part of the processing decision tree. Might actually clutter the screen with too many options.

Chrystinne:
@kepaik Ok. So, if there is no need for the "Training Status" column on the Credentialing table, we can close this issue, right? And I'll close the PR (https://github.com/MIT-LCP/physionet-build/pull/1689).

PR #1689, opened 2022-10-08
Adding column Training Status to the credentialing table

Chrystinne:
It solves the #1679 issue.

Adding column Training Status to the credentialing table (ID Check tab). 

List of the training status:
- Under Review;
- Withdrawn;
- Rejected;
- Active;

### DIFF
--- a/physionet-django/user/models.py
+++ b/physionet-django/user/models.py
@@ -778,6 +778,21 @@ class CredentialApplication(models.Model):
     class Meta:
         default_permissions = ('change',)
 
+    def get_training_status(self):
+        training_status = self.user.trainings.all().filter(
+            user__username__icontains=self.user.username).values_list('status', flat=True).first()
+        return self.get_training_status_name(training_status)
+
+    def get_training_status_name(self, training_status):
+        if training_status == 0:
+            return 'Under Review'
+        elif training_status == 1:
+            return 'Withdrawn'
+        elif training_status == 2:
+            return 'Rejected'
+        elif training_status == 3:
+            return 'Active'
+
     def file_root(self):
         """Location for storing files associated with the application"""
         return os.path.join(CredentialApplication.FILE_ROOT, self.slug)
@@ -1074,7 +1089,6 @@ class Training(models.Model):
 
     class Meta:
         default_permissions = ()
-
     def delete(self, *args, **kwargs):
         if self.completion_report is not None:
             self.completion_report.delete()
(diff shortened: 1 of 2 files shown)


physionet-django/user/models.py
```diff
@@ -778,6 +778,21 @@ class CredentialApplication(models.Model):
     class Meta:
         default_permissions = ('change',)
 
+    def get_training_status(self):
+        training_status = self.user.trainings.all().filter(
+            user__username__icontains=self.user.username).values_list('status', flat=True).first()
+        return self.get_training_status_name(training_status)
+
+    def get_training_status_name(self, training_status):
```
tompollard: You could use the Enum class to get the text representation of the status, e.g.:

```
status = 1
print(TrainingStatus(status).name)
>> WITHDRAWN
```

physionet-django/user/models.py
```diff
@@ -778,6 +778,21 @@ class CredentialApplication(models.Model):
     class Meta:
         default_permissions = ('change',)
 
+    def get_training_status(self):
```
tompollard, line 781: It is possible that there are multiple training applications associated with a user, so this doesn't seem quite right to me. By selecting the first(), don't we end up ignoring training applications?

tompollard (commented):
Thanks for looking at this Chrystinne. I'm not clear exactly what we want to display in the new "Training Status" column, but the current behavior doesn't seem right.

e.g. If a user has multiple CITI training applications (for example a rejected application and an application under review), what status should be displayed? (and does the code display this?).

We also have the ability to support multiple types of training (e.g. CITI training,  Datathon). How does the status listed in the column relate to these different types of training?

physionet-django/user/models.py
Chrystinne: That's a good point, @tompollard. When I wrote the code, I saw that - for the Review tab on the training list -, there was only the "CITI Data or Specimens Only Research" training type for all users. So I added the columns with only the "CITI Data or Specimens Only Research" in mind. 

But you are right, we need to figure out whether it makes sense to keep this information because it's still not clear to me what we should show on the Training Status column when we have different types of training.

Chrystinne: Thanks, @tompollard.

physionet-django/user/models.py
Chrystinne: @tompollard what about turning the two columns Training Type and Training Status into one (Training Status). So Training Status would list every training available for each user along with its status. (e.g. CITI program: Active, Datathon training: Under review).
Would that make sense?

tompollard: Makes sense to me. It might end up leading to lots of information in the single cell within the column, but perhaps not a big deal? Might be worth checking to see whether @kepaik is okay with this (or if he has better ideas!).

Issue #1690, opened 2022-10-11
Distinguish/Rename "credentialed" with "identity verified"


kepaik:
Lots of confusion when the first part of credentialing (identity verification) is completed, but the training report hasn't been processed.

Credentialing = (identity verification) + (training certification)

Chrystinne:
@tompollard Do we need another pull request for this change? Or may I add this change to the currently opened pull request (https://github.com/MIT-LCP/physionet-build/pull/1689)?

Chrystinne:
@tompollard I've created a new pull request. It might be better, even when it is a minor fix.

tompollard:
> Credentialing = (identity verification) + (training certification)

@kepaik This means that being "credentialed" is dataset specific because we support multiple training types. Fine with me but I want to check whether this is what you intend?

kepaik:
This has less to do with the backend, but the front end for the users.

We've gotten many emails where they are confused that they have been "credentialed", but they actually do not have the credentials to access teh database.

kepaik:
![5F30CF2C@5261FB04 04234563 png](https://user-images.githubusercontent.com/6868450/195189510-f2186ea3-d99d-49ef-85a1-6378ea49467a.jpg)
![C96800E1@E5DA011A 04234563 png](https://user-images.githubusercontent.com/6868450/195189514-36d8b61e-0756-4d9d-8417-290da457c8e0.jpg)

kepaik:
Maybe if we change it to "verified" for identity, and then "credentialed" for each database.

tompollard:
> Maybe if we change it to "verified" for identity, and then "credentialed" for each database.

Yeah, I like the idea of something like this.

Chrystinne:
@kepaik I've renamed the column "Credentialed" to "Identity Verified". It's available at https://physionet.org/console/training/review/

kepaik:
The column name is not the issue.  It's the informational page and emails to the user.

https://physionet.org/settings/credentialing/

See above images for reference.

Chrystinne:
We need to replace the current text with the following:

"To access certain datasets, we require your identity to be "verified". This means that you must formally submit your personal details for review, so that we can confirm your identity.

Certain datasets also require you to complete [training](https://physionet.org/settings/training/) and/or to sign a Data Use Agreement. You can find specific requirements in the "Files" section of the project description."

To raise a support request, please [click here](https://physionet.freshdesk.com/).

[...]

PR #1691, opened 2022-10-11
Changing column name on Training list 

Chrystinne:
Changing column name on Training list (from "Credentialed" to "Identity Verified"). It's related to the #1690 issue.


tompollard:
Looks good to me, thanks Chrystinne!

PR #1693, opened 2022-10-19
Updates allowed domains in views.

Mira-MM:
This pull request verifies domain additions in a more secure way. For example, if the allowed domains from the instructors includes "mit.edu" an "@" is enforced before the email domain when matching the email address with the allowed domains to avoid users with fake emails such as "mit.edu@wobble.com"  to register for the event.


danamouk:
thanks @Mira-MM looks good.

tompollard:
Thanks, looks good to me too!

PR #1694, opened 2022-10-24
Merge to Dev: Story 609 - add dynamic page functionality for Static Pages

harryface:
Allows dynamic static pages to be created on the fly with a dynamic URL too

- Levergaed middleware
- No model change
- View tests included
- Former static page URL modified (urls to existing static pages tested and still working as expected)

NB: If the url clashes with a hard-coded url, hard-coded would take precedence.

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -761,21 +761,22 @@ class SectionForm(forms.ModelForm):
 
 
 class StaticPageForm(forms.ModelForm):
-    """
-    Form for creating a dynamic static page."""
+    """ Form for creating a dynamic static page."""
 
     url = forms.RegexField(
         label="URL",
         max_length=100,
-        regex=r"^[-\w/\.~]+$",
+        regex="^\/about\/([-\w\.~]+\/)+$",
         help_text=(
-            "Url should be unique. If it clashes with a static url, "
-            "The static url will take precedence."
+            "URL should be unique. If the new URL clashes with a static url, "
+            "the static url will take precedence. URL must start with /about/ "
+            "for example /about/publish/"
         ),
         error_messages={
             "invalid": (
-                "This value must contain only letters, numbers, dots, "
-                "underscores, dashes, slashes or tildes."
+                "Must be in format /about/value/value/value/ "
+                "and value must contain only letters, numbers, dots, "
+                "underscores, dashes or tildes."
             ),
         },
     )
@@ -785,12 +786,16 @@ class StaticPageForm(forms.ModelForm):
         fields = "__all__"
 
     def clean_url(self):
-        """ Validate if it has a leading and trailing slashes """
+        """ This is redundant, regex in regex field does the same.
+        Validate that URL starts with /about/, append trailing / if not exist
+        """
 
         url = self.cleaned_data.get("url")
-        if not url.startswith("/") or not url.endswith("/"):
+        if not url.startswith("/about/"):
             raise forms.ValidationError(
-                "URL needs to have both leading and trailing slashes.")
+                "URL must start with /about/, (eg) /about/publish/")
+        if not url.endswith("/"):
+            url = f"{url}/"
         return url
 
     def clean(self):
@@ -803,7 +808,7 @@ class StaticPageForm(forms.ModelForm):
             same_url = same_url.exclude(pk=self.instance.pk)
 
         if same_url.exists():
-            raise forms.ValidationError(f"Static page with url {url} already exists")
+            raise forms.ValidationError(f"Static page with URL: {url} already exists")
 
         return super().clean()
 
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -766,7 +766,7 @@ class StaticPageForm(forms.ModelForm):
     url = forms.RegexField(
         label="URL",
         max_length=100,
-        regex="^\/about\/([-\w\.~]+\/)+$",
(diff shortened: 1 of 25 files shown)


tompollard:
Thanks @harryface, this will be useful! 

We have just merged https://github.com/MIT-LCP/physionet-build/pull/1695 which adds testing functionality to URLs. Please could you rebase to deal with conflicts and then take a look at the failing tests?

Most likely there will need to be some additional tweaks to urls.py to address the tests added in https://github.com/MIT-LCP/physionet-build/pull/1695. We can chat about this tomorrow if anything isn't clear.

harryface:
Rebased and refactored.
The `test_url` assumes that every url that receives a positional argument should also be called without one, this one does not and as a result fails...

bemoody:
I think we want to restrict these pages to the /about/ hierarchy or other well-defined hierarchies.  I *don't* want to have new pages created all over the place that we then have to worry about conflicting with.

> The test_url assumes that every url that receives a positional argument should also be called without one, this one does not and as a result fails...

`test_url` doesn't make any such assumptions.  The `views.static_view` function is invoked both with and without an argument because there are two separate `path()`s for it (one that has an argument and one that doesn't.)

harryface:
> I think we want to restrict these pages to the /about/ hierarchy or other well-defined hierarchies. I _don't_ want to have new pages created all over the place that we then have to worry about conflicting with.
> 
> > The test_url assumes that every url that receives a positional argument should also be called without one, this one does not and as a result fails...
> 
> `test_url` doesn't make any such assumptions. The `views.static_view` function is invoked both with and without an argument because there are two separate `path()`s for it (one that has an argument and one that doesn't.)

With the rebase, I forgot to remove the about URL, I was wrong.
Should I go ahead and restrict it to about or still allow it as a catch all?

tompollard:
> Should I go ahead and restrict it to about or still allow it as a catch all?

Yeah, I think this is a good idea.

alistairewj:
Agreed!

bemoody:
>> Should I go ahead and restrict it to about or still allow it as a catch all?

> Yeah, I think this is a good idea.

Which one?

tompollard:
> > > Should I go ahead and restrict it to about or still allow it as a catch all?
> 
> > Yeah, I think this is a good idea.
> 
> Which one?

Whoops, the first one!

physionet-django/console/forms.py
```diff
@@ -760,6 +760,54 @@ def save(self):
...
+    """
+    Form for creating a dynamic static page."""
+
+    url = forms.RegexField(
+        label="URL",
+        max_length=100,
+        regex=r"^[-\w/\.~]+$",
+        help_text=(
+            "Url should be unique. If it clashes with a static url, "
+            "The static url will take precedence."
```
tompollard: Perhaps tweak slightly to:

"URL should be unique. If the new URL clashes with a static url, "
"the static url will take precedence."

Ideally, I think we should also 
- include an example of the kind of URL path that we are expecting (e.g. `/about/path`)
- specify that the path must begin with `about`.

physionet-django/console/forms.py
```diff
@@ -760,6 +760,54 @@ def save(self):
...
+        """ Validate that the new url does not exists"""
+
+        url = self.cleaned_data.get("url")
+
+        same_url = StaticPage.objects.filter(url=url)
+        if self.instance.pk:
+            same_url = same_url.exclude(pk=self.instance.pk)
+
+        if same_url.exists():
+            raise forms.ValidationError(f"Static page with url {url} already exists")
```
tompollard: Tweak to:

`Static page with URL {url} already exists.`

physionet-django/console/views.py
```diff
@@ -2363,6 +2363,54 @@ def static_pages(request):
...
+
+@permission_required('physionet.change_staticpage', raise_exception=True)
+def static_page_edit(request, page_pk):
+
+    static_page = get_object_or_404(StaticPage, pk=page_pk)
+    if request.method == 'POST':
+        static_page_form = forms.StaticPageForm(instance=static_page, data=request.POST)
+        if static_page_form.is_valid():
+            static_page_form.save()
+            messages.success(request, "The static page was successfully editted.")
```
tompollard: typo ("editted")

physionet-django/console/forms.py
```diff
@@ -760,6 +760,54 @@ def save(self):
...
 
+class StaticPageForm(forms.ModelForm):
+    """
+    Form for creating a dynamic static page."""
+
+    url = forms.RegexField(
+        label="URL",
+        max_length=100,
+        regex=r"^[-\w/\.~]+$",
+        help_text=(
```
tompollard, line 770: If possible, it would be good if the help text was also displayed on the "edit" page. Currently I only see it on the creation page.

tompollard (commented):
Thanks @harryface, this is looking good to me. I've annotated the code with a few comments, and also listed a few thoughts below:

- Do any of the fixtures need updating? i.e. are the pages/paths listed in https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/fixtures/demo-physionet.json correct?
- Currently we have an about page at https://physionet.org/about/. Will this path be tricky to deal with or can allow this by simply supporting an empty string after "/about/"?

I assume you are working on the changes so that paths are restricted to `/about/`. I'll do some testing once this is pushed.

tompollard:
Looks good to me, thanks @harryface!

PR #1695, opened 2022-10-24
Automated URL testing

bemoody:
This creates a framework for systematic testing of URL patterns.

What this means:

- The test suite will automatically load every URL defined in
  physionet/urls.py and check that it returns an HTTP status between
  200 and 399.

- When you, as a developer, define a new URL pattern in
  physionet/urls.py, you will need to define some suitable test
  arguments (which may mean that you'll need to add some suitable
  fixture data to demo-*.json.)

This currently applies to `physionet.urls`, `notification.urls`,
`search.urls`, and `export.urls`; eventually I intend to expand it to
all URLs in the site.

There are a few reasons for this, as explained in the commit message:

- ensuring that every view gets a bare minimum of testing

- ability to diff and grep the resulting html

- future use of vnu or other validators (which are not written in
  python and don't play well with the python unittest framework)

Every view *should be possible to test in this way.*  In many cases it
will require adding additional demo fixture data.  This generally
isn't difficult, we just have to work through the views one by one.

### DIFF
--- a/physionet-django/export/urls.py
+++ b/physionet-django/export/urls.py
@@ -15,3 +15,6 @@ urlpatterns = [
     path('rest/published-stats-list/', views.published_stats_list,
         name='published_stats_list'),
 ]
+
+# Parameters for testing URLs (see physionet/test_urls.py)
+TEST_DEFAULTS = {}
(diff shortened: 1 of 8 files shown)


tompollard:
This looks great to me Benjamin, thanks!

PR #1696, opened 2022-11-02
JSON export of AWS user accounts

bemoody:
This provides a JSON file of AWS user accounts that can be downloaded by administrators.

The data export code was provided by Chrystinne - I've just taken this code and turned it into a view.

One issue is that this will duplicate the list of users many times over, as there are hundreds of published projects to which *every* user has access.  Instead, you probably only want to include *restricted/credentialed* projects, and more specifically you probably only want to include projects that are *currently mirrored on AWS*.


Chrystinne:
@bemoody Thank you so much for the support! Downloading the JSON file will help a lot in uploading the users' list if needed. I already updated the code to restrict the JSON to contain only the specific datasets for the Datathon. I will push my commit with the change tomorrow when I have the final list of the datasets.

tompollard:
Looks good to me!

PR #1699, opened 2022-11-09
Display reviewer details for training in the console

tompollard:
This is a minor change to display more information in the console about training reports. Specifically:

- On the "training detail" page for training reports (e.g. http://localhost:8000/console/training/view/1/) we currently display information like submission date, valid until, etc. This pull request adds the date of the review and the reviewer.
- On the approved/valid and rejected lists, the pull request adds the date of the review and the reviewer.

PR #1700, opened 2022-11-09
Order training details by review date

tompollard:
This is a minor change to follow https://github.com/MIT-LCP/physionet-build/pull/1699. Currently the training reports are displayed in the console at http://localhost:8000/console/training/review/ in a random-ish order. This pull request:

- Displays all approved and rejected training reports in the order in which they were reviewed (newest to oldest).
- Makes some minor style changes to the training detail page at (e.g.) http://localhost:8000/console/training/view/3/


tompollard:
@briangow sneaked in a change to display datetimes as dates, just because the times aren't necessary and it looks a little cleaner.

briangow:
@tompollard looks good to me!

PR #1706, opened 2022-11-14
Add ability to add dynamic buttons to front page

harryface:
This allows for the addition of dynamic buttons to front page
- Allows ordering of the buttons from the front page index page on console.


tompollard:
Thanks @harryface, looks great to me. Once we've sorted out the migration issue I think this should be good to merge.

> I was also thinking we remove the existing hard coded ones and put them dynamically (yet to be implemented)...

Yeah, this would be nice to do. Handling the dynamically rendered content (e.g. list of projects at http://localhost:8000/about/database/) needs some thought.

bemoody:
In general, when you add a new field to a model, it must have `null=True`.  That's how we ensure that the server doesn't crash while it's being upgraded.

I'm really not convinced, though, that we want the front page buttons to be permanently tied to StaticPages.

Also, the buttons need to have an order, and reusing `nav_order` for this makes it awkward to manage.

Also, the buttons need *labels*, which are not necessarily the same as the title of the linked page.

Why not just create a new model `FrontPageButton`, which has an order, a label, and a URL?

tompollard:
> Also, the buttons need labels, which are not necessarily the same as the title of the linked page.
> Why not just create a new model FrontPageButton, which has an order, a label, and a URL?

Sounds good to me. All of the current links are to /about/ but your suggested approach gives us more flexibility.

harryface:
Sure, this gives us more flexibility.
Though for this to work, the section model needs to be adjusted to use a GenericForeignKey instead of a Foreignkey(hoping this would not create more migration issues)

I will take this approach and as well modify the static_page view to search both model.

tompollard:
Thanks @harryface 

> GenericForeignKey instead of a Foreignkey(hoping this would not create more migration issues)

Is is necessary to have a dependency on another model? It feels like we could make the  `FrontPageButton` completely independent. i.e. in the console, we have a new admin page that allows people to add front page buttons. Those buttons can link to any URL provided.

harryface:
> Thanks @harryface
> 
> > GenericForeignKey instead of a Foreignkey(hoping this would not create more migration issues)
> 
> Is is necessary to have a dependency on another model? It feels like we could make the `FrontPageButton` completely independent. i.e. in the console, we have a new admin page that allows people to add front page buttons. Those buttons can link to any URL provided.

It needs to be linked to the section model, each of the pages has multiple sections attached to them.

tompollard:
> It needs to be linked to the section model, each of the pages has multiple sections attached to them.

It might be easiest to chat about this! All we need is the ability to create a button and add a URL that it points to, isn't it? In this case, it feels like the buttons could be treated completely independently of the staticpage model.

@bemoody please could you explain what you were thinking?

bemoody:
The current front page buttons are not associated with StaticPages.  For example, `/about/database/` is not a StaticPage.  It's not a database object at all.

I don't think there is any reason to want to limit this feature to StaticPages.  If we wanted to make `/lightwave/` a front-page button, or `/content/`, or `/data/`, those would be perfectly reasonable things to do.  None of those are database objects.

So you shouldn't think about this feature as an extension of the StaticPage model (or the Section model - I don't understand what Sections have to do with any of this.)

You might be thinking that adding a field to an existing model is less work than creating a new model.  As you can see from this pull request, that's not true at all - quite the contrary!

Also, because @harryface mentioned GenericForeignKeys: *never use GenericForeignKeys.*  They are used in a few places in the code currently, and *every one of them is a mistake.*

tompollard:
> Also, because @harryface mentioned GenericForeignKeys: never use GenericForeignKeys. They are used in a few places in the code currently, and every one of them is a mistake.

I didn't have the confidence to hate on GenericForeignKeys for fear of having to defend my opinion, but I'm glad you say this!

physionet-django/physionet/views.py
```diff
@@ -28,6 +28,7 @@ def home(request):
     featured = PublishedProject.objects.filter(featured__isnull=False).order_by('featured')[:6]
     latest = PublishedProject.objects.filter(is_latest_version=True).order_by('-publish_datetime')[:6]
     news_pieces = News.objects.all().order_by('-publish_datetime')[:5]
+    fp_buttons = FrontPageButton.objects.all()
```
tompollard: as we have `front_page_banner` below, perhaps use `front_page_buttons` instead?

physionet-django/physionet/test_context_processor.py
```diff
@@ -0,0 +1,17 @@
+from django.test import TestCase
```
tompollard: The "static page" changes now seem unrelated to this pull request. Could these be moved to a new pull request?

physionet-django/physionet/context_processors.py
```diff
@@ -20,3 +21,10 @@ def platform_config(request):
         'STRAPLINE': settings.STRAPLINE,
         'SOURCE_CODE_REPOSITORY_LINK': settings.SOURCE_CODE_REPOSITORY_LINK
     }
+
+
+def static_pages(_):
```
tompollard: Perhaps move static page related improvements to a separate pull request?

physionet-django/console/templates/console/frontpage_button/index.html
```diff
@@ -0,0 +1,58 @@
+{% extends "console/base_console.html" %}
+
+{% load project_templatetags %}
+
+{% block title %}Frontpage Buttons{% endblock %}
+
+{% block content %}
+ <div class="card mb-3">
+  <div class="card-header">
+    Frontpage Buttonss <span class="badge badge-pill badge-info">{{ pages|length }}</span>
```
tompollard: Typo here ("Frontpage Buttonss")

physionet-django/console/templates/console/console_navbar.html
```diff
@@ -322,6 +322,16 @@
         </li>
       {% endif %}
 
+      <!-- front pages -->
```
tompollard: I'm not super keen on the way this appears in the console menu, but we can clean this up in a later pull request. I think probably "static page" and "frontpage buttons" could be grouped into a submenu?

tompollard (approved):
Thanks @harryface, looks good to me. This PR focuses on the front page buttons, but there is still some left over changes to static pages. Please could those be moved to a separate pull request?

tompollard:
Thanks Harry!

PR #1709, opened 2022-11-16
Change credentialing navbar to identity check

lamawmouk:
This code base changes the 'credentialing' navbar to 'identity check' for clarity about the credentialing process which is a two step process including identity check and training.


tompollard:
Thanks Lama, looks good to me. If we say "Identity check", shouldn't we also say "Training check"?

lamawmouk:
Thanks @tompollard I will update the other tab as well.

tompollard (approved):

PR #1710, opened 2022-11-16
Fix security issue with direct GCP uploads

kshalot:
`filter` does not work in place. The new queryset not being assigned means that the project will always be found and any user can change the content of any ActiveProject by generating a signed url.


tompollard:
Thanks @kshalot

PR #1711, opened 2022-11-16
Prompt for affiliation when accepting authorship invitation

bemoody:
All authors of a project have to provide their name and (one or more) affiliations.  Sometimes people forget to enter their affiliation or don't understand that it's required (remember, this person may not be the submitting author and may not know anything at all about PhysioNet or the platform.)

Putting explanatory text like "After accepting, please add your affiliation to the project author page" is better than nothing but not a great solution.

Here we try to make it simpler by ensuring that they *have* to enter an affiliation (or confirm the affiliation, if they previously set an affiliation in their user profile) as part of the same form where they accept the authorship invitation.

### DIFF
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -42,6 +42,7 @@ from project.models import (
 )
 from project.projectfiles import ProjectFiles
 from user.models import User, TrainingType
+from user.validators import validate_affiliation
 
 INVITATION_CHOICES = (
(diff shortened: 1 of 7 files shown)


bemoody:
We might also want to get rid of Author.import_profile_info (it's still used when the *submitting author* creates a new project) or make the purpose of that function clearer.

tompollard:
Thanks Benjamin, this looks great .

PR #1713, opened 2022-11-21
Login ux upgrade

superryeti:
Fixed Login Instructions under the "Login" tab on Login page

**Context**
Around 1 months ago  looks like Harry added dynamic page feature and made it so that all the static page begin with this format `/about/static-page-url`. This seems to have broken the Login instruction static page and it was no longer appearing under the SSO login page.[I found this when working on a task to add an editable Login Instructions under the SSO Login Container]

**Changes**

- I only fixed the URL to login static page and We already seem to have the  rest if the feature that's let's us add login instructions under the Login Section for SSO login. A fixture `login-instruction-static-page.json` already exists which automatically creates a static page called Login instructions.
-  To add login instructions, Admin can login to admin console and add the instruction content on the Static page called Login instructions. The content will be displayed on the login page.

### DIFF
--- a/physionet-django/physionet/fixtures/login-instruction-static-page.json
+++ b/physionet-django/physionet/fixtures/login-instruction-static-page.json
@@ -4,7 +4,7 @@
     "pk": 3,
     "fields": {
       "title": "Login Instructions",
-      "url": "/login/"
+      "url": "/about/login/"
     }
   }
 ]
--- a/physionet-django/user/views.py
+++ b/physionet-django/user/views.py
@@ -79,7 +79,7 @@ class SSOLoginView(auth_views.LoginView):
     def get_context_data(self, *args, **kwargs):
         context = super().get_context_data(*args, **kwargs)
         try:
-            login_static_page = StaticPage.objects.get(url='/login/')
+            login_static_page = StaticPage.objects.get(url='/about/login/')
             instruction_sections = Section.objects.filter(static_page=login_static_page)
         except StaticPage.DoesNotExist:
             instruction_sections = []


tompollard:
@amitupreti I think I need to chat with you about this. I don't think I understand the problem well enough to know whether the fix is working.

On the main branch, when I go to http://localhost:8000/login/ (with `ENABLE_SSO="True"` in the .env file) I see the single sign on prompt. I don't understand how static page "login instructions" could also be displayed at this URL.

superryeti:
@tompollard 

**Problem:** on `SSOLoginView` , there is a section of code that looks for a static page called login and it will pull out the content of that static page and send it to the template `sso/login.html`

This `sso/login.html` is responsible to show the SSO Login button and  also the static page with login instructions
This is currently broken as the code is looking for a static page called `/login` but it should be looking for a static page called `/about/login` (because after Harry's update all static page now start with `/about/static-page-slug` )

```
#login.html

{% block content %}
  <div class="container d-flex justify-content-center" >
    <div class="card text-center" style="width: 36rem;">
      <div class="card-body">
        <h5 class="card-title">Single Sign-On</h5>
        <h6 class="card-subtitle mb-2 text-muted">Login through an external institute</h6>
        <a type="button" class="btn btn-primary center" href="{% url 'sso_login' %}" aria-disabled="true">
          {{ sso_login_button_text }}
        </a>
      </div>
    </div>
  </div>
  {% if login_instruction_sections|length > 0 %}
    {% include "sso/sso_instruction.html" with sections=login_instruction_sections %}
  {% endif %}
  <br>
{% endblock %} 

```

```
#sso view get_context_data function

    def get_context_data(self, *args, **kwargs):
        context = super().get_context_data(*args, **kwargs)
        try:
            login_static_page = StaticPage.objects.get(url='/login/')
            instruction_sections = Section.objects.filter(static_page=login_static_page)
        except StaticPage.DoesNotExist:
            instruction_sections = []

        sso_extra_context = {
            'sso_login_button_text': settings.SSO_LOGIN_BUTTON_TEXT,
            'login_instruction_sections': instruction_sections,
        }
        return {**context, **sso_extra_context}
```


**Summary:**

**Before fix:** we cannot show login instructions under the SSO box

Before Fix


**After fix:**  we can show login instructions under the sso box(by adding content to the static page called `/about/login`)

AfterFix

tompollard:
Thanks for taking the time to explain @amitupreti. Looks good to me!

PR #1718, opened 2022-11-28
 WIP: Refactor is_admin to use proper permission and permission groups

superryeti:
Context:

There is a user.is_admin flag. In the past, this was used to control access to administrator privileges. A PR was made which added a permission system: https://github.com/MIT-LCP/physionet-build/pull/1487. is_admin flag now used to allow users to view the admin console page

The Goal of this PR is to refactor the use of is_admin by introducing  a new permission called `can_view_admin_console`. This permission is linked to groups.

Summary of changes
1. Added a new permission `can_view_admin_console`
2. Updated the fixtures to include this permission in user groups
3. Updates templates,views and forms to use the new permission or check user's access to Admin group
4. added a method on User model that checks if the current user has this new permission(the method is used in the django templates)


tompollard:
@amitupreti it looked to me like the tests are failing because they are looking for an `is_admin` flag to be set (but the flag  either isn't set or doesn't exist). It might be that the tests need to be updated?

superryeti:
Thanks @tompollard . checking that

bemoody:
It looks like you're trying to do a lot of different things here.  Please break it down and address one issue at a time.

For example, "I want a permission to control whether a person sees the admin console menu, instead of using the is_admin flag."  That would be a single issue and a single pull request.

Whether a person sees the admin console menu, whether a person can be assigned as an editor, and whether a person can approve credentialing applications (for example) are entirely different questions.  Don't try to change all of these things at once, you are just making things confusing.

Please pick one issue to address, squash your relevant changes into a single patch, and open a new pull request.

superryeti:
> 

Thank you. @bemoody  Those are valid points. I will split the PR and focus on 1 issue at a time.

superryeti:
Opened a small PR for just controlling if the user sees the admin console and its content. https://github.com/MIT-LCP/physionet-build/pull/1719

superryeti:
pushed commit on wrong branch. Please ignore

superryeti:
Closing this in favour of smaller PR.(all small PR's raised)

PR #1719, opened 2022-11-29
Refactored the use of is_admin for access to admin console in Templates

superryeti:
**Context:**

There is a user.is_admin flag. In the past, this was used to control access to administrator privileges. A PR was made which added a permission system: https://github.com/MIT-LCP/physionet-build/pull/1487. is_admin flag now used to allow users to view the admin console page

The Goal of this PR is to refactor the use of is_admin by introducing a new permission called can_view_admin_console. This permission is linked to groups.

on this PR, i only updated the templates to replace the use of is_admin with permission on templates. I will open other smaller PR's to address views and tests

Summary of changes

    Added a new permission can_view_admin_console
    Updated the fixtures to include this permission in user groups
    Updates templates to use the new permission or check user's access to Admin group
    added a method on User model that checks if the current user has this new permission(the method is used in the django templates)


physionet-django/user/fixtures/demo-user.json
```diff
@@ -14332,6 +14365,11 @@
         "view_event_menu",
         "user",
         "event"
+      ],
```
tompollard: The "Host" is not an admin role (the group simply allows people able to create events) so I don't think this group should be able to view the admin console.

tompollard (commented):
Thanks @amitupreti. I think there is one update that needs making to the fixtures (comment inline) but aside from this everything looks good to me.

physionet-django/user/fixtures/demo-user.json
superryeti: Noted. Thanks. Will make the changes to fixtures

tompollard:
Thanks!

PR #1721, opened 2022-11-30
Refactoring of GenerateSignedUrl(file upload process)

superryeti:
**Context:** Currently the file upload process allows any user who has access to admin console to upload the files. This is a security issue. We need to restrict the file upload process to only the users who have access to  either of the following roles:
1. Project Author
2. Project Editor
3. System Admin

My change will restrict the access to unauthorized users. i also added a test case that checks and verifies the access to the file upload process.

Few further concerns not addressed on this PR but something to think about (from the TRA)
1. Should we prevent the  authors/editor from uploading file if the project has been published?
2. Should we prevent the authors from uploading file after the project has been submitted for review?


tompollard:

PR #1723, opened 2022-12-01
Updated the File Upload View to make it consistent with the file upload process in UI + minor bug fix

superryeti:
**Context1:** This is a follow up to pr #1721. Currently, the file upload view doesnot check the project submission status and allows the user to upload the file even if the project is not in the submission phase where upload is permitted. The change aims to increase the security of view in case a user with access to the platform tries to maliciously upload a file to the project.

**Explanation:**
On the UI, Authors can only upload files if the project is not submitted or in revision state. and Editors can only upload files if the project is accepted.

But this is not the case with the file upload view. Before this change a malicious user can bypass the UI and upload files by sending the request directly.(See TRA report for more details)

**Context 2(minor bug):** It seems sometime, django filters with return duplicate objects when we have or filter, Q and one to many relation This causes the file upload view to throw a `MultipleObjectsReturned` exception on the get_object_or_404 call.So adding a distinct() call to the queryset fixes the issue. Here is a related issue on stackoverflow: https://stackoverflow.com/questions/34468505/why-is-django-queryset-with-q-expression-returning-duplicate-values

To replicate the bug, please call the generate_signed_url view for project ` Demo software project` by user `tpollard@mit.edu`
In Summary(context2): `get_object_or_404(queryset, slug=project_slug)` throws `MultipleObjectsReturned` error because the output of `queryset.filter(Q(authors__user=request.user) | Q(editor=request.user))` had duplicate projects

Note: Although it would have been a good idea to open 2 PR's, i did both changes in single PR because on of the test case(test_invalid_access) i added will fail with out the bug fix

### DIFF
--- a/physionet-django/project/modelcomponents/activeproject.py
+++ b/physionet-django/project/modelcomponents/activeproject.py
@@ -605,3 +605,20 @@ class ActiveProject(Metadata, UnpublishedProject, SubmissionInfo):
         ProjectFiles().publish_complete(self, published_project)
 
         return published_project
+
+
+    def is_editable_by(self, user):
+        """
+        Whether the user can edit the project
+        """
+        if not user.has_perm('project.change_activeproject'):
+            project_authors = [author.user for author in self.author_list()]
+            if user in project_authors:
+                if user != self.submitting_author().user or not self.author_editable():
+                    return False
+            elif user == self.editor:
+                if not self.copyeditable():
+                    return False
+            else:
+                return False
+        return True
--- a/physionet-django/project/modelcomponents/activeproject.py
+++ b/physionet-django/project/modelcomponents/activeproject.py
@@ -606,7 +606,6 @@ class ActiveProject(Metadata, UnpublishedProject, SubmissionInfo):
 
         return published_project
 
-
     def is_editable_by(self, user):
         """
         Whether the user can edit the project
--- a/physionet-django/project/test_views.py
+++ b/physionet-django/project/test_views.py
@@ -1228,8 +1228,8 @@ class TestGenerateSignedUrl(TestMixin):
                         }
                         ),
                 self.valid_data,
-                format='json'
-                )
+                format='json')
+
             self.assertEqual(response.status_code, HTTPStatus.FORBIDDEN)
 
         # awaiting editor decision
--- a/physionet-django/project/test_views.py
+++ b/physionet-django/project/test_views.py
@@ -1219,12 +1219,12 @@ class TestGenerateSignedUrl(TestMixin):
         self.client.login(**self.user_credentials)
 
         # Non-submitting author
-        self.client.login(username='aewj@mit.edu', password='Tester11!')
+        self.client.login(username='george@mit.edu', password='Tester11!')
         with self.subTest('Non Submitting author can not upload files.'):
             response = self.client.post(
                 reverse('generate_signed_url',
                         kwargs={
-                            "project_slug": ActiveProject.objects.get(title='MIMIC-III Clinical Database').slug
+                            "project_slug": ActiveProject.objects.get(title='Demo software project').slug
                         }
                         ),
                 self.valid_data,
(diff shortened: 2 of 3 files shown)


superryeti:
Update: Instead of using distinct() on the queryset, i made changes to the workflow(because there was another error caused with the use of distinct(), so instead of chasing down the rabbit hole, i tried to make it simple)

Context:
I tried to fix the error(link is attached below) that appeared when using distinct() in queryset. I couldn't fix it/narrow its cause, so i changed the workflow to avoid the error entirely.
At the end the view works same as before(with 1 improvement, now in case the user is not authorized we will throw 403 error, earlier for all the cases such as project not found, unauthorized user we used to send 404 error)
https://github.com/MIT-LCP/physionet-build/actions/runs/3593868625/jobs/6051387850

Explanation of change:
Earlier, we were listing all the projects the current user has access to as either author or editor, and then trying to find the project.
Now we are first checking if project exists, if it exists, then we check if the current user has access to it.

To make sure, this change doesn't break anything, i tested by running the test, and also tried to come up with scenarios that this change might cause a bug(i couldn't think of any)

bemoody:
Where did this come from?

> if not request.user.groups.filter(name='Admin').exists():

This shouldn't be here.  Python code shouldn't be looking for user groups with particular names.

There should be a specific *permission* for "this person is allowed to edit files of active projects other than their own", and perhaps another permission for "this person is allowed to edit files of published projects" (though arguably we don't want to permit that for anybody.)

superryeti:
> Where did this come from?
> 
> > if not request.user.groups.filter(name='Admin').exists():
> 
> This shouldn't be here. Python code shouldn't be looking for user groups with particular names.
> 
> There should be a specific _permission_ for "this person is allowed to edit files of active projects other than their own", and perhaps another permission for "this person is allowed to edit files of published projects" (though arguably we don't want to permit that for anybody.)

Noted. The code before this was checking for is_admin, So i thought we would want to check the same thing(i.e if user belongs to the Admin group).  My assumption was that this code is from before the time we decided use is_admin to just check for access to admin console.

But i got your point of checking if the user has change_access to ActiveProject. We already have a permission. i will use that and push a commit.)

bemoody:
In any event, we don't want to duplicate the logic for deciding if the project is editable.  This function should probably be calling the `author_editable` and/or `copyeditable` functions, or maybe there should be a single function in `ActiveProject` that's used by this function as well as by `project_auth`.

(Actually I'm not sure why `generate_signed_url` isn't using `project_auth` itself.)

superryeti:
> In any event, we don't want to duplicate the logic for deciding if the project is editable. This function should probably be calling the `author_editable` and/or `copyeditable` functions, or maybe there should be a single function in `ActiveProject` that's used by this function as well as by `project_auth`.

That's a very good point. 

let me look into this.

superryeti:
@bemoody  i made the suggested changes but opted out of generalizing the logic for deciding if the project is editable at the moment

I opted out of adding a function that checks if the current user can edit the project because i think that function is not very useful to file upload process as uploads are different compared to editing abstract, or other project details.
For example for uploads only submitting author can upload files, but outside of uploads, any author can edit the project.

Also i opted out of making changes to project_auth, because i have another PR which is changing the project_auth to work with the new permission we added. So it would probably get confusing to merge or maybe the tests will fail when we merge the changes

bemoody:
> For example for uploads only submitting author can upload files, but outside
> of uploads, any author can edit the project.

That doesn't sound right to me.  We've talked in the past about the
possibility of shared editing, or allowing different people to edit
text vs. file content, but AFAIK at present, only the submitting
author is allowed to edit either the text content or the file content.

Allowing multiple people to edit the text would require some mechanism
for resolving conflicts, which is a hard problem!  Have I missed
something?

superryeti:
@bemoody  i think i might have misunderstood.

to confirm if i understood,among authors of project, only submitting author can make changes to project content(edit/update), right?

bemoody:
Yes, that's right, and I think that's the safest thing to do.

(If we do want to allow multiple authors to upload I would rather have
that be explicitly opt-in.)

superryeti:
Didn't update the `project_auth` decorator. as i am not sure how much it might impact other parts of project. I think its best to open another PR for project_auth if needed, so that we dont overlook anything

tompollard:
Thanks Amit!

PR #1725, opened 2022-12-01
Refactoring of is_admin to use appropriate permission on views for all Apps

superryeti:
Context: This is a simplification of #1718 and a follow up to #1719

The current changes summary are:
1. replace use of is_admin with appropriate permission on views
2. Fix test case to work with the changes
3. Update the User model to create helper functions to create_admin(this was needed as one of the existing test case creates an Admin user and uses that user to access training)


physionet-django/user/views.py
```diff
@@ -766,7 +766,7 @@ def training_report(request, training_id):
     Serve a training report file
     """
     trainings = Training.objects.all()
-    if not request.user.is_admin:
+    if not request.user.has_perm('user.change_credentialapplication'):
```
tompollard, line 769: Not sure it really matters, but should this be `user.change_training`?

physionet-django/console/views.py
```diff
@@ -1667,7 +1667,7 @@ def training_process(request, pk):
 
 
 @login_required
-@user_passes_test(is_admin, redirect_field_name='project_home')
+@user_passes_test(has_change_credentialapplication_permission, redirect_field_name='project_home')
```
tompollard, line 1670: Should this be checking something like `user.change_training_permission`?

physionet-django/console/views.py
```diff
@@ -1590,7 +1590,7 @@ def search_training_applications(request, display_training):
...
```
tompollard, line 1593: Should this be checking something like `user.change_training_permission`?

physionet-django/console/views.py
```diff
@@ -1513,7 +1513,7 @@ def credentialed_user_info(request, username):
...
```
tompollard, line 1516: Should this be checking something like `user.change_training_permission`?

tompollard (commented):
Thanks, looks good to me! I think we have a permission specifically for training. There are a few places where I wondered whether we should using this instead of the credentialing permission?

physionet-django/user/views.py
superryeti: Thank you.
Thats a good point. I remember we were using credentialing along with training, and want to separate it by adding new permission.

Do you think it would be a good idea to add the permission for training in this PR? as we would have to go ahead and make changes to templates and maybe few tests. Would it make more sense to open a separate PR for the training permissions?

I remember Ben suggested to not introduce multiple changes in a single PR and keep PR's short.

Please let me know what you recommend @tompollard , i can then (1)  push changes here or (2) open another PR after this or (3) open a PR side by side and squash some of the changes on training here.

tompollard: > Do you think it would be a good idea to add the permission for training in this PR? 

Hmm, does the permission not already exist? If not, then I think it is best to leave it like this for now. 

In our live database the `user. change_training` exists, but maybe this is a default that was added when the training model was introduced.

superryeti: Oh sorry, i missed that. let me make changes to use it. Thank you :)

tompollard: Okay, looking at the model it seems that the training permissions do not currently exist because we set `default_permissions = ()`.

```
class Training(models.Model):
    ...
    class Meta:
        default_permissions = ()
```

See: https://docs.djangoproject.com/en/4.1/ref/models/options/#default-permissions for an explanation. I'm not sure why the default permissions (add, change, delete, view) appear on our live server, but I guess we should assume that they don't exist for now.

superryeti: Noted. resolving Comments in favour of opening another PR for trainings

physionet-django/console/views.py
superryeti: Thank you. resolving Comments in favour of opening another PR for trainings(After discussion with Tom)

tompollard:
After our discussion, I think this is good to merge. It turns out that the training permissions don't currently exist (see: https://github.com/MIT-LCP/physionet-build/pull/1725#discussion_r1041287631). We could perhaps add training permissions in a new PR.

physionet-django/console/views.py
superryeti: Thank you. resolving Comments in favour of opening another PR for trainings(After discussion with Tom)

physionet-django/project/test_views.py
```diff
@@ -1233,12 +1233,12 @@ def test_invalid_access(self):
             self.assertEqual(response.status_code, HTTPStatus.FORBIDDEN)
 
         # awaiting editor decision
-        self.client.login(username='tpollard@mit.edu', password='Tester11!')
+        self.client.login(username='rgmark@mit.edu', password='Tester11!')
```
tompollard, line 1236: Sorry @amitupreti, please could you explain why it is necessary to change the test?

tompollard: Just noticed your comment on Slack ("the test was failing because one of the test user had admin permission and had direct access to the project edit"). Makes sense!

superryeti: Hi, So the user tpollard previously had is_admin permission, and after we added the permission for console. I had updated the demo-user.json(fixtures) to add the tpollard user to Admin group.

To followup with the new permission, i update the views in PR #1723 to fix the file upload process(which was also using is_admin flag), for that change i had created a generic function is_editable_by which checks if the current user should be allowed to edit the project.(and that function lets anyone with `project.change_project` permission change the project anytime).

So the user tpollard had the `project.change_project` permission, and the test was failing as the test was trying to make sure that editor should not be able to make changes to project(but because tpollard had the `project.change_project` permission, he could edit the project and the test failed)

PR #1727, opened 2022-12-03
Added a new permission to determine who gets to be assigned as a project editor and Supporting Changes

superryeti:
Context:
Previously anyone who had is_admin flag was able to be assigned as project editor.  Following #1719, we are removing the use of is_admin flag. This permission are assigned to relevant groups, and will be used to determine who can be assigned as project editor. If for a  deployment of PhysioNet has a different set of groups(or groups have been), they just need to add the new permission to the relevant groups.

Changes
1. Added a custom permission on activeproject `can_edit_activeprojects`
2. Fixed the forms to check for the users who have this permission. Only those users can be assigned as project editors.
3. Fixed test cases as with the test, the old fixtures were not working(as the logic of who can be assigned as project editor has changed). So, I have updated the fixtures to add a new user with correct permission, and updated an existing user with permission.
4. Updated fixtures to Provide both Editor groups ability to be assigned as project editors.

NOTE: On the active deployments, The system administrator will need to update the groups with the new permission, so that editors can be assigned.


bemoody:
You cannot (and you don't want to) evaluate any database queries at the time 'physionet-django/console/forms.py' is imported.  (The fact that this breaks 'resetdb' is just one symptom of a much larger problem.)  The query needs to be performed in `__init__` instead.

Look at AssociatedEmailChoiceForm for an example of how to use a ModelChoiceField with a QuerySet that is created at runtime.

I don't know if django.contrib.auth provides any function to list (or obtain a QuerySet for) "all users who have a specified permission".  If that function doesn't exist, you should write a function to do that.

superryeti:
Thank you so much. This was giving me a lot of headache. @bemoody

superryeti:
Waiting for #1725 to merge as this PR has the same test failure after rebasing as #1725.

superryeti:
Created a new PR because i messed this PR when trying to Rebase 

https://github.com/MIT-LCP/physionet-build/pull/1741

PR #1728, opened 2022-12-06
Update the is_editable_by_func to make sure only authorized user can edit the project

superryeti:
After discussion with Tom, it looks like only submitting author, editors should be able to edit the project(when the project is on correct submission status i.e is before submission and during revision for author; and copyedit stage for editor) The previous implementation was not correct, it let Admin make changes to project even when they were not assigned to project

### DIFF
--- a/physionet-django/project/modelcomponents/activeproject.py
+++ b/physionet-django/project/modelcomponents/activeproject.py
@@ -610,7 +610,7 @@ class ActiveProject(Metadata, UnpublishedProject, SubmissionInfo):
         """
         Whether the user can edit the project
         """
-        if user == self.submitting_author().user:
+        if self.authors.filter(is_submitting=True, user=user).exists():
             if self.author_editable():
                 return True
         elif user == self.editor:
--- a/physionet-django/project/modelcomponents/activeproject.py
+++ b/physionet-django/project/modelcomponents/activeproject.py
@@ -611,11 +611,9 @@ class ActiveProject(Metadata, UnpublishedProject, SubmissionInfo):
         Whether the user can edit the project
         """
         if user == self.submitting_author().user:
-            if not self.author_editable():
-                return False
+            if self.author_editable():
+                return True
         elif user == self.editor:
-            if not self.copyeditable():
-                return False
-        else:
-            return False
-        return True
+            if self.copyeditable():
+                return True
+        return False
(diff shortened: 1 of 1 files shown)


physionet-django/project/modelcomponents/activeproject.py
```diff
@@ -610,14 +610,12 @@ def is_editable_by(self, user):
...
-            else:
+        if user == self.submitting_author().user:
+            if not self.author_editable():
+                return False
+        elif user == self.editor:
+            if not self.copyeditable():
                 return False
+        else:
+            return False
         return True
```
tompollard: Can this final `return True` be removed? Doesn't look like it would ever get called.

superryeti: Ah that's right. How stupid i missed that. As backup, normally IDE would highlight stuff like this.

let me push the change.

superryeti: let me check again, it might be reachable now that i look at it. and probably needed

superryeti: @tompollard  I think we need it as the logic is that

If the user is  submitting author and project is on the right stage ---> allow edit
or if the user is editor and project is on right stage ---> allow edit
otherwise dont allow edit


and the block of code should return False if the right conditions are not met. and if all conditions are met return true.
I think if i refactor the func like this, it will be more readable

```
        if user == self.submitting_author().user:
            if self.author_editable():
                return True
        elif user == self.editor:
            if self.copyeditable():
                return True
        return False
```

or 

```
        if user == self.submitting_author().user:
            if self.author_editable():
                return True
        elif user == self.editor:
            if self.copyeditable():
                return True
        else:
            return False
```

superryeti: Let me know what you think. i can submit the change

bemoody:
I think you should be able to do something like

    if self.authors.filter(is_submitting=True, user=user).exists():

which is less readable but more efficient.

superryeti:
> I think you should be able to do something like
> 
> ```
> if self.authors.filter(is_submitting=True, user=user).exists():
> ```
> 
> which is less readable but more efficient.

Thank you. let me update.  is it because this(your suggestion) would generate a query that directly does the check in SQL(database level), while the previous one pulled information from the database, and checked it in python/back-end(which definitely inefficient) ?

bemoody:
Yes, exactly.  Normally not a big deal but in this case the function might be called very frequently to handle bulk file uploads.

superryeti:
That makes sense. Thank you. Really Appreciate your feedback's.

bemoody:
It looks good to me.  I don't have a strong feeling about whether admins should be able to edit but if @tompollard is happy then I am.

tompollard:
Looks good to me, thanks!

PR #1729, opened 2022-12-06
Adds edit events functionality on events home page for active events.

danamouk:
Hosts have the ability to add events when creating new events. This pull request adds an edit event functionality to active events to enable editing of the form content including Event Name, Description, Start Date, End Date, Category and Allowed fields.


bemoody:
This seems incomplete.

Really, though, the events code is rather scattered at the moment, and I think it would be better to work on getting things better organized *before* trying to add new features such as this.

tompollard:
Superseded by https://github.com/MIT-LCP/physionet-build/pull/1826

PR #1733, opened 2022-12-07
Fix some broken demo projects

bemoody:
Fix some demo projects to be consistent with how the system actually works.


tompollard:
Looks good to me, thanks!

Issue #1623, opened 2022-07-06
Wrong corresponding email when creating a new version


bemoody:
physionet-django/project/forms.py line 418:
```
        # Copy over the author/affiliation objects
        for p_author in self.latest_project.authors.all():
            author = Author.objects.create(project=project, user=p_author.user,
                display_order=p_author.display_order,
                is_submitting=p_author.is_submitting,
                is_corresponding=p_author.is_corresponding,
                corresponding_email=self.user.get_primary_email(),)
```

This is bad.  `p_author.user` is not the same person as `self.user`.

Currently there are 63 `Author`s whose `corresponding_email.user` does not equal `user`.  I don't know how many published projects there are where the `contact.email` and/or the corresponding author `corresponding_email` are set to somebody else's email address.

There should be something (maybe a "clean" function? maybe a database constraint?) to prevent you from saving an `Author` where `self.corresponding_email is not None and self.corresponding_email.user != self.user`.

bemoody:
There are currently 21 PublishedAuthors whose email address is not an address belonging to that User.

Three with None (issue #1159/#1419), four that might plausibly be an older email address for the same person, two that are the same as their current address but capitalized differently.  Twelve that appear to be a completely different person.

bemoody:
```
import project.models
import user.models
pps = project.models.PublishedProject.objects.all()
for pp in pps:
    if pp.is_legacy or pp.slug.startswith('challenge-'):
        continue
    co = pp.contact
    ca = pp.corresponding_author()
    coes = user.models.AssociatedEmail.objects.filter(email__iexact=co.email).first()
    caes = user.models.AssociatedEmail.objects.filter(email__iexact=ca.corresponding_email).first()
    if co.email != ca.corresponding_email or not coes or not caes or coes.user != ca.user or caes.user != ca.user:
        print("https://physionet.org/content/%s/%s/" % (pp.slug, pp.version))
        print("  %20s = %s | %s" % ('<contact>', co.name, co.email))
        print("  %20s = %s | %s" % ('<author>', ca.get_full_name(), ca.corresponding_email))
        for em in ca.user.associated_emails.all():
             print("  %20s = %s | %s %s%s" % (ca.user.username, ca.user.get_full_name(), em.email, ' *'[em.is_primary_email], ' P'[em.is_public]))
        if coes:
            print("  %20s = %s | %s" % (coes.user.username, coes.user.get_full_name(), coes.email))
        if caes:
[...]

Issue #1419, opened 2021-10-27
Error when publishing new version of project: 'NoneType' object has no attribute 'email'


tompollard:
The following error was raised when attempting to publish the new version of a project by clicking "Publish submission" at https://physionet.org/console/submitted-projects/SLUG/publish/:

```
Internal Server Error: /console/submitted-projects/SLUG/publish/

AttributeError at /console/submitted-projects/SLUG/publish/
'NoneType' object has no attribute 'email'

Traceback:

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/handlers/exception.py" in inner
 34.             response = get_response(request)

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
 115.                 response = self.process_exception_by_middleware(e, request)

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
 113.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/physionet/python-env/physionet/lib/python3.7/site-packages/django/contrib/auth/decorators.py" in _wrapped_view
 21.                 return view_func(request, *args, **kwargs)

File "./console/views.py" in handling_view
 90.                 return base_view(request, *args, **kwargs)

File "./console/views.py" in publish_submission
 535.                 make_zip=int(publish_form.cleaned_data['make_zip']))

File "./project/modelcomponents/activeproject.py" in publish
 533.                         published_author.corresponding_email = author.corresponding_email.email

Exception Type: AttributeError at /console/submitted-projects/SLUG/publish/
Exception Value: 'NoneType' object has no attribute 'email'
```

bemoody:
At a guess, somebody deleted their associated email address while that
address was still marked as corresponding for an active project?

At a basic level, if the corresponding author has a null corresponding
email for any reason, then check_integrity() should fail and it
shouldn't be possible to submit or publish.

But the weird relationship between user-associated-email and
author-corresponding-email and project-corresponding-email is all very
messy and confusing.

Possibly the (corresponding) author should be able to change their
corresponding address up until the time the project is published, or at
least up until the time they approve the project for publication.

If the project is in a state where the author isn't allowed to change
their address, then they also shouldn't be allowed to *delete* that
address from their account.

(Possibly authors shouldn't be allowed to delete their former
corresponding address even after publication.)

tompollard:
[...]

Issue #1159, opened 2020-11-10
Inconsistencies in handling the corresponding email address


bemoody:
You can't change your corresponding email address while a project is under submission, but you can change your primary address, and that changes the corresponding addresses for all your projects.

You can delete the address marked as corresponding, if it's not your primary address, at which point the corresponding address becomes "None".  The project can still be submitted in this state.  I think it will crash if the editor tries to publish the project in this state but I haven't checked.

Issue #1158, opened 2020-11-10
Better selection of corresponding email address


bemoody:
Currently:

- when creating a new project, your primary email address is used as the corresponding address.

- when invited as a coauthor, your primary address is used as the corresponding address.

- when creating a new version of a published project, each author's current primary address is used as their new corresponding address.

(All authors have a "corresponding address" but this doesn't have any effect except for the corresponding author.)

If someone has selected a different email address as "public" rather than "primary", the public one should be preferred by default.

If someone is invited as a coauthor and they don't have a public address, it might be better to use the invitation address rather than the primary address.

When creating a new version of a project, with the same corresponding author, it would probably be better to use their previous corresponding address rather than their current public (or primary) address.

PR #1735, opened 2022-12-07
Set correct corresponding email when creating a new version

bemoody:
If you are the submitting author of a project and not the corresponding author, then when you create a new version, the new version's corresponding author will have the name of the original corresponding author, and *the submitting author's primary email address*.  This is issue #1623.

This pull also prevents such broken projects from being submitted/published.  This is issue #1419.

There are 21 published projects, and likely many active projects as well, with nonsensical corresponding authors, which will have to be fixed manually.

This does **not** fix the other problems with corresponding emails, such as the initial selection of corresponding email (#1158) or the fact that people can delete addresses while the project is in submission (#1159).


tompollard:
Thanks Benjamin, looks good!

PR #1741, opened 2022-12-09
Added a new permission to determine who gets to be assigned as a project editor and Supporting Changes

superryeti:
SAME PR as https://github.com/MIT-LCP/physionet-build/pull/1727 . I closed the previous PR because i messed up the PR when trying to rebase my changes.

Context:
Previously anyone who had is_admin flag was able to be assigned as project editor. Following https://github.com/MIT-LCP/physionet-build/pull/1719, we are removing the use of is_admin flag. This permission are assigned to relevant groups, and will be used to determine who can be assigned as project editor. If for a deployment of PhysioNet has a different set of groups(or groups have been), they just need to add the new permission to the relevant groups.

Changes

    Added a custom permission on activeproject can_edit_activeprojects
    Fixed the forms to check for the users who have this permission. Only those users can be assigned as project editors.
    Fixed test cases as with the test, the old fixtures were not working(as the logic of who can be assigned as project editor has changed). So, I have updated the fixtures to add a new user with correct permission, and updated an existing user with permission.
    Updated fixtures to Provide both Editor groups ability to be assigned as project editors.

NOTE: On the active deployments, The system administrator will need to update the groups with the new permission, so that editors can be assigned.


superryeti:
I am not sure how to fix this. I tried to avoid making any unnecessary changes, starting a fresh PR from the recent dev branch


Its for sure got to do with the form because i had added a custom permission in one of my previous merged PRs. 

The only difference on this PR is that we have edited the form, and all the tests that are failing on the upgrade are the ones that are using the modified forms.

Error Breakdown:
The error says something like `TypeError: 'NoneType' object is not subscriptable` and the code is trying to sort the users(i assume maybe when the form is instantiated, it is trying to load the list of users but failing)
So the reason for no users found could probably be because there are no users who have this new permission(for whatever reason the resetdb, loaddemo step might not have worked properly),But i can't find out why.


The resetdb is running the makemigrations command and the permission should be created in the database.

@bemoody  could you please help? Thank you

superryeti:
I also tried to run the test on my local setup

./test-upgrade.sh -v origin/production

but get this error 
image

bemoody:
OK, so the bug here is in physionet.db.chaotic.compiler.  It's not immediately obvious why django.db.models.sql.compiler does what it does, and the way it does so is positively wacky.  But, y'know, it's not clear that the kludge we're using here was really intended to work in the first place.

>> virtualenv: No such file or directory

You need to install virtualenv.

bemoody:
Incidentally - not strictly related to this pull request - there should be an `order_by` applied to the list of editors in AssignEditorForm.  Currently, the form just lists usernames, so sorting by username would make sense.

But ideally I'd rather display each editor's full name, or full name and email address, and sort by name.

superryeti:
Regarding the compiler, 

I was trying to debug this locally, and i found that on the `django/db/models/sql/compiler.py, line 407,` the code was failing because it is trying to look for `order_by` stuff like `ASC or DESC` in the SQL query, but the sql query that was causing this error was `RANDOM()`

To debug, i kind of commented some lines that installed the pip requirements in the .test_upgrade.sh file, and modified the  `django/db/models/sql/compiler.py` to add my own print statements.( i know its the worst way to debug, but it was faster this way then to figure out how to trigger the execution of the compiler and do the steps in the upgrade_test)


So the RANDOM() is not from the django but from our  `physionet-django/physionet/db/chaotic/compiler.py` line 127.

Because we have overridden this method `get_order_by`, this method is later called a bunch of times in the `django/db/models/sql/compiler.py`. So i guess the error is throwing because we added couple of Extra SQL statements `RANDOM()`.

The function below is from  `django/db/models/sql/compiler.py", line 398 to 407`, which is trying to find if there are any additional select clauses on the SQL statement by using regular express(which looks for `ASC` or `DESC`

```

    def get_extra_select(self, order_by, select):
        extra_select = []
        if self.query.distinct and not self.query.distinct_fields:
            select_sql = [t[1] for t in select]
            for expr, (sql, params, is_ref) in order_by:
                without_ordering = self.ordering_parts.search(sql)[1]
                if not is_ref and (without_ordering, params) not in select_sql:
                    extra_select.append((expr, (without_ordering, params), None))
        return extra_select

```

---------------------
From what i understand the purpose of this compiler is  update for sqlite3 that adds a random ordering to the results of every SELECT query. By default, sqlite3 and the standard Django sqlite3 backend return rows in a predictable order, typically the same order that the rows were inserted. So hope is that this will make tests consistent with other databases which donot return results in an order by default.

In summary, the compiler stuff is supposed to add Random at the end of sql queries to randomize results if there is no order_by specified.


Questions
1. since we have an order_by specified in our form, the random thing should not be added to the query anyway right?(Seems like the implementation of this randomizing results if order not specified is not correct)
2. I tried to log all the values of result in `physionet-django/physionet/db/chaotic/compiler.py` and it looks like the Random is added to all queries. Here is the result for our order_by of users by username

```results =   [(OrderBy(Col(user_user, user.User.username), descending=False), ('"user_user"."username" ASC', [], False))]

results  after adding the random by the compiler =   [(OrderBy(Col(user_user, user.User.username), descending=False), ('"user_user"."username" ASC', [], False)), (OrderBy(Random(), descending=False), ('RANDOM()', [], False))]
```


So my thought was the django compiler code is most likely not the one to blame, to me it looks like 

we should modify compiler this way to fix the order_by crashing the test

```
    def get_order_by(self):
        result = super().get_order_by()
        result.append((OrderBy(Random()), ('RANDOM()', [], False)))
        return result
```

```
    def get_order_by(self):
        result = super().get_order_by()
        if len(result) == 0:
            result.append((OrderBy(Random()), ('RANDOM()', [], False)))
        return result
```

This kind seems to fix the error from before,

but i still dont understand why we got the error specifically on the code in forms for the parts we made changes  on this PR?(because there are a lot of other cases where we have order_by specified and alot of other case with no order_by specified in queries) 

if this was an issue, should this have happened before?

bemoody:
We didn't see this error before because nothing else uses `distinct`.

Tom must have run into this previously, because he made a comment in dbc6649498d8b74889097e29e8c9544335d8bd86 that "sqlite doesn't support distinct()".  But the problem isn't sqlite, it's our wrapper.  If you switch the "engine" in `physionet/settings/development/sqlite.py` back to `'django.db.backends.sqlite3'`, then distinct queries should work just fine.

Of course, that code also shows an alternative to `distinct` - load the results into a python set.  That wouldn't quite work in this situation, though, and I'd rather solve the underlying problem now that we know what it is. :)

Your solution, unfortunately, doesn't solve the problem completely, as you'll see if you try running `str(user.models.User.objects.all().distinct().query)`.

The issue is really independent of this pull request, so I'm planning to make a separate pull request to fix it, once I have time to test it.

*This* pull request looks fine to me as-is, and we should be able to merge it after the sqlite issue is fixed.  I see that this is still marked as "draft" - is there anything else that you intended to change?

superryeti:
Thank you. That was the only part i was confused. Let me run the tests locally by switching the db engine for sqlite for development and make the pr ready for review.

Really appreciate your help

tompollard:
@bemoody I'd like to merge this after a final check. I'm assuming that I don't need to wait for your `distinct` fix before doing so?

tompollard:
@bemoody okay to merge or prefer to wait for your fix?

bemoody:
OK to merge.

tompollard:
@amitupreti please could you raise an issue to remind us that we need to implement a fix for the sqlite3 backend problem?

superryeti:
@tompollard  Thank you. I will do that

superryeti:
@tompollard  i think we already have one https://github.com/MIT-LCP/physionet-build/issues/1743

tompollard:
Thanks Amit, in that case we are all good!

Issue #1747, opened 2022-12-14
Author guidelines section update


lamawmouk:
As part of the author guidelines, the Background section needs to be adjusted to remove the potential reuse of the data as it is duplicated in the Usage notes section which contains the potential use of the data.

PR #1749, opened 2022-12-14
Updates the author guidelines.

lamawmouk:
This fixes the issue #1747. This pull request updates the author guidelines to prompt users to read sections carefully and updates the background section to avoid duplication in the usage notes section of the project.


tompollard:
Thanks @lamawmouk, your changes look good, but this template is no longer used (it needs to be removed!). If you test on your local server, you should see that the changes that you make to this page aren't rendered. 

Instead, static pages like this are stored in the database and are editable from https://localhost:8000/console/static-pages/.  I have incorporated your edits into the live site.

PR #1751, opened 2022-12-15
Remove templates for about pages that are now handled as "static pages"

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/pull/1749, pages such as /about/publish/ and /about/ are now handled in the database using the "StaticPages" model.

There are several hardcoded templates left in the codebase that are no longer needed. This pull request removes three of them: 

- physionet-django/templates/about/about.html
- physionet-django/templates/about/author_guidelines.html
- physionet-django/templates/about/share.html


lamawmouk:
Thanks @tompollard looks good. I have checked that the static pages in the database hold the information for the three files- about, share, and author guidelines and these files are removed.
about_sharee
console_static

briangow:
Thanks @tompollard , looks good!

PR #1752, opened 2022-12-15
Remove hardcoded "quality guidelines"

tompollard:
When an author is submitting a project, they see a section titled "Quality guidelines". 

This quality guidelines are repeated twice, once at http://localhost:8000/projects/SLUG/overview/ and once at http://localhost:8000/projects/SLUG/proofread/

The guidelines differ by project type (model, data, software, etc). The content is stored in a hardcoded template file. An example is shown in the screenshot below:

---

Screen Shot 2022-12-15 at 4 14 33 PM

---

I think it makes sense to remove this section for several reasons, including:
1. we already refer people to the author guidelines, which contain a more detailed version of the recommendations;
2. the important parts appear in the submission process in more relevant places, e.g. on the file upload page;
3. this kind of hardcoded content is tricky to update, so it becomes stale quickly;
4. the hardcoded templates are inflexible, making it trickier to introduce improvements like managing data types using the database

### DIFF
--- a/physionet-django/project/templates/project/project_overview.html
+++ b/physionet-django/project/templates/project/project_overview.html
@@ -37,14 +37,6 @@
    <a href="{% url 'static_view' static_url='publish' %}#author_guidelines"> author
 guidelines</a>.</p>
 <hr>
-<h3>Quality Guidelines</h3>
-<p>Your contribution must satisfy the following guidelines:</p>
-{% if project.resource_type.id == 0 %}
-  {% include "about/data_quality_guidelines.html" %}
-{% else %}
-  {% include "about/software_quality_guidelines.html" %}
-{% endif %}
-<hr>
 {% if is_submitting %}
   <div class="btn-container-rsp btn-left">
     <button type="button" class="btn btn-danger btn-rsp" data-toggle="modal" data-target="#delete-project-modal" {% if under_submission %}disabled title="Cannot delete while under submission"{% endif %}>Delete Project</button>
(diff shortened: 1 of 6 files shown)


briangow:
@tompollard , I see what you're saying but I also think it would be nice to show a summary of key reminders before submission, at least conceptually. Having said that, I'm not sure if the current Quality Guidelines are capturing anything that isn't being clearly presented elsewhere in the submission process. 

I'm fine with merging this this but would support creating an improved version of the Quality Guidelines in the future (especially if we can find a way to do it in a sustainable manner).

tompollard:
Thanks for taking a look!

> I see what you're saying but I also think it would be nice to show a summary of key reminders before submission, at least conceptually

I basically agree, but right now I don't the hardcoded text is saying anything particularly worthwhile (other than things that are already highlighted during the process).

- "The data files are in an open format" and "No PHI is contained" are both displayed on the upload page (and these issues are being more directly addressed in the upload form under development).
- "The content is suitable for XX" is a pretty pointless thing to say.
- "All the information needed for reuse" is kind of empty too.
- etc.

If we are going to display a checklist, then one option would be to pull out the fields that we check for in the reviewer form [edit: I mean we have a reviewer checklist. We could display the reviewer checklist].

I added a new issue at https://github.com/MIT-LCP/physionet-build/issues/1753 with the idea that we could implement an improved version of these questions.

briangow:
@tompollard , I agree, I think some of the points in the Quality Guidelines are too general to be helpful. The others are being covered elsewhere. 

Your suggestions in #1753 make sense to me.

tompollard:
Thanks!

Issue #1734, opened 2022-12-07
I01: Limit number of emails attached to an account / make the limit configurable


superryeti:
Currently, Users are able to add a large number of associated email to their account,  On Test Cycura showed an example of adding 100 associated emails to account.

It  might make sense to limit a max number of emails that can be associated to user(maybe we can have a something in env file for this which can let each deployment decide if they want to implement the email limit.

bemoody:
Ten should be enough for anybody, right? :P

Well, I could imagine wanting to set up a separate contact address for each project, but nobody's done that as far as I know.

superryeti:
Default 10 makes sense to me.

Also in case if someone needs to associate a lot of emails because they have a lot of projects published, i suppose we could make it possible by using some kind of integer somewhere in the User model(so when user signs up, they would get the default email limit), and if someone needs to add more email as this is a special case, maybe the user could ask admin to increase that number for them.

But maybe it will be too much work and something we dont want to pursue.


Please let me know if you think we want to have the ability to update the max email quota for specific users. or just implement the hard email limit for all users.

superryeti:
We decided to not make changes to model and add a simple check on the views to limit users from adding unlimited emails

PR #1754, opened 2022-12-16
Fix/limit associated emails per account

superryeti:
PR followed up by issue: https://github.com/MIT-LCP/physionet-build/issues/1734

Context: 
Currently, Users are able to add a large\unlimited number of associated email to their account.

So we added a configurable limit which is added to each user. So each user has a default number of emails they can associate to their account. For some reason if the user needs the limit increased, they can ask admin to increase the limit for them.

NOTE: Although, i can understand that modifying the user model or any models is probably not desirable in most cases as it would make changes to database. I will leave it to the reviewers to decide if this is the best way to go about this.
If we don't want to make it configurable per user, i can remove the associated_emails_allowed field and just make it a constant in the settings file.


tompollard:
Thanks @amitupreti, personally I don't think this deserves an extra field in the database (at least not until we have a clear use case). My preference is that we simply set the policy in a config file. 

If someone demonstrates that a policy is inadequate (e.g. because the policy says 4 emails and someone makes good argument for needing more) then the policy could be updated for all users.

superryeti:
> Thanks @amitupreti, personally I don't think this deserves an extra field in the database (at least not until we have a clear use case). My preference is that we simply set the policy in a config file.
> 
> If someone demonstrates that a policy is inadequate (e.g. because the policy says 4 emails and someone makes good argument for needing more) then the policy could be updated for all users.

Thank you @tompollard . let me update the code to remove the changes in model.

physionet-django/user/migrations/0050_auto_20221216_1547.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.1.14 on 2022-12-16 20:47
```
tompollard: This migration can be removed

physionet-django/user/migrations/0051_remove_user_associated_emails_allowed.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 3.1.14 on 2022-12-16 21:00
```
tompollard: This migration can be removed

.env.example
```diff
@@ -130,3 +130,7 @@ SOURCE_CODE_REPOSITORY_LINK='https://github.com/MIT-LCP/physionet-build'
 # These variables are used to set the theme colors for the site
 DARK='#343A40'
 PRIMARY='#002A5C'
+
+# Users settings
+# maximum number of emails that can be associated to a user model
+ASSOCIATED_EMAILS_ALLOWED =  10
```
tompollard: Instead of `ASSOCIATED_EMAILS_ALLOWED`, how about something like `MAX_ASSOCIATED_EMAILS`? Nitpick but there is an extra space before the `10`

physionet-django/user/views.py
```diff
@@ -330,8 +330,15 @@ def edit_emails(request):
             set_public_email(request, public_email_form)
 
         elif 'add_email' in request.POST:
-            add_email_form = forms.AddEmailForm(request.POST)
-            add_email(request, add_email_form)
+            if associated_emails.count() >= settings.ASSOCIATED_EMAILS_ALLOWED:
+                messages.error(
+                    request,
+                    f'You cannot add more than {settings.ASSOCIATED_EMAILS_ALLOWED} email addresses.'
+                    f' If you need to more emails, please contact {settings.CONTACT_EMAIL}.'
```
tompollard: Perhaps change to: `If you need to link more emails to your account, ...`

tompollard (commented):
Thanks @amitupreti, I added a few comments inline.

physionet-django/user/migrations/0050_auto_20221216_1547.py
superryeti: Thank you. Removing them. i was also wondering the same and asked Harry if that would be okay

physionet-django/user/migrations/0051_remove_user_associated_emails_allowed.py
superryeti: Thank you. Removing them. i was also wondering the same and asked Harry if that would be okay

physionet-django/user/views.py
superryeti: My bad. Sorry about the sentence. Fixing it rn :)

.env.example
superryeti: Thank you. That's a way better name

superryeti:
@tompollard  resolved the comments. Thank you
1. remove the migration files
2. fixed variable names and sentences

superryeti:
@tompollard  Refactored the variable to `MAX_EMAILS_PER_USER`

tompollard:
@amitupreti Currently my .env file contains:

`CONTACT_EMAIL='PhysioNet Contact <contact@dev.physionet.org>'`

The error message that is rendered does not contain this address,  presumably something to do with the input variable being a string?:

> You cannot add more than 1 email addresses. If you need to link more emails to your account, please contact PhysioNet Contact .

Quick fix would be to remove the `If you need to link more emails to your account, please contact PhysioNet Contact .` sentence altogether. Not sure it is needed!

superryeti:
> @amitupreti Currently my .env file contains:
> 
> `CONTACT_EMAIL='PhysioNet Contact <contact@dev.physionet.org>'`
> 
> The error message that is rendered does not contain this address, presumably something to do with the input variable being a string?:
> 
> > You cannot add more than 1 email addresses. If you need to link more emails to your account, please contact PhysioNet Contact .
> 
> Quick fix would be to remove the `If you need to link more emails to your account, please contact PhysioNet Contact .` sentence altogether. Not sure it is needed!

Thank you. let me remove the last sentence with the Contact part. That makes sense

tompollard:
Thanks @amitupreti I think this is okay for now, so I will go ahead and merge. Having a bit more of a play around, it would be an improvement to display the max number of emails before the user tries to add another.

e.g. We could deactivate the Add Email form and display a message saying "Max number of email addresses has been reached". 

The "You cannot add more than X email addresses." is also a little ambiguous. I think it might be better to say "The maximum number of email addresses has been reached.". 

Perhaps these points can be addressed in a new pull request.

superryeti:
@tompollard  Those are very good ideas. let me open another PR for those.

Thank you

Issue #1732, opened 2022-12-07
L01: Arbitrary file upload in training


superryeti:
Any file can be uploaded to validate training by changing its extension to .pdf. The file's header should be checked to ensure it matches a PDF file's signature.

Solution would probably be similar to the https://github.com/MIT-LCP/physionet-build/issues/1730  as its the same file validation task

bemoody:
Indeed, if it's required to be a .pdf file we should check that it's actually a PDF file.

(On the other hand I think it would be better to require it to be a genuine training report from CITI, and not get into handling untrusted PDF files at all.)

alistairewj:
One of the training courses we require is TCPCS-2, which is a simple PDF. Although CITI has an API for their certificates, the group behind TCPCS-2 are a long way from having that kind of service, so validating PDFs are actually what they say they are is helpful for us.

bemoody:
There's validating file formats and then there's verifying/authenticating the file contents.  The latter is worth doing but worth keeping as a separate issue (#1180).

Issue #1180, opened 2021-01-06
Validate the training certificate (relates to #1179)


tompollard:
When a training certificate is uploaded, we should add a validation step to check that it appears to be correct. This relates to #1179 and https://github.com/MIT-LCP/physionet-build/pull/1026

Lucas-Mc:
If we want to make sure they passed (not sure if they get the report if they don't though?), it may be worth exploring the [Tika](https://github.com/chrismattmann/tika-python) package... we can run this code which would parse the PDF and give us the raw text:
```
from tika import parser

raw = parser.from_file('citiCompletionReport.pdf')
data = raw['content']
```
We can then search for the minimum score (group 1) and achieved score (group 2) using the following regex pattern:
`Minimum\\xa0Passing:\s(.+?)\\.+?Reported\\xa0Score\*:\s(.+?)\\`

I haven't tested this on the production servers though...

Lucas-Mc:
Alternatively, we can do regex search for required HIPAA packages and training completion date / expiration date... maybe something like this:
1. `HIPAA`
2. `Completion\\xa0Date:\s(.+?)\\`

PR #1755, opened 2022-12-19
Fix/arbitary file upload in training

superryeti:
PR for issue: https://github.com/MIT-LCP/physionet-build/issues/1732 

Currently users are able to upload any file type by changing the file extension in the training form.
This PR adds a utility function to verify if the uploaded file is a pdf file or not.

CHANGES

1. Update view
Since we are already using pdfminer.six to extract text from pdf files. So I  used the same library to verify if the uploaded file is a pdf file or not.

2. Fix Test
There was an existing test case which tested valid upload of training  certificate, the test was using django core `SimpleUploadedFile` to create a `text/plain` file and uploading that which is not a valid pdf(even if we change the content type).
So the solution to this would be to either create pdf on the fly during the test or use a static valid pdf file.
I chose the latter option, because creating pdf on the fly would require us to add a new dependency to the project which might introduce a security risk and need more discussion.

Questions:

1. I see the `TrainingType` model has two options (a) upload pdf as training certificate (b) upload pdf url or some kind of url as training certificate. So i thought of adding some more changes so that the code would work with urls also. But when i checked the code, and it looks like it is probably some kind of verification url that someone can obtain directly from the training provider. I didnot see any kind of extra code that used this and did the verification.(I assume this is the next step, of where we actually verify the content inside the training file automatically on another PR).

So the questions is, is my assumption correct? or should i add more changes to download pdf from the url(`completion_report_url`) and validate the pdf.

2. Also shouldn't we limit the upload size of the PDF, currently locally i was able to upload 100 MB file without any issues, I cant imagine any certificate being that big.


superryeti:
Will open another PR for verifying/authenticating the pdf file contents the as mentioned here

bemoody:
Validation of form data should be done in the form classes, not in views.  Look at the existing validation of the training report, in user.forms.TrainingForm.clean().

I'm very hesitant about adding features to automatically parse complex file formats (and PDF is a very, very complex format.)  I don't know 'pdfminer' at all, I don't know how secure it is.  I don't know who added it to the console or what sort of security review they did.  So simply saying "this library is already used somewhere else in the site" is not enough to convince me that using it here is a good idea.

I'm not sure what the *exact* bug is you're trying to solve.  At a basic level, checking if it's a PDF file just means checking whether it starts with `%PDF-1.` followed by a digit.

superryeti:
> Validation of form data should be done in the form classes, not in views. Look at the existing validation of the training report, in user.forms.TrainingForm.clean().

Thank you. Noted. let me work on the changes

superryeti:
> I'm not sure what the _exact_ bug is you're trying to solve. At a basic level, checking if it's a PDF file just means checking whether it starts with `%PDF-1.` followed by a digit.

You are right, the PR is to just validate if its a proper pdf file. I saw the existing library and went on with it.
Noted about the `%PDF-1`. That sounds like a much simpler approach to do the same thing. Looking into this :)

superryeti:
> I'm very hesitant about adding features to automatically parse complex file formats (and PDF is a very, very complex format.) I don't know 'pdfminer' at all, I don't know how secure it is. I don't know who added it to the console or what sort of security review they did. So simply saying "this library is already used somewhere else in the site" is not enough to convince me that using it here is a good idea.

--> Noted, for next step(after this PR) , we still want to verify/authenticate the file contents in training certificate, right? For that should i look for a more robust and proven library in case pdfminer doesnot look like a good option(assuming we still need to use some sort of 3rd party package to read the content of pdfs properly like PyPDF2) https://github.com/MIT-LCP/physionet-build/issues/1180.

superryeti:
> > I'm very hesitant about adding features to automatically parse complex file formats (and PDF is a very, very complex format.) I don't know 'pdfminer' at all, I don't know how secure it is. I don't know who added it to the console or what sort of security review they did. So simply saying "this library is already used somewhere else in the site" is not enough to convince me that using it here is a good idea.
> 
> --> Noted, for next step(after this PR) , we still want to verify/authenticate the file contents in training certificate, right? For that should i look for a more robust and proven library in case pdfminer doesnot look like a good option(assuming we still need to use some sort of 3rd party package to read the content of pdfs properly like PyPDF2) #1180.

Also, looks like we already tried to parse details from pdf at one point https://github.com/MIT-LCP/physionet-build/commit/493d9997#diff-c9532897c607f52a3040933942f31cb762f654d98acd8105ea72776efe2334fa

and this seems to be now done by adding regex patterns to the db(pretty smart). There is a model `TrainingRegex` linked to `TrainingType`

bemoody:
> Also shouldn't we limit the upload size of the PDF, currently locally i was able to upload 100 MB file without any issues, I cant imagine any certificate being that big.

Yep, looks like yet another thing that was broken in splitting the old credentialing form into two forms.  Those changes were done poorly and a lot of existing functionality was broken.

bemoody:
> I see the TrainingType model has two options (a) upload pdf as training certificate (b) upload pdf url or some kind of url as training certificate.

Indeed, it looks like the idea here is that a training type could require a URL instead of a PDF file.  I don't know how exactly this was supposed to work.

superryeti:
Added the following changes 

Fixing the PDF validator function,Moved file validation to forms, and added a file size limit for pdf arbitary_file_upload_in_training

1. Fixing the PDF validator function
    Instead of using the pdfminer.six library, i implemented a simple function that check if the pdf starts with
    the %PDF- header as suggested by Benjamin. 

2. Moved file validation to forms
    The file validation was done in the view(by me stupidly on the previous commit), i moved it to the form

3. Added a file size limit for pdf
    Also, earlier there was no limit on the file size, i added a limit of 1MB Default(but customizable from .env) for the pdf file
    For the file size validation function, i added it in the user.validators instead of adding it somewhere else like physionet.utility because, i thought 
    we might to implement different sizes for different applications, so i didnot try to make it generic.


Also, removed the static pdf and used a bytes object character instead of a pdf file for testing.

superryeti:
Regarding the validating of contents of the pdf during upload, I will write a plan and discuss with Benjamin, Alistair, Tom and start another PR after discussion.

Limiting this PR to just PDF type validation upon upload.

physionet-django/user/views.py
```diff
@@ -729,7 +729,6 @@ def edit_training(request):
             training_form = forms.TrainingForm(user=request.user)
         else:
             messages.error(request, 'Invalid submission. Check the errors below.')
-
```
tompollard: no change needed here.

.flake8
```diff
@@ -1,3 +1,3 @@
 [flake8]
 max-line-length = 119
```
tompollard: no change needed here.

superryeti: Oh yes, let me remove that. It is from one of my previous comment. i was trying to add a static pdf file. and was having issues because flake8 was trying to scan it for formatting and syntax

physionet-django/user/views.py
superryeti: Good catch. Let me push commit for both

superryeti:
Regarding , the PDF size for citi, i couldn't find out the file size(because i couldn't find the citi sample training PDF)
Should i increase the default size limit to 10MB @tompollard ?

alistairewj:
1MB should be fine. I checked 2-3 examples and they are all around 80kb. Seems very unlikely to ever get near 1MB for these files.

tompollard:
> 1MB should be fine.

I agree! Sorry @amitupreti, it looks like we have a conflict. Please could you rebase?

superryeti:
Thank you. Doing that right now

tompollard:
@amitupreti please could you squash the commits? A single "Fix/arbitary file upload in training" commit would be fine with me.

tompollard:
(actually, probably the message should make it clear that the changes are setting a limit on upload filesize)

bemoody:
If `validate_file_size` is specific to training reports, then the name of the function should mention that.  There are lots of different "things" that people upload that have different size limits on them.

I'm not sure it makes sense to define the limit as "MAX_TRAINING_REPORT_UPLOAD_SIZE_MB", why not just "MAX_TRAINING_REPORT_UPLOAD_SIZE" in bytes?

Please rebase & squash, don't merge dev into a feature branch.

superryeti:
Sorry for all the mess with the merge. i was trying to rebase on top of dev earlier. but messed up and ended up merging somehow. 

i finally figured it out and could push to a new branch.

Also implemented Benjamin's feedback on that branch

superryeti:
opened a new PR here https://github.com/MIT-LCP/physionet-build/pull/1785

tompollard:
@amitupreti Were you not able to force push to this branch? (that's what I'd expected you to do). Let me know if not and I'll take a look at the branch rules to make sure you can do this next time.

superryeti:
@tompollard 
HI Yes, To fix my mess, i made a backup branch(to not lose the changes).and on this current branch i reset the git head to my last commit before i tried to rebase. 

I then squashed my commits, rebased and fixed conflicts locally. But git kept saying `Head is detached from the current branch`. So after checking few posts on stackoverflow, i found the suggestion was that i should create a new branch because something i did detached my git from the current branch and i couldnot push to current branch.

So i checked out a new branch from the current state and pushed to a new branch

I also tried to use the new branch on this PR. i found few posts saying i could edit the PR(but i was unable to do probably because i am pushing through a fork)

PR #1756, opened 2022-12-19
Updates the research summary on the credential settings page.

lamawmouk:
This change updates the research summary description on the credentialing application page to encourage users to provide more context about the data they want to use and for what purposes the data will be used for e.g. research or commercial etc.

### DIFF
--- a/physionet-django/user/forms.py
+++ b/physionet-django/user/forms.py
@@ -477,8 +477,9 @@ class ResearchCAF(forms.ModelForm):
         model = CredentialApplication
         fields = ('research_summary',)
         help_texts = {
-            'research_summary': """Brief description of your research. If you
-                will be using the data for a class, please include course name
+            'research_summary': """Please provide a detailed description of how you would like
++                to use the data and which data you would like to use. If you
+                will be using the data for a class, please include the course name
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks Lama, (assuming you agree!) please could you tweak the message to:

`Please provide a detailed description of how you plan to use the data, including the name of any specific dataset(s) you intend to use. If you will be using the data for a class, please also include the name and number of the course.`

tompollard:
@lamawmouk another quick thought. We should probably increase the allowed word length for this section, because otherwise people will struggle to give any kind of detail (I think you mentioned that the limit is 150 characters?).

lamawmouk:
The limit is 150 words (1000 characters).

tompollard:
> The limit is 150 words (1000 characters).

Got it, so okay for now?

lamawmouk:
I think we should add a minimum word limit and raise an error if less than say  100 words?

tompollard:
> I think we should add a minimum word limit

This may need a bit of discussion with the team. Might be best to save for later to avoid holding up this PR!

tompollard:

Issue #1701, opened 2022-11-09
Training reports


lamawmouk:
We should make it more clear and highlight on the training page that users :
1) Upload training report not training certificate
2) Complete the data and specimens course (with HIPAA module) as suggested on the training submission page.

lamawmouk:
This issue is resolved at  #1757

PR #1757, opened 2022-12-20
Updates the training upload prompt on the training settings page.

lamawmouk:
This pull request solves the issue #1701 and updates the training prompt to clarify to the users what we look for in the training reports.

### DIFF
--- a/physionet-django/user/templates/user/edit_training.html
+++ b/physionet-django/user/templates/user/edit_training.html
@@ -50,8 +50,10 @@
     {% endif %}
   {% endfor %}
 
-  <p>Please use the form below to submit a new completion report. 
-    For CITI training, please refer to our <a href="{% url "citi_course" %}" target="_blank">step-by-step instructions</a>. 
+  <p>Please use the form below to submit a new completion report.</p>
+  <p>For CITI training, please refer to our <a href="{% url "citi_course" %}" target="_blank">step-by-step instructions</a> and make sure to:
+    <ul>1) Upload the training report and not the training certificate.</ul>
+    <ul>2) Complete the Data or Specimens Only Research course (with HIPAA module).</ul>
   {% if ticket_system_url %}To raise a support request, please <a href="{{ ticket_system_url }}">click here</a>.
   {% endif %}</p>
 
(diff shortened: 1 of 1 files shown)


physionet-django/user/templates/user/edit_training.html
```diff
@@ -50,8 +50,10 @@ <h4>Submit evidence of a completed course</h4>
     {% endif %}
   {% endfor %}
 
-  <p>Please use the form below to submit a new completion report. 
-    For CITI training, please refer to our <a href="{% url "citi_course" %}" target="_blank">step-by-step instructions</a>. 
+  <p>Please use the form below to submit a new completion report.</p>
+  <p>For CITI training, please refer to our <a href="{% url "citi_course" %}" target="_blank">step-by-step instructions</a> and make sure to:
+    <ul>1) Upload the training report and not the training certificate.</ul>
```
tompollard: Here you could change the `<ul>` to `<ol>` and remove the `1)` and `2)`

tompollard (commented):
@lamawmouk it would be good to have a quick chat about this. We need to be a bit careful about adding PhysioNet-specific requirements in this way, because they may not apply to other people using the same codebase (e.g. not everyone may require the Data or Specimens Only Research course).

We used to parse the PDF to prevent upload of the certificate, and at some point this functionality was broken, so the first point (please upload your completion report, not certificate) shouldn't really need saying (we just need to fix the functionality we broke!).

physionet-django/user/templates/user/edit_training.html
lamawmouk: Yes for the ordered list, but since we want to remove the certificate aspect can I leave point 2 and keep it unordered and for point 2 say "if  you're requesting access to MIMIC datasets please complete the Data or Specimens Only Research course" as majority of the people ask for MIMIC and upload Biomedical Investigators or others, so we could clarify it there upon upload of the pdf?

tompollard: Thanks Lama. I think if we are keeping just one point (which sounds good) then perhaps no need to use a list at all? I think it's okay to mention the Data or Specimen's course for now, but I don't think we should discuss MIMIC (which would be confusing on other platforms).

@alistairewj are you okay with this being added to the training page?:

`For CITI training, please refer to our <a href="{% url "citi_course" %}" target="_blank">step-by-step instructions</a> and be sure to <strong>upload the training report<strong>, not the certificate.`

lamawmouk: Yes I agree I think we should mention “Data or Specimens Only Research” course as people who have any CITI training wouldn't click on the link to read CITI instructions and end uploading any course for CITI. 
 If we say CITI training we should also add "complete Data or Specimens Only Research course with HIPAA module" somewhere upon report upload and so we avoid sending that rejection message.

tompollard:

Issue #1737, opened 2022-12-07
M05: Project File Upload Request Can Be Replayed


superryeti:
Context: The generate_signed_url view is responsible to create a google storage signed url and upload the files to google storage. it looks like the PUT request that uploads files to the signed url can be replayed to upload another file.
This can be used to bypass the manual review process

For example: let say author submits valid files, and copies the PUT request that uploaded the files. Later once the Editor has approved the project and before other author approve it, the submitting author can use that PUT request and upload another file(as the current view lets the link to be valid for 24 hours). All of this is assuming that the editor review and approves project before 24 hours.

Recommendation: The signed URL should only be usable once, or if not the link should only remain active for ~60  minutes or something like that.

Also the security report also mentions, that the file size is not verified if the user directly submits the modified PUT request to change the uploaded file.

Here is the exact line from the report(just in case i misunderstood)

         
<img width="732" alt="image" src="https://user-images.githubusercontent.com/24412619/206302142-ed4ea4bb-7d9f-4558-8888-4be4fd741d53.png">

bemoody:
Clearly needs somebody with a deep understanding of GCS - not me.

generate_signed_url_v4 documentation is none too clear (and, uhh... are you importing that function from a non-public module?!)

One idea to deal with the "adding files after project submission" issue would be to see whether you can create an *ephemeral set of credentials* for each project.  That wouldn't help with "repeatedly uploading huge files to exceed your storage allowance", though.

superryeti:
Discussed with Alistair
Here is the summary

1. We cant limit the upload size in GCP buckets, so we decided to just limit the signed url and make it ephemeral
2. Regarding the use of private module, replacing this with https://github.com/googleapis/python-storage/blob/HEAD/samples/snippets/storage_generate_upload_signed_url_v4.py.  we can see it being used in a test in googles python library https://github.com/googleapis/python-storage/blob/7dfeb622bb966e368786e3c9be67ad77b3150725/tests/system/test__signing.py#L38

superryeti:
Will open 1 PR for replacing the private module and another PR for making updating the signed url expiration.

bemoody:
I would add a few things, following our discussion this morning.

If I understood correctly what Karol was saying, the client is required to *begin uploading* the file before the signed URL expires, but the client can *continue* transmitting data after that point in time.

This would mean that expiration of the signed URL *does not guarantee* that project files can no longer be modified.

So I have three questions:

[...]

PR #1763, opened 2022-12-21
 Removed the use of private function to generate the signed url  and replaced it with our own wrapper function

superryeti:
**Context:** On project.views , `generate_signed_url` function, we were using a private google cloud function.
  This commit removes that by using a recommendation from google cloud documentation. https://cloud.google.com/storage/docs/samples/storage-generate-signed-url-v4#storage_generate_signed_url_v4-python

  The docs seem to recommend that we create a blog instance from google storage bucket, assign the blob name and use the `generate_signed_url` function from the blob instance.

  I added that to the `generate_signed_url` function and removed the private function call.

  1. Why did i create a wrapper function(`generate_signed_url`)?

  So turns out we need to mock call to the the `generate_signed_url` function and the MediaStorage class in the test(probably because we can actually call the during test as we dont have a gcp setup that can run during tests)
  So if i didnot create a wrapper, i was not sure how to mock calls to the `blob.generate_signed_url` function. So the easiest way was to create a wrapper function and mock that instead and edit the test case.

  I am not sure if that is the best way or even a decent way. I leave that to the reviewer to decide. :)

  2. Why i think the change should work same(i.e upload file in the right directory like 'active-project/slug/filename.txt')?

      The `MediaStorage` is just a wrapper to `GoogleCloudStorage` and to get the blob object we are doing the following:
          storage = MediaStorage()
          bucket = storage.bucket()
          blob = bucket.blob(blob_name)

          which should lead us to /google/cloud/storage/bucket.py#L680
          Here the blob_name is assigned as name for the `Blob` class.

          when we do blob.generate_signed_url, we are calling the `Blob` class's `generate_signed_url` function.
          this `generate_signed_url` function is defined in /google/cloud/storage/blob.py#L420 and on L596 we can see that,  its using this code to generate the resource

          resource = "/{bucket_name}/{quoted_name}".format(
              bucket_name=self.bucket.name, quoted_name=quoted_name
          )

          which should make the resource same as it was before this change.(i.e /bucket_name/blob_name)
          canonical_resource = f'/{storage.bucket.name}/active-projects/{project_slug}/{filename.strip("/")}'

      Other than that previously we were using the `from google.cloud.storage._signing import generate_signed_url_v4` function, which is the same function the `Blob.generate_signed_url` function is using.
      So it should be the same thing

Reference: Issue https://github.com/MIT-LCP/physionet-build/issues/1737

### DIFF
--- a/physionet-django/physionet/storage.py
+++ b/physionet-django/physionet/storage.py
@@ -14,9 +14,9 @@ class StaticStorage(GoogleCloudStorage):
     location = ''
 
 
-def generate_signed_url_v4(blob_name, size, expiration=dt.timedelta(days=1)) -> str:
+def generate_signed_url(blob_name, size, expiration=dt.timedelta(days=1), version='v4') -> str:
     """
-    Generate a v4 signed URL to access project files on GCS
+    Generate a signed URL to access project files on GCS
 
     Parameters:
         blob_name (str): The name/path of the blob to be uploaded.
@@ -34,6 +34,7 @@ def generate_signed_url_v4(blob_name, size, expiration=dt.timedelta(days=1)) -> 
         api_access_endpoint='https://storage.googleapis.com',
         expiration=expiration,
         method='PUT',
-        headers={'X-Upload-Content-Length': str(size)}
+        headers={'X-Upload-Content-Length': str(size)},
+        version=version
     )
     return url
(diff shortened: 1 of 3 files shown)


superryeti:
will wait for this PR to merge, before opening the PR to limit the expiration of signed url(as we would want to make use of this new function)

tompollard:
Thanks, looks good to me. As Benjamin mentioned, we should check that the expiration time doesn't prevent people from uploading large files (please open a new PR to fix this if needed).

superryeti:
Thank you. I will look into that and open another PR

Issue #1760, opened 2022-12-20
Confusing "reminder" message


bemoody:
Today I received an email that reads:

```
From: PhysioNet Automated System <noreply@physionet.org>
Subject: Reminder - Revisions requested for project [...]

Project title: [...]
Submission ID: [...]
Submitting author: [...]

Dear Benjamin Moody,

Thank you for submitting your project entitled "[...]" to PhysioNet.

A summary of our standard set of editorial checks is listed below:

We ask that you address the following comments before we proceed:
Did you mean to resubmit this?

Once these comments have been addressed, please resubmit your
project.
```

However, I'm not the submitting author.  I didn't submit anything and I couldn't resubmit even if I wanted to.  I don't know what I'm being "reminded" of.

(Of course this message was intended for the submitting author, but the message is written as if it's directed at me personally.)

briangow:
I agree, I discussed this with @tompollard  today after someone brought it to our attention. Thanks for raising the issue!

Do we want to keep the message like it is and only send it to the submitting author or revise it and send a message to all of the authors?

alistairewj:
I'd only send it to the submitting author. No use sending a message to someone when they have no meaningful action to take from it.

tompollard:
In general my preference is to copy emails to all co-authors because it is good for them to know what is happening with a project that they are co-authoring. 

Although they cannot take direct action (i.e. they can't edit the project themselves) they can still take action (e.g. email the submitting author to say that they are happy to take on some of the tasks outlined by the editors).

To avoid confusion, we could add some text to the non-submitting authors saying something like: "Only the submitting author is able to edit the project. You are receiving this message as a co-author and no direct action is required".

Saying this, I'd be happy with just sending to the submitting author (and it's probably easier to set this up!).

alistairewj:
Counterpoint is having the submitting author receive a nagging e-mail to resubmit followed by their co-authors forwarding a very similar message to them gives me major Office Space vibes ;)

Since it's non-critical to submission I feel like we should be less noisy! The system already sends out a ton of e-mails during the publication process. Adding a no-op one is going to train people to ignore all of the messages.

tompollard:
> Since it's non-critical to submission I feel like we should be less noisy! The system already sends out a ton of e-mails during the publication process. Adding a no-op one is going to train people to ignore all of the messages.

[...]

PR #1766, opened 2022-12-22
Only send the reminder email for a project in "Awaiting Author Revisions" to the submitting author

briangow:
This PR changes the current behavior of emailing all authors of a project when a reminder is sent, to instead only email the submitting author. As explained in https://github.com/MIT-LCP/physionet-build/issues/1760 the email being sent was confusing authors who were not the submitter.

### DIFF
--- a/physionet-django/notification/utility.py
+++ b/physionet-django/notification/utility.py
@@ -291,8 +291,11 @@ def edit_decision_notify(request, project, edit_log, reminder=False):
     # Prepend reminder to the subject if needed
     if reminder:
         subject = "Reminder - {}".format(subject)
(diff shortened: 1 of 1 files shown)


tompollard:
Looks good to me, thanks! This only changes the reminder emails. Do we want to change the behaviour for all "edit_decision_notify" emails?

briangow:
@tompollard , I don't think we want to change some of the other "edit_decision_notify" emails. For example, if the project is accepted we probably want to notify all authors. As mentioned in #1760 , the phrasing in the reminder email was particularly confusing unless you are the submitting author. When we revisit the phrasing across all of the emails we can decide if there are any others that should only go to the submitting author.

PR #1767, opened 2022-12-22
 Update View and Form for better User experience with adding edit_emails

superryeti:
Context: Followup to https://github.com/MIT-LCP/physionet-build/pull/1754 

Changes
Updated view and forms, to show users how many emails they have currently attached, and the total limit.

if the user has reached the limit, the form will not be shown.Instead a warning message will be shown.

Screenshots:
UI before email limit has been reached
image

UI after email limit has been reached
image

### DIFF
--- a/physionet-django/user/templates/user/edit_emails.html
+++ b/physionet-django/user/templates/user/edit_emails.html
@@ -91,15 +91,20 @@
 
 <hr>
 <h4>Add Email</h4>
-<p>Add a new email address.</p>
-<form action="{% url 'edit_emails' %}" method="post" class="form-signin row no-pd" name="add_email">
-  {% csrf_token %}
-  <div class='col-md-9'>
-    {% include "form_snippet_no_labels.html" with form=add_email_form %}
-  </div>
-  <div class="btn-container-rsp mg-left">
-    <button class="btn btn-primary btn-rsp" name="add_email" type="submit">Add Email</button>
-  </div>
+    {% if total_associated_emails < max_associated_emails_allowed %}
+        <p>Add a new email address. <br> <small>Maximum Emails  allowed: {{total_associated_emails}}/{{max_associated_emails_allowed}}</small></p>
+        <form action="{% url 'edit_emails' %}" method="post" class="form-signin row no-pd" name="add_email">
+          {% csrf_token %}
+
+          <div class='col-md-9'>
+            {% include "form_snippet_no_labels.html" with form=add_email_form %}
+          </div>
+          <div class="btn-container-rsp mg-left">
+            <button class="btn btn-primary btn-rsp" name="add_email" type="submit">Add Email</button>
+          </div>
+    {% else %}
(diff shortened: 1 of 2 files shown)


tompollard:
Thanks @amitupreti, I have a few quick thoughts:
- instead of making the "add email" form completely disappear, what do you think about leaving the form but setting it to disabled?
- I'm not super keen on the yellow text. It is difficult to see on my screen. I think keeping the text in plain `<p>` format is probably okay.
- Instead of "You have reached the maximum number of emails allowed. Please remove one or more of existing emails to continue." how about "You have reached the maximum number of emails. Please remove an existing email to add another."
- I also find that the size of "Maximum Emails allowed: x/x" makes it difficult to read. Perhaps remove this sentence and then modify the current "Add a new email address." text to say "Add a new email address (Max emails: X)."

superryeti:
> Thanks @amitupreti, I have a few quick thoughts:
> 
>     * instead of making the "add email" form completely disappear, what do you think about leaving the form but setting it to disabled?
> 
>     * I'm not super keen on the yellow text. It is difficult to see on my screen. I think keeping the text in plain `<p>` format is probably okay.
> 
>     * Instead of "You have reached the maximum number of emails allowed. Please remove one or more of existing emails to continue." how about "You have reached the maximum number of emails. Please remove an existing email to add another."
> 
>     * I also find that the size of "Maximum Emails allowed: x/x" makes it difficult to read. Perhaps remove this sentence and then modify the current "Add a new email address." text to say "Add a new email address (Max emails: X)."

Excellent feedback. Thank you so much @tompollard for the feedback. Let me make the suggested changes and push commits

superryeti:
Made the following changes




1. disabling the form button if the email limit is reached(i didn't disable the input because it uses a snippet which we dont want to edit. So the alternative i know is using javascript. For that i created an anonymous function and executed it on page load directly)
2. removed the yellow color from the warning message
3. refactored the warning message to make it informative
4. removed the small email limit text and merged it with the Add a New Email label.



Before the Email Limit

image


After the email limit has been reached

image

physionet-django/user/templates/user/edit_emails.html
```diff
@@ -91,15 +91,19 @@ <h4>Public Email</h4>
 
 <hr>
 <h4>Add Email</h4>
-<p>Add a new email address.</p>
+<p>Add a new email address(Max emails: {{ max_associated_emails_allowed }}).</p>
```
tompollard, line 94: Missing space here. `Add a new email address(Max emails:` should be `Add a new email address (Max emails:`

tompollard:
Thanks, this looks like an improvement to me! There's a missing space in the text (as noted inline), but not a big deal. @amitupreti if you haven't fixed this within the next ~10min then I'll go ahead and merge and we can tidy up later!

superryeti:
> Thanks, this looks like an improvement to me! There's a missing space in the text (as noted inline), but not a big deal. @amitupreti if you haven't fixed this within the next ~10min then I'll go ahead and merge and we can tidy up later!

Hi Sorry for the delay,just saw the comment. i will open a quick small PR shortly. Thank you

Issue #1731, opened 2022-12-07
M04: Project File Upload Lack of Validation


superryeti:
There is a lack of data validation when generating a signed URL for a file upload, allowing bypass of constraints. We should validate the filename parameter when preparing to generate a signed URL (letters, numbers, dashes, dots, underscores, forward slashes). Note: this is only possible if someone users javascript to mimic file upload(our UI will prevent this by using the validators `validate_filename_or_parent`)

Cycura showed that it is possible to upload files with names like `{{constructor.constructor('alert(1)')(){}}}` with javascript.
With this method, users can also upload filenames and generate signed_urls that can have ?param=value to control paramaters, or use ../ to traverse folder.

Recommendation:
Implement the same validation on the view that generates the signed urls to prevent this.

PR #1768, opened 2022-12-22
Making the filename validation consistent across view and form

superryeti:
From the TRA, we found out that the api `generate_signed_url` did not have the same filename validation as the form. This PR makes the validation consistent across the two by using the same validation function `validate_filename`.

I also removed two existing validation statements on the view as it was already covered by the existing validation function `validate_filename` I manually tested this by passing filename with space and non ascii characters and both were covered by the validation function `validate_filename`.

Followup to https://github.com/MIT-LCP/physionet-build/issues/1731


tompollard:
thanks!

Issue #1739, opened 2022-12-08
The message for deprecated files in a project could be more informative


briangow:
The current message given when the files for a particular version of a project had been deprecated is: "The files for this project are no longer available". If this is on the only version of the project that may be fine. However, if there are multiple versions of a project it may confuse the user. We could potentially remedy this with one of these approaches:

1. Update the message to say "The files for this version of this project are no longer available. If another version exists please check for the files there."
2. Dynamically check to see if other versions of the project are available and if they are update the second line: "The files may exist in v1.1, v1.2". The v1.1, v1.2 could include hyperlinks to those project versions. 
3. Dynamically check to see if the files are available in other versions and if they are update the second line: "Files for this project do exist in v1.2".

@tompollard , @bemoody please let me know your thoughts.

tompollard:
Good point, thanks for raising this. 

My vote would be for 3 (Dynamically check to see if the files are available in other versions and if they are update the second line: "Files for this project do exist in v1.2".). 

We could implement 1 or 2 as a quick fix and then introduce 3 later?

bemoody:
Maybe add a field to explicitly link to the "replacement" version?  I don't like the idea of saying "the files you're looking for MIGHT be located in one of these places".

In general, it's not safe to assume that if somebody is looking for file X in version 1.0, then file X in version 2.0 (even if it exists) is an appropriate replacement.  So I think you need to be very careful about what the message implies.

PR #1771, opened 2022-12-23
Update deprecated files message

briangow:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/1739, the message indicating that files have been deprecated wasn't very informative when a newer version of the project is available. 

This PR updates the previous message in cases when a newer version of the project is available. In this case it now states: "The files for this version of the project (<old version number>) are no longer available. The latest version of this project is <latest version number>", where <latest version number> is a hyperlink to the latest version of the project.

### DIFF
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -324,9 +324,15 @@
   <h2 id="files">Files</h2>
   {% if project.allow_file_downloads %}
   {% if project.deprecated_files %}
-    <div class="alert alert-danger col-md-8" role="alert">
-      The files for this project are no longer available.
-    </div>
+    {% if project.is_latest_version %}
+      <div class="alert alert-danger col-md-8" role="alert">
+        The files for this project are no longer available.
+      </div>
+    {% else %}
+      <div class="alert alert-danger col-md-8" role="alert">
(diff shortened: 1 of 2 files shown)


physionet-django/project/templates/project/published_project.html
```diff
@@ -324,9 +324,15 @@ <h5 class="card-header">Versions</h5>
   <h2 id="files">Files</h2>
   {% if project.allow_file_downloads %}
   {% if project.deprecated_files %}
-    <div class="alert alert-danger col-md-8" role="alert">
-      The files for this project are no longer available.
-    </div>
+    {% if project.is_latest_version %}
```
tompollard: As the div is used in both cases, it might be a bit cleaner (avoids duplication) to move this outside the if clause? e.g. 

```
<div class="alert alert-danger col-md-8" role="alert">
  {% if project.is_latest_version %}
      The files for this project are no longer available.
  {% else %}
      The files for this version of the project ({{ project.version }}) are no longer available. The latest version of this project is <a href="{% url 'published_project' latest_version.slug latest_version.version %}" target="_blank">{{ latest_version.version }}</a>
  {% endif %}
</div>
```

tompollard:
Thanks, looks good to me!

physionet-django/project/templates/project/published_project.html
briangow: @tompollard thanks for the suggestion. I did think about that also. When searching online earlier to see if this was a good idea, I got confused by a post that discussed putting a conditional statement inside the tag itself - which isn't what we do here. I should have just tried it, because it works just fine.

Issue #1770, opened 2022-12-23
Hide LightWave for projects under embargo?


tompollard:
@briangow I notice that we display a "Visualize waveforms" link next to the entry for https://physionet.org/content/mimic-iv-ecg/ on the front page of PhysioNet. When I click the link, I am taken to LightWave at: https://physionet.org/lightwave/?db=mimic-iv-ecg/0.1. 

LightWave then gets stuck at: "Reading list of records". Presumably this is because LightWave is unable to read the records because of the embargo. I think we should probably hide LightWave for people in this case?

briangow:
@tompollard , good catch! I didn't realize we displayed a Visualize Waveforms link by default on the front page if a project has waveforms.

briangow:
@tompollard , I just submitted a PR to remedy this in the case of embargo. However, I'd assume we'd see similar behavior in Lightwave for any project that isn't open access since access isn't granted until the DUA, credentialing, etc is satisfied.

briangow:
This link is now only displayed if the user has access to the waveform files.

PR #1774, opened 2022-12-23
Hide the lightwave link on the front page when a project is under embargo

briangow:
As noted in https://github.com/MIT-LCP/physionet-build/issues/1770 , the announcement on the front page for a new project will show a link to Lightwave if the project has compatible files. This isn't desired when the files in a project are under embargo. This PR hides the link to Lightwave if the project files are under embargo.

### DIFF
--- a/physionet-django/project/templatetags/project_templatetags.py
+++ b/physionet-django/project/templatetags/project_templatetags.py
@@ -189,3 +189,20 @@ def mailto_link(*recipients, **params):
 @register.filter
 def filename(value):
     return os.path.basename(value.name)
+
+
+@register.simple_tag
+def call_method(obj, method_name, *args):
+    """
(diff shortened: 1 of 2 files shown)


briangow:
After discussion, we decided that it is better to base showing the Lightwave Visualize waveforms link on whether or not a user has access to the files. I have revised this PR to check to see if the user has access to the WFDB files in a given project. If they do, the Visualize waveforms link is displayed. 

I also removed the `not published_project.deprecated_files` condition. This was being used to hide the Visualize waveforms link for projects with files that have been deprecated. I have confirmed that this update still hides the Visualize waveforms link for projects with deprecated files.

tompollard:
Looks good to me! @bemoody, okay with you to merge?

Issue #1764, opened 2022-12-22
Flickering modals that prevent the user clicking on buttons


tompollard:
There is a bug with some of the modals that causes them to flicker and prevents the user from clicking on buttons. @briangow has noticed this too.

e.g. As an author, when I am trying to approve a project for publication at: https://physionet.org/projects/SLUG/submission/, I see the following modal:

![Screen Shot 2022-12-22 at 9 52 51 AM](https://user-images.githubusercontent.com/822601/209161272-bc7b862b-7c9b-4f44-bfb4-efd8128eba1a.png)

The approve button flickers and I am unable to click the "Approve publication" button.

superryeti:
also found on on Delete Email confirmation Modal popup
[Modal Flikering.webm](https://user-images.githubusercontent.com/24412619/209365301-dd80f4d3-c2cf-409d-ab5e-157d595cb015.webm)

superryeti:
If no one is working on this. i can take this over. @tompollard

tompollard:
thanks @amitupreti we would definitely appreciate your help with this!

superryeti:
Sounds great. I will get started. Thank you

superryeti:
Interestingly, the flickering of modals seems to be happening because of use of this class `list-group-item` with a modal inside it.

So the two solutions discussed in the issue linked below was

1. Place the model outside the `<li>` with class `list-group-item` OR
2. Override the hover behaviour of the class `list-group-item`


```css
.list-group-item, .list-group-item:hover{ z-index: auto; }
```

The class `list-group-item` is used in total 51 places on the physionet model, out of that, i could find the similar conditions(i.e modal inside `<li>` with class `list-group-item`) in x places

1. manage_published_project.html - 2 places
2. active_submission_timeline.html - 1 place
3. author_list.html - 1 place(the issue Tom found)
4. edit_emails.html - 1 place

I kind of like solution 2 because it is quicK, less change on codebase, makes sure the flickering issue is fixed throughout the project and i could not find any code implementation that has complex stacking of modals or `<li>` elements.

But i might have missed something or maybe the project will evolve in future to include complex stacking. So solution1 seems like the right thing to do. 

For now i will go ahead and submit a PR with solution1 and wait for comments from @tompollard  on what he thinks might be the best solution( if solution2 is preferred, then i can discard the solution 1 and open another PR if needed with solution 2 or other recommendation.)


Related issue: https://github.com/twbs/bootstrap/issues/25206

[...]

PR #1776, opened 2023-01-02
Fix for flickering throughout project

superryeti:
Context: Several modals were flickering on physionet. Please check the issue for details https://github.com/MIT-LCP/physionet-build/issues/1764

In Summary, the flickering was caused by the modal being present inside a li tag with class `list-group-item`. The value of z-index on hover, was causing the flickering. The fix was to remove the z-index on hover.

The simple fix was to override the z-index on hover for the modal. We can do this by adding the following to the physionet.css file. This should make sure, that the override is applied throughout the site(as the physionet.css is loaded in base_css.html which is loaded in base.html)

The PR implements solution2 from the discussion

### DIFF
--- a/physionet-django/static/custom/css/physionet.css
+++ b/physionet-django/static/custom/css/physionet.css
@@ -593,3 +593,5 @@ a, code{
   border-radius: 0.25rem;
   border: 1px solid rgba(0, 0, 0, 0.15);
 }
+
+.list-group-item, .list-group-item:hover{ z-index: auto; }
--- a/physionet-django/static/custom/css/physionet.css
+++ b/physionet-django/static/custom/css/physionet.css
@@ -594,4 +594,9 @@ a, code{
   border: 1px solid rgba(0, 0, 0, 0.15);
 }
 
-.list-group-item, .list-group-item:hover{ z-index: auto; }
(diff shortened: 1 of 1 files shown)


bemoody:
Nice work locating the bug!

A bunch of minor things:

- Use the same style as the existing code.  Space before `{`, line break after it, one property per line, properties indented by two spaces.

- Please add a brief comment in the CSS file to explain what the rule is for (even just a pointer to the upstream issue would be sufficient.)

- We probably don't want to override the style of `.list-group-item`, only `.list-group-item:hover`.

- However, what you've done here creates two style rules with equal specificity (one in bootstrap.css, one in physionet.css, with the same selector `.list-group-item:hover`.)  This is not a great idea because it's then dependent on the order of rules/stylesheets.  Instead, try to write a rule that is in some way *more specific* than the rule you're trying to override.  For example, instead of `.list-group-item:hover` you could write `body .list-group-item:hover`.

All together, this is what I would write:
```
/* fix layering of modal inside list-group-item
   (https://github.com/twbs/bootstrap/issues/25206) */
body .list-group-item:hover {
  z-index: auto;
}
```

tompollard:
This is great, thanks Amit!

PR #1777, opened 2023-01-03
Adds events app to reorganize the events code.

danamouk:
This pull request creates an events app to move all of the events-related templates, views, urls, forms, and widgets.


bemoody:
Thanks, Dana!  I haven't tested it but looks good code-wise.

A couple important things to note (for eveyone else's benefit):

- The existing *models* (and user permissions) are left in the user app for now.  Ideally these will also be moved over eventually (which could be a slightly tricky migration.)  Any *new* models can and should be defined in `events/models.py`.

- There are currently no test cases for the events features (except for TestURLs).  This should be addressed in another pull request.

tompollard:
@bemoody thanks for checking. So good to merge?

@danamouk please could you raise issues for the following two points?

1. The existing models (and user permissions) are left in the user app for now. Ideally these will also be moved over eventually (which could be a slightly tricky migration.) Any new models can and should be defined in events/models.py.

2. There are currently no test cases for the events features (except for TestURLs). This should be addressed in another pull request.

tompollard:
Thanks @danamouk!

PR #1781, opened 2023-01-03
Adds Requirements met column (via traffic light)

lamawmouk:
This PR adds a traffic light column to the credentialing table (ID Check tab). It adds header cell which contains the header Requirement met. It also adds the data cells which are highlighted in amber (the color is default now but will change based on whether the requirement is met or not).

### DIFF
--- a/physionet-django/console/static/console/css/console.css
+++ b/physionet-django/console/static/console/css/console.css
@@ -1,7 +1,25 @@
 /* Based on sb-admin.css from start-bootstrap */
 
-.application-ok {
-  background: orange;
+.traffic-circle {
+  border: rgb(99, 99, 99) solid 1px;
+  width: 19px;
+  height: 19px;
+  border-radius: 10px;
+  display: block;
+  margin-left: auto;
+  margin-right: auto
+}
+
+.traffic-red {
+  background: rgb(204, 50, 50);
+}
+
+.traffic-orange {
+  background: rgb(231,180,22);
+}
+
+.traffic-green {
+  background: rgb(45, 201, 55);
 }
 
 h1 {
--- a/physionet-django/console/static/console/css/console.css
+++ b/physionet-django/console/static/console/css/console.css
@@ -1,5 +1,9 @@
 /* Based on sb-admin.css from start-bootstrap */
 
+.application-ok {
+  background: orange;
+}
+
 h1 {
   margin-top: 10px;
   font-size: 36px;
--- a/physionet-django/console/templates/console/credential_processing.html
+++ b/physionet-django/console/templates/console/credential_processing.html
@@ -57,7 +57,7 @@
                 <td>{{ user.email }}</td>
                 <td>{{ application.reference_email }}</td>
                 <td>{{ application.application_datetime|date }}</td>
-                <td><i class="traffic-circle traffic-{{application.get_traffic_status}}"></i></td>
+                <td><i class="traffic-circle traffic-{{ application.get_traffic_status }}"></i></td>
                 <td>
                   <form action="{% url 'process_credential_application' application.slug %}">
                     <input class="btn btn-success" type="submit" value="Process"/>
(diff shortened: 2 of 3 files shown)


physionet-django/console/static/console/css/console.css
```diff
@@ -1,5 +1,9 @@
 /* Based on sb-admin.css from start-bootstrap */
 
+.application-ok {
```
tompollard: I think using "traffic" in the style name might be clearer. Instead of filling a cell with color, you could render a circle (this makes the lights easier to reuse elsewhere too).

```css
.traffic-circle {
  display: inline-block;
  border: rgb(99, 99, 99) solid 1px;
  width: 19px;
  height: 19px;
  border-radius: 10px; /* half of width + borders */ 
}

.traffic-red {
  background: rgb(204, 50, 50);
}

.traffic-orange {
  background: rgb(231,180,22);
}

.traffic-green {
  background: rgb(45, 201, 55);
}
```

physionet-django/console/templates/console/credential_processing.html
```diff
@@ -44,6 +44,7 @@
                 <th onclick="sortTable(2,'personalt')">Email <i class="fas fa-sort" id="icon_2_personalt"></i></th>
                 <th onclick="sortTable(3,'personalt')">Reference Email <i class="fas fa-sort" id="icon_3_personalt"></i></th>
                 <th class="table-elapsed" onclick="sortTable(4,'personalt')">Application <i class="fas fa-sort" id="icon_4_personalt"></i></th>
+                <th>Requirements met</th>
```
tompollard: Maybe just "Status" for brevity?

physionet-django/console/templates/console/credential_processing.html
```diff
@@ -56,6 +57,7 @@
                 <td>{{ user.email }}</td>
                 <td>{{ application.reference_email }}</td>
                 <td>{{ application.application_datetime|date }}</td>
+                <td class="{{application.get_trafficlight_status}}">{{""}}</td>
```
tompollard: Here you can render the colored circle with:

```
<td><i class="traffic-circle traffic-{{ application.get_traffic_status }}"></i></td>
```

physionet-django/user/models.py
```diff
@@ -793,6 +793,9 @@ class CredentialApplication(models.Model):
 
     class Meta:
         default_permissions = ('change',)
+    
+    def get_trafficlight_status(self):
```
tompollard: Perhaps shorten the name, and return the color?

```
    def get_traffic_status(self):
        return 'orange'
```

tompollard (commented):
Thanks @lamawmouk, I like the approach. I've added a few suggestions inline.

physionet-django/console/templates/console/credential_processing.html
lamawmouk: Agree!

physionet-django/console/static/console/css/console.css
lamawmouk: I've added the centering to the circle to lie in the middle of the cell container.

physionet-django/console/templates/console/credential_processing.html
```diff
@@ -57,7 +57,7 @@
                 <td>{{ user.email }}</td>
                 <td>{{ application.reference_email }}</td>
                 <td>{{ application.application_datetime|date }}</td>
-                <td class="{{application.get_trafficlight_status}}">{{""}}</td>
+                <td><i class="traffic-circle traffic-{{application.get_traffic_status}}"></i></td>
```
tompollard: Nitpick, but please could you add the space on either side of the variable name? `{{ application.get_traffic_status }}`. See: https://docs.djangoproject.com/en/4.1/topics/templates/#variables

tompollard:
Thanks Lama!

PR #1785, opened 2023-01-05
 Fix to verify pdf file type and limit pdf size

superryeti:
1. Added a utility function to verify if the uploaded file is a pdf file or not.

    Currently users are able to upload any file type by changing the file extension in the training form.
    This commit adds a utility function to verify if the uploaded file is a pdf file or not.

2. Update test to work with the training pdf validation The new change makes sure that the file uploaded is a valid pdf and is readable. Our previous test was using django core SimpleUploadedFile to create a plain/txt file and uploading that which is not a valid pdf(even if we change the content type). So the solution to this would be to either create pdf on the fly during the test or use a static valid pdf file.

    I chose the later option, because creating pdf on the fly would require us to add a new dependency to the project which might introduce a security risk and need more discussion.

3. Added a file size limit for pdf Also, earlier there was no limit on the file size, i added a limit of 1MB Default(but customizable from .env) for the pdf file For the file size validation function, i added it in the user.validators instead of adding it somewhere else like physionet.utility because, i thought we might to implement different sizes for different applications, so i didnot try to make it generic.

### DIFF
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -534,7 +534,7 @@ PLATFORM_WIDE_CITATION = {
 
 SOURCE_CODE_REPOSITORY_LINK = config('SOURCE_CODE_REPOSITORY_LINK',
                                      default='https://github.com/MIT-LCP/physionet-build')
-MAX_TRAINING_REPORT_UPLOAD_SIZE = config('MAX_TRAINING_REPORT_UPLOAD_SIZE', cast=int, default=1)
+MAX_TRAINING_REPORT_UPLOAD_SIZE = config('MAX_TRAINING_REPORT_UPLOAD_SIZE', cast=int, default=1048576)
 
 # User model configurable settings
 MAX_EMAILS_PER_USER = config('MAX_EMAILS_PER_USER', cast=int, default=10)
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -534,6 +534,7 @@ PLATFORM_WIDE_CITATION = {
 
 SOURCE_CODE_REPOSITORY_LINK = config('SOURCE_CODE_REPOSITORY_LINK',
                                      default='https://github.com/MIT-LCP/physionet-build')
+MAX_TRAINING_REPORT_UPLOAD_SIZE_MB = config('MAX_TRAINING_REPORT_UPLOAD_SIZE_MB', cast=int, default=1)
 
 # User model configurable settings
 MAX_EMAILS_PER_USER = config('MAX_EMAILS_PER_USER', cast=int, default=10)
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -534,7 +534,7 @@ PLATFORM_WIDE_CITATION = {
 
 SOURCE_CODE_REPOSITORY_LINK = config('SOURCE_CODE_REPOSITORY_LINK',
                                      default='https://github.com/MIT-LCP/physionet-build')
-MAX_TRAINING_REPORT_UPLOAD_SIZE_MB = config('MAX_TRAINING_REPORT_UPLOAD_SIZE_MB', cast=int, default=1)
+MAX_TRAINING_REPORT_UPLOAD_SIZE = config('MAX_TRAINING_REPORT_UPLOAD_SIZE', cast=int, default=1)
 
 # User model configurable settings
 MAX_EMAILS_PER_USER = config('MAX_EMAILS_PER_USER', cast=int, default=10)
--- a/physionet-django/user/test_views.py
+++ b/physionet-django/user/test_views.py
@@ -15,7 +15,7 @@ from django.contrib.auth.models import AnonymousUser
 from django.contrib.messages.storage.fallback import FallbackStorage
 from django.core import mail
 from django.core.management import call_command
-from django.core.files.uploadedfile import SimpleUploadedFile, UploadedFile
+from django.core.files.uploadedfile import SimpleUploadedFile
 from django.test import RequestFactory, TestCase
 from django.urls import reverse
 from django.utils import timezone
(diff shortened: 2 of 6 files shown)


physionet-django/physionet/settings/base.py
```diff
@@ -534,6 +534,7 @@ class StorageTypes:
 
 SOURCE_CODE_REPOSITORY_LINK = config('SOURCE_CODE_REPOSITORY_LINK',
                                      default='https://github.com/MIT-LCP/physionet-build')
+MAX_TRAINING_REPORT_UPLOAD_SIZE_MB = config('MAX_TRAINING_REPORT_UPLOAD_SIZE_MB', cast=int, default=1)
```
tompollard: The config file says `MAX_TRAINING_REPORT_UPLOAD_SIZE` but this is referencing `MAX_TRAINING_REPORT_UPLOAD_SIZE_MB`.

tompollard: Is there a reason to convert to MB here? If we receive the number in bytes, why not keep the bytes? (especially when we convert the MB back to bytes anyway for the comparison in `validate_training_file_size()`.

physionet-django/user/validators.py
```diff
@@ -255,3 +256,11 @@ def validate_domain_list(value):
     """
     if not re.fullmatch(r'(\w+\.\w+,*\s*)*', value):
         raise ValidationError('Must be separated with commas.')
+
+
+def validate_training_file_size(value):
+    """
+    Validate the file size of a file.
+    """
+    if value.size > settings.MAX_TRAINING_REPORT_UPLOAD_SIZE_MB * 1024 * 1024:
```
tompollard: If you are taking the number in bytes, this conversion wouldn't be necessary (i.e. instead of `> settings.MAX_TRAINING_REPORT_UPLOAD_SIZE_MB * 1024 * 1024` why not `> settings.MAX_TRAINING_REPORT_UPLOAD_SIZE`?

physionet-django/user/validators.py
```diff
@@ -255,3 +256,11 @@ def validate_domain_list(value):
...
+        raise ValidationError('The maximum file size that can be uploaded is 5MB.')
```
tompollard: I don't think we want to hardcode `5MB` here. I guess we may want to do the byte to KB/MB conversion though? 

If we are rounding, we need to avoid small numbers being rounded to 0. @bemoody copying you in in case you have idea about how best to deal with this? Could be an argument for requiring the file size to be specified in KB/MB?

physionet-django/physionet/settings/base.py
superryeti: HI, This seems some kind of mistake, i remember removing the MB stuff and doing bytes as Benjamin suggested. 

it might happened because i was poking around multiple branches and lost my latest changes. Pushing the changes with the fix asap

physionet-django/user/validators.py
superryeti: HI, This seems some kind of mistake, i remember removing the MB stuff and doing bytes as Benjamin suggested. 

it might happened because i was poking around multiple branches and lost my latest changes. Pushing the changes with the fix asap

physionet-django/user/validators.py
superryeti: Ah yes, my bad. Thank you.

Benjamin mention that we should use KB earlier. i had that changes but it seems i lost it when pushing here.

Doing that asap

physionet-django/physionet/settings/base.py
superryeti: But yes, we want everything in KB as per discussion on the previous closed PR

physionet-django/user/validators.py
superryeti: i see you meant to talk about rounding it down.

It should be a nice whole number even when we convert i think, as our predefined size limit is a 1048576 Bytes, which should be 1024 KB.

So even if we update the limit in future, we should be fine(it wouldnot be rounded to 0 unless the upload limit is less than 1MB) as long a we use a whole number limit like 1 MB, 2 MB etc.

superryeti: pushed the changes, but not marking this as resolved as we are waiting to hear from Benjamin

physionet-django/physionet/settings/base.py
```diff
@@ -534,6 +534,7 @@ class StorageTypes:
 
 SOURCE_CODE_REPOSITORY_LINK = config('SOURCE_CODE_REPOSITORY_LINK',
                                      default='https://github.com/MIT-LCP/physionet-build')
+MAX_TRAINING_REPORT_UPLOAD_SIZE = config('MAX_TRAINING_REPORT_UPLOAD_SIZE', cast=int, default=1)
```
tompollard: sorry, i should have spotted this before. default of 1 doesn't make sense if we are now using bytes.

superryeti: ah yes. my bad i missed this before. Updating asap

superryeti: @tompollard  do you think `default=1 * 1024 * 1024` would look better than `default= 1048576 ` 
Although ultimately both are the same , but the first one is much easier to change in future.

tompollard: hmm, i think i'd just go for `1048576`

superryeti: alright. sounds good

physionet-django/user/test_views.py
```diff
@@ -15,7 +15,7 @@
 from django.contrib.messages.storage.fallback import FallbackStorage
 from django.core import mail
 from django.core.management import call_command
-from django.core.files.uploadedfile import SimpleUploadedFile
+from django.core.files.uploadedfile import SimpleUploadedFile, UploadedFile
```
tompollard: This line is changed but the change doesn't seem necessary. I don't see UploadedFile being used?

superryeti: Ah yes, that should not be there, let me make changes.

tompollard:
Thanks @amitupreti, looks good to me!

PR #1788, opened 2023-01-09
Limit access to static page and front page button to writters and admins.

harryface:
Added permission check on the navbar console.


tompollard:
@harryface we fixed the migration issue in https://github.com/MIT-LCP/physionet-build/pull/1790. Please could you remove the merge migration commit?

Issue #1789, opened 2023-01-09
Conflicting migrations (0064_auto_20221209_1108.py and 0064_auto_20221209_1300.py)


tompollard:
I have not looked into the details yet, but it looks like conflicting migrations were merged (probably by me):

[0064_auto_20221209_1108.py](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/migrations/0064_auto_20221209_1108.py)
[add migration for the embargo files feature](https://github.com/MIT-LCP/physionet-build/commit/db0daf1b918b5ef64c18abe6d638bf2213cdb4cd)

[0064_auto_20221209_1300.py](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/migrations/0064_auto_20221209_1300.py)
[fix linter concerns](https://github.com/MIT-LCP/physionet-build/commit/29d40e47a05c0fb9a38995b9c654efeca2b71a37)

This needs fixing before we can do anything else :/

tompollard:
The last migration to be applied on our production server is: [0064_auto_20221209_1108.py](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/migrations/0064_auto_20221209_1108.py)

```
from django.db.migrations.recorder import MigrationRecorder

last_migration = MigrationRecorder.Migration.objects.latest('id')
print(last_migration.app)   
print(last_migration.name)

>> project
>> 0064_auto_20221209_1108
```

I have checked the previous migration objects and [0064_auto_20221209_1300.py](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/migrations/0064_auto_20221209_1300.py) was not applied.

tompollard:
Fixed in https://github.com/MIT-LCP/physionet-build/pull/1790 and https://github.com/MIT-LCP/physionet-build/pull/1792

PR #1790, opened 2023-01-09
Fix migrations project0065

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/1789, two conflicting migrations were merged. I will need to look into this wasn't picked up by our tests. Perhaps because they aren't set to re-run after the main branch is updated?

This pull request fixes the migration history by:

1) Removing the most recent migration (https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/migrations/0064_auto_20221209_1300.py), which conflicts with the previous one (https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/migrations/0064_auto_20221209_1108.py).
2) Adds a new auto-generated migration.

The last migration to be applied on our production server is: [0064_auto_20221209_1108.py](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/migrations/0064_auto_20221209_1108.py).  I have checked the previous migration objects and [0064_auto_20221209_1300.py](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/migrations/0064_auto_20221209_1300.py) was not applied.


tompollard:
(upgrade test is expected to fail, because the production branch has the faulty migration)

bemoody:
Looks reasonable.  Could you do me a favor and rename `0065_auto_20230109_1019.py` to `0065_editor_permissions.py`?

> (upgrade test is expected to fail, because the production branch has the faulty migration)

Right, I guess you had pushed the conflicting commit to github but not to physionet.  So not a problem, I've just reset github to where physionet is currently:
```
To ssh://github.com/MIT-LCP/physionet-build
 + 369408bca...03ba32202 production -> production (forced update)
```

bemoody:
It is weird to me that the `update` script on physionet-staging didn't reject the push.  `getmigrationtargets` seemingly reports an error if neither of the conflicting migrations are applied yet, but it doesn't report an error if one is applied and the other isn't.

tompollard:
Thanks @bemoody, good idea to rename the migration to something human readable. I'll try to get into the habit of doing this in future.

bemoody:
It should be possible to push this branch to staging server.  Then to production server.  Then try to re-run outstanding pull requests (e.g by amending the last commit and force pushing.)

tompollard:
Thanks Benjamin!

PR #1791, opened 2023-01-09
Fix keyerror on production

superryeti:
Looks like because of an improperly handled dictionary access, there was an error when someone tried to access their training report url. Incase where user has training url instead of a training pdf, the training form, sends us `completion_report_url`. so the code i added last week was to check for validation of pdf, and the key for training pdf was `completion_report` Since the user didnot have this key in their trainingform, the code threw error.

A simple get statement will fix the issue


superryeti:
Will  rebase and push the changes after https://github.com/MIT-LCP/physionet-build/pull/1790 is merged

tompollard:
Thanks Amit!

PR #1793, opened 2023-01-09
Feature: Grant access to projects via temporary training validity.

harryface:
This PR allows short term trainings to be enabled on projects such that if a short term training is amongst the group of trainings required for a project access, a user with ONLY a validity of the short term training gains access to the project.

1. Added a role a which grants only access to the training section of the admin console and added it to a group in demo-user.json
2. Streamlined change_credentialapplication permission to credentialing
3. Added ability to filter the training page by the type of training from console
4. While creating a training type, there exists a boolean field that allows it to be marked as a short term training
5. On the Projects page, there is a toggle button that when On, allows short term trainings to take precedence (i.e if a short term training is selected as one of the required trainings, its validity grants users access)


alistairewj:
Replaced by PRs #1813 (filter training types), #1799 (new training role), and #1808 (temporary training permission)

PR #1794, opened 2023-01-10
Update to Django 3.2 / django-background-tasks -> django-background-tasks-updated

bemoody:
`django-background-tasks-updated` is a fork of `django-background-tasks` to provide support for django 4.0.

The only actual code changes are these:

https://github.com/iamjonmiller/django-background-tasks/commit/36c5c2b928d53a66ec6b9a190d16528bb4904171
https://github.com/iamjonmiller/django-background-tasks/commit/b234e1e1fa7bbb6999b52bab8065ba6b79c69ac3

which amounts to:

- replacing `compat.close_connection` with `django.db.close_old_connections`
- replacing `compat.StringIO` with `io.StringIO`
- replacing `compat.models.GenericForeignKey` with `django.contrib.contenttypes.fields.GenericForeignKey`
- replacing `django.dispatch.Signal(providing_args=X)` with `django.dispatch.Signal(X)`

For no reason I can see, `django-background-tasks-updated` is marked as requiring `Django>=3.2.12`.  The code is compatible with Django 3.1.14 as far as I can tell, and works if you install it using `pip3 --no-deps`.


tompollard:
@bemoody I have added the following changes:

- Bumped Django from 3.1.14 to 3.2.16
- Cast `SITE_ID` to `int` to fix a failing configuration check added in Django 3.2. "The [`SITE_ID`](https://docs.djangoproject.com/en/3.2/ref/settings/#std-setting-SITE_ID) setting must be an integer."
- To maintain historical behavior (and avoid a migration) I set the default type for primary key fields to `AutoField`. See: https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys.

PR #1796, opened 2023-01-10
CONSOLE ACCESS: Role control for access to static pages and front page button..

harryface:
Added permission check on the navbar console.


tompollard:
Thanks @harryface, looks good to me (cleaned up PR of https://github.com/MIT-LCP/physionet-build/pull/1788).

PR #1798, opened 2023-01-10
BUG: remove fixture from migration

harryface:
Loading Fixtures from migrations means that the model cannot be edited.
This PR is meant to allow the migrations stand alone.


tompollard:
thanks @harryface, looks good to me!

PR #1799, opened 2023-01-10
Feature: Add training reviewer role and adjust accordingly

harryface:
Add a new permission to the training type, that grants a user access to the training check of console.


tompollard:
Looks good to me, thanks Harry. Please could you regenerate the migrations?

tompollard (commented):
This update changes the behavior for people going to (e.g.) https://localhost:8000/console/training/review/. 

Previously the user would be redirected to https://localhost:8000/. Now they see a 404 error. 

I think this is okay, but it would be helpful to highlight changes like this in the comment description in future.

tompollard:
@alistairewj after this change is merged, there is a new `can_review_training` permission. This permission is needed to process training applications. You'll need to add the permission to the relevant permission groups (which you can do through the Django admin tool).

harryface:
> This update changes the behavior for people going to (e.g.) https://localhost:8000/console/training/review/.
> 
> Previously the user would be redirected to https://localhost:8000/. Now they see a 404 error.
> 
> I think this is okay, but it would be helpful to highlight changes like this in the comment description in future.

Noted Tom, will start effecting that.

PR #1802, opened 2023-01-11
Removes independent researcher category 

lamawmouk:
This PR remove the category of an independent researcher from the credentialing application page by making change to the model by deleting the category in the field 'RESEARCHER_CATEGORIES'. A new migration file is created and the migration is applied to the database.


tompollard:
@lamawmouk When we discussed this today, I don't think we said we would remove "independent researcher". I thought the outcome of the discussion was:

1) Independent researchers would be required to provide a reference.
2) Anyone listed as a reference should have an institutional email address.

lamawmouk:
Thanks @tompollard for clarification! I will close the PR and raise the issue with those two points

PR #1805, opened 2023-01-12
Bump to Django==4.1.5 and update dependencies (bump to django-ckeditor==6.5.1, django-js-asset==2.0.0, psycopg2==2.9.5, Python >=3.8).

tompollard:
This pull request bumps versions of the following packages:

- Django==4.1.5
- django-ckeditor==6.5.1
- django-js-asset==2.0.0
- psycopg2==2.9.5

For compatibility with Django 4, I have also changed the minimum Python version from 3.7 to 3.8.

As discussed in the [Django release notes](https://docs.djangoproject.com/en/3.1/releases/3.1/#id2), the `HttpRequest.is_ajax()` method is deprecated

> The HttpRequest.is_ajax() method is deprecated as it relied on a jQuery-specific way of signifying AJAX calls, while current usage tends to use the JavaScript [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). Depending on your use case, you can either write your own AJAX detection method, or use the new [HttpRequest.accepts()](https://docs.djangoproject.com/en/3.1/ref/request-response/#django.http.HttpRequest.accepts) method if your code depends on the client Accept HTTP header.

I have exactly reimplemented the `is_ajax()` method, following instructions in the documentation.


bemoody:
> I have exactly reimplemented the is_ajax() method, following instructions in the documentation.

We should just get rid of is_ajax.  All of the existing uses are buggy and were pointless to begin with.

tompollard:
Thanks @bemoody

> We should just get rid of is_ajax. All of the existing uses are buggy and were pointless to begin with.

Removed `HttpRequest.is_ajax()` in https://github.com/MIT-LCP/physionet-build/pull/1806/

tompollard:
I have done a fair amount of manual testing after these upgrades (submitting projects, adding and removing news items, credentialing people) and I've not noticed any obvious issues.

bemoody:
What are all the migrations for?

tompollard:
> What are all the migrations for?

The migrations are an artifact of a change introduced in Django 4.0 to the migrations autodetector, which "[now uses model states instead of model classes](https://docs.djangoproject.com/en/4.1/releases/4.0/#migrations-autodetector-changes)". 

> ### Migrations autodetector changes[¶](https://docs.djangoproject.com/en/4.1/releases/4.0/#migrations-autodetector-changes)
> 
> The migrations autodetector now uses model states instead of model classes. Also, migration operations for ForeignKey and ManyToManyField fields no longer specify attributes which were not passed to the fields during initialization.

> As a side-effect, running makemigrations might generate no-op AlterField operations for ManyToManyField and ForeignKey fields in some cases.

So we are seeing a series of no-op AlterField migrations for ManyToManyField and ForeignKey fields. The response at https://code.djangoproject.com/ticket/33538#comment:1 suggests they are not easily avoidable (and I couldn't find a way).

tompollard:
@bemoody okay to merge?

tompollard:
@bemoody need to regenerate the migrations after merging the last couple of PRs. Will do shortly.

tompollard:
Hmm, looks like there is also deprecated code in the one of the merged commits.

```
ImportError: cannot import name 'url' from 'django.conf.urls' (/Users/tompollard/projects/physionet-build/env/lib/python3.9/site-packages/django/conf/urls/__init__.py)
```

Will fix that, then regenerate migrations.

bemoody:
> --- a/Dockerfile
> +++ b/Dockerfile
> @@ -1,4 +1,4 @@
> -FROM python:3.7-slim-buster
> +FROM python:3.8-slim-buster

I don't know about this "slim" stuff, but shouldn't it be python 3.9?  And bullseye rather than buster?

tompollard:
@bemoody I have changed to Python 3.9 Bullseye. 

> "The slim image is a paired down version of the full image. This image generally only installs the minimal packages needed to run your particular tool. In the case of python, that’s the minimum packages to run python and the same for node.js."

I have switched to the standard image for now (i.e. removed the "slim"). If we want to switch back later, it's easy to do.

alistairewj:
The slim packages are experimental ([according to Debian's docker page](https://hub.docker.com/_/debian)) but they don't seem so bad, mostly excluding documentation. [Slimify script here](https://github.com/debuerreotype/debuerreotype/blob/master/scripts/debuerreotype-slimify) and the [include](https://github.com/debuerreotype/debuerreotype/blob/master/scripts/.slimify-includes)/[exclude](https://github.com/debuerreotype/debuerreotype/blob/master/scripts/.slimify-excludes) manifests seem innocuous enough.

tompollard:
@alistairewj please could you take a look at this if you have the opportunity?

tompollard:
Gah, tests now failing after rebasing on dev. @alistairewj @bemoody @amitupreti if the cause is obvious to anyone, please let me know. It's not immediately clear to me where the `get_random_string` is being called (and why it isn't being given a length).

```
ERROR: test_add_event_invalid (events.tests_views.TestEvents)
tests the view that adds an invalid event(event with duplicate title for same host)
  File "/__w/physionet-build/physionet-build/physionet-django/events/tests_views.py", line 49, in test_add_event_invalid
    self.test_add_event_valid()
  File "/__w/physionet-build/physionet-build/physionet-django/events/tests_views.py", line 30, in test_add_event_valid
    response = self.client.post(
  File "/env3/lib/python3.9/site-packages/django/test/client.py", line 852, in post
    response = super().post(
  File "/env3/lib/python3.9/site-packages/django/test/client.py", line 441, in post
    return self.generic(
  File "/env3/lib/python3.9/site-packages/django/test/client.py", line 541, in generic
    return self.request(**r)
  File "/env3/lib/python3.9/site-packages/django/test/client.py", line 810, in request
    self.check_exception(response)
  File "/env3/lib/python3.9/site-packages/django/test/client.py", line 663, in check_exception
    raise exc_value
  File "/env3/lib/python3.9/site-packages/django/core/handlers/exception.py", line 55, in inner
  File "/env3/lib/python3.9/site-packages/django/core/handlers/base.py", line 197, in _get_response
  File "/env3/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapped_view
  File "/__w/physionet-build/physionet-build/physionet-django/events/views.py", line 52, in event_home
    event_form = AddEventForm(user=user)
  File "/__w/physionet-build/physionet-build/physionet-django/events/forms.py", line 24, in __init__
    super(AddEventForm, self).__init__(*args, **kwargs)
  File "/env3/lib/python3.9/site-packages/django/forms/models.py", line 359, in __init__
    self.instance = opts.model()
  File "/env3/lib/python3.9/site-packages/django/db/models/base.py", line 533, in __init__
    val = field.get_default()
  File "/env3/lib/python3.9/site-packages/django/db/models/fields/__init__.py", line 933, in get_default
    return self._get_default()
TypeError: get_random_string() missing 1 required positional argument: 'length'
```

superryeti:
image
The error seems to be coming from the slug in the Models.py. 

The test was failing when trying to create a new event. (i am trying to reproduce the error locally)

tompollard:
Thanks @amitupreti, well spotted. Somehow missed this when I did a quick search earlier.   

@bemoody please could you check the fix in the latest commit? In newer versions of Django, a length is required by `get_random_string`. 

As far as I can see, the default can not be included in the model slugfield itself (or at least this causes issues with migrations), so I am instead overriding the save method.

tompollard:
@bemoody this should be good to go! please could you take a look?

alistairewj:
I just tested it on docker and everything seems to be working smoothly!

bemoody:
Great work, looks good to me.

tompollard:
Thanks everyone, it sounds like I'm safe to merge!

PR #1806, opened 2023-01-12
Remove calls to HttpRequest.is_ajax() method

tompollard:
In future releases of Django, [`HttpRequest.is_ajax()` is deprecated](https://docs.djangoproject.com/en/3.1/releases/3.1/#id2):

> The HttpRequest.is_ajax() method is deprecated as it relied on a jQuery-specific way of signifying AJAX calls, while current usage tends to use the JavaScript [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). Depending on your use case, you can either write your own AJAX detection method, or use the new [HttpRequest.accepts()](https://docs.djangoproject.com/en/3.1/ref/request-response/#django.http.HttpRequest.accepts) method if your code depends on the client Accept HTTP header.

As noted by @bemoody in https://github.com/MIT-LCP/physionet-build/pull/1805, our calls to `HttpRequest.is_ajax()` are unnecessary and can be removed anyway. Doing this will help us to upgrade to the latest version of Django.

### DIFF
--- a/physionet-django/project/views.py
+++ b/physionet-django/project/views.py
@@ -922,9 +922,6 @@ def project_files_panel(request, project_slug, **kwargs):
     else:
         files_editable = False
 
-    if not request.is_ajax():
-        return redirect('project_files', project_slug=project_slug)
-
     (display_files, display_dirs, dir_breadcrumbs, parent_dir,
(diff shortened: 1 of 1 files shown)


bemoody:
Looks good!

PR #1807, opened 2023-01-12
Credentialing: Require academic or institutional email addresses for references

lamawmouk:
This PR ensures that all users provide a reference for identity check including the independent researchers.
It also ensures that users references are academic or an institutional email or at least the references are credentialed.

### DIFF
--- a/physionet-django/user/forms.py
+++ b/physionet-django/user/forms.py
@@ -29,7 +29,7 @@ from user.models import (
 from user.trainingreport import TrainingCertificateError, find_training_report_url
 from user.userfiles import UserFiles
 from user.validators import UsernameValidator, validate_name, validate_training_file_size
-from user.validators import UsernameValidator, validate_institutional_email
+from user.validators import validate_institutional_email
 from user.widgets import ProfilePhotoInput
 
 from django.db.models import OuterRef, Exists
--- a/physionet-django/user/forms.py
+++ b/physionet-django/user/forms.py
@@ -507,8 +507,7 @@ class ReferenceCAF(forms.ModelForm):
                 respond promptly, as long response times will prevent approval
                 of your application.""",
             'reference_name': 'The full name of your reference.',
-            'reference_email': """The email address of your reference. It is
-                strongly recommended that this be an institutional email address.""",
+            'reference_email': """The academic or institutional email address of your reference.""",
             'reference_organization': """Your reference's employer or primary
                 affiliation.""",
             'reference_title': "Your reference's professional title or position."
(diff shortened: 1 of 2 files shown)


physionet-django/user/forms.py
```diff
@@ -585,7 +586,7 @@ def clean(self):
                        data['reference_organization'],
                        data['reference_title']]
 
-        ref_required = data['researcher_category'] in [0, 1, 6, 7]
+        ref_required = data['researcher_category'] in range(8)
```
tompollard: This looks like it adds a requirement for a reference for all categories? I think the only change we wanted to make was to require a reference for independent researchers?

physionet-django/user/forms.py
```diff
@@ -535,6 +534,8 @@ def clean_reference_email(self):
             if reference_email in self.user.get_emails():
                 raise forms.ValidationError("""You can not put yourself
                     as a reference.""")
+            if reference_email.split('@')[-1] in ["yahoo.com", "163.com", "126.com", "outlook.com", "gmail.com", "qq.com", "foxmail.com"]:
```
tompollard: We should move this validation rule to one of the validation modules (`validators.py`). Ideally we would have something like an `is_institutional_email()` validator that you would call here.

physionet-django/user/forms.py
tompollard: See: https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/user/models.py#L707-L720 for the list of categories:

```python
    """
    An application to become credentialed
    """
    RESEARCHER_CATEGORIES = (
        (0, 'Student'),
        (7, 'Graduate Student'),
        (1, 'Postdoc'),
        (2, 'Academic Researcher'),
        (3, 'Hospital Researcher'),
        (4, 'Industry Researcher'),
        (5, 'Government Researcher'),
        (6, 'Independent Researcher'),
    )
```

tompollard (commented):
Perhaps more than you want to take on here, but this might be a nice opportunity to switch to a Django Enum for the researcher categories? 

See: 
- Django docs: https://docs.djangoproject.com/en/4.1/ref/models/fields/#enumeration-types
- Stackoverflow:  https://stackoverflow.com/questions/54802616/how-to-use-enums-as-a-choice-field-in-django-model

Currently we have 

```
    """
    An application to become credentialed
    """
    RESEARCHER_CATEGORIES = (
        (0, 'Student'),
        (7, 'Graduate Student'),
        (1, 'Postdoc'),
        (2, 'Academic Researcher'),
        (3, 'Hospital Researcher'),
        (4, 'Industry Researcher'),
        (5, 'Government Researcher'),
        (6, 'Independent Researcher'),
    )
```

tompollard:
@bemoody please could you take a quick look at this and add your thoughts?

bemoody:
In `validate_institutional_email` you have:

```
    if value.split('@')[-1] in domains:
```

This won't catch "John.Doe@Yahoo.COM" (unless there's something else that converts to lowercase beforehand) so we probably want a `.lower()` in there.

I also think that `validate_institutional_email` should validate that the inpyut *is* a valid email address, which you can do by importing `validate_email` from  `django.core.validators` and calling it in `validate_institutional_email`.

lamawmouk:
Thanks Benjamin, good point!

bemoody:
Thanks, Lama.  I think this is a good start which we can continue to improve on over time.

PR #1808, opened 2023-01-13
Feature: Access to projects via temporary training validity

harryface:
This PR allows short term trainings to be enabled on projects such that if a short term training is amongst the group of trainings required for a project access, a user with ONLY a validity of the short term training gains access to the project.

- Utilised already existing django created many-to-many fields, and added temporary field to it


alistairewj:
Closing this PR as we are pursuing the events app for temporary dataset access.

PR #1800, opened 2023-01-10
Reorganizes models in the events app.

danamouk:
As referenced in #1782, this pull request organizes the models code in the events app. As part of the transition of the models from the users app to the events app, two migrations for both the user and events app are created. In the `user/model.py` to avoid fields.E304 Reverse accessor clashes in Django, old related names were added as part of foreign key and in the `events/model.py` User in the foreign key was modified to "user.User".

Other aspects that are reorganizated as part of the events app include `events/enums.py` and `events/validators.py`, `events/admin.py`.

PR #1809, opened 2023-01-13
Handle newly installed apps in test-upgrade.sh

bemoody:
Correctly handle newly-installed apps (apps that previously didn't contain any migrations) in `test-upgrade.sh`, by explicitly handling the "zero" state in getmigrationtargets.

This is a prerequisite for pull #1800.

Tested with django 3.1, but I'd like to test also with 3.2 and 4.1 before merging.


tompollard:
thanks benjamin!

PR #1813, opened 2023-01-16
Feature: Filter Training by training type

harryface:
- Utilised the search functionality already in existence
- Added iexact for the training type.


tompollard:
This PR adds a dropdown filter menu to the training page that allows admin users to search for training by training type (http://localhost:8000/console/training/review/).

Screenshot 2023-01-25 at 11 31 11 PM

We don't need this for PhysioNet (at least not right now) because we only support a single training type, but it is presumably useful for HDN.

PR #1814, opened 2023-01-17
Feature: Make lightwave switchable

rafalkosla101:
Context of those changes:
We would like to separate lightwave to a package that is pip installable, after doing that we also should be adding the app conditionally to the django itself. Doing that would solve potential vulnerability.


tompollard:
@rafalkosla101 please could you rebase on dev so that we can merge?

rafalkosla101:
Rebased @tompollard

tompollard:

PR #1817, opened 2023-01-18
Fix for reorder references

superryeti:
Context: Followup fix for  PR #1602

This PR adds a small fix to let us preserve order of references during editing, On https://github.com/MIT-LCP/physionet-build/issues/1557 we found out that we were unable to change the order of the references in cases where we wanted to switch references from position a to b and b to a in the same edit. 
On PR https://github.com/MIT-LCP/physionet-build/pull/1602, Tom added a  fix for that problem by adding new column called order which was to be then used to sort the references in the UI and this PR completes that change

1. project/forms.py 
 This change adds a method on `save` on ReferenceFormSet, it then loops through the existing forms, and assigns the order value to the references
2. project/modelcomponents/metadata.py
    Also, i found one case where the existing unique_together constraint on the Reference model was not working as expected because we were using content_type to make unique id which resulted in a case where we were enabling duplicates, Details here https://github.com/MIT-LCP/physionet-build/pull/1602#discussion_r1066288881

In Summary, We are now able to edit the order of references


tompollard:
Very cool, thanks Amit! I'll take a look at this later today.

tompollard:
Looks good to me, thanks Amit!

PR #1820, opened 2023-01-19
Feature: Add session management

rafalkosla101:
This PR adds idle timeout which can be configured by ENV variables.  The PR also introduces a warning screen when user is idle for too long when logged in - example:
image

All the times can be adjusted by using ENV variables provided in env.example
All the default are defined so the behaviour of the current app will not be changing.


physionet-django/notification/urls.py
```diff
@@ -1,4 +1,4 @@
-from django.urls import path
+from django.urls import path, include
```
tompollard: It looks like this import isn't used?

tompollard (commented):
Thanks Rafal, looks good to me (though I think @bemoody should review before we merge, because he understands this better than I do).

Is there anything we do (or can do) to prevent people from losing information when midway through a form? The project descriptions can take a significant amount of time to write (e.g. at http://localhost:8000/projects/p7TCIMkltNswuOB9FZH1/content/).

bemoody:
I think it's a lousy idea.  Setting a timeout of 24 hours or so might be reasonable, though you do need to give serious thought to large transfers.  Setting a timeout of ten minutes is insane, and I would never voluntarily use your site.

If you don't trust people to be competent about security (setting a screen lock, not letting other people use their account, not entering their password on somebody else's computer) then why would you trust their system isn't *already* compromised?

tompollard:
> Setting a timeout of 24 hours or so might be reasonable, though you do need to give serious thought to large transfers. Setting a timeout of ten minutes is insane, and I would never voluntarily use your site.

The features are optional, so we are not under any obligation to use them. Let's say we we decide not to use them for PhysioNet, are you okay with the code being merged so that the feature is available to Health Data Nexus?

.env.example
```diff
@@ -137,3 +137,10 @@ MAX_EMAILS_PER_USER = 10
 
 # Max training report size in bytes
 MAX_TRAINING_REPORT_UPLOAD_SIZE = 1048576
+
+# Session settings
```
alistairewj: worth commenting what the unit of measurement is on the expiry (I assume seconds)

bemoody:
Well, in a way I am concerned that inflicting a lousy UX on HDN users, for no benefit, also reflects badly on us.

Implementation-wise, this would need a lot more documentation.  If you want to add a bunch of new configuration options and global variables, you need to document precisely what each of them means.  "Session security" is a meaningless phrase.  You also need to write better commit messages.

alistairewj:
Fair points. I agree with the sentiment that a very short idle time is painful and feels more like security theatre than anything else. @rafalkosla101 is following the defaults recommended to us by a security firm, but they tend to be overly conservative (and would probably recommend we don't share data at all since it would be more secure! :joy:).

Can we maybe enumerate some of the edge cases this might cause issues for? It's obviously annoying to have to login every day, but that's something we'll accept on the HealthDataNexus side. In terms of potential implications (if this was enabled on PhysioNet with a 24 hour timeout, for example), I can think of two major cases:

- Large downloads via the platform
    - We should test if the idle timeout breaks downloads
- Large downloads via wget
    - Seems like this operates outside of Django, so it would be fine?

bemoody:
Thanks.  Rafal, I'm sorry for my earlier tone; I'm skeptical about
this idea, but it's not a criticism of you personally.

> - Large downloads via the platform
>     - We should test if the idle timeout breaks downloads
> - Large downloads via wget
>     - Seems like this operates outside of Django, so it would be fine?

This is one of the things I'm not clear about and why I'd like more
technical explanation.  What would "session-security" do differently,
compared to how django currently handles session expiration?

rafalkosla101:
@bemoody No need for any apology - constructive criticism is always welcome since I can learn out of it. Giving a bit more of explanation of above changes - Django by default can be configured to renew sessions on every request made to the Django app itself by adding one of the settings(SESSION_SAVE_EVERY_REQUEST variable to be exact). Session security is quite simple package that adds a few views and jquery based mechanism to prolong the session time. It is using django in-build session mechanism, but invalidates the session if user has not done anything in a certain period of time. The advantage of it is that you do not have to make any requests to the server to prolong your session since it is working with the frontend part. Any action made by user is prolonging that time(mouse movements, clicks etc). In my opinion it is a little bit more convenient to have it working that way since when user reads quite long article or does other things that actually do not make requests to the server itself, he/she is being logged out. It also introduces warning screen which can remind user to be active.(it has no other effect) Also based on the documentation it should work even if the user work on the other tab.

Also its worth noting - session security auto-reloads page the the session expires, django itself doesn't

Putting those global/env variables on the spotlight:

ENABLE_SESSION_SECURITY - enables the package and all the changes introduced with this pr (it adds a jquery to pages that checks for the user activity and eventually prolongs the session +  sets the below explained variables to the values provided with .env file)

SESSION_SECURITY_WARN_AFTER - sets time of the warning pop up(the one that a screenshot is present in the PR message)

SESSION_SECURITY_EXPIRE_AFTER - sets actual session time expiration when user is idle(is not even moving his/her mouse)

SESSION_COOKIE_AGE - Django default setting that is responsible for the absolute time of the session, if it is not set explicitly the default value should be 2 weeks

rafalkosla101:
@bemoody @alistairewj There is also a middle ground between being secure(in some way) and UX. Setting only SESSION_COOKIE_AGE as env variable could change the absolute timeout. We could go with 24h, which seems reasonable for me since the default(2 weeks) is a little bit long in my opinion. Setting absolute timeout would actually get rid of some security risks. The whole idle timeout idea is more of a prevention of user bad behaviour than actual security issue.

rafalkosla101:
Separated absolute timeout to new PR(https://github.com/MIT-LCP/physionet-build/pull/1847)

alistairewj:
You probably need to rebase this since we try to avoid having a merge commit from dev into feature branches.

alistairewj:
Decided not to merge this as neither deployments would use session security currently.

PR #1821, opened 2023-01-19
Adds ref requirement for all users

lamawmouk:
This code base changes the credentialing process to require all users to submit reference as ref in #1804.

### DIFF
--- a/physionet-django/user/forms.py
+++ b/physionet-django/user/forms.py
@@ -585,7 +585,7 @@ class CredentialApplicationForm(forms.ModelForm):
                        data['reference_organization'],
                        data['reference_title']]
 
-        ref_required = data['researcher_category'] in [0, 1, 6, 7]
+        ref_required = True
         supervisor_required = data['researcher_category'] in [0, 1, 7]
         state_required = data['country'] in ['US', 'CA']
 
--- a/physionet-django/user/templates/user/credential_application.html
+++ b/physionet-django/user/templates/user/credential_application.html
@@ -24,7 +24,7 @@
   <hr>
   <p>The following steps will help to ensure a timely review of your application:</p>
   <ul>
-    <li>You are required to provide a reference, who we may contact with regard to your application.</li>
+    <li>You are required to provide a reference, whom we may contact with regard to your application.</li>
     <li>It is also helpful if your reference is a registered user and they have linked their ORCID iD to their profile.</li>
     <li>Check the form carefully before submission. <strong>Incomplete or erroneous applications will be rejected.</strong></li>
   </ul>
--- a/physionet-django/user/templates/user/credential_application.html
+++ b/physionet-django/user/templates/user/credential_application.html
@@ -24,8 +24,8 @@
   <hr>
   <p>The following steps will help to ensure a timely review of your application:</p>
   <ul>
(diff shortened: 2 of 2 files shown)


physionet-django/user/forms.py
```diff
@@ -585,7 +585,7 @@ def clean(self):
                        data['reference_organization'],
                        data['reference_title']]
 
-        ref_required = data['researcher_category'] in [0, 1, 6, 7]
+        ref_required = data['researcher_category'] in range(8)
```
tompollard: If everyone requires a reference, why check against range at all? Why not just set to True?

tompollard (commented):
Do you think we need to update any of the text on the form to make it clear that a reference is always required?

physionet-django/user/forms.py
lamawmouk: Thanks Tom good point. I can update ref_required=True

We would need to update text for the text. It is currently:

The following steps will help to ensure a timely review of your application:
- Listing a reference is always helpful. Remember to prompt your reference to expect an email about your application.
- It is also helpful if your reference is a registered user and they have linked their ORCID iD to their profile. Please provide the email associated with their account.

Update:
The following steps will help to ensure a timely review of your application:
- All users should list a reference. Remember to prompt your reference to expect an email about your application.
- It is also helpful if your reference is a registered user and they have linked their ORCID iD to their profile. 

I don't think we need "Please provide the email associated with their account." as we required an institutional/academic in the other PR.

Let me know if that works!

tompollard: @lamawmouk 

The following change sounds good to me:

> All users should list a reference. Remember to prompt your reference to expect an email about your application.

I'm not sure I agree with the following comment. Wouldn't we still prefer to have the email address linked with the PhysioNet account? (e.g. otherwise how would we find the linked ORCID?).

> I don't think we need "Please provide the email associated with their account." as we required an institutional/academic in the other PR.

tompollard: > I'm not sure I agree with the following comment. Wouldn't we still prefer to have the email address linked with the PhysioNet account? (e.g. otherwise how would we find the linked ORCID?).

I guess they may end up in a conflicted state (use gmail connected to account or institutional address?) so I can see why you are suggesting this change!

physionet-django/user/templates/user/credential_application.html
```diff
@@ -24,8 +24,8 @@ <h2>Recommendations for a timely review</h2>
   <hr>
   <p>The following steps will help to ensure a timely review of your application:</p>
   <ul>
-    <li>Listing a reference is always helpful. Remember to prompt your reference to expect an email about your application.</li>
-    <li>It is also helpful if your reference is a registered user and they have linked their ORCID iD to their profile. Please provide the email associated with their account.</li>
+    <li>All users should list a reference. Please provide their academic or institutional email. Remember to prompt your reference to expect an email about your application.</li>
```
tompollard: Please could you change this to: "You are required to provide a reference, who we may contact with regard to your application."

tompollard: Sorry Lama, should have thought more about this before making the suggestion. It should be "whom"!

Screenshot 2023-01-23 at 5 19 54 PM

lamawmouk: no problem, will update it

tompollard:

PR #1822, opened 2023-01-19
Fix HttpError 412 error when provisioning access to GCP resources for users

alistairewj:
Recently, we had a few error messages along the lines of:

```
Exception Type: HttpError at /projects/<project-slug>/<project-version>/request_access/4
Exception Value: <HttpError 412 when requesting https://admin.googleapis.com/admin/directory/v1/groups/<project-gcp-group-email>/members?alt=json returned "Condition not met">
```

These seem to occur when a user tries to request access to a project on GCP when their cloud e-mail is not a valid Google e-mail. This PR fixes this by:

1. Catching http errors from the Google Admin API rather than raising a 500
2. E-mailing the user that the request failed, and notifying them with an error message on the site
3. Add a special case for the 412 response code - the notification tells the user to verify their cloud e-mail is valid

### DIFF
--- a/physionet-django/notification/templates/notification/email/notify_gcp_access_request.html
+++ b/physionet-django/notification/templates/notification/email/notify_gcp_access_request.html
@@ -1,5 +1,6 @@
 Dear {{ user.get_full_name }},
 
+{% if successful %}
 {% if data_access.platform == 3 %}
 You have requested access to use {{ project }} in GCP Storage.
 
@@ -22,6 +23,18 @@ For information on how to access this resource please follow these steps
 3. You should now see the {{ SITE_NAME }} datasets in the left sidebar.
    - If you don't have access to a particular dataset, no information will be displayed when you click on the dataset name (and you may see a spinning circle icon).
    - If you cannot add the project "physionet-data", please access it through this link: https://console.cloud.google.com/bigquery?project=physionet-data&page=dataset
+{% endif %}
+{% else %}
+We are sorry to say that we were unable to grant access to {{ project }} in GCP.
+
+Reasons might include:
+ - An invalid cloud e-mail was specified. Please ensure the e-mail in your cloud profile is a valid
+ Google e-mail. Only Google e-mails may be used for GCP access.
+ - The service is no longer available.
+
+Please reapply for access if you are able to address the issue.
+
+If you think this was an error on our part, please contact {{ contact_email }}.
 {% endif %}
 
 {{ signature }}
--- a/physionet-django/project/utility.py
+++ b/physionet-django/project/utility.py
@@ -345,7 +345,9 @@ def grant_gcp_group_access(user, project, data_access):
             granted_access = True
             return '{0} was previously awarded to {1} for project: {2}'.format(
                 access, email, project), granted_access
-        raise error
+        else:
+            return 'Unable to grant {0} access to {1} for {2}'.format(
+                access, email, project), granted_access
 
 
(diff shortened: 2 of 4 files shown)


physionet-django/notification/templates/notification/email/notify_gcp_access_request.html
```diff
@@ -23,5 +24,17 @@
    - If you don't have access to a particular dataset, no information will be displayed when you click on the dataset name (and you may see a spinning circle icon).
    - If you cannot add the project "physionet-data", please access it through this link: https://console.cloud.google.com/bigquery?project=physionet-data&page=dataset
```
tompollard, line 25: In another PR we should fix this hardcoded "physionet" content in the template.

physionet-django/project/utility.py
```diff
@@ -315,6 +319,8 @@ def grant_gcp_group_access(user, project, data_access):
     email = user.cloud_information.gcp_email.email
     service = create_directory_service(settings.GCP_DELEGATION_EMAIL)
     access = ""
+    # Indicate whether we successfully granted access to the user
+    granted_access = False
```
tompollard, line 341: Minor thing but for readability it would be nice if `granted_access = False` wasn't separating `access = ""` from its assignment in the if clause below.

physionet-django/project/utility.py
```diff
@@ -329,15 +335,27 @@ def grant_gcp_group_access(user, project, data_access):
...
     except HttpError as error:
+        if error.resp.status == 412:
+            # Google has a somewhat cryptic 412 error: "Condition not met"
+            # In our experience, this occurred when the user specified a non-Google
+            # e-mail in their cloud profile which could not be added to the group.
+            return (
+                'Unable to provision access, please verify '
+                '{0} is a valid Google account'.format(email)
+            ), granted_access
         if json.loads(error.content)['error']['message'] == 'Member already exists.':
```
tompollard: Doesn't really matter, but perhaps `elif` instead of `if`? No need to check that member already exists if we've already seen the 412 error

tompollard (approved):
Thanks Alistair, looks good to me! Added a couple of minor comments. Do you want to address either of these before we merge?

physionet-django/project/utility.py
alistairewj: In that case the `return` statement would exit before the following `if`, so this is mostly a style choice

tompollard: oh yep sorry

physionet-django/project/utility.py
alistairewj: More than readability, this needs a fix!

alistairewj:
I refactored the utility functions for clarity and to fix the bug you spotted. The last commit has the changes done in the refactor if you are interested.

physionet-django/notification/templates/notification/email/notify_gcp_access_request.html
alistairewj: Agreed - though admittedly low priority as we don't use this in our deployment. Probably needs an env var as the GCP project is not specified elsewhere in the platform.

tompollard: Raised issue at: https://github.com/MIT-LCP/physionet-build/issues/1830

tompollard:
Thanks Alistair, looks good! Hopefully @Chrystinne can do some more cleaning up of this later as part of the AWS integration.

PR #1823, opened 2023-01-19
Fix/refactor published project template

superryeti:
Context: We wanted to clean the published project template and remove redundancy and possibly remove the if else logic from template and  add those logic to views.

Summary of Change
1. I added a boolean variable  `has_access_to_files` in view which will replace the different block for open and credential block on template.
The boolean will be true if the current user should be able to access to the project.

2. Also formatted the part of template i made changes(to make it readable)

If the user has access to project, they can see the files, and will have same access to all the download/access options before this change.

Notes:
1. For the previous other access stuff, i removed the duplicated code across the credentialed and open block. and merged all the code inside the if boolean variable block.

2. For the gcp access stuff, It seems to be something only open access project are concerned about(this was very clear in the template before this commit). However its now not possible to make that assumption from the template(However the UI should look and work exactly the same)

4. Also, there is a some kind of redundancy in the template for Open access projects for the download of zipped file(check code below)

<li><a href="{% static project.zip_url %}">Download the ZIP file</a> ({{ compressed_size }})</li>

image

PS1 : The current code does remove redundancy, and makes the template cleaner. But one might argue that it takes away readability from the template.

PS2 : This is just a test commit to get feedback. i will later work on this properly and squash commits, and open a PR.(Didnot want to spend more time on this as we might not end up refactoring the credentialed and open block)

### DIFF
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -372,7 +372,43 @@
                     {% endif %}
 
                     <ul>
-
+                    {% if project.access_policy %}
+                        {% if project.compressed_storage_size %}
+                            <li><a href="{% url 'serve_published_project_zip' project.slug project.version %}">Download
+                                the ZIP file</a> ({{ compressed_size }})
+                            </li>
+                        {% endif %}
+                    {% else %}
+                        {% if project.gcp %}
+                            {% if project.gcp.sent_zip %}
+                                <!-- Temporary fix: Change default location to local servers -->
+                                <!--
+              <li><a href="https://storage.googleapis.com/{{ project.gcp.bucket_name }}/{{ project.zip_name }}">
+                  Download the ZIP file</a> ({{ compressed_size }}).</a></li>
+             -->
+                                <li><a href="{% static project.zip_url %}">Download the ZIP file</a>
+                                    ({{ compressed_size }})
+                                </li>
+                            {% elif project.compressed_storage_size %}
+                                <li><a href="{% static project.zip_url %}">Download the ZIP file</a>
+                                    ({{ compressed_size }})
+                                </li>
+                            {% endif %}
+                            {% if project.gcp.sent_files %}
+                                <li>Access the files using the Google Cloud Storage Browser <a
+                                        href="https://console.cloud.google.com/storage/browser/{{ project.gcp.bucket_name }}/">here</a>.
+                                    Login with a Google account is required.
+                                <li>Access the data using the Google Cloud command line tools (please refer to the <a
+                                        href="https://cloud.google.com/storage/docs/gsutil_install">gsutil</a>
+                                    documentation for guidance):
+                                    <pre class="shell-command">gsutil -m -u YOUR_PROJECT_ID cp -r gs://{{ project.gcp.bucket_name }} DESTINATION</pre>
+                                </li>
+                            {% endif %}
+                        {% elif project.compressed_storage_size %}
+                            <li><a href="{% static project.zip_url %}">Download the ZIP file</a> ({{ compressed_size }})
+                            </li>
+                        {% endif %}
+                    {% endif %}
                     {% include "project/published_project_data_access.html" %}
                     {% if is_wget_supported %}
                             <li>Download the files using your terminal:
@@ -405,11 +441,6 @@
                         {# ZIP START #}
 
                         <ul>
-                            {% if project.compressed_storage_size %}
-                                <li><a href="{% url 'serve_published_project_zip' project.slug project.version %}">Download
-                                    the ZIP file</a> ({{ compressed_size }})
-                                </li>
-                            {% endif %}
                         </ul>
                         {# ZIP END #}
                     {% else %}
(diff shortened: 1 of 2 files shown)


alistairewj:
There's some reformatting that you did (further indentation) that makes the diff log a bit harder to read - might be worth it to undo the reformatting.

> 2\. For the gcp access stuff, It seems to be something only open access project are concerned about(this was very clear in the template before this commit). However its now not possible to make that assumption from the template(However the UI should look and work exactly the same)

The open access projects which have a zip file sent to GCP seem to have a temporary fix from you @tompollard - https://github.com/MIT-LCP/physionet-build/commit/eec7147df

Since it doesn't seem like we want to push folks to GCP by default anytime soon, it seems OK to drop that specific functionality now. It's better in my mind anyway - if pushing folks to GCP, it would be better if they can natively access the data files themselves, rather than only being able to download a zip. We can use the existing request_access functionality to grant users access, and improve the GCP data send feature on the project management page so it's easier for administrators to create the buckets.

So, my vote is that the subtemplate is definitely worth including


> * I added a boolean variable  `has_access_to_files` in view which will replace the different block for open and credential block on template.
>   The boolean will be true if the current user should be able to access to the project.
> *

This I'm less sure about. I don't really like having `has_access` and `has_access_to_files`. The logic of the template could be moved into the view or the method without adding another variable. It does seem like `has_access` can be used to gate all the access (rather than using ifs with deprecated files, embargo, and access_policy is not None).

While on the topic, @briangow - was there a reason that you put the embargo gate-keeping in the template and not in the project.has_access method?

superryeti:
> There's some reformatting that you did (further indentation) that makes the diff log a bit harder to read - might be worth it to undo the reformatting.

@alistairewj  Good point, Let me undo that.

bemoody:
Sorry, I cannot follow what changes you are making in the templates.  Please break down your changes into small pieces (one change per commit) and explain each change.

For the change in project/views.py, I do not see what the distinction is between `has_access` and `has_access_to_files`.

superryeti:
> Sorry, I cannot follow what changes you are making in the templates. Please break down your changes into small pieces (one change per commit) and explain each change.
> 
> For the change in project/views.py, I do not see what the distinction is between `has_access` and `has_access_to_files`.

@bemoody  Good point.  Sorry about the mess. I will do that and do a force push with smaller commits and better explanation

Quick answer is that we had a lot of repeated code in the template(between the if statement and else statement for open access and credentialed access). My idea was to add another boolean that decides if the current user should be able to access the files and do that check in the view.

So in summary, with `has_access_to_files`, i was trying to bring all the code in template under 1  if statement instead of if and  else statement which had a lot of repeated code
```
 {% if project.compressed_storage_size %}
                    <li><a href="{% url 'serve_published_project_zip' project.slug project.version %}">Download the
                        ZIP file</a> ({{ compressed_size }})
                    </li>
                {% endif %}
```

or the wget stuff or the data access stuff

bemoody:
> My idea was to add another boolean that decides if the current user should be able to access the files and do that check in the view.

But that's what has_access is... isn't it?

superryeti:
> > My idea was to add another boolean that decides if the current user should be able to access the files and do that check in the view.
> 
> But that's what has_access is... isn't it?

Yes, That is correct about what you said about  `has_access`. The idea was that between the open and credentialed type of project a lot of code is repeated, and we wanted to try to make it simpler

**I  just got what you are trying to say (Duh -  stupid me) . i should have used the `has_access` (its already doing that, i should have  just used it in the view instead of adding a new boolean.**

image

amitupreti:
@bemoody , I finished refactoring the project template smaller and understandable commits

Quick Note, I did reformat the whole html before starting to refactor code(just to make it easier for me to read the code and also for the reviewer to easily follow commits, since the the proper formatting should help to avoid mistakes when moving code around)

Please follow the commits to understand the changes(if we see the overall changes, its pretty weird, confusing and scary)

superryeti:
@bemoody  Could you please take a look?
I refactored the code in small and understandable commits.

In Summary, I only refactored the template and merged the scattered code between (if `credentialed project`and else `open access project`  block into a single if `has_access` block and removed the redundant code.

Please check the commits to follow the changes.

physionet-django/project/templates/project/published_project.html
```diff
@@ -7,452 +7,462 @@
...
+                            {% endif %}
+                        {% elif project.compressed_storage_size %}
+                            <li><a href="{% static project.zip_url %}">Download the ZIP file</a> ({{ compressed_size }})
+                            </li>
+                        {% endif %}
+                    {% endif %}
+                    {% include "project/published_project_data_access.html" %}
+                    {% if is_wget_supported %}
+                            <li>Download the files using your terminal:
+                                <pre class="shell-command">wget -r -N -c -np {% if project.access_policy %} --user {{ user }} --ask-password {% endif %}
```
tompollard: Something has changed in the formatting of this chunk (though I can't immediately see what!).

Compare current:

Screenshot 2023-02-08 at 11 49 25 AM

... with new:

Screenshot 2023-02-08 at 11 57 29 AM

tompollard (commented):
This looks good to me, other than the formatting issue mentioned inline.

bemoody:
Yeah, inside a `pre` you can't just ignore line breaks (or spaces).  The whole command should be written on one line - I think it should be:
```
<pre class="shell-command">wget -r -N -c -np{% if project.access_policy %} --user {{ user }} --ask-password{% endif %} {{ bulk_url_prefix }}{% url 'serve_published_project_file' project.slug project.version '' %}</pre>
```

I hate to say it, but it would be better if you didn't re-indent the entire file.  (Did you do that all by hand or use a tool to indent?)  This project uses (more or less) a standard indentation style of two spaces both for HTML elements and for template block tags.  We don't want to have different indentation styles for different files if we can avoid it.  In this case it'd be easiest to just add another commit to re-indent, after fixing the `pre` issue.

physionet-django/project/templates/project/published_project.html
superryeti: Thank you, checking, thats so weird.

superryeti:
> to have different indentation styles for different files if we

Thats a good point, i used pycharm's autoindent, Let me fix the `pre` tag issue and reindent with the standard 2 space indent. (i think i did the auto indent for a new html i had created in another pr i will reindent over there as well)

Thank you @bemoody.

physionet-django/project/templates/project/published_project.html
superryeti: fixed this as per Benjamin's suggestion

superryeti:
Thank you @tompollard  @bemoody , i reindent the template with 2 spaces after fixing the `pre` tag issue.

tompollard:
Thanks Amit, looks good to me! @bemoody?

tompollard:
Thanks Amit, it looks like you've addressed Benjamin's feedback, so I'm going to go ahead and merge.

PR #1825, opened 2023-01-19
Cleanup Training Templates

superryeti:
There are two identical template files in physionet-django/project/templates/project:

published_project_required_training.html
published_project_required_trainings.html

It looks like the plural one was used in the code, However the singular one looked like a better name. So I removed the plural one and changed code to use the singular named template


physionet-django/project/views.py
```diff
@@ -1732,15 +1732,15 @@ def published_project_license(request, project_slug, version):
...
     """Displays a published project's required training"""
     project = get_object_or_404(PublishedProject, slug=project_slug, version=version)
 
     required_trainings = project.required_trainings.all()
 
     return render(
         request,
-        'project/published_project_required_trainings.html',
+        'project/published_project_required_training.html',
         {'project': project, 'required_trainings': required_trainings},
```
tompollard: I think this line needs updating to match the new template?

`{'project': project, 'required_trainings': required_trainings}` should be:
 `{'project': project, 'required_training': required_trainings}`

tompollard (commented):
Thanks @amitupreti! It looks like there is a slight difference between the templates (`{% for training in required_training %}` vs `{% for training in required_trainings %}`). Other than that, looks good to me.

physionet-django/project/views.py
superryeti: Thanks Tom. I will make the changes

Good Catch

tompollard:

PR #1826, opened 2023-01-20
Add Event Update Features

superryeti:
On PR https://github.com/MIT-LCP/physionet-build/pull/1729, Dana started working on the event update feature.

This PR adds on top of that work.

The following changes are added:

1. Event update form Updated the `save` method in `AddEventForm` so that it will work for both creating and updating events. The form will now check if the event is being created or updated. If the event is being updated, it will update the event instead of creating a new one.

2. Event Home template(event_home.html) I updated the modal for event update, and added javascript codes that renders the data in the update event modal by sending a get request to url `get_event_details`. The url loads the event details and returns a json response. The javascript then renders the data in the modal.

3. Event Views and urls updated views and urls.py to support the event update feature.

Question: Do we add a test for event create and update event now?(or should we wait till we add other features for events). probably now sound like a good idea, as it will be a simple test now and we can add more cases as we add features

### DIFF
--- a/physionet-django/events/forms.py
+++ b/physionet-django/events/forms.py
@@ -24,11 +24,15 @@ class AddEventForm(forms.ModelForm):
         super(AddEventForm, self).__init__(*args, **kwargs)
 
     def save(self):
-        Event.objects.create(title=self.cleaned_data['title'],
-                             category=self.cleaned_data['category'],
-                             host=self.host,
-                             description=self.cleaned_data['description'],
-                             start_date=self.cleaned_data['start_date'],
-                             end_date=self.cleaned_data['end_date'],
-                             allowed_domains=self.cleaned_data['allowed_domains']
-                             )
+        # Handle updating the event
+        if self.initial:
+            self.instance.save()
+        else:
+            Event.objects.create(title=self.cleaned_data['title'],
+                                 category=self.cleaned_data['category'],
+                                 host=self.host,
+                                 description=self.cleaned_data['description'],
+                                 start_date=self.cleaned_data['start_date'],
+                                 end_date=self.cleaned_data['end_date'],
+                                 allowed_domains=self.cleaned_data['allowed_domains']
+                                 )
--- a/physionet-django/events/forms.py
+++ b/physionet-django/events/forms.py
@@ -26,7 +26,13 @@ class AddEventForm(forms.ModelForm):
     def clean(self):
         cleaned_data = super(AddEventForm, self).clean()
         if Event.objects.filter(title=cleaned_data['title'], host=self.host).exists():
-            raise forms.ValidationError({"title": ["Event with this title already exists"]})
+            # in case of update, we don't want to raise an error if the title is the same
+            if self.initial.get('title'):  # in case of update we have an initial value
+                # if the instance title is different from the new title, then we have a duplicate with another event
+                if self.instance.title != cleaned_data['title']:
+                    raise forms.ValidationError({"title": ["Event with this title already exists"]})
+            else:
+                raise forms.ValidationError({"title": ["Event with this title already exists"]})
 
     def save(self):
         # Handle updating the event
--- a/physionet-django/events/forms.py
+++ b/physionet-django/events/forms.py
@@ -23,6 +23,11 @@ class AddEventForm(forms.ModelForm):
         self.host = user
         super(AddEventForm, self).__init__(*args, **kwargs)
 
+    def clean(self):
+        cleaned_data = super(AddEventForm, self).clean()
+        if Event.objects.filter(title=cleaned_data['title'], host=self.host).exists():
+            raise forms.ValidationError({"title": ["Event with this title already exists"]})
+
     def save(self):
         # Handle updating the event
         if self.initial:
--- a/physionet-django/events/templates/events/event_home.html
+++ b/physionet-django/events/templates/events/event_home.html
@@ -55,6 +55,26 @@
     </div>
   </div>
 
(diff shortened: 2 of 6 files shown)


physionet-django/events/views.py
```diff
@@ -10,6 +11,32 @@
...
+@login_required
+def update_event(request, event_slug, **kwargs):
+    user = request.user
+    is_instructor = user.has_perm('user.add_event')
+    if request.method == 'POST':
+        event = Event.objects.get(slug=event_slug)
+        event_form = AddEventForm(user=user, data=request.POST, instance=event)
+        if event_form.is_valid() and is_instructor:
+            event_form.save()
+            messages.success(request, "Updated Event  Successfully")
```
tompollard: `"Updated Event  Successfully"` -> `"Updated Event Successfully"`

physionet-django/events/templates/events/event_home.html
```diff
@@ -148,4 +169,32 @@ <h5 class="modal-title">Participants</h5>
...
+   $('.edit-btn').click(function(e) {
+       $('#edit-event-modal').modal('hide');
+       let url = $(this).data('event-url');
+       let form_url = $(this).data('form-url');
+        $.ajax({
+           url: url,
+           type: 'GET',
+           success: function(data) {
+                console.log(data);
+                let event = data[0];
```
tompollard, line 185: `console.log` for debugging?

physionet-django/events/templates/events/event_home.html
```diff
@@ -74,6 +94,7 @@ <h3>{{ event.title }}</h3>
               <small>Share the class code: {{ url_prefix }}{% url 'event_add_participant' event.slug %} </small><br>
               </p>
               <button class="btn btn-sm btn-primary" data-toggle ="modal" data-target="#add-participant-modal-{{ event.id }}">View participants</button>
+              <button class="btn btn-sm btn-secondary edit-btn" data-toggle ="modal" data-target="#edit-event-modal"  data-event-url="{% url 'get_event_details' event.slug %}" data-form-url="{% url 'update_event' event.slug %}" >Edit Event</button>
```
tompollard, line 97: not big deal but there is some odd spacing here (e.g. `data-toggle ="modal"`)

tompollard (commented):
This is great, thanks Amit!

Currently an error is raised when someone attempts to create/edit an event
with an existing title (`IntegrityError at /events/. UNIQUE constraint failed: events_event.title, events_event.host_id`). Are you able to catch this and prevent the form being saved until the title is unique?



> Do we add a test for event create and update event now?(or should we wait till we add other features for events). probably now sound like a good idea, as it will be a simple test now and we can add more cases as we add features

Yeah, I think it would be good to add one or two tests now. Please could you add to the PR?

physionet-django/events/templates/events/event_home.html
```diff
@@ -55,6 +55,26 @@ <h5 class="modal-title">Add Event</h5>
...
+              <h5 class="modal-title">Edit Event</h5>
+              <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                <span aria-hidden="true">&times;</span>
+              </button>
+            </div>
+            <div class="container">
+              <form action="" method="post" class="form-signin edit-event" >
+                {% csrf_token %}
+                {% include "form_snippet.html" with form=event_form %}
+                <button class="btn btn-primary btn-fixed" name="edit-event" type="submit">Edit event</button>
```
tompollard: I find the "Edit event" text on the button slightly odd, though can't explain why! Should we change it to something like "Save" instead?

superryeti: > This is great, thanks Amit!
> 
> Currently an error is raised when someone attempts to create/edit an event with an existing title (`IntegrityError at /events/. UNIQUE constraint failed: events_event.title, events_event.host_id`). Are you able to catch this and prevent the form being saved until the title is unique?
> 
> > Do we add a test for event create and update event now?(or should we wait till we add other features for events). probably now sound like a good idea, as it will be a simple test now and we can add more cases as we add features
> 
> Yeah, I think it would be good to add one or two tests now. Please could you add to the PR?

Thank you for the feedback. Working on it.

superryeti:
@tompollard  Thank you for your feedbacks. I made the following changes

1. Fix minor issues like leftover debugging statements, weird spaces etc
2. Fix the integrity error(please check the [commit message](https://github.com/MIT-LCP/physionet-build/pull/1826/commits/29870ca899314f7e0f69f85b124896328ce3eb30) for explanation of changes)
3. Add test case for add, edit views

physionet-django/events/forms.py
```diff
@@ -23,12 +23,21 @@ def __init__(self, user, *args, **kwargs):
         self.host = user
         super(AddEventForm, self).__init__(*args, **kwargs)
 
+    def clean(self):
+        cleaned_data = super(AddEventForm, self).clean()
+        if Event.objects.filter(title=cleaned_data['title'], host=self.host).exists():
```
tompollard, line 28: this clause raises an error if the title matches the event being edited. this shouldn't happen. 

i.e. currently if i click "edit event", i cannot save the the content unless i change the title.

tompollard (commented):
Thanks Amit, one small issue as mentioned inline (our event form validation prevents events being edited unless the title changes). Other than this, looks good to me.

physionet-django/events/forms.py
superryeti: Ah thats a very good point. Let me fix that,  Good catch

tompollard:
Looks great, thanks @amitupreti

PR #1827, opened 2023-01-23
Use app label when searching for permissions

bemoody:
Django "permission" objects are used to identify actions that a particular person or group is allowed to perform.

A permission is identified by a "codename" (like "can_edit_activeprojects" or "view_event_menu") as well as a "content type" (an internal Django object that corresponds to a particular model class, such as project.models.ActiveProject or events.models.Event.)

Although the "codename" should ideally be descriptive enough to be unique across the site, there's nothing preventing two Django apps from using the same codename; and as we found with the event migration, there's nothing to automatically remove or disable obsolete permissions or content types.

The general Django style is to identify permissions by app label *and* codename (see e.g. the `permission_required` decorator or the `perms` built-in template variable.)  So our `User.get_users_with_permission` function should do the same thing, as should `Event.enroll_user`.

This should solve the immediate issue with people being unable to join events.

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -86,7 +86,9 @@ class AssignEditorForm(forms.Form):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


tompollard:
Looks good, thanks Benjamin

PR #1831, opened 2023-01-24
Feature: Add logging module for actions performed inside admin panel

rafalkosla101:
This PR adds logging configuration and the module itself to post admin activity to the standard output, so for example google platform could collect them. The log format is. as below:
image
The whole implementation is based on the LogEntry model that is django build-in model, which is saving any changes to the database. I am simply redirecting those entries to the output with a more specific format.

The main question is - Should it be an optional feature or logging those events is actually important and we should be adding this permanently?

### DIFF
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -428,6 +428,9 @@ LOGGING = {
             'format': '%(created)s %(name)s %(levelname)s %(message)s %(sinfo)s',
             '()': 'physionet.log.UwsgiJsonFormatter',
         },
+        'admin': {
+            'format': '%(levelname)s %(asctime)-15s %(user)s %(message)s'
+        }
     },
     'handlers': {
         'console': {
@@ -451,6 +454,12 @@ LOGGING = {
             'filters': ['require_debug_false'],
             'class': 'physionet.log.SaferAdminEmailHandler',
         },
+        'admin_logging': {
+            'level': 'INFO',
+            'class': _class,
+            'formatter': 'admin',
+            'stream': _logfile,
+        },
     },
     'loggers': {
         '': {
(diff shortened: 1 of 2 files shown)


tompollard:
Currently, deleting a Group called BLAH displays:

> `INFO 2023-01-25 23:43:14,945 admin Deleted “BLAH.”`

Adding a new TrainingType called blah displays:

> `INFO 2023-01-25 23:44:02,584 admin Added “BLAH”.`

Could we also capture the following details in the log?

1. the app (e.g. project, user, event, etc) and;
2. the model (e.g. ActiveProject, TrainingType, etc)? 

If so, I think the logs would be more useful.

bemoody:
Hmm.  Looks like it uses `str()` of the model.

https://github.com/django/django/blob/4.1.5/django/contrib/admin/models.py#L96
```
            return gettext("Deleted “%(object)s.”") % {"object": self.object_repr}
```

https://github.com/django/django/blob/4.1.5/django/contrib/admin/models.py#L67
```
    # Translators: 'repr' means representation
    # (https://docs.python.org/library/functions.html#repr)
```

https://github.com/django/django/blob/4.1.5/django/contrib/admin/options.py#L946
```
    def log_deletion(self, request, obj, object_repr):
```

but...

https://github.com/django/django/blob/4.1.5/django/contrib/admin/options.py#L2139
```
            obj_display = str(obj)
            attr = str(to_field) if to_field else opts.pk.attname
            obj_id = obj.serializable_value(attr)
            self.log_deletion(request, obj, obj_display)
```

Arguably using `repr()` would make more sense, though I can also see that if this information is getting stored in the database then it is risky to *keep* potentially sensitive information that the administrator meant to remove.

We could, and maybe should, change `__str__` of our models to return something more descriptive (and/or get rid of `__str__` entirely.)  There's stuff in templates that does things like `{{ project }}` when it would be better to do `{{ project.title }}` instead.

I don't think we can change the behavior for Django builtin models like Group, though.

tompollard:
> Hmm. Looks like it uses str() of the model.

Hmm, thanks for looking into this. Perhaps this PR is good to merge then, and we should raise an issue to fix the `str()` methods and templates? @bemoody

tompollard:
Thanks Rafal!

PR #1832, opened 2023-01-24
Fix File uploads on gcp uploads

superryeti:
Context: The `generate_signed_url` view is responsible to generate the signed url needed to upload the file to the gcp bucket. This view also checks the file names and rejects files names with non-ascii characters, files names with spaces etc. On this PR(https://github.com/MIT-LCP/physionet-build/pull/1768) Amit(me :D) started using `validate_filename` function to validate the file names. (Details on the need for that change is on that PR).

Problem: The above PR broke the GCP file uploads because gcp allows file names with `/` in them (https://cloud.google.com/storage/docs/objects) . The `validate_filename` function rejects file names with `/` in them.

Fix: The current change uses another existing function `validate_subdir` which has similar validation logic(the `validate_subdir` doesnot check for filename length) as `validate_filename` but doesn't reject file names with `/` in them.

Question: Do we want to add file length validation like filenames can have 100 characters only? I didnot go ahead and add such limit because gcp allows you to specify path like `folder/subfolder/subsubfolder/myfile.txt`. So that might sometime stop someone from uploading files in case they have a lot of nested folders.

### DIFF
--- a/physionet-django/project/test_views.py
+++ b/physionet-django/project/test_views.py
@@ -1150,7 +1150,8 @@ class TestGenerateSignedUrl(TestMixin):
         cls.invalid_size_data_3 = {'filename': 'random.txt'}
         cls.invalid_filename_data_1 = {'size': 250000, 'filename': 'ran dom.txt'}
         cls.invalid_filename_data_2 = {'size': 250000, 'filename': 'random§.txt'}
-        cls.valid_data = {'size': 250000, 'filename': 'random.txt'}
+        cls.invalid_filename_length_data_1 = {'size': 250000, 'filename': 'invalid' * 100 + '.txt'}
+        cls.valid_data = {'size': 250000, 'filename': 'folder1/folder2/random.txt'}
 
     def test_invalid_size(self):
         self.client.login(**self.user_credentials)
@@ -1185,6 +1186,11 @@ class TestGenerateSignedUrl(TestMixin):
 
         with self.subTest('Non-numeric file size returns a bad request.'):
             response = self.client.post(self.url, self.invalid_size_data_2, format='json')
+
+            self.assertEqual(response.status_code, HTTPStatus.BAD_REQUEST)
+
+        with self.subTest('A filename cannot be longer than 256 characters.'):
+            response = self.client.post(self.url, self.invalid_filename_length_data_1, format='json')
 
             self.assertEqual(response.status_code, HTTPStatus.BAD_REQUEST)
 
--- a/physionet-django/project/validators.py
+++ b/physionet-django/project/validators.py
@@ -5,6 +5,7 @@ from django.utils.translation import gettext as _
 
 MAX_FILENAME_LENGTH = 100
 MAX_PROJECT_SLUG_LENGTH = 30
+MAX_GCS_OBJECT_NAME_LENGTH = 256
 
 _good_name_pattern = re.compile(r'\w+([\w\-\.]*\w+)?', re.ASCII)
 _bad_name_pattern = re.compile(r'^(?:con|nul|aux|prn|com\d|lpt\d)(?:\.|$)',
@@ -127,3 +128,33 @@ def validate_topic(value):
     """
     if not re.fullmatch(r'[a-zA-Z\d][\w -]*', value):
         raise ValidationError('Letters, numbers, spaces, underscores, and hyphens only. Must begin with a letter or number.')
+
+
+def validate_gcs_bucket_object(filename, size):
+    """
+    Validate a filename and size for upload to gcs bucket objects.
+
+    File names may be at most 256 characters long.
+    Only letters, numbers, dashes, underscores, dots, and / allowed,
+    or empty. No consecutive dots or fwd slashes.
+
+    File size cannot be negative.
+    """
+    if not filename or not size:
+        raise ValidationError('You must provide the filename and the size.')
+
+    if not size.isnumeric():
+        raise ValidationError('The size must be a numeric value.')
+
+    size = int(size)
+    if size <= 0:
+        raise ValidationError('The file size cannot be a negative value.')
+
+    if len(filename) > MAX_GCS_OBJECT_NAME_LENGTH:
+        raise ValidationError(
+            f'Invalid file name "{filename}". '
+            f'File names may be at most {MAX_GCS_OBJECT_NAME_LENGTH} characters long.'
+        )
+
+    for path_segment in filename.split('/'):
+        validate_filename(path_segment)
--- a/physionet-django/project/views.py
+++ b/physionet-django/project/views.py
@@ -58,7 +58,6 @@ from user.models import CloudInformation, CredentialApplication, LegacyCredentia
 from django.db.models import F, DateTimeField, ExpressionWrapper
 
 LOGGER = logging.getLogger(__name__)
-MAX_GCS_OBJECT_NAME_LENGTH = 256
 
 def project_auth(auth_mode=0, post_auth_mode=0):
     """
--- a/physionet-django/project/views.py
+++ b/physionet-django/project/views.py
@@ -51,13 +51,14 @@ from project.models import (
     UploadedDocument,
 )
 from project.projectfiles import ProjectFiles
-from project.validators import validate_filename
+from project.validators import validate_filename, validate_gcs_bucket_object
 from user.forms import AssociatedEmailChoiceForm
 from user.models import CloudInformation, CredentialApplication, LegacyCredential, User, Training
 
 from django.db.models import F, DateTimeField, ExpressionWrapper
 
 LOGGER = logging.getLogger(__name__)
+MAX_GCS_OBJECT_NAME_LENGTH = 256
 
 def project_auth(auth_mode=0, post_auth_mode=0):
     """
@@ -2285,21 +2286,12 @@ def generate_signed_url(request, project_slug):
     filename = request.POST.get('filename')
     size = request.POST.get('size')
 
-    if not filename or not size:
-        return JsonResponse({'detail': 'You must provide the filename and the size.'}, status=400)
-
-    if not size.isnumeric():
-        return JsonResponse({'detail': 'The file size must be a numeric value.'}, status=400)
-
-    size = int(size)
-    if size <= 0:
-        return JsonResponse({'detail': 'The file size cannot be a negative value.'}, status=400)
-
     try:
-        validate_filename(filename)
+        validate_gcs_bucket_object(filename, size)
     except ValidationError as e:
         return JsonResponse({'detail': e.messages}, status=400)
 
+    size = int(size)
     queryset = ActiveProject.objects.all()
     project = get_object_or_404(queryset, slug=project_slug)
 
--- a/physionet-django/physionet/storage.py
+++ b/physionet-django/physionet/storage.py
@@ -27,8 +27,7 @@ def generate_signed_url_helper(blob_name, size, expiration=dt.timedelta(days=1),
         str: The signed URL.
     """
     storage = MediaStorage()
-    bucket = storage.bucket()
-    blob = bucket.blob(blob_name)
+    blob = storage.bucket.blob(blob_name)
 
     url = blob.generate_signed_url(
         api_access_endpoint='https://storage.googleapis.com',


kshalot (changes requested):
@amitupreti the changes themself look good to me ✅ The max filesize of a GCS object is 1024 bytes (when UTF-8 encoded) which makes the lower bound of 256 Unicode characters (unless I forgot how encoding works). I took a look and I don't see any preflight validation in `google.cloud.storage` itself. It won't hurt to enforce a sensible limit here.

On a different note - could you modify the tests for this view just so we prevent such errors from happening in the future? If you add the name length validation then that should be tested as well. There is already a test case defined for this function. adding those new conditions will be a breeze

bemoody:
> @amitupreti the changes themself look good to me ✅ The max filesize of a GCS
> object is 1024 bytes (when UTF-8 encoded) which makes the lower bound of 256
> Unicode characters (unless I forgot how encoding works, but AFAIR a UTF-8
> character is 1-4 bytes long).

Does that 1024-byte limit include project slug and/or version, in
addition to the file path?

superryeti:
> On a different note - could you modify the tests for this view just so we prevent such errors from happening in the future? If you add the name length validation then that should be tested as well. There is already a test case defined for this function. adding those new conditions will be a breeze

Noted, Updating the tests.

kshalot:
> @amitupreti the changes themself look good to me ✅ The max filesize of a GCS object is 1024 bytes (when UTF-8 encoded) which makes the lower bound of 256 Unicode characters (unless I forgot how encoding works, but AFAIR a UTF-8 character is 1-4 bytes long).
> Does that 1024-byte limit include project slug and/or version, in addition to the file path?

It's a limit on the object name itself so it does not take project slug/version into account (as these are in the bucket name in our implementation)

kshalot:
@amitupreti There seems to be another issue with the upload:
https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/storage.py#L30

`storage.bucket` returns an **instance** of `Bucket`, it's not callable. The parenthesis have to be removed.

superryeti:
> @amitupreti There seems to be another issue with the upload: https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/storage.py#L30
> 
> `storage.bucket` returns an **instance** of `Bucket`, it's not callable. The parenthesis have to be removed.

@kshalot  Thank you. I seem to have missed that. i remember checking the code in google python library. but missed it somehow. 

checking and Fixing that as well

superryeti:
> returns an **instance** of `Bucket`

@kshalot  quick question about this.

I was following the code and the `storage.bucket` seems to lead to a method. So shouldn't there be parenthesis as we want to call that method which seems to lead to the `Bucket` Instance of google cloud ? 

If we dont pass the parenthesis, wont we just get a reference or pointer the the bucket function on `django-storage`

image
image
image

I think this `GoogleCloudStorage` comes from the `django-storage` module

kshalot:
@amitupreti The `bucket` method in `GoogleCloudStorage` is decorated with the [property](https://docs.python.org/3/library/functions.html#property) decorator.

superryeti:
> @amitupreti The `bucket` method in `GoogleCloudStorage` is decorated with the [property](https://docs.python.org/3/library/functions.html#property) decorator.

@kshalot  Wow Thanks. I had no idea that the property decorator does that. Thats so cool.

superryeti:
@kshalot  Thank you so much for all the help and feedback. Could you please review the PR. i have made the recommended changes

physionet-django/project/test_views.py
```diff
@@ -1150,7 +1150,8 @@ def setUpTestData(cls):
         cls.invalid_size_data_3 = {'filename': 'random.txt'}
         cls.invalid_filename_data_1 = {'size': 250000, 'filename': 'ran dom.txt'}
         cls.invalid_filename_data_2 = {'size': 250000, 'filename': 'random§.txt'}
-        cls.valid_data = {'size': 250000, 'filename': 'random.txt'}
+        cls.invalid_filename_length_data_2 = {'size': 250000, 'filename': 'invalid' * 100 + '.txt'}
```
kshalot: Should probably be called `...data_1`, not `...data_2`.

physionet-django/project/views.py
```diff
@@ -51,13 +51,14 @@
...
 from project.projectfiles import ProjectFiles
-from project.validators import validate_filename
+from project.validators import validate_filename, validate_subdir
 from user.forms import AssociatedEmailChoiceForm
 from user.models import CloudInformation, CredentialApplication, LegacyCredential, User, Training
 
 from django.db.models import F, DateTimeField, ExpressionWrapper
 
 LOGGER = logging.getLogger(__name__)
+MAX_FILENAME_LENGTH_GCP_BUCKET = 256
```
kshalot: This name is misleading - it's not related to a GCS bucket, but an object inside a GCS bucket.

Now that I think of it, we should opt for a new validator function here. For example `validate_gcs_object_name` or `validate_bucket_object_name`. I think it's clearer this way, since now we also have to check for the length. What do you think?

kshalot (changes requested):

physionet-django/project/test_views.py
superryeti: oh sorry. yes. that is correct. fixing this

physionet-django/project/views.py
superryeti: I think thats a good idea. We can have gcs related validation over there and we can just add more logic if we need to update the validation in future.

kshalot: Exactly 👍

physionet-django/project/validators.py
```diff
@@ -127,3 +128,19 @@ def validate_topic(value):
...
+    Only letters, numbers, dashes, underscores, dots, and / allowed,
+    or empty. No consecutive dots or fwd slashes.
+    """
+    if len(filename) > MAX_FILENAME_LENGTH_GCS_BUCKET_OBJECT:
+        raise ValidationError(
+            f'Invalid file name "{filename}". '
+            f'File names may be at most {MAX_FILENAME_LENGTH_GCS_BUCKET_OBJECT} characters long.'
+        )
+    if not re.fullmatch(r'[\w\-\./]*', filename) or '..' in filename or filename.startswith('/') or '//' in filename:
+        raise ValidationError('Invalid path')
```
kshalot: I'd go for something detailed (since this is a error message that is shown to the client):
```suggestion
        raise ValidationError('The filename cannot start with a forward slash, or contain contain consecutive forward slashes or dots.')
```

kshalot (changes requested):

physionet-django/project/validators.py
superryeti: Oh yes, Thats true my bad, Good Point. 

`Inavalid path`  is not a good message and doesnot even give any context to the users

physionet-django/project/validators.py
```diff
@@ -5,6 +5,7 @@
 
 MAX_FILENAME_LENGTH = 100
 MAX_PROJECT_SLUG_LENGTH = 30
+MAX_FILENAME_LENGTH_GCS_BUCKET_OBJECT = 256
```
kshalot: Last tiny thing - I'd unpack this name a little.

```suggestion
MAX_GCS_OBJECT_NAME_LENGTH = 256
```

kshalot (changes requested):
One last comment. Otherwise LGTM

kshalot:
@amitupreti Maybe also squash the commits into one to not introduce too much clutter.

I'd go for two commits:
1. Fix GCS signed URL filename validation <- this would contain the validation, specs etc.
2. Fix calling `GoogleCloudStorage` property method <- this would contain the single line fix for calling the bucket instance

The commit messages are just an example, go for whatever you feel is right

physionet-django/project/validators.py
superryeti: `MAX_GCS_OBJECT_NAME_LENGTH` is a way better name. i like it. Its consistent with the others and easy on the eye and not too long :D

superryeti:
> @amitupreti Maybe also squash the commits into one to not introduce too much clutter.
> 
> I'd go for two commits:
> 
> 1. Fix GCS signed URL filename validation <- this would contain the validation, specs etc.
> 2. Fix calling `GoogleCloudStorage` property method <- this would contain the single line fix for calling the bucket instance
> 
> The commit messages are just an example, go for whatever you feel is right

Good idea, Let me squash the comments.

superryeti:
Thanks @kshalot  i squashed the commits into two. Thank you for all the feedback. Really appreciate it

kshalot (approved):
Cool, looking good!

superryeti:
@tompollard  could you please review the PR

physionet-django/project/views.py
```diff
@@ -2294,7 +2294,7 @@ def generate_signed_url(request, project_slug):
         return JsonResponse({'detail': 'The file size cannot be a negative value.'}, status=400)
 
     try:
-        validate_filename(filename)
+        validate_gcs_bucket_object_name(filename)
```
tompollard: I haven't looked at this PR in detail, so I might be missing something, but it looks like this change means that file validation will now vary between platforms, depending on whether or not they are using the `generate_signed_url` method? 

If possible, I think it would be best to keep all platforms in sync, using the strictest validation rule in all cases. This helps to maintain consistent platforms and would help to ensure it is possible to switch between approaches over time.

superryeti: That is correct @tompollard  The reason for validation being separate here is that gcs allows file names with `/` in the names(i think slashes are used to determine directory). For example if we upload files with name `folder1/folder2/file.txt`, then the file.txt will be uploaded inside the folder1/folder2 directory.

For file validation during upload in project, i think there are two methods 1)gcs 2) upload to the server directly.

for the method (2), i think the same validation is done on the forms(except the / part)


@kshalot  might be able to provider a better explanation

kshalot: The GCS upload is different because of the fact that `subdir/file` is not a file inside a subdirectory, but a single object which's name is literally `subdir/file` (so there is no such thing as a directory). So the underlying mechanisms imply a slight different way to validate them. Having said that, I think what @tompollard said has value - we can make the GCS validation overlap with the regular filename validation as much as possible to make potential migrations easy.

Since both directories and files in the local storage type are validated with `validate_filename`, we could implement `validate_gcs_bucket_object_name` as:
```python
# The full length still has to be checked since `validate_filename` only checks the length of a single segment
if len(value) > MAX_GCS_OBJECT_NAME_LENGTH:
  raise ValidationError(...)

for path_segment in filename.split('/'):
  validate_filename(path_segment)
```

This way the GCS backend should be compatible with the local one, as it's stricter (enforces shorter path in project directory).

superryeti: Thats a good idea @kshalot . Thanks

@tompollard , i will make the changes and push them. 

@tompollard  would you like me to squash the changes? I am a bit torn between keeping the record of this decision in git history, vs  not keeping it.

superryeti: @tompollard  i pushed the change  with 1 extra commit message, Will squash the last two commits if that is better.

Please let me know

tompollard:
Thanks @amitupreti and @kshalot 

@bemoody any idea what's going on here?:

```
Run ./test-upgrade.sh -v origin/production
fatal: detected dubious ownership in repository at '/__w/physionet-build/physionet-build'
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/physionet-build/physionet-build
Error: Process completed with exit code 12[8](https://github.com/MIT-LCP/physionet-build/actions/runs/4045502781/jobs/6957097553#step:8:9).
```

superryeti:
> Thanks @amitupreti and @kshalot
> 
> @bemoody any idea what's going on here?:
> 
> ```
> Run ./test-upgrade.sh -v origin/production
> fatal: detected dubious ownership in repository at '/__w/physionet-build/physionet-build'
> To add an exception for this directory, call:
> 
> 	git config --global --add safe.directory /__w/physionet-build/physionet-build
> Error: Process completed with exit code 12[8](https://github.com/MIT-LCP/physionet-build/actions/runs/4045502781/jobs/6957097553#step:8:9).
> ```
I was checking stackoverflow for the error. Quick observation was that commit has mentioned in error is ` ae035feaa636eebb50c81b29dc242896cdf8ff28 ` which i tracked down to the bump version PR. i am trying to rebase the changes in case that fixes this

physionet-django/project/views.py
```diff
@@ -2295,10 +2296,16 @@ def generate_signed_url(request, project_slug):
     if size <= 0:
         return JsonResponse({'detail': 'The file size cannot be a negative value.'}, status=400)
 
-    try:
-        validate_filename(filename)
-    except ValidationError as e:
-        return JsonResponse({'detail': e.messages}, status=400)
+    if len(filename) > MAX_GCS_OBJECT_NAME_LENGTH:
```
kshalot: To make the view lighter this could have stayed the `validate_gcs_object_name` function that would contain the validation logic.

tompollard:
@amitupreti in case you didn't see, the dev branch is failing with the same error: https://github.com/MIT-LCP/physionet-build/commits/dev. Possibly related to the switch in Docker image?

kshalot (changes requested):
I'd hide the validation in a function. Let's not bloat the view IMHO.  Then there's fewer paths for returning response (if validation fails then return failure). Much cleaner.

kshalot:
@tompollard My guess would be that the new Docker image has a newer version of Git that now includes the safe dir feature.

superryeti:
> I'd hide the validation in a function. Let's not bloat the view IMHO. Then there's fewer paths for returning response (if validation fails then return failure). Much cleaner.

Duh! thats a really good idea. I like it very much. That way we can keep the validation logic separated from view, and can keep adding stuff on the validation function. 

There is a bunch of existing validation logic, view will look way cleaner this way.

Will make the changes and squash commit :D

physionet-django/project/views.py
```diff
@@ -2285,21 +2286,12 @@ def generate_signed_url(request, project_slug):
...
-    if size <= 0:
-        return JsonResponse({'detail': 'The file size cannot be a negative value.'}, status=400)
-
     try:
-        validate_filename(filename)
+        validate_gcs_bucket_object(filename, size)
     except ValidationError as e:
         return JsonResponse({'detail': e.messages}, status=400)
 
+    size = int(size)
```
kshalot, line 2293: Why is this cast necessary?

kshalot: I think I get it. This makes things a bit awkward but seems hard to avoid. I actually did not notice that we can create a single function both for size and filename validation but it does seem cleaner this way.

superryeti: Oh yes, because there is 

1. `if size > project.get_storage_info().remaining:` which checks if the project has space remaining and expects and integer
2. `generate_signed_url_helper` also expects an integer.

So i thought typecasting once would be better than typecasting twice.

or another option was to bring the size validation on the view and it will do typecasting before validation.

superryeti: oh yes, i took your suggestion one step further. 

its weird to avoid type casting unless we want to bring the size validation inside the view, or do typecasting directly on the code when needed

kshalot: It's weird it's not sent as a number 🤷 Your solution makes sense though.

superryeti: Let me know if you think separating validation function for size and name makes more sense

superryeti: > number

Ah yes, we also have a validation that does `size.isnumeric()` , so it looks like size is always received  as string.

kshalot (approved):
It's a ✅ from my side.

tompollard:
@amitupreti, thanks to Benjamin the failing tests should now be fixed. Please could you rebase on dev?

superryeti:
> @amitupreti, thanks to Benjamin the failing tests should now be fixed. Please could you rebase on dev?

Thank you. Rebasing now

superryeti:
> @amitupreti, thanks to Benjamin the failing tests should now be fixed. Please could you rebase on dev?

@tompollard  rebased and pushed the changes

physionet-django/project/views.py
```diff
@@ -51,13 +51,14 @@
...
 from project.projectfiles import ProjectFiles
-from project.validators import validate_filename
+from project.validators import validate_filename, validate_gcs_bucket_object
 from user.forms import AssociatedEmailChoiceForm
 from user.models import CloudInformation, CredentialApplication, LegacyCredential, User, Training
 
 from django.db.models import F, DateTimeField, ExpressionWrapper
 
 LOGGER = logging.getLogger(__name__)
+MAX_GCS_OBJECT_NAME_LENGTH = 256
```
tompollard: `MAX_GCS_OBJECT_NAME_LENGTH` is also defined in physionet-django/project/validators.py. If it is needed here, it could be imported (though I don't see it being used).

superryeti: Ah, i missed that. just pushed the change removing that.

tompollard:
Thanks @amitupreti, looks good to me. I think we agreed to do some refactoring later (generalize function calls so that filename length is consistent; make cloud provider agnostic where possible).

superryeti:
Thanks @tompollard , i have opened an issue for the new required changes. I will add your comment there and work on this sometime this week.

PR #1834, opened 2023-01-25
Use of get_random_string without an explicit length is deprecated

tompollard:
Use of `get_random_string` without an explicit length is deprecated: https://code.djangoproject.com/ticket/31359. This pull request changes the approach used to generate slugs on the Event model to address this issue.

Currently `get_random_string` is called without an argument, generating slugs with a default length of 12 characters. In this pull request, I move the call to `get_random_string` to the `save()` method and set the default length to 20 characters.

This work is a step towards upgrading to the latest major release of Django (Django 4). See: #1805.


bemoody:
You don't want to call save twice.

tompollard:
> You don't want to call save twice.

This is true, I do not!

bemoody:
Is there a reason to change from 12 -> 20? Seems like overkill & also can be confused with project slugs.

tompollard:
> Is there a reason to change from 12 -> 20? Seems like overkill & also can be confused with project slugs.

No special reason, just seemed like a nice number. Now switched to 12.

alistairewj (approved):

tompollard:
Thanks Alistair and Benjamin, going to go ahead and merge!

Issue #1712, opened 2022-11-17
Add traffic lights to support identity checks


tompollard:
At the moment, reviewing identity/credentialing applications involves subjective decisions based on loosely defined rules. This has various problems, such as: 

- Slow, draining work for people reviewing applications.
- Unclear process for applicants.
- Inconsistent decisions.

We could consider adding a traffic light next to each identity/credentialing submission in the admin console. The idea would be that the color (e.g. red, amber, green) would help reviewers to make an objective decision.

- Green: likely good to approve
- Amber: needs manual review (and/or contact reference)
- Red: needs more information

We would begin by identifying characteristics of applications that can be verified as true/false by a function. We would then use these checks to  determine whether or not the identity meets our requirement for approval, e.g.: the applicant has

- .edu email domain linked to their profile
- ORCID linked to their profile
- @xxxx.xxxx consortium email domain linked to their profile?
- NIH commons account linked to their profile
- UToronto account linked to their profile
- etc

Flag application as green (~most requirements met), amber (~few requirements met), red (~no requirements met).

tompollard:
@lamawmouk do you have any thoughts on this?

lamawmouk:
Thanks @tompollard the traffic idea sound interesting for the review process. I think adding some of the list of characteristics that aren't available on the submission page to the user e.g. NIH commons account and character max limit to research topic section etc. would be enable us to flag applications upon review.

lamawmouk:
Workflow/pipeline suggestions:

Green:
- Users who provide an academic email/institutional email.

Orange:
- Reference with an academic email 
-  Users provide a webpage link (e.g. google scholar, papers, github.io websites ). If none of these are available use references papers (provide a DOI link).  
- Detailed summary topic description.

Red:
- No institutional email/ academic email
- No webpage links
- Incomplete fields e.g. institution location
- One line summary description or one word description.

tompollard:
Thanks Lama, this is helpful! How do you suggest we aggregate these items to come up with a single recommendation for an application?  For example, are you saying that for "Orange" the user must check all of the following boxes?

> - Reference with an academic email
> - Users provide a webpage link (e.g. google scholar, papers, github.io websites ). If none of these are available use references papers (provide a DOI link).
> - Detailed summary topic description.

If someone has yes to "Reference with an academic email." but a "One line summary description.", should we mark the application as orange or red?

lamawmouk:
[...]

PR #1838, opened 2023-01-26
Traffic light identity check status 

lamawmouk:
This codebase implements rules from discussion of the traffic status pipeline as in #1712.

This adds a function `is_institutional_email()` which takes in a string value representing an email address and returns a Boolean indicating whether the email address is from an institutional domain. The function first tries to validate the email address using the `validate_institutional_email(value)`. If  no exception is raised, the function returns True. If the ValidationError exception is raised, the function returns False.

It also adds `has_orcid()`  method of the User object  that returns a Boolean indicating whether the user has an ORCID. The function first checks if the user object has an attribute orcid and if it is not None. If the attribute exists, the function returns True. If the attribute 'orcid' does not exist, the function raises an exception, and the function returns False. If the attribute does not exist and no exception is raised, the function returns False.

It also adds `get_traffic_status()` method for `CredentialApplication` object  that returns a string representing the traffic status of the user based on their email, ORCID, and webpage.
The function first checks if the user has an institutional email address, by calling the `is_institutional_email()` function from the validators module on every email address of the user. If any email is institutional, the variable has_inst_email is set to True.  It then calls the `has_orcid()` method on the user object to check if the user has an ORCID and the variable `has_orcid`  is set to True if the user has an ORCID. The function also checks if the `CredentialApplication` object has a webpage attribute, if it exists and is not None then `has_webpage` is set to True. Therefore:
- If the user has both an institutional email and an ORCID, the function returns 'green'.
- If the user has an institutional email or an ORCID or a webpage, the function returns 'orange'
- If the user does not have an institutional email or an ORCID or a webpage, the function returns 'red'

### DIFF
--- a/physionet-django/user/models.py
+++ b/physionet-django/user/models.py
@@ -433,6 +433,17 @@ class User(AbstractBaseUser, PermissionsMixin):
         """
         return self.is_superuser or self.has_perm('user.can_view_admin_console')
 
+    def has_orcid(self):
+        """
+        Returns True if the user has an orcid.
+        """
+        try:
+            if self.orcid:
+                return True
+        except Orcid.DoesNotExist:
+            pass
+        return False
+
     @staticmethod
     def get_users_with_permission(app_label, permission_codename):
         """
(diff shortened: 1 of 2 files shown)


tompollard:
Thanks @lamawmouk and @GitScanlan, looks good to me.  I'm not convinced it is a good idea to put weight on an unvalidated webpage (`has_webpage`) but I can see how this might be useful when doing your manual checks. I'm happy to merge this PR and we can see how things go.

PR #1840, opened 2023-01-27
Feature: Make signed url expiration time configurable

rafalkosla101:
As the issue of having 24 hours long expiration time of signed urls rose I would like to introduce this PR  that makes the time value configurable using simple ENV variable. By adding those changes we can prevent unwanted uploads of already existing files or changes to the files of a project after the before-mentioned project has already been accepted by the revisor. I have also removed he default value of signed url inside of generate_signed_url_helper - we should be passing that value explicitly for better understanding of code.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -92,6 +92,9 @@ GCP_STATIC_BUCKET_NAME=
 GCP_PROJECT_ID=
 GCP_BUCKET_LOCATION=
 
+# Expiration time of signed urls used to upload files to Google Cloud Platform
+GCS_SIGNED_URL_LIFETIME_IN_MINUTES=1440
+
 # Site-specific content
 SITE_NAME="DataShare"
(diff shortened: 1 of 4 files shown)


.env.example
```diff
@@ -92,6 +92,9 @@ GCP_STATIC_BUCKET_NAME=
 GCP_PROJECT_ID=
 GCP_BUCKET_LOCATION=
 
+# Expiration time of signed urls used to upload files to Google Cloud Platform
+GCS_SIGNED_URL_LIFETIME_IN_MINUTES=60
```
alistairewj: change to 1440

bemoody:
See also discussion at issue #1737.

tompollard:
Thanks Rafal, I think we agreed that this could also be merged!

PR #1842, opened 2023-01-29
API: Add "project" to the URL path

### LINKED ISSUE #1812: Project API: Add "project" to the URL path
The API that was added in In https://github.com/MIT-LCP/physionet-build/pull/1697 has two endpoints: /api/v1/published and /api/v1/published/SLUG/VERSION.

In future we may want to serve non-project information (e.g. platform details; user details; license details; news items; etc), so should `project` appear in these path? 

e.g. should the URLs be `/api/v1/project/published` and `/api/v1/project/published/SLUG/VERSION`?

Rutvikrj26:
Pull Request for Issue #1812 Project API: Add "project" to the URL path 

Current Implementation : 

- [x] `/api/v1/project/published `- all published projects
- [x] `/api/v1/project/published/<slug>` - all versions of the project with `slug=<slug>`
- [x] /`api/v1/project/published/<slug>/<version>` - <version> of the project with `slug=<slug>`

Pending : 

- [ ] `/api/v1/project/published/<slug>/latest` - latest version of the project with `slug=<slug>`

The aim is to provide access to the metadata for the Published Projects in directory-style access via the APIs, thus improving the user experience with the API querying.

### DIFF
--- a/physionet-django/export/serializers.py
+++ b/physionet-django/export/serializers.py
@@ -22,8 +22,14 @@ class PublishedProjectSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = PublishedProject
-        fields = ('id', 'title', 'abstract', 'license', 'dua', 'main_storage_size',
+        fields = ('slug', 'title', 'abstract', 'license', 'dua', 'main_storage_size',
                   'compressed_storage_size')
+
+
+class ProjectVersionsSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = PublishedProject
+        fields = ('slug', 'title', 'version', 'abstract')
 
 
 class PublishedProjectDetailSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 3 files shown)


tompollard:
Looks good to me, thanks Rutvik [sorry]! If possible please could you add your initials to the beginning of your branch name (e.g. `au/api-paths`?). This helps (me at least!) to keep a clean list of branches.

tompollard:
Should we be returning 404 for slugs that aren't found (e.g. https://localhost:8000/api/v1/project/published/blahblahblah)? If so, please incorporate into the next PR.

Rutvikrj26:
Sure!
The 404 is the code for the HTTP Response. The webpage was found `../published` , and so the HTTP request was a success, thus 200. However, the query that is being made, the resource requested via the parameter `/wewef` was not there, and thus, you will get an empty response. This shows that the API is working, but the requested item is not available. 
We can make it custom as well and send out the 404 code if you prefer sending the 404 code in case the requested item is not available.

Issue #1844, opened 2023-01-30
Tests failing on new Docker image due to git's safe dir feature.


tompollard:
Tests are now failing with the following error (blocking all new PRs):

```
Run ./test-upgrade.sh -v origin/production
fatal: detected dubious ownership in repository at '/__w/physionet-build/physionet-build'
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/physionet-build/physionet-build
Error: Process completed with exit code 12[8](https://github.com/MIT-LCP/physionet-build/actions/runs/4045502781/jobs/6957097553#step:8:9).
```

As noted by @kshalot in https://github.com/MIT-LCP/physionet-build/issues/1832#issuecomment-1408917087:

> My guess would be that the new Docker image has a newer version of Git that now includes the safe dir feature.

bemoody:
That bug happened ages ago, weird that it's just being pushed to bullseye-security *now* (looks like it was build last week and presumably just uploaded today, there's still no Debian security advisory yet.)

Presumably we should do what it says. :)

bemoody:
(I don't follow the details entirely, but looks like actions/checkout has its own workaround that involves creating a temporary home directory.)

tompollard:
Fixed in https://github.com/MIT-LCP/physionet-build/pull/1845

PR #1845, opened 2023-01-30
.github/workflows: mark git repository as a safe directory

bemoody:
Fix bug #1844:

Modern versions of git will complain if trying to operate on a repository directory not owned by you (which is, indeed, unsafe.)

In the GitHub workflows, this error is triggered for some reason, maybe because of something wrong in how the container is set up or maybe because of something weird about how "actions/checkout" works.

Regardless of the reason, there doesn't seem to be any reason not to mark the repository directory as safe, and doing this should be harmless if GitHub later fixes the issue.

### DIFF
--- a/.github/workflows/physionet-build-test.yml
+++ b/.github/workflows/physionet-build-test.yml
@@ -53,6 +53,10 @@ jobs:
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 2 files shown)


tompollard:
This is great, thanks Benjamin

Issue #1833, opened 2023-01-25
Credentialing: Automatically reject applications if a reference check is pending for >1 month


lamawmouk:
Based on the discussion in the PhysioNet meeting, the current interface of the identity/credentialing submission in the admin console has an Awaiting reference response tab with references who haven't responded for a long time (e.g. response on hold for more > 4 months) .

One suggestion would be to remove users from the system after a certain period of time. @tompollard, is there a reasonable time frame that we could use to remove pending users, maybe have a 2 months threshold.

superryeti:
Hi @tompollard , 

And i could think of the following for this issue

1. Do the check for the duration of pending reference  validation just before the content is rendered on the console(on the views.py), and then reject the application. (This sound like a terrible idea as we would be waiting until someone loads the console app and opens the credentialing section)

2. Do some kind of periodic check  with some kind of scheduling tool, as this might help reduce the work load on the server during busy hours(when people are using the console app) compared to option1
Also it kind of makes sense to this check in batch as its not something that is super important to do exactly after like 120 days and 0. Maybe every day at the time when the site has less visitors(like after midnight)[Kind of like how netflix does with subscriptions

tompollard:
Thanks @amitupreti my vote would be for option 2! Sound okay to you @bemoody?

lamawmouk:
We could possibly apply a method in the form, by creating a CredentialApplication form and then using a clean method in the form to check the response time of the reference after the user has submitted the form.
For example, we can add the clean method in the form which will be called after the user submits the form and before the form's data is saved to the database. If the response time is greater than 2 months, the application will stay or be deleted from the table.  We could use  to set the user attribute of the CredentialApplication object to None, but the itself will still exist in the database or use  CredentialApplication.objects.filter(user=self.user).delete() to remove the CredentialApplication object from the database.

superryeti:
@lamawmouk Sorry if i misunderstood the original issue or if i am misunderstanding your latest comment.

From my understanding the issue is, there is a lot of credentialed applications submission, where references have not responded. So we want to reject the application automatically after x months.

But based on your recent comment, we seem to be discussing to check some kind of response time when a user is trying to submit an Application from here http://localhost:8000/credential-application/.

Wouldn't  the checking of response time be done in the future, after the user submits form, the approver approves id and requests reference verification?

tompollard:
[...]

PR #1846, opened 2023-01-30
Auto Reject pending credentialing applications whose references have not responded after a set number of days

superryeti:
Solves https://github.com/MIT-LCP/physionet-build/issues/1833

Problem: 

Based on the issue https://github.com/MIT-LCP/physionet-build/issues/1833  the current interface of the identity/credentialing submission in the admin console has an Awaiting reference response tab with references who haven't responded for a long time (e.g. response on hold for more > 4 months) .

Solution:
We decided to add a cronjob that will trigger near midnight(during off-peak hours, with the hope that the servers are in north America).
In this PR,

1. We have added a management command called `reject_pending_credentialing_applications` which will automatically reject all pending credentialing applications whose references have not responded after a set number of days.
2. An email will also be sent to applicants
3. The number of days is set in .env file, and had default value of 30 days.
4. The cronjob is set to start at 11:45 PM(there is another cronjob that starts at 11:35 PM, so i started the new cronjob after 10 minutes)
5. We also made the auto-reject configurable by adding a boolean variable on .env file. By default the autoreject is turned off(just so that the current deployments will work as it is unless they enable the auto reject themselves)

Questions:
1. Do we add test for the management command? I see the management command `accept_pending_requests` has a test, but other management commands donot have any test?
2. For the deployment with containerization, will the cronjob run per container(that might result to bunch of containers trying to change the database at same time)?
- If yes, then we will be sending bunch of emails to same applicant from different containers


alistairewj:
> 1. Do we add test for the management command? I see the management command `accept_pending_requests` has a test, but other management commands donot have any test?

If you have an idea for a test, I think it's a good idea.

> 2. For the deployment with containerization, will the cronjob run per container(that might result to bunch of containers trying to change the database at same time)?

I believe cron jobs run in only one pod.. but there's no guarantee only one call of the cronjob runs in some edge cases. Try to write it so the job is idempotent. The database should handle concurrent writes so long as the changes are consistent. We can also specify in the kubernetes version of the cron job to skip running the job if another one is currently running to avoid concurrency issues to some extent (though I can't imagine this job taking 24+ hours).

superryeti:
Because we need a credentialed application that has 

> > 1. Do we add test for the management command? I see the management command `accept_pending_requests` has a test, but other management commands donot have any test?
> 
> If you have an idea for a test, I think it's a good idea.
> 
> > 2. For the deployment with containerization, will the cronjob run per container(that might result to bunch of containers trying to change the database at same time)?
> 
> I believe cron jobs run in only one pod.. but there's no guarantee only one call of the cronjob runs in some edge cases. Try to write it so the job is idempotent. The database should handle concurrent writes so long as the changes are consistent. We can also specify in the kubernetes version of the cron job to skip running the job if another one is currently running to avoid concurrency issues to some extent (though I can't imagine this job taking 24+ hours).

Thank you @alistairewj  for the quick feedback. I added a test for management command

Summary of change: 
1. I added an CredentialedApplication  with the reference request date  of 1 year ago on the `dem-user.json` fixtures, so that it is eligible to be automatically rejected
2. Added a test case inspired by `test_accept_pending_requests.py`(Quick Note, the test case is a bit different than what i found suggested in django docs, but i went ahead and copied the style in anyway, just in case there was a reason in the past to do it this way `test_accept_pending_requests.py`

bemoody:
This looks like a great start (though I haven't tested it.)  There are a number of small but significant changes I would make.

- As Tom pointed out to me, we don't want to set the responder to "admin" (in fact there may not be an "admin" user.)  Here I think we should simply set the responder to None (is there any reason we can't do that?)

- We do want to record the auto-rejection in the database.  One way to do it would be to add a new field to CredentialApplication:
```
    ...

    class AutoRejectionReason(models.IntegerChoices):
        NO_RESPONSE_FROM_REFERENCE = 1

    auto_rejection_reason = models.PositiveSmallIntegerField(
        blank=True, null=True, choices=AutoRejectionReason.choices
    )
```
An alternative would be to add an additional `status` value, but I think this is cleaner.

- Put the explanatory message in the template (`physionet-django/notification/templates/notification/email/process_credential_complete.html`) rather than in `responder_comments`.  For example...
```
...
{% if application.status == 1 %}Thank you for applying for credentialed access to {{ SITE_NAME }}. Your application was not approved. This may be for the following reason(s):
{% if application.auto_rejection_reason == 1 %}

- We were unable to contact the reference you provided.

{% elif include_comments %}

{{ application.responder_comments }}

{% else %}
...
```

- I would wrap the auto-rejection functionality in a new method
```
    def auto_reject(self, reason):
        """
        Automatically reject this application.
        A reason for rejection (AutoRejectionReason) must be provided.
        """
        self.auto_rejection_reason = reason
        self.reject(responder=None)
```

- We want to heavily rate-limit these rejections, and *not* to send hundreds of emails at once.  As far as I'm concerned, it's perfectly fine if it takes a week or two to clear the current backlog.

The command should run more than once per day, and process a small number of rejections at once.  To pick some more-or-less arbitrary numbers, perhaps a maximum of five auto-rejections, once per hour.

- If sending the message fails (because the mail server died, DNS is on fire, whatever), then the application shouldn't be saved.  Probably best is to do this:
```
    with transaction.atomic():
        application.auto_reject(
            CredentialApplication.AutoRejectionReason.NO_RESPONSE_FROM_REFERENCE
        )
        notification.process_credential_complete(request, application)
```

physionet-django/user/management/commands/reject_pending_credentialing_applications.py
```diff
@@ -0,0 +1,64 @@
...
+from django.utils import timezone
+
+import notification.utility as notification
+from user.models import CredentialApplication, User
+
+LOGGER = logging.getLogger(__name__)
+REJECTION_MESSAGE = 'Your Reference {reference_name} did not respond to the reference check request within '\
+                    f'{settings.MAX_REFERENCE_VERIFICATION_DAYS_BEFORE_AUTO_REJECTION} days.'
+
+DFAULT_NUMBER_OF_APPLICATIONS_TO_REJECT = 5
```
tompollard: Typo: DFAULT_NUMBER_OF_APPLICATIONS_TO_REJECT -> DEFAULT_NUMBER_OF_APPLICATIONS_TO_REJECT

physionet-django/user/management/commands/reject_pending_credentialing_applications.py
```diff
@@ -0,0 +1,64 @@
...
+
+        # only perform auto rejection if the setting.ENABLE_CREDENTIALING_AUTO_REJECTION is True
+        if not settings.ENABLE_CREDENTIALING_AUTO_REJECTION:
+            LOGGER.info('Auto rejection of credentialing applications is disabled. Exiting.')
+            LOGGER.warning('If this was unintentional, please set '
+                           'MAX_REFERENCE_VERIFICATION_DAYS_BEFORE_AUTO_REJECTION to True in .env file.')
+            return
+
+        # total number of applications to be rejected
+        total_applications_to_reject = options['number'] or DFAULT_NUMBER_OF_APPLICATIONS_TO_REJECT
```
tompollard: As above (typo!)

physionet-django/user/management/commands/reject_pending_credentialing_applications.py
superryeti: Thanks @tompollard, pushing change.
My english skills is failing me. My english teacher from high school is disappointed in me rn :D

tompollard: I make typos like this all the time. You're too hard on yourself Amit!

superryeti:
@tompollard @bemoody  I made the suggested changes.

To test locally please enable the `ENABLE_CREDENTIALING_AUTO_REJECTION` to true in .env

physionet-django/user/models.py
```diff
@@ -757,6 +757,12 @@ class CredentialApplication(models.Model):
         (4, 'Revoked')
     )
 
+    class AutoRejectionReason(models.IntegerChoices):
```
tompollard: Can we use the Django enum here for readability (so in the email template, we can call this with a human readable value, instead of an integer?)

superryeti: For sure. let me make an update

physionet-django/user/management/commands/reject_pending_credentialing_applications.py
```diff
@@ -0,0 +1,70 @@
...
+from django.db import transaction
+from django.http import HttpRequest
+from django.utils import timezone
+
+import notification.utility as notification
+from user.models import CredentialApplication, User
+
+LOGGER = logging.getLogger(__name__)
+
+DEFAULT_NUMBER_OF_APPLICATIONS_TO_REJECT = 5
```
tompollard: Is there a reason why we don't treat this as.a configurable variable in the .env file?

physionet-django/user/management/commands/reject_pending_credentialing_applications.py
```diff
@@ -0,0 +1,70 @@
...
+        if not settings.ENABLE_CREDENTIALING_AUTO_REJECTION:
+            LOGGER.info('Auto rejection of credentialing applications is disabled. Exiting.')
+            LOGGER.warning('If this was unintentional, please set '
+                           'MAX_REFERENCE_VERIFICATION_DAYS_BEFORE_AUTO_REJECTION to True in .env file.')
+            return
+
+        # total number of applications to be rejected
+        total_applications_to_reject = options['number'] or DEFAULT_NUMBER_OF_APPLICATIONS_TO_REJECT
+
+        LOGGER.info(f'Total number of applications to be rejected: {total_applications_to_reject}')
```
tompollard: I find this log message confusing. It suggests to me that X applications are going to be rejected, when actually this is just the maximum possible number of applications that would be rejected. 

We could reword the message, but I think it's probably okay to just remove it. I don't think it is saying anything particularly useful.

physionet-django/user/management/commands/reject_pending_credentialing_applications.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # creating an instance of HttpRequest to be used in the notification utility
+        request = HttpRequest()
+
+        # get applications to be rejected
+        filtered_applications = get_application_to_be_rejected(total_applications_to_reject)
+
+        LOGGER.info(f'Found {len(filtered_applications)} applications to be rejected.')
```
tompollard: Perhaps make it clear we are talking about credentialing applications? e.g. 

`f'{len(filtered_applications)} credentialing applications selected for rejection. No ref response. .'`

physionet-django/user/management/commands/reject_pending_credentialing_applications.py
```diff
@@ -0,0 +1,70 @@
...
+        for application in filtered_applications:
+            with transaction.atomic():
+                application.auto_reject(reason=CredentialApplication.AutoRejectionReason.NO_RESPONSE_FROM_REFERENCE)
+                LOGGER.info(f'Application {application.id} rejected')
```
tompollard: Perhaps remove this log message, then add `application.id` to the next?

physionet-django/user/management/commands/reject_pending_credentialing_applications.py
```diff
@@ -0,0 +1,70 @@
...
+
+                # send notification to applicant
+                notification.process_credential_complete(request, application)
+                LOGGER.info(f'Notification sent to applicant {application.get_full_name()}')
```
tompollard: Add `application.id` here.

`LOGGER.info(f'Rejected ApplicationID: {application.id}. Notification sent to applicant: {application.get_full_name()}')`

physionet-django/user/management/commands/reject_pending_credentialing_applications.py
```diff
@@ -0,0 +1,70 @@
...
+        # get applications to be rejected
+        filtered_applications = get_application_to_be_rejected(total_applications_to_reject)
+
+        LOGGER.info(f'Found {len(filtered_applications)} applications to be rejected.')
+        for application in filtered_applications:
+            with transaction.atomic():
+                application.auto_reject(reason=CredentialApplication.AutoRejectionReason.NO_RESPONSE_FROM_REFERENCE)
+                LOGGER.info(f'Application {application.id} rejected')
+
+                # send notification to applicant
```
tompollard: The code is clear, so you could remove this comment ("# send notification to applicant").

physionet-django/user/management/commands/reject_pending_credentialing_applications.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # total number of applications to be rejected
+        total_applications_to_reject = options['number'] or DEFAULT_NUMBER_OF_APPLICATIONS_TO_REJECT
+
+        LOGGER.info(f'Total number of applications to be rejected: {total_applications_to_reject}')
+
+        # creating an instance of HttpRequest to be used in the notification utility
+        request = HttpRequest()
+
+        # get applications to be rejected
```
tompollard: This comment also feels unnecessary ("# get applications to be rejected")

tompollard (commented):
Thanks Amit, this looks good to me (though I haven't tested that the cron task is actually running as expected). I added a few minor comments inline.

physionet-django/user/management/commands/reject_pending_credentialing_applications.py
superryeti: Thats an excellent point, making it configurable makes a lot of sense. Let me do that.

physionet-django/user/management/commands/reject_pending_credentialing_applications.py
superryeti: Thats a good point. I like the removing idea. Let me push changes

physionet-django/user/management/commands/reject_pending_credentialing_applications.py
superryeti: I agree, It will be helpful to have credentialing app, and the reason in logs.

.env.example
```diff
@@ -146,8 +146,13 @@ ENABLE_LIGHTWAVE=True
 # In-built django variable that logouts user if cookie expiration time has been exceeded
 SESSION_COOKIE_AGE=43200
 
+#### Auto rejection of credentialing applications Start ####
```
superryeti, line 149: I thought it might be a good idea to make a block of settings for autorejection and group them together

tompollard:
Looks good to me, thanks Amit! Let's give it a try...

tompollard:
Deployed on our live server!

PR #1847, opened 2023-01-31
Feature: Add absolute timeout to sessions

rafalkosla101:
This PR was separated from the main one with the whole session management. It adds a config value that is being used in django to set the absolute session time.

Using django native SESSION_COOKIE_AGE session timeout can be set to a certain period of time by setting the value using cookies stored in user browser. Then when the requests are made by server the before-mentioned cookie time is checked. If the cookie already expired user is logged out. It is client-side timeout only

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -137,3 +137,7 @@ MAX_EMAILS_PER_USER = 10
 
 # Max training report size in bytes
 MAX_TRAINING_REPORT_UPLOAD_SIZE = 1048576
+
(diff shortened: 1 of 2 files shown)


bemoody:
To be clear, is this *literally* only the cookie maximum age (client-side timeout)?  Or does this also control a server-side timeout?  Please document it.

tompollard:
Thanks Rafal!

PR #1848, opened 2023-01-31
Add instructions on how to add cronjobs

superryeti:
Context: While working on https://github.com/MIT-LCP/physionet-build/pull/1846, i realized that we hadnot documented about how PhysioNet  performs scheduled tasks. So after discussing with @tompollard , we decided to add a quick note, so that future developers can find this information quickly. 

Change:
The PR adds an extra section on the PhysioNet Readme.MD and points out how the existing cronjobs are added, and displays an example of one of the cronjobs.


bemoody:
Thanks, Amit.  That DJANGO_SETTINGS_MODULE thing is a real pain, isn't it?

PR #1851, opened 2023-01-31
URLs representing "directories" should end in a slash

bemoody:
A URL that represents a "directory" should end in a slash.  URLs that don't end in a slash are "files", and in that case the trailing component should be an appropriate filename for the resource.       

On the one hand, you can say this is just a stylistic convention.  But, aside from the fact that we want to be internally consistent, a lot of existing software makes assumptions that are broken if you don't do things the conventional way.

Note that Django will, in any event, redirect clients from `foo` to `foo/` or from `foo/` to `foo` as appropriate.


tompollard:
Good catch, thanks Benjamin!

Issue #1839, opened 2023-01-27
Using events functionality to allow temporary access to datasets


tompollard:
@alistairewj this is a suggested set of PRs to move towards a system that allows temporary access to datasets for datathons.

Steps 1 to 4 should provide all functionality needed for the Toronto datathon (i.e. hosts can create events, and temporary access to datasets can be granted for events). Step 5 could be implemented later.

1. **Feature: Hosts can approve/deny requests to join their event (required)**. Hosts must explicitly approve event participants (by clicking approve/deny for join requests, which appear in their events page).
2. **Feature: Event management list in the console (required)**. In the console, admin users are able to access a searchable page that lists all events (similar to the “active users” page).
3. **Feature: Event detail pages in the console (required)**. In the console, events listed in feature 2 are linked to individual event detail pages (similar to the user management page for individual users). Admin users can view a page listing details about an individual event.
4. **Feature: Event detail pages include a “request temporary access to dataset” form (required)**. On the event detail pages, admin users can grant temporary access to specific datasets. Adding a dataset to an event means that cloud access is temporarily granted to all participants for a fixed period. This approach broadly follows the approach that we have used for datathons in the past (i.e. create temporary permissions group for the event and attach to relevant datasets).
5. **Feature: Enable Hosts to request temporary access to datasets for a datathon (nice to have)**. Hosts now have the ability to select a dataset that they would like to make available for an event. Selecting the dataset sends a request to either (1) the platform admin team, who manually seek approval or (2) the data contributor who receives and handles the request.

Other random thoughts:

- Feature: Introduce some kind of DUA or code of conduct for people signing up for an event.
- Feature: Allow hosts to upload a list of participants for an event. When this done, individual email invites are sent out (in a similar way to author invitations).
- Feature: Allow participants to be promoted to "co-host" or "event admin" or whatever, giving them host privileges (e.g. ability to approve users into the event).
- Feature: Provide additional event management tools for Hosts (though not sure what these would be). e.g. assistance with managing participant mailgroup.

PR #1853, opened 2023-02-02
Events: Implement a waitlist and approval for events(instead of accepting the users directly to the event)

superryeti:
Context:
This PR aims to implement the feature 1 from https://github.com/MIT-LCP/physionet-build/issues/1839

Hosts can approve/deny requests to join their event (required). Hosts must explicitly approve event participants (by clicking approve/deny for join requests, which appear in their events page).

When users visit the event page, they are prompted to request access to project, and once they request access they are added to a waitlist.

Now the event host/creator, will see the notifications on the events page. and can approve, reject participants to the event.


superryeti:
So far, students can join the waitlist, and the event host, can approve, reject the students.

Students, can see the status of their request on the event page.

The PR might get too big, So i was thinking to now just add test case

And on another PR after this,  add code to send email notification once someone requests to join the event, and once a decision has been made

physionet-django/events/templates/events/event_participant.html
```diff
@@ -13,10 +13,10 @@
 {% if event %}
     <h2 class="form-signin-heading">Event Registration</h2>
     <p>You have been invited to join {{ event.host.get_full_name }}'s {{ event.title }} from {{ event.start_date }} to {{ event.end_date }}.</p>
-    <p>Click confirm if you are participating in this event.</p>
+    <p>Click request access if you want to participate in this event.</p>
```
tompollard: "Click "confirm" to join the waiting list for this event."

physionet-django/events/templates/events/event_participant.html
```diff
@@ -13,10 +13,10 @@
...
     <form  method="post">
         {% csrf_token %}
-        <button type="submit" class="btn btn-sm btn-primary" name="confirm_event" value="confirm" role="button">Confirm</button>
+        <button type="submit" class="btn btn-sm btn-primary" name="confirm_event" value="confirm" role="button">Request Access</button>
```
tompollard: "Request Access" doesn't seem quite right. Perhaps just leave as "confirm"?

physionet-django/events/views.py
```diff
@@ -101,8 +134,8 @@ def event_add_participant(request, event_slug):
 
     if request.method == 'POST':
         if request.POST.get('confirm_event') == 'confirm':
-            event.enroll_user(user)
-            messages.success(request, "You have been enrolled")
+            event.waitlist(user)
+            messages.success(request, "You have successfully requested to join this event")
```
tompollard, line 148: Perhaps -> "You have joined the waiting list for this event."

physionet-django/events/models.py
```diff
@@ -37,27 +62,107 @@ def save(self, *args, **kwargs):
     def __str__(self):
         return self.title
 
-    def enroll_user(self, user):
+    def waitlist(self, user, comment=None):
```
tompollard: maybe `join_waitlist`?

physionet-django/events/models.py
```diff
@@ -37,27 +62,107 @@ def save(self, *args, **kwargs):
...
-        Adds a participant to an event.
+        Adds a participant to an event as wait listed. This is also the first time a participant is added to an event.
+
+        Args:
+            user (User): The user to add to the event.
+            comment (str): The comment to add to the participant about the status.
         """
         with transaction.atomic():
-            EventParticipant.objects.create(user=user, event=self)
+            participant = EventParticipant.objects.create(user=user, event=self)
```
tompollard: Instead of the `if comment` could you pass comment to the create call above? `create(user=user, event=self, decision_comment=comment)`

physionet-django/events/models.py
```diff
@@ -7,7 +9,30 @@
 from events import validators
 
 
-# Create your models here.
+class EventParticipationStatus(IntEnum):
```
tompollard: Perhaps use a Django enumeration type here instead: https://docs.djangoproject.com/en/4.1/ref/models/fields/#enumeration-types. I think they help to make more human readable code.

physionet-django/events/models.py
```diff
@@ -37,27 +62,107 @@ def save(self, *args, **kwargs):
     def __str__(self):
         return self.title
 
-    def enroll_user(self, user):
+    def waitlist(self, user, comment=None):
         """
-        Adds a participant to an event.
+        Adds a participant to an event as wait listed. This is also the first time a participant is added to an event.
```
tompollard: Although you are changing the name of the function, the behaviour looks basically the same as before. i.e. someone gets added to the participant list regardless of whether or not their application is approved.

superryeti:
Working on failing tests

tompollard (commented):
Thanks @amitupreti, this is going in the right direction. I have taken a look and added a few notes below:

- We should avoid saying "Reject". It's not a very friendly word! ("You have been rejected from the event"). "Participation Status: Reject". We can discuss alternative language on Slack.
- The default form response when responding to a request to join should be "Approve", not "Reject".
- When someone clicks the registration link to join the event, they are immediately added to the participant list. This doesn't seem right. Shouldn't they only be added after the host approves them?  If someone is rejected then they still appear in the participant list. Maybe we need a separate model to track applications?
- If I am rejected from an event and then try re-registering, I see a message saying "You have already been accepted to this event" (but the status says rejected).
- Perhaps only display "Number of participants: X" to the host for now.

superryeti:
Thank you for the feedbacks @tompollard , Working on them

physionet-django/events/models.py
superryeti: Here, the idea for `waitlist` was  more like "waitlist this user to the event, since the method is in the Event`

Wouldn't `join_waitlist` be more appropriate if the method was in `EventParticipant`?

physionet-django/events/models.py
superryeti: oh yes, that is a good idea, since '' is also a valid comment, but i suppose we will need to pass a comment on views.

physionet-django/events/models.py
superryeti: Thank you. i didnot know about these.These are excellent.

superryeti:
@tompollard I added the new changes, fixed test and fixed the upgrade issue.
Could you please take a look?

physionet-django/events/views.py
```diff
@@ -101,8 +142,8 @@ def event_add_participant(request, event_slug):
 
     if request.method == 'POST':
         if request.POST.get('confirm_event') == 'confirm':
-            event.enroll_user(user)
-            messages.success(request, "You have been enrolled")
+            event.join_waitlist(user=user, comment='Automatic')
```
tompollard: What is the purpose of the "Automatic" comment here?

physionet-django/events/views.py
```diff
@@ -54,21 +59,45 @@ def event_home(request):
...
+    # handle notifications to join an event
+    if request.method == 'POST' and 'participation_response' in request.POST.keys():
+
+        formset = EventApplicationResponseFormSet(request.POST)
+        # only process the form that was submitted
+        for form in formset:
+            if form.instance.id == int(request.POST['participation_response']):
+                if form.is_valid():
+                    event_application = form.save(commit=False)
+                    event_application.accept(comment=form.cleaned_data.get('decision_comment'))
```
tompollard: Is this the correct place to be calling "accept"? It looks like people are being added to the list of participants even when applications are rejected.

physionet-django/events/models.py
```diff
@@ -61,3 +64,79 @@ class Meta:
...
     def __str__(self):
         return self.user.get_full_name()
+
+
+class EventApplications(models.Model):
+    """
+    Captures information about applications for events.
+    """
+    class EventApplicationStatus(models.IntegerChoices):
+        WAITLISTED = 0
```
tompollard: Wouldn't text choices be a better choice here, since the integer carries no meaning? e.g. see: https://stackoverflow.com/questions/54802616/how-can-one-use-enums-as-a-choice-field-in-a-django-model

e.g. something like:

```
    class EventApplicationStatus(models.TextChoices):
        WAITLISTED = 'WL', _('Waitlisted')
        APPROVED = 'AP', _('Approved')
        NOT_APPROVED = 'NA', _('Not Approved')
        WITHDRAWN = 'WD', _('Withdrawn')
```

tompollard (commented):
Thanks @amitupreti, looks pretty good. I added a few more comments inline. One other thought: While an application is awaiting a response, we should display a "Awaiting response" message to the applicant in their Events home.

superryeti:
> Thanks @amitupreti, looks pretty good. I added a few more comments inline. One other thought: While an application is awaiting a response, we should display a "Awaiting response" message to the applicant in their Events home.

Thank you, awesome working on these right now :)

physionet-django/events/views.py
superryeti: Because the waitlist method needs us to pass a comment, i thought 'Automatic' would make sense(as in the participant was added automatically to waitlist)

Maybe something like "Approval needed as per Event Settings" would be better?

physionet-django/events/views.py
superryeti: Oh yes, you are right, i am all over the place.  There should be a if and else statement her.

Submitting a commit asap

physionet-django/events/models.py
superryeti: hmm. thats true, integer here has no significance. working on this asap

physionet-django/events/views.py
tompollard: It doesn't feel right that we are adding a comment simply because the model expects one. In this case, maybe the comment should be optional.

It makes me realize that we should be clear of the purpose of this comment (both for ourselves, and to the host). Is the comment meant to be public (i.e. sent to the applicant) or is it a "note to self"?

superryeti: Noted. for optional comment, its already optional by default on modal, i think i could just edit the `join_waitlist` method with a if statement(or pass an empty comment on the view)

I am not sure if the comment is meant to be public(but do we want to make it public). it might be helpful for the participants to know why they were rejected.

Also i had assumed that we would be sending the comment in the e-mail to the participant on the rejection email(in case we decide to add the email)

What do you think @tompollard ?

tompollard: > Also i had assumed that we would be sending the comment in the e-mail to the participant on the rejection email(in case we decide to add the email)

 I think it makes sense for the host to give an explanation to the applicant. Perhaps we should rename this something like "comment_to_applicant" in this case, to make this clear?

superryeti: that sounds good. i will make the changes

Thank you

superryeti:
Hi @tompollard , Thank you for the feedback. i implemented the changes.

physionet-django/events/forms.py
```diff
@@ -1,7 +1,7 @@
 from django import forms
 
 from events.widgets import DatePickerInput
-from events.models import Event
+from events.models import Event, EventApplications
```
tompollard: Sorry, should have picked this up before, but for consistency with other models, should `EventApplications` be `EventApplication`?

superryeti: Yes, i agree `EventApplication` is correct. My bad. For some weird reason, Plural made sense to me. Pushing changes right away

tompollard:
Nice work, thanks Amit! I'm good with merging this. Later I think we will need to:

- Clean up some of the language that we use across the events.
- Add some kind of "You are on the waitlist" message on the Events page for people waiting for the host to respond.
- [Edited]: Notification emails

superryeti:
Thanks @tompollard , I will open small PR for those, adding this to my list

PR #1854, opened 2023-02-03
Add Event Detail page in console

superryeti:
As discussed here https://github.com/MIT-LCP/physionet-build/issues/1839, we will later provide access to a dataset to an event. So this PR lays foundation for that by adding a simple Event detail page on the console.

Currently the following changes are added

An event detail page on console(same as credentialed application detail page) and admins can view the project details, and participants list on console

image


tompollard:
Thanks @amitupreti this is a good start.

PR #1857, opened 2023-02-07
DRAFT - Feature: on platform training and quiz

harryface:
Allows creation of on platform training with quizzes that users can take.
- Creation of this training and quiz is done via the admin url

To undergo a platform training, visit [Undergo a Platform Training](http://localhost:8000/settings/training/) and select a Training type from **Complete training course on Platform** section
- A quiz failed, will take you back to the beginning of the training section(content block)
- When a correct answer is given for the last quiz, it automatically takes you back to http://localhost:8000/settings/training/ and populates the database (Training Model) with an ACCEPTED status

To create an On Platform Training, visit [Create On Platform Training](http://localhost:8000/admin/user/onplatformtraining/add/;) The content block and quiz contains an **order** field. This field determines how they would be displayed to the user taking the training.
**NB**: This page does not yet allow adding of Multi-choices to the quiz, this can be done by editing the quizzes from [Add Quiz Multi Choices](http://localhost:8000/admin/user/quiz/).


bemoody:
I would *strongly* encourage you to completely forget about the admin interface for now.

1. Administrators don't need and don't really want a web interface.

2. We *really* don't want to give people access to the Django admin system.

3. You're spending a lot of effort and adding a lot of complexity for comparatively little benefit.

As an administrator myself, I would far prefer to have the training course and quiz stored as a single JSON file, which I can edit by hand, test locally, and upload to the site once I'm satisfied with it.

bemoody:
In terms of technical design, I think it would make sense to put this functionality into its own Django app, rather than mixing its models, forms, views, etc. into the `user` app.  It's much easier to do this now rather than later.

I don't know how much thought you've put into *versioning* of training courses.  But we should assume that we will want to make changes to the training course over time.  We will probably want to keep a record of *which version of the course* each person completes.  And we need to be sure to present the correct quiz questions and check the quiz answers against the correct version.

(Ensuring that we don't accidentally make changes to an existing training course, after the course is made public, is one of the reasons that I think creating an on-site web admin interface is far more effort than it's worth.)

tompollard:
It would be nice if there was some form of randomness in the questions/answers, so the course is different each time someone takes it.

It feels slightly awkward to me that the training is being taken from http://localhost:8000/settings/training/. There's a lot in this page now. I wonder whether we should be reorganising the training content somehow. Need to think about this.

> I would far prefer to have the training course and quiz stored as a single JSON file, which I can edit by hand, test locally, and upload to the site once I'm satisfied with it.

I agree, I like this suggestion!

harryface:
Please @bemoody and @tompollard did I capture all your comments below?

- Move Training to its App: Should I move just the on-platform training app or should I also move  Training and TrainingType model etc
- Upload on-platform training as well as its content block and quiz as a Json file (with a validator in place)
- Remove the on-platform training from http://localhost:8000/settings/training/ and find a new home for it
- Randomise the Quiz Multi-Choice

bemoody:
Tell me about how you plan to handle versioning.  There are at least two distinct scenarios:

(1) You want to make a *minor* change to the training course.  People who have already completed training v1 are allowed to access the data.  People who have started (but not completed) v1 are allowed to complete it, and will subsequently be allowed to access the data.  People who have not yet started the course will be directed to v2.

(2) You want to make a *major* change.  People who already completed training v1 or v2 are *not* allowed to use the data, and will be told they need to complete training v3 first.  People who started v1 or v2 might (or might not) be allowed to complete it, but it won't do them any good because it won't be accepted for project access.  People who haven't yet started will be directed to v3.

How would you, as the administrator, handle each of these two scenarios?

alistairewj:
> Please @bemoody and @tompollard did I capture all your comments below?
> 
> * Move Training to its App: Should I move just the on-platform training app or should I also move  Training and TrainingType model etc

Asked @tompollard for his thoughts on this. Personally I'd say move it all at once, but sometimes it's nice to incrementally add features.

> * Upload on-platform training as well as its content block and quiz as a Json file (with a validator in place)

Sounds good.

> * Remove the on-platform training from http://localhost:8000/settings/training/ and find a new home for it

Maybe we can leave it there for now? The page isn't *too* cluttered right now, and gives us time to see how this works. @tompollard thoughts?

> * Randomise the Quiz Multi-Choice

If I were to guess one way to do this, maybe you could have a list of N answers in the JSON, a field indicating which is the correct answer, and a field indicating how many answers to include in the display. You could then randomly show up to N-1 questions, with the correct answer also there.

> Tell me about how you plan to handle versioning. There are at least two distinct scenarios:
> 
> (1) You want to make a _minor_ change to the training course. People who have already completed training v1 are allowed to access the data. People who have started (but not completed) v1 are allowed to complete it, and will subsequently be allowed to access the data. People who have not yet started the course will be directed to v2.
>
> (2) You want to make a _major_ change. People who already completed training v1 or v2 are _not_ allowed to use the data, and will be told they need to complete training v3 first. People who started v1 or v2 might (or might not) be allowed to complete it, but it won't do them any good because it won't be accepted for project access. People who haven't yet started will be directed to v3.
> 
> How would you, as the administrator, handle each of these two scenarios?

Might be a good use for semantic versioning?

* Add a training version attribute, v1.X
* When a training is updated with a new minor version, we add the new training object but don't force users to take the new training.
* When a training is updated with a new major version, the old training objects are set to expire 90 days from that date. Users are notified that they must update their training within 90 days.
* For users in the middle of completing a training, the training object would either (1) be invalidated or (2) have the expiry date set to a stricter value, say 30 days.

Updating the training expiry date seems to be a nice way to gracefully transition to the new version. Although it is a bit of a workaround as we wouldn't be explicitly checking training version with the view, it's probably worth keeping `has_access` simpler.

tompollard:
Something that is making me a bit confused: If we are offering new training, should this be entirely independent from the "training" checks that we have currently?

i.e. the current training functionality is focused around training requirements. Now we are introducing new functionality for running courses. Are these two completely different things?

harryface:
> Something that is making me a bit confused: If we are offering new training, should this be entirely independent from the "training" checks that we have currently?
> 
> i.e. the current training functionality is focused around training requirements. Now we are introducing new functionality for running courses. Are these two completely different things?

@tompollard it is not independent of Training and all the checks been ran, a successful on-platform training actually saves it in the Training Table, and OnPlatformTraining is a proxy to TrainingType.

harryface:
This has been replaced by 
- [On Platform Training 1](https://github.com/MIT-LCP/physionet-build/pull/1887) and
- [On Platform Training 2](https://github.com/MIT-LCP/physionet-build/pull/1889)

Issue #1858, opened 2023-02-07
Remove Host's name from the "Event Registration" page


tompollard:
When someone is invited to an event, they see the following information:

![Screenshot 2023-02-07 at 1 41 01 PM](https://user-images.githubusercontent.com/822601/217336502-e77e0ef3-8c79-4fcc-bbe2-4f7aaa0e420d.png)

We should reconsider what we display on this page. e.g.: we probably want to remove the Host's name from the page. In general, I think the title is sufficient.

tompollard:
We might also want to add a short sentence that explains the registration process (e.g. "After clicking confirm, you will be added to a waiting list for the event. You will receive a notification to let you know when you have been successfully registered.").

tompollard:
- Should the description be displayed on this page too?
- Also, is "You have been invited" the right language to use? Is it true that the person has been invited?

superryeti:
Noted. I also remember we had a discussion about the registration page being separate from the event detail page(public view)

1. Invitation/event registration page  - the page where users and request to join the waiting list for event or join the event directly in case of public event
2. Event detail page - a page that displays details of events for participants, organizers, host etc. If i remember correctly, this page should show details depending on the access(like if the participant is on waitlist, they can see only minimal information, and once approved they can see the details of the event, like maybe host name, dua, etc

@tompollard  Do you think it makes sense to have
1. For the registration page, something like /events/slug/invitation
2. For the event detail page to have /events/slug
3. 
or maybe we use the current url and once someone joins the waiting list, they can see the event detail page. (this sound unnecessarily complicated as then we will have to think about what do we do if someone had already applied and got rejected, do we show them the registration page or the detail page

superryeti:
> * Should the description be displayed on this page too?
> * Also, is "You have been invited" the right language to use? Is it true that the person has been invited?

I had the same questions,

Regarding the first one,  if we plan to have a separate event detail page, maybe we could say something like. You can find the event detail here.(and also add a button in the event detail page, so that people can request access to the event)

Or if we plan to just have the description here, I think its a good idea to show the information in table like format, like the console event detail page(In this case i think we show the start, end date , description, maybe something like x days  before the registration closes)

superryeti:
> * Also, is "You have been invited" the right language to use? Is it true that the person has been invited?

[...]

PR #1861, opened 2023-02-08
update event registration page

superryeti:
Address https://github.com/MIT-LCP/physionet-build/issues/1858

Context: The current Event Registration page is very minimal and doesnot provide details about the event. It also shows who is the organizer of the event(on the issue https://github.com/MIT-LCP/physionet-build/issues/1858 we discussed that we dont want to always show the event host as in some cases events will have event managers)

Similarly, the current page doesnot explain what the participants should expect after registration.

The PR adds more details about the event in a more organized manner, and also adds note on what to expect

Before:
image

After(first look)

image

### DIFF
--- a/physionet-django/events/templates/events/event_detail.html
+++ b/physionet-django/events/templates/events/event_detail.html
@@ -2,7 +2,7 @@
 
 {% load static %}
 
-{% block title %}Events Home{% endblock %}
+{% block title %}Event - {{ event.title }}{% endblock %}
 
 {% block local_css %}
 <link rel="stylesheet" type="text/css" href="{% static 'project/css/project-home.css' %}">
@@ -11,18 +11,11 @@
 {% block content %}
 <div class="container">
 {% if event %}
-    <h2 class="form-signin-heading">Event Registration</h2>
+    <h2 class="form-signin-heading">{{ event.title }}</h2>
     <div class="col-md-9 no-pd">
         <br/>
+        <h4>Event Details</h4>
         <hr>
-        <div class="row mb-1">
-                <div class="col-md-3">
-                    Event name:
-                </div>
-                <div class="col-md-9">
-                     {{ event.title }}
-                </div>
-        </div>
 {#        <div class="row mb-1">#}
 {#            <div class="col-md-3">#}
 {#                Hosted by:#}
--- a/physionet-django/events/templates/events/event_detail.html
+++ b/physionet-django/events/templates/events/event_detail.html
@@ -34,15 +34,6 @@
         </div>
         <div class="row mb-1">
             <div class="col-md-3">
-                Created on:
-            </div>
-            <div class="col-md-9">
-                {{ event.added_datetime | date:"M. d, Y" }}
-            </div>
-        </div>
-
-        <div class="row mb-1">
-            <div class="col-md-3">
                 Start Date:
             </div>
             <div class="col-md-9">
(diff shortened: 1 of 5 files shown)


physionet-django/events/templates/events/event_detail.html
```diff
@@ -0,0 +1,88 @@
...
+        </div>
+
+
+
+    </div>
+        <hr>
+
+
+    <br>
+    <p>You may request to join this event below.</p>
```
tompollard: Please remove "<p>You may request to join this event below.</p>". It's clear from the following text/button that you can do this.

physionet-django/events/templates/events/event_detail.html
```diff
@@ -0,0 +1,88 @@
...
+{#        </div>#}
+        <div class="row mb-1">
+            <div class="col-md-3">
+                Category:
+            </div>
+            <div class="col-md-9">
+                {{ event.category }}
+            </div>
+        </div>
+        <div class="row mb-1">
```
tompollard, line 35: Please remove "Created on". This field has no meaning to the registrant.

physionet-django/events/templates/events/event_detail.html
```diff
@@ -0,0 +1,88 @@
...
+
+
+    <br>
+    <p>You may request to join this event below.</p>
+
+    <strong><p>Requesting participation adds your details to a waiting list. You will receive a notification when your request has been processed.</p></strong>
+
+    <form  method="post">
+        {% csrf_token %}
+        <button type="submit" class="btn btn-sm btn-primary" name="confirm_event" value="confirm" role="button">Request Participation</button>
```
tompollard: "Request Participation" sounds slightly awkward. Perhaps replace with "Join waitlist".

physionet-django/events/templates/events/event_detail.html
```diff
@@ -0,0 +1,88 @@
...
+
+
+    </div>
+        <hr>
+
+
+    <br>
+    <p>You may request to join this event below.</p>
+
+    <strong><p>Requesting participation adds your details to a waiting list. You will receive a notification when your request has been processed.</p></strong>
```
tompollard: Replace "Requesting participation adds your details to a waiting list. You will receive a notification when your request has been processed." with:

"Click the button below to join the waiting list. You will receive a notification when your registration is confirmed."

tompollard (commented):
I have suggested a few tweaks to the content. "Join waitlist" might not be quite right, but I think this is the best option for now. We can improve later.

physionet-django/events/templates/events/event_detail.html
superryeti: Thats a valid point. Will do that.

physionet-django/events/templates/events/event_detail.html
superryeti: I agree. "Join waitlist" is to the point and consistent to what happens.(The participants join the waitlist after clicking the button)

superryeti:
Thanks @tompollard , The page looks like this now(please see screenshot)

Do we want to merge this change, so that i can open two different PR
1. Work on Event Detail page(with different options depending on different cases)
2. Send notifications to Host, participants after approval or when someone joins a waitlist
3. Allow Hosts to assign participants as event managers, and display contact information on the event detail page

In this PR, i started renaming the template and views and was thinking it would be a good idea to avoid weird merge conflict(in case if i start working on the above 3 from the current dev)


This is how the page looks so far


image

tompollard:
The event_detail template is a little messy (unnecessary blank lines, not sure we want the comment section etc), but I'm happy to merge as (1) it's an improvement on the previous template and (2) the improvements are ongoing.

tompollard:
(Just because it isn't clear from the persnickety comment above...what I was thinking was: this is great, a big improvement on what we had before. Thanks again Amit!).

superryeti:
Thanks Tom. I will work on the other minor PRs. This PR getting merge unblocks bunch of parallel changes.

Issue #1381, opened 2021-08-24
API to determine latest version of a published project


bemoody:
The wfdb-python package wants to be able to read data from a published project, specifying only the slug and requesting the "latest version".

This is... kinda messy, since the latest version might not be what you expect.  But it is a convenience for programmers.

The way this is currently implemented in wfdb-python, though, is **very** fragile:

    # url is the project URL such as 'https://physionet.org/content/mitdb/'
    response = requests.get(url)
    contents = [line.decode('utf-8').strip() for line in response.content.splitlines()]
    version_number = [v for v in contents if 'Version:' in v]
    version_number = version_number[0].split(':')[-1].strip().split('<')[0]

Various alternatives come to mind even without making any changes to the server:
- version could be extracted from the redirected URL (`/content/mitdb/1.0.0/`)
- version could be extracted from the LD-JSON data
- version could be extracted from the content URL in the LD-JSON data
- version could be found by parsing the entire database list at `/rest/database-list/`

But whatever we do needs to be robust, efficient, and supported by the server in the long term (hence why I'm raising this issue here and not in `wfdb-python`).

Ideally the programmer should be required to supply a *major version number* along with the project slug so that incompatible future versions won't be used automatically.  (The version number might default to 1; note that all currently published waveform DBs are at version 1 except for `nch-sleep/0.1.0`.)

So one idea would be to add a URL such as `/rest/content/mitdb/1/`, that would return data similar to `/rest/database-list/`, but only for the latest version of mitdb 1.x.

Another idea I've tossed around at times would be to add a URL such as `/content/mitdb/1.*/` that would redirect to `/content/mitdb/1.0.0/`.  Parsing the location header would be simpler for the client than parsing a JSON string, but reduces long-term flexibility for the server.

alistairewj:
Looks like this relates to #821 - briefly a user thought there were some browser usability advantages to having a "latest" or "current" as a slug redirecting to the latest version. You raised some good points about exactly resolving versions in that issue.

(Side note: I actually think #821 is already dealt with to some extent...? In that if I go to https://physionet.org/content/mimiciv/ I am automatically redirected to the latest version.)

[...]

Issue #821, opened 2019-11-18
Redirect base URL to https://physionet.org/content/project_slug/current 


tompollard:
> I have another request to add to the to-do pile ahead of next year’s challenge. When I navigate to a project page, the new PhysioNet website redirects to URL with the current version number for that project, e.g., https://physionet.org/content/challenge-2019/1.0.0/

> However, when we update the project, the version number in the URL (presumably) changes. If someone bookmarks this URL or uses autocomplete in their browser to return to this page, then they’ll won’t see updates or announcements that we’ve made since then. Even worse, it will look like we haven’t updated the webpage at all.

> Is it possible for https://physionet.org/content/challenge-2019/ not to redirect to another URL, or for it to redirect to something like https://physionet.org/content/challenge-2019/current/ instead of https://physionet.org/content/challenge-2019/x.y.z/ — like the master branch on git? If not, then we’ll probably want to keep the same version number for next year’s Challenge.

tompollard:
This seems like it could work (and avoids the problem of people bookmarking specific versions):

- create a `/current` URL which displays the latest version.
- redirect the base URL to `/current` (instead of the latest version number).

Can anyone think of a reason why this isn't a good idea? @bemoody @elfeto @alistairewj

elfeto:
We can add `/current`, but the easiest would be to redirect current to latest version. The same action as `https://physionet.org/content/challenge-2019/` .

To change EVERYTHING to current instead of version, I think it might be troublesome.

tompollard:
Thanks @elfeto, I think redirecting current to the latest version doesn't really solve the problem because it has the same bookmarking issue:

1. you go to: https://physionet.org/content/challenge-2019/
2. you are forwarded to (perhaps via current): https://physionet.org/content/challenge-2019/1.0.0/
3. you bookmark the page
4. the project is updated, and your bookmark is now out of date. 

As most people just want the latest version, I think we want https://physionet.org/content/challenge-2019/current to be a proxy for the latest version.

bemoody:
Mm.  So, personally, I am in favor of directing people to *using*
versioned URLs as much as possible.  If you cite PN in another work,
you should be using a versioned URL.  If you are downloading files in
a script, you should be using a versioned URL.  If you send a message
to your colleague "hey, take a look at this funky waveform", you
should be using a versioned URL.

Which is to say that I think the current behavior is the right one in
almost all cases.  I also like that the sloppy /content/foo/ redirects
[...]

PR #1866, opened 2023-02-10
Ru/latest paths

Rutvikrj26:
Added `/latest/` Paths for API and actual project. The point was referenced in several outstanding issues. 

1. Issue #821 
2. Issue #1381 

The Current implementation adds `content/<project_slug>/latest/` & `content/<project_slug>/latest/sub-dir`.

The idea behind the current implementation is to have a website that can be bookmarked by the general public for browsing, that will provide the latest data. Thus, to serve the purpose, the latest is not redirected but serves the exact latest version of the project.

There are several other functions, such as get-zip, files-panel, etc that are not yet implemented.

### DIFF
--- a/physionet-django/export/urls.py
+++ b/physionet-django/export/urls.py
@@ -8,6 +8,8 @@ urlpatterns = [
     path('v1/project/published/', views.PublishedProjectList.as_view(), name='Published_project_list'),
     path('v1/project/published/<str:project_slug>/', views.ProjectVersionList.as_view(),
          name='Published_Project_versions'),
+    path('v1/project/published/<str:project_slug>/latest/', views.PublishedProjectLatestDetail.as_view(),
+         name='Published_Project_latest_detail'),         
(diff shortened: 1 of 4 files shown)


bemoody:
I'm closing this because I don't think there is any consensus about how these features should work.  We need to agree on that first.

Personally, I've commented on these issues a lot and I don't feel like my comments have been addressed.

bemoody:
(To be clear, not blaming you, Rutvik, or anybody else.  We just need to talk this through before trying to implement it.)

Rutvikrj26:
Yeah, No Worries @bemoody . I was going to close this PR eitherways. 

I had a talk with Tom about this. After reading all the previous comments and talking with @tompollard, I understood that there was a need for `/latest/` that could be bookmarked by general public viewing the website. Hence, I made paths that would serve with `/latest/`, rather than redirecting. We also came to the same consensus that this could create several problems, and hence was not a good idea, at least before it is well discussed and thought out.  

Thanks for closing!

PR #1867, opened 2023-02-11
Add support for `hdn-research-environment` app

kshalot:
I closed https://github.com/MIT-LCP/physionet-build/pull/1538 in favor of this PR.

This is PR adds conditional support for the `hdn-research-environment` package (the dependency used by T-CAIREM to run the cloud workbenches for researchers).

To sum up the changes:
* `pyproject.toml` and `poetry.lock` were updated to include the `hdn-research-environment@^1.2.0` dependency (+ `requirements.txt` was exported from `poetry.lock`)
* New environment variables were added + settings based on those variables
* A link to the application's dashboard was added to the navbar (appears only when the app is enabled)
* The application's URLs are included (only when the app is enabled)
* A manager for the `PublishedProject` model was created (more detail in the comments)


physionet-django/project/managers/publishedproject.py
```diff
@@ -0,0 +1,49 @@
+from django.db.models import Manager, Q
+from project.models import AccessPolicy, DUASignature, DataAccessRequest
+from user.models import Training, TrainingType
+
+
+class PublishedProjectManager(Manager):
+    def accessible_by(self, user):
```
kshalot, line 7: We need a way to fetch all the projects that a specified user has access to (part of the contract required by the `hdn-research-environment` package). There is the [has_access](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/modelcomponents/publishedproject.py#L206) method available to instances of `PublishedProject`, but using it will be very inefficient (N+1 queries).

The `accessible_by` method only uses a few queries to fetch this information.

bemoody:
> We need a way to fetch all the projects that a specified user has access to (part of the contract required by the hdn-research-environment package). There is the has_access method available to instances of PublishedProject, but using it will be very inefficient (N+1 queries).
>
> The accessible_by method only uses a few queries to fetch this information.

Makes sense.  It's awkward to have two methods that are supposed to give the same answer in very different ways, but I don't see a clear alternative.  Would you please add a note in the docstring of `accessible_by`, saying this logic needs to be kept in sync with `has_access`, and vice versa?

kshalot:
@bemoody Agreed. I updated the docstrings for both methods - let me know whether the wording works for you.

tompollard:
Note on how requirements were generated here:

```
poetry export -f requirements.txt --output requirements.txt
```

requirements.txt
```diff
@@ -500,5 +309,181 @@ libsass==0.21.0 \
...
+setuptools==65.5.1 ; python_version >= "3.8" and python_version < "4.0" \
+    --hash=sha256:d0b9a8433464d5800cbe05094acf5c6d52a91bfac9b52bcfc4d41382be5d5d31 \
+    --hash=sha256:e197a19aa8ec9722928f2206f8de752def0e4c9fc6953527360d1c36d94ddb2f
+six==1.16.0 ; python_version >= "3.8" and python_version < "4.0" \
+    --hash=sha256:1e61c37477a1626458e36f7b1d82aa5c9b094fa4802892072e49de9c60c4c926 \
+    --hash=sha256:8abb2f1d86890a2dfb989f9a77cfcfd3e47c2a354b01111771326f8aa26e0254
+sqlparse==0.4.2 ; python_version >= "3.8" and python_version < "4.0" \
+    --hash=sha256:0c00730c74263a94e5a9919ade150dfc3b19c574389985446148402998287dae \
+    --hash=sha256:48719e356bb8b42991bdbb1e8b83223757b93789c00910a616a071910ca4a64d
+tzdata==2022.7 ; python_version >= "3.8" and python_version < "4.0" and sys_platform == "win32" \
```
alistairewj, line 474: @kshalot - is this all syntactic sugar?

tompollard: These environment markers are defined at https://peps.python.org/pep-0508/ and https://peps.python.org/pep-0345/#environment-markers. 

If I understand correctly `sys_platform == "win32"` is essentially an if clause, saying that the package should only be installed on win32 environments (skip otherwise).

tompollard:
This looks good to me. @bemoody are you okay with the updated/reorganized package installation files?

requirements.txt
kshalot: For the record from the meeting - this is generated from `poetry.lock`. This particular change (tzdata) is also present in the `dev` branch.

kshalot:
In the context of https://github.com/MIT-LCP/physionet-build/issues/1870 - it seems that `poetry export -f requirements.txt --output requirements.txt` does a better job at making sure that both `requirements.txt` and `poetry.lock` are consistent. But if you rely on `requirements.txt` in your deployment you'd essentially be bumping a bunch of subdependencies when this PR is merged and deployed.

Semantic versioning should have our backs here, but if you prefer to only change what's necessary for the `hdn-research-environment` package to work in `requirements.txt` and worry about aligning the files later, we can do that too.

tompollard:
@kshalot I like the idea of using `poetry.lock` as the source of truth and building `requirements.txt` from it. I'm happy with this PR as it is, but will double check with Benjamin before we merge.

PR #1869, opened 2023-02-13
allow admins(people with all access permission to events) to manage event data set

superryeti:
Related to https://github.com/MIT-LCP/physionet-build/issues/1839

Context:
For events app, we want the admin to add datasets to the event, so that event participant can skip the usual verifications like credentialing, training and have speedy access to the datasets. The dataset access will expire after the event

This PR is the first step to the adding of dataset. The changes will allow admin to link datasets to event and remove the datasets if needed.

image

image


physionet-django/console/views.py
```diff
@@ -2885,9 +2886,40 @@ def event_management(request, event_slug):
     Admin page for managing an individual Event.
     """
     selected_event = get_object_or_404(Event, slug=event_slug)
+
+    event_dataset_form = EventDatasetForm()
```
alistairewj: Most of the repo adopts an `if ... else` structure for initializing forms rather `EmptyForm() .. if POST`. IMO better to keep it that way and do `if ... else`

physionet-django/console/templates/console/event_management.html
```diff
@@ -1,149 +1,79 @@
...
+       aria-hidden="true">
+    <div class="modal-dialog modal-lg" role="document">
+      <div class="modal-content">
+        <div class="modal-header">
+          <h5 class="modal-title">Participants</h5>
+          <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+            <span aria-hidden="true">&times;</span>
+          </button>
         </div>
+        {% include "events/event_entries.html" %}
```
alistairewj: single quotes (to match convention)

physionet-django/console/views.py
```diff
@@ -2885,9 +2886,40 @@ def event_management(request, event_slug):
...
+            event_dataset_form = EventDatasetForm(request.POST)
+            if event_dataset_form.is_valid():
+                if selected_event.datasets.filter(
+                        dataset=event_dataset_form.cleaned_data['dataset'],
+                        status=EventDataset.EventDatasetStatus.Active).count() == 0:
+                    event_dataset_form.instance.event = selected_event
+                    event_dataset_form.save()
+                    messages.success(request, "The dataset has been added to the event.")
+                else:
+                    messages.error(request, "The dataset is already added to the event.")
```
alistairewj: *has already been

physionet-django/console/templates/console/event_management_manage_dataset.html
```diff
@@ -0,0 +1,81 @@
+<div class="card mb-3">
+  <div class="card-header">Project Dataset</div>
+  <div class="card-body">
+    <ul class="list-group list-group-flush">
+      <li class="list-group-item">
+        <h5>Add Dataset</h5>
```
alistairewj: Add a dataset to the event

physionet-django/console/templates/console/event_management_manage_dataset.html
```diff
@@ -0,0 +1,81 @@
...
+        <form action="" method="post" class="form-event-dataset">
+          {% csrf_token %}
+          {% include "form_snippet.html" with form=event_dataset_form %}
+          <button class="btn btn-primary btn-fixed"
+                  name="add-event-dataset"
+                  type="submit">Add Event</button>
```
alistairewj: Add to event

physionet-django/console/templates/console/event_management_manage_dataset.html
```diff
@@ -0,0 +1,81 @@
...
+                      <div class="modal-dialog modal-lg" role="document">
+                        <div class="modal-content">
+                          <div class="modal-header">
+                            <button class="close" data-dismiss="modal" aria-label="Close">
+                              <span aria-hidden="true">x</span>
+                            </button>
+                          </div>
+                          <div class="modal-body">
+                            <div class="row">
+                              <p>Are you sure you want to delete the dataset {{ dataset.dataset.title }}</p>
```
alistairewj: Not quite right since we aren't deleting the dataset

```suggestion
                              <p>"Remove {{ dataset.dataset.title }} from the event?</p>
```

physionet-django/console/templates/console/event_management_manage_dataset.html
```diff
@@ -0,0 +1,81 @@
...
+                            </div>
+                          </div>
+                          <div class="modal-footer">
+                            <button class="btn btn-group btn-danger"
+                                    type="submit"
+                                    value="{{ dataset.id }}"
+                                    name="remove-event-dataset">
+                              Yes
+                            </button>
+                            <button class="btn btn-primary" data-dismiss="modal">Cancel</button>
```
alistairewj: I think yes/no is simpler here

```suggestion
                            <button class="btn btn-primary" data-dismiss="modal">No</button>
```

physionet-django/events/models.py
```diff
@@ -138,3 +138,49 @@ def withdraw(self, comment_to_applicant=None):
...
+        Active = 'AC', _('Active')
+        Inactive = 'IN', _('Inactive')
+
+    class EventDatasetAccessType(models.TextChoices):
+        GOOGLE_BIG_QUERY = 'GBQ', _('Google BigQuery')
+        RESEARCH_ENVIRONMENT = 'RE', _('Research Environment')
+
+    event = models.ForeignKey(Event, on_delete=models.CASCADE, related_name='datasets')
+    dataset = models.ForeignKey("project.PublishedProject", on_delete=models.CASCADE)
+    access_type = models.CharField(default=EventDatasetAccessType.GOOGLE_BIG_QUERY, max_length=3,
```
alistairewj: 3 characters seem short. I'd make this 10 just to save effort in the future. For example we might want `AWS-S3`, etc.

physionet-django/console/views.py
```diff
@@ -2885,9 +2886,40 @@ def event_management(request, event_slug):
...
+    event_dataset_form = EventDatasetForm()
+
+    # handle the add dataset form(s)
+    if request.method == "POST":
+        if 'add-event-dataset' in request.POST.keys():
+            event_dataset_form = EventDatasetForm(request.POST)
+            if event_dataset_form.is_valid():
+                if selected_event.datasets.filter(
+                        dataset=event_dataset_form.cleaned_data['dataset'],
+                        status=EventDataset.EventDatasetStatus.Active).count() == 0:
```
alistairewj: This works for our immediate aim but isn't future proof. We'll probably want the ability add datasets through multiple access mechanisms, e.g. you are allowed to use this dataset via BigQuery and Google Storage (GBQ / GCS). This will only allow us to add a dataset for a single access mechanism. I'd change this to check against access type as well.

alistairewj (changes requested):
Mostly very very minor comments on the wording - the only real change I'd say is determining whether an event is valid to add.

superryeti:
> Mostly very very minor comments on the wording - the only real change I'd say is determining whether an event is valid to add.

Thank you @alistairewj . I will address the comments and update the PR early morning.

physionet-django/console/views.py
superryeti: Good point. I originally did the code to check against access type. but reverted it to this(just in case someone might suggest that we just want the single access mechanism)

Let me fix this.

physionet-django/events/models.py
```diff
@@ -153,3 +153,49 @@ def withdraw(self, comment_to_applicant=None):
...
+        """
+        checks if the dataset is accessible to the user
+        Dataset is accessible if:
+        1. The dataset is active
+        2. The event has not ended
+        """
+        if self.status == self.EventDatasetStatus.Inactive:
+            return False
+
+        if self.event.end_date < timezone.now().date():
```
alistairewj: This will prevent users accessing data on the last day of an event, as 02-25-2022 < 02-25-2022. Should change this to `<=`

superryeti: Good catch. i think this issue might be in Event as well. i will check and open a separate PR for that.

I made the suggested change in this PR also

superryeti:
Thanks @alistairewj, @tompollard   for the review. i have implemented the suggested changes/fixes

physionet-django/events/models.py
superryeti: Upon further inspection, I think the current code implementation is correct.

Our goal is to let users access dataset till the last day

Case 1
Say last day of event is 25 Feb. 2023, and the current date is also  25 Feb. 2023,
This means the if statement will be False. and the function will return True. hence the data will be accessible.

Case 2
Say last day  of event is 25 Feb. 2023, and the current date is 26 Feb. 2023,
This means the if statement will be True. and the function will return False. hence the data will not be accessible.

image


image

superryeti: Because, the function works like this
Check all the cases where we might reject the access. if we can reject access for any reason, reject it rightaway
if not keep checking, and finally after everything is checked, and we have no reason to reject access, Grant the access.

amitupreti: I might be confused here. So let me do a quick test on the local setup

superryeti: Resolved after discussing with @alistairewj  on slack.

We decided to improve the readibiliy by using
`if now_date > end_date -> return False`

superryeti:
@alistairewj  Thank you for the review. I updated the code with the readability fix on event model.

 https://github.com/MIT-LCP/physionet-build/pull/1869/commits/1d2eaebdd71157d19db0935168e023b00382157b

physionet-django/events/models.py
```diff
@@ -153,3 +153,49 @@ def withdraw(self, comment_to_applicant=None):
...
         """
         self._apply_decision(self.EventApplicationStatus.WITHDRAWN, comment_to_applicant)
+
+
+class EventDataset(models.Model):
+    """
+    Captures information about datasets for events.
+    """
+    class EventDatasetStatus(models.TextChoices):
+        Active = 'AC', _('Active')
```
alistairewj: Thinking about it more, maybe we should remove the Active/Inactive status? We don't actually provide a mechanism of activating/deactivating event datasets, and the functionality seems duplicative of removing an event dataset.

alistairewj:
This looks good to me overall - @tompollard do you think we should leave the "Active/Inactive" status for EventDataset? I can't think of a use for this that isn't covered by simply removing the dataset from the event, but maybe I missed something.

physionet-django/events/models.py
superryeti: For sure, did you mean to remove this class and use status as a simple boolean? @alistairewj

superryeti: But yes, this is used to set status as inactive once someone removes the dataset from event, and we never let them reactivate it, so this looks unnecessary(unless we want the option to allow them to reactivate a dataset, but that sounds like an extra feature/work and maybe its just better with a simple boolean)

alistairewj: Yeah I was actually wondering if we need the status at all. We don't seem to use it anywhere. Seems the add/remove dataset covers this well enough!

alistairewj: So, perhaps the best resolution is, why don't you make a commit where you remove all references to the `EventDatasetStatus`, and we can keep the commit in the history. If we decide to add the feature in later, it will be easy to bring it back. Then I think we are good to merge.

superryeti: Sounds good. doing that right now

superryeti:
Thanks @alistairewj. I added a commit here(based on Tom's Feedback), i can remove it and rebase or we can merge it as it is.

https://github.com/MIT-LCP/physionet-build/pull/1869/commits/3547ae6c60183c42d11d1df59e32ff387d661f8b

alistairewj:
Happy for that approach - though looks like the commit message got cut off half way through so you may want to amend it.

superryeti:
Thanks, @alistairewj . I amended the commit.

alistairewj (approved):
I think `status` would be better named `active` (or `is_active` like the form) but that's a minor point.

tompollard:
> I think status would be better named active (or is_active like the form) but that's a minor point.

I agree that making it clear this is a boolean field would be better. Wouldn't normally worry too much about variable names, but this is a database field so better to fix now than have to migrate again later! @amitupreti please could you update as suggested (e.g. `as_active`).

superryeti:
> it clear this is a boolean field would be better. Wouldn't normally worry too much about variable names

Thank you. doing it right now

tompollard:
@amitupreti do you prefer `active` or `is_active`? I see that we use `is_active` in other places (e.g. https://github.com/MIT-LCP/physionet-build/blob/d82d6f9a8be9acc2978595c6abbec0229c044679/physionet-django/user/models.py#L357). I haven't checked to see whether `active` is used elsewhere.

superryeti:
> it clear this is a boolean field would be better. Wouldn't normally worry too much about variable names

Thank you. doing it right now

> @amitupreti do you prefer `active` or `is_active`? I see that we use `is_active` in other places (e.g.
> 
> https://github.com/MIT-LCP/physionet-build/blob/d82d6f9a8be9acc2978595c6abbec0229c044679/physionet-django/user/models.py#L357
> 
> ). I haven't checked to see whether `active` is also used elsewhere.

Yup just realized is_active is better and force pushed the change already

superryeti:
Thanks @tompollard .i updated the commit with `is_active`. There was test failure when installing python packages but it went away after i reran the test

tompollard:
Sorry Amit, please could you clean up the commits a little? e.g. 40f86a00b76676e4f635165817070f10dec5eb28 is an unused migration so doesn't need to be there.

tompollard:
(if it isn't too much hassle, please could you rebase on dev at the same time?)

superryeti:
> [40f86a0](https://github.com/MIT-LCP/physionet-build/commit/40f86a00b76676e4f635165817070f10dec5eb28)

Doing that. I think i rebased to dev yesterday. Let me do it again just to be sure.

superryeti:
> Sorry Amit, please could you clean up the commits a little? e.g. [40f86a0](https://github.com/MIT-LCP/physionet-build/commit/40f86a00b76676e4f635165817070f10dec5eb28) is an unused migration so doesn't need to be there.

checking all the commits and commits with migration

Sorry about the one with unused migrations. We were waiting on this just in case you wanted to use the Active/Inactive stuff)

tompollard:
Thanks @amitupreti, this looks good to me. I think it's better for the migration to go into a separate commit. ("remove EventDatasetStatus from EventDataset Model") doesn't seem to have much to do with the migration, but not a big deal/

superryeti:
> Thanks @amitupreti, this looks good to me. I think it's better for the migration to go into a separate commit ("remove EventDatasetStatus from EventDataset Model") doesn't seem to have much to do with the migration, but not a big deal/

Good call. Let me do it right now. it should take couple of minutes only.

superryeti:
Thanks @tompollard . I separated the migration file into a last commit(its super convenient to have the migration file separately as a last commit)

tompollard:

PR #1871, opened 2023-02-14
Update fixtures for event so that we have a testable event in the Events tab

superryeti:
Context:
Currently, the fixtures for event has just 1 event(but that event is in past, already finished). So that makes it difficult to test the Event app(unless we create a new Event).

So this PR should add a new event to the fixtures, so that we can quickly do our tests after doing `loaddemo`

1. Also, the event fixture had one small bug, where we hadn't assigned the `Event Host` permission to the existing event's host `rgmark`, this PR also adds the user to the `Event Host` auth group.
2. Also, refactored the fixture to move the `auth group` definition to demo-user.json fixture.

Conclusion: After this change, once you do `loaddemo`, you should be able to see the `Events` tab when you login as `User` `rgmark` , and can play around with an Active Event.

### DIFF
--- a/physionet-django/events/fixtures/demo-events.json
+++ b/physionet-django/events/fixtures/demo-events.json
@@ -15,23 +15,5 @@
       "slug": "iLII4L9jSDFh",
       "allowed_domains": null
     }
-  },
-  {
-    "model": "auth.group",
-    "fields": {
-      "name": "Host",
-      "permissions": [
-        [
-          "add_event",
-          "events",
-          "event"
-        ],
-        [
-          "view_event_menu",
-          "events",
-          "event"
-        ]
-      ]
-    }
   }
 ]
--- a/physionet-django/events/fixtures/demo-events.json
+++ b/physionet-django/events/fixtures/demo-events.json
@@ -15,5 +15,22 @@
       "slug": "iLII4L9jSDFh",
       "allowed_domains": null
     }
+  },
+  {
+    "model": "events.event",
+    "pk": 2,
+    "fields": {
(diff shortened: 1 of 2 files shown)


tompollard:
Unless I made a mistake somewhere, it looks like rgmark does not have permission to (1) create events or (2) edit events. Possibly `events` should be `event` in the fixtures somewhere? I think this may have changed at some point. 

These are the permissions that should be assigned to host:

Screenshot 2023-02-14 at 4 02 06 PM

superryeti:
> events



> Unless I made a mistake somewhere, it looks like rgmark does not have permission to (1) create events or (2) edit events. Possibly `events` should be `event` in the fixtures somewhere? I think this may have changed at some point.
> 
> These are the permissions that should be assigned to host:
> 
> Screenshot 2023-02-14 at 4 02 06 PM

Good catch. it seems like when we moved the events from `user` app.

we forgot to change the permission in the view.

So i think we need to change the views
```
is_instructor = request.user.has_perm('user.add_event')

to
is_instructor = request.user.has_perm('events.add_event')
```

superryeti:
Let me open another small PR for that, and then i will rebase this after that is merged

tompollard:
sounds good, thanks amit.

superryeti:
Opened a Small PR here for the Event Host not being able to create/edit events

https://github.com/MIT-LCP/physionet-build/pull/1872

superryeti:
@tompollard  rebased this and forced push

tompollard:
Looks good, thanks!

PR #1872, opened 2023-02-14
Refactor old event permission

superryeti:
Looks like when we moved the event app out of the user app, we forgot to refactor the permission check in the view.

This PR fixes that. Credits to @tompollard  for finding this.


tompollard:
Thanks for picking this up Amit, looks good to me!

PR #1873, opened 2023-02-14
Project version for wfdb-python

bemoody:
wfdb-python does some silly and fragile parsing to try to extract the version number from the HTML project page.  This is really a bug in wfdb-python, and I knew about this issue ages ago and forgot about it, *but* I still don't want to break existing scripts.

This was broken I think due to pull #1823.

I *think* this should fix it but yeah, it's a terrible kludge, and I haven't tested it.


tompollard:
Thanks Benjamin, let's give it a try!

PR #1874, opened 2023-02-15
Allow Event Hosts to add CoHosts

superryeti:
Context:  https://github.com/MIT-LCP/physionet-build/issues/1839 As mentioned here, for Events app, we want to allow Event Hosts to add cohosts to the event.

This PR updates the Model with a boolean to toggle the cohost status. As discussed on the meetings, we add a checkbox which sends an ajax requests and updates the cohost status for participant

image


superryeti:
I will add two small PRs after this

1. To send email notifications to the cohosts
2. To allow Event hosts to approve/reject other participants and see the list of participants

physionet-django/events/templates/events/event_home.html
```diff
@@ -197,5 +197,28 @@ <h5 class="modal-title">Participants</h5>
...
+                success: function (data) {
+                  console.log(data);
+                  if(data[0].status===true){
+                    checkbox.prop('checked', !checkbox.prop('checked'));
+                  }
+                },
+            });
+        });
+    });
+
```
superryeti, line 266: Here i had to add the javascript code outside the template `event_entries.html` because if i add this code inside the `event_entries.html` the same block of javascript will get called multiple times(once for each event) because the `event_entries.html` is included multiple times inside the `event_home.html` template.

So adding the code directly in `event_home.html` template will work

physionet-django/events/templates/events/event_entries.html
```diff
@@ -6,6 +6,7 @@
           <th>Full name</th>
           <th>Email</th>
           <th>Credentialed</th>
+          <th>CoHost</th>
```
tompollard: H is not capitalised.

superryeti: Thanks @tompollard , I made a new commit and squashed the change

tompollard (commented):
Thanks Amit, I think I'm good with this (going to sit on it for an hour to be sure!). If you have a chance in the meantime, please could you add the Cohost name/s below the Host's name on the events page?

Screenshot 2023-02-16 at 11 10 31 AM

superryeti:
> the Cohost name/s below the Host's name on the events page
@tompollard 

Sounds good. Let me do that, i will rebase this with the dev and add the Cohost in the Event Listing.

Slightly related:
I remember discussing about having an option shown to Event Host about whether to show contact details of cohosts in the Event Detail page or not. Do we want to do that here?

I was thinking if its as simple as i mentioned above, then we could just add a boolean in the `Event` model. or maybe we need to discuss more about what we want here, and move this to Nice to have list and discuss after the HDN datathon

tompollard:
> I remember discussing about having an option shown to Event Host about whether to show contact details of cohosts in the Event Detail page or not. Do we want to do that here?

I don't think it is something we need right now, so save this thought for later!

tompollard:
On reflection I think I might prefer this to be managed through some kind of "Invite co-host" button that would (1) send an invitation to co-host (2) allow the co-host to accept/decline the invitation. Let's keep the current javascript implementation for now and review/discuss later. It's easy to change in a subsequent pull request if needed.

superryeti:
> On reflection I think I might prefer this to be managed through some kind of "Invite co-host" button that would (1) send an invitation to co-host (2) allow the co-host to accept/decline the invitation. Let's keep the current javascript implementation for now and review/discuss later. It's easy to change in a subsequent pull request if needed.

Noted. I will take a note, and bring this up with you and Alistair when catching up on the remaining features we want for the event(After the HDN Datathon)

superryeti:
Thanks @tompollard , I added a commit for the Cohost on the Event List page

Image for reference

image

superryeti:
I figured out the issue, it was a bit weird one

So the test was failing in case of

`the upgrade test was doing "Run old tests, on old database, after early migrations"`

So this means the `cohost` boolean  on `EventParticipant `model was migrated but the old code and tests were being run.

For some reason when the test was running, and `EventParticipant` was being created(i.e someone was approved to the event), the `default=False` in cohost boolean was not respected, and django orm was throwing error says null not allowed.

tompollard:
While reviewing these changes, I noticed that we are using `is_instructor` and `is_host`. If I understand correctly the difference is that:

: `is_instructor` is someone who has permission to create events
: `is_host` is someone who is host of an existing event

I think this terminology appeared in an earlier PR, but it is kind of confusing. We should clear up the terminology in a separate PR.

superryeti:
> earlier

Yes about the purpose of both. You are correct

Let me take a note. I will open another PR for that today.

PR #1875, opened 2023-02-15
Bump `hdn-research-environment` to `1.2.2`

kshalot:
We need a small bump, I figured we'll upstream it immediately.

On a related note: I'm thinking about ways of keeping dependencies like this up-to-date. Maybe some GitHub Action in the future?

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -743,13 +743,13 @@ protobuf = ">=3.12.0"
 
 [[package]]
 name = "hdn-research-environment"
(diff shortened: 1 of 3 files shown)


tompollard:
> On a related note: I'm thinking about ways of keeping dependencies like this up-to-date. Maybe some GitHub Action in the future?

Yeah a GitHub action could work. How does Dependabot get its instructions to make updates? Could we piggyback on that?

PR #1876, opened 2023-02-15
allow dua creation for events

superryeti:
Context: As discussed with Tom and Alistair, We want event hosts to be able to create and assign Event Agreements(or Event DUAs to events). And the participants will have to read the agreement and sign it before they have full access to the Event.

Details:

 What is Event Agreements?
    -->Event Agreements are like DUAs but specific to Event.
    
    Idea behind creating a new model on this PR: My understanding is that each event will be assigned a Event Agreements and the event host can choose which Event Agreements to link to event.  One assumption i made was we dont want two different event editors to be able to edit each others Event Agreements(So i tied each Event Agreements to the Event Agreements creator)
    
     So the reason i decided to create a new model instead of using the existing DUA are
     1. To keep things simple
     2. The existing DUAs are tied to project type like `database`,`software` etc. So to make them work for events, we would have to come up with weird hack. So i thought using a new model will be cleaner and simpler.
     3. The need to tie each Event Agreement to event host(otherwise one event host might end up editing another event hosts Agreement and things might become confusing.

Questions:
Because the changes were getting super big, i didnot add any test case(but i feel like i should). So will leave upto @tompollard  if i should add the test case in this PR or open another PR for test cases.

image


tompollard:
> Because the changes were getting super big, i didnot add any test case(but i feel like i should). So will leave upto @tompollard if i should add the test case in this PR or open another PR for test cases.

Having tests alongside related code is a good thing. What I think we should try to avoid is unrelated code being grouped together into a single PR.

superryeti:
> > Because the changes were getting super big, i didnot add any test case(but i feel like i should). So will leave upto @tompollard if i should add the test case in this PR or open another PR for test cases.
> 
> Having tests alongside related code is a good thing. What I think we should try to avoid is unrelated code being grouped together into a single PR.

Alright that sounds good. let me add tests here.

This PR should be clean of unrelated stuff. I think we only have DUA Crud stuff here.

superryeti:
Thank you @tompollard.

I added tests, rebased and also fixed dependency conflict after rebasing.

physionet-django/console/templates/console/console_navbar.html
```diff
@@ -147,7 +147,7 @@
               <span class="nav-link-text">Legal</span>
             </a>
             <!-- submenu -->
-          {% if license_nav or dua_nav or code_of_conduct_nav %}
+          {% if license_nav or dua_nav or code_of_conduct_nav or perms.events.add_event %}
```
alistairewj: This makes the menu uncollapsed by default. Need to change perms.events.add_event to events_nav or the equivalent

alistairewj: This makes the menu uncollapsed by default. Need to change from the permission to the nav.

```suggestion
          {% if license_nav or dua_nav or code_of_conduct_nav or event_agreement_nav %}
```

physionet-django/console/templates/console/console_navbar.html
```diff
@@ -136,7 +136,7 @@
       {% endif %}
 
       <!-- legal -->
-      {% if perms.project.add_dua or perms.project.add_codeofconduct or perms.project.add_license %}
+      {% if perms.project.add_dua or perms.project.add_codeofconduct or perms.project.add_license or  perms.events.add_event %}
```
alistairewj: Extra whitespace here

```suggestion
      {% if perms.project.add_dua or perms.project.add_codeofconduct or perms.project.add_license or perms.events.add_event %}
```

physionet-django/console/templates/console/event_agreement_list.html
```diff
@@ -0,0 +1,93 @@
...
+                      <div class="modal-dialog" role="document">
+                        <div class="modal-content">
+                          <div class="modal-header">
+                            <h5 class="modal-title">Delete DUA</h5>
+                            <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                              <span aria-hidden="true">&times;</span>
+                            </button>
+                          </div>
+                          <div class="modal-body">
+                              <p>Are you sure you want to delete the DUA?</p>
```
alistairewj:
```suggestion
                              <p>Are you sure you want to delete the agreement?</p>
```

physionet-django/console/templates/console/event_agreement_list.html
```diff
@@ -0,0 +1,93 @@
...
+                  </div>
+                </td>
+            </tr>
+        {% endfor %}
+        </tbody>
+      </table>
+      {% include "console/pagination.html" with pagination=event_agreements %}
+    </div>
+    {% else %}
+      <p>There aren't any data Event agreements.</p>
```
alistairewj:
```suggestion
      <p>No event agreements available.</p>
```

physionet-django/console/templates/console/event_agreement_list.html
```diff
@@ -0,0 +1,93 @@
...
+                    <span class="badge badge-success">Active</span>
+                  {% else %}
+                    <span class="badge badge-danger">Inactive</span>
+                  {% endif %}
+                </td>
+                <td>{{ event_agreement.version }}</td>
+                <td>
+                  <a href="{% url 'event_agreement_detail' event_agreement.pk %}" class="btn btn-sm btn-primary" role="button">Edit</a>
+                  <a href="{% url 'event_agreement_new_version' event_agreement.pk %}" class="btn btn-sm btn-success" role="button">New version</a>
+                  <button type="button" class="btn btn-sm btn-danger " data-toggle="modal" data-target="#remove-event-agreement-{{ event_agreement.pk }}">Delete</button>
```
alistairewj: trailing whitespace

```suggestion
                  <button type="button" class="btn btn-sm btn-danger" data-toggle="modal" data-target="#remove-event-agreement-{{ event_agreement.pk }}">Delete</button>
```

physionet-django/console/views.py
```diff
@@ -2891,3 +2892,96 @@ def event_management(request, event_slug):
...
+def event_agreement_delete(request, pk):
+    if request.method == 'POST':
+        event_agreement = get_object_or_404(EventAgreement, pk=pk)
+        event_agreement.delete()
+
+    return redirect("event_agreement_list")
+
+
+@permission_required('events.add_event', raise_exception=True)
+def event_agreement_new_version(request, pk):
```
alistairewj: Since `_new_version` and `_detail` are more or less identical, it would be good to move this code block above the `_detail` one.

would really like to share the code in some way, but it's non-trivial to do that, so putting them side by side will help for now

physionet-django/events/forms.py
```diff
@@ -60,3 +60,17 @@ class EventApplicationResponseForm(forms.ModelForm):
...
+        model = EventAgreement
+        fields = (
+            'name',
+            'version',
+            'slug',
+            'is_active',
+            'html_content',
+            'access_template',
+        )
+        labels = {'html_content': 'Content'}
```
alistairewj, line 89: Add help text, e.g.

```suggestion
        labels = {'html_content': 'Content'}
        help_texts = {
            'name': '* The displayed name of the agreement.',
            'version': '* The version number of the agreement.',
            'slug': '* A simple string for use in the URL displaying the agreement. Should include the version number.',
            'is_active': '* Only active agreements are usable in future events.',
            'html_content': '* The agreement text displayed to the participant.',
            'access_template': '* Instructions on accessing the dataset.'
        }
```

Unsure if I guessed correctly at access_template.

physionet-django/console/test_views.py
```diff
@@ -614,3 +615,159 @@ def test_frontpage_button_delete(self):
...
+                'html_content': self.event_agreement_html_content,
+                'access_template': self.event_agreement_access_template
+            })
+
+        self.assertEqual(response.status_code, 200)
+        event_agreement = EventAgreement.objects.get(slug=self.event_agreement_slug)
+        self.assertEqual(event_agreement.name, self.updated_event_agreement_name)
+
+    def test_edit_event_agreement_invalid(self):
+        """tests the view that edits an invalid event agreement"""
```
alistairewj: specify you are only testing for invalid version here

physionet-django/console/test_views.py
```diff
@@ -614,3 +615,159 @@ def test_frontpage_button_delete(self):
...
+                'html_content': self.event_agreement_html_content,
+                'access_template': self.event_agreement_access_template
+            })
+        self.assertEqual(response.status_code, 200)
+        event_agreement = EventAgreement.objects.get(slug=self.event_agreement_slug)
+        self.assertEqual(event_agreement.name, self.event_agreement_name)
+        return event_agreement
+
+    def test_add_event_agreement_invalid(self):
+        """tests the view that adds an invalid event agreement"""
```
alistairewj, line 659: specify in the test name / boilerplate you are checking for invalid version

physionet-django/console/test_views.py
```diff
@@ -614,3 +615,159 @@ def test_frontpage_button_delete(self):
...
+                'is_active': True,
+                'html_content': self.event_agreement_html_content,
+                'access_template': self.event_agreement_access_template
+            })
+
+        self.assertEqual(response.status_code, 302)
+        event_agreement = EventAgreement.objects.get(slug=self.event_agreement_slug_new_version)
+        self.assertEqual(event_agreement.version, self.event_agreement_version_new_version)
+
+    def test_event_agreement_new_version_invalid(self):
```
alistairewj: rename this as you are testing the slug is invalid

physionet-django/console/test_views.py
```diff
@@ -614,3 +615,159 @@ def test_frontpage_button_delete(self):
...
+                'version': self.event_agreement_version_new_version,
+                'slug': self.event_agreement_slug,
+                'is_active': True,
+                'html_content': self.event_agreement_html_content,
+                'access_template': self.event_agreement_access_template
+            })
+
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, 'console/event_agreement_new_version.html')
+        self.assertContains(response, 'Event agreement with this Slug already exists.')
```
alistairewj: another test it would be good to add:

- try to add a project with a new slug but an existing version
- system raises a 200 and tells the user that the version already exists

physionet-django/console/views.py
```diff
@@ -2891,3 +2892,96 @@ def event_management(request, event_slug):
...
+            'event_agreement_form': event_agreement_form
+        }
+    )
+
+
+@permission_required('events.add_event', raise_exception=True)
+def event_agreement_delete(request, pk):
+    if request.method == 'POST':
+        event_agreement = get_object_or_404(EventAgreement, pk=pk)
+        event_agreement.delete()
```
alistairewj: add a message telling the user the event is deleted.
```suggestion
        event_agreement.delete()
        messages.success(request, "The Event Agreement has been deleted.")
```

alistairewj (changes requested):
Looks pretty good but I have some suggested changes.

Some general notes:

- I don't like that we can edit the agreement *after* it has potentially been signed, but this is consistent with the DUA interface. So we can leave it as is for now, but something to discuss in the future.
- Agreement slugs being unique is kind of weird, as the user has to manually include the version, but again this is how the other parts of the system work and it's a fairly minor thing to keep track of. Adding the help text will be enough for this.

superryeti:
> Looks pretty good but I have some suggested changes.
> 
> Some general notes:
> 
> * I don't like that we can edit the agreement _after_ it has potentially been signed, but this is consistent with the DUA interface. So we can leave it as is for now, but something to discuss in the future.
> * Agreement slugs being unique is kind of weird, as the user has to manually include the version, but again this is how the other parts of the system work and it's a fairly minor thing to keep track of. Adding the help text will be enough for this.

Awesome Thanks. Working on it right now

superryeti:
> * I don't like that we can edit the agreement _after_ it has potentially been signed, but this is consistent with the DUA interface. So we can leave it as is for now, but something to discuss in the future

Noted. I will bug you and Tom about this. I think there are bunch of other stuff we need to discuss about the Event(after the datathon). i will bring it up then

alistairewj (changes requested):
Looks great so far. Made some notes for changes. General comments not for this PR but for consideration:

- I don't like that we can edit the agreement *after* it has potentially been signed, but this is consistent with the DUA interface. So we can leave it as is for now, but something to discuss in the future.
- The slug being unique is weird (and requires the user to remember to put the version number in the slug). Adding the help text will help avoid user error here, and since this is an admin interface mostly it should be OK.

physionet-django/events/forms.py
```diff
@@ -60,3 +60,17 @@ class EventApplicationResponseForm(forms.ModelForm):
...
         model = EventApplication
         fields = ('status', 'comment_to_applicant')
+
+
+class EventAgreementForm(forms.ModelForm):
+    class Meta:
+        model = EventAgreement
+        fields = (
+            'name',
+            'version',
```
alistairewj, line 83: We need to verify the given name/version doesn't already exist, otherwise the server will 500.

Can borrow some code from here:

https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/forms.py#L387-L392

physionet-django/events/models.py
```diff
@@ -153,3 +155,33 @@ def withdraw(self, comment_to_applicant=None):
             comment_to_applicant (str): The comment to add to the participant about the status.
         """
         self._apply_decision(self.EventApplicationStatus.WITHDRAWN, comment_to_applicant)
+
+
+class EventAgreement(models.Model):
+    name = models.CharField(max_length=100)
+    slug = models.SlugField(max_length=120, unique=True)
```
alistairewj: This should probably use the slug validator

```python
from project.validators import validate_slug
slug = models.SlugField(max_length=120, unique=True, validators=[validate_slug])
```

physionet-django/events/models.py
```diff
@@ -153,3 +155,33 @@ def withdraw(self, comment_to_applicant=None):
...
+    version = models.CharField(max_length=15, default='', validators=[validate_version])
+    is_active = models.BooleanField(default=True)
+    html_content = SafeHTMLField(default='')
+    access_template = SafeHTMLField(default='')
+    creator = models.ForeignKey("user.User", on_delete=models.CASCADE)
+
+    class Meta:
+        default_permissions = ('add',)
+        unique_together = (('name', 'version', 'creator'),)
```
alistairewj: I suggest changing this to `(('name', 'version'))`. This is more consistent with the DUA model, and we don't want to have to start worrying about "Alistair's" agreement vs. "Amit's" agreement.

physionet-django/console/templates/console/console_navbar.html
superryeti: Oh yes, I need to use the `event_agreement_nav` here, my bad.  Good catch.

physionet-django/console/templates/console/event_agreement_list.html
superryeti: oh man. Thats an embarrassing mistake. i will remember to review my own PR first going forward.

physionet-django/console/views.py
superryeti: Regarding the first point, i like that, it looks neat to have them together.

Noted about the 2nd point. I added it to the Nice to have for event

physionet-django/events/forms.py
superryeti: Thank you. 

I am also not sure about `access_template`. on the fixtures there is no example for `access_template`.

I check with January on what HDN has for `access_template` and turns out we just have the sentence `I accept`.

But your explanation/help text is the best so far.

In any case, i will go with your suggestion for now and ping Tom about the `access_template`'s purpose on DUA

physionet-django/events/forms.py
superryeti: Solved this when adding a new test case for version.
I did a simple validation like this

```
    def clean(self):
        cleaned_data = super(EventAgreementForm, self).clean()
        if EventAgreement.objects.filter(name=cleaned_data['name'], version=cleaned_data['version'],
                                         creator=self.creator).exists():
            raise forms.ValidationError(
                {"name": ["Agreement with this name and version already exists for this user."]})
        return cleaned_data
```

superryeti: In case of projects, we had loaded all the previous project for that slug  on the variable `previous_projects` 

But for Event Agreement, that might not be the best idea to follow that implementation because our `unique_together` criteria has title, version, creator as unique. So we need to check for the title+version+creator or make different validation function for each.(the suggested method worked for the published project because we only cared about not having duplicate version inside same projects)

physionet-django/events/models.py
superryeti: For sure.  quick question, wouldn't this lead to a case where Event Host "A" might endup editing an Event Agreement that is being used by user "B" on another Event.(but i suppose for someone with access to group "Event Host", we would expect them to be responsible.)

alistairewj: Yes, that's true, but we likely wouldn't give event hosts permission to create agreements in the same way that project publishers cannot create data use agreements.

superryeti: Okay. i see. That makes sense. Admins will create the Event Agreement.

superryeti:
@alistairewj @tompollard  I addressed the comments.

Sorry it took longer than expected.

physionet-django/console/templates/console/console_navbar.html
```diff
@@ -167,6 +167,11 @@
                   <a id="code_of_conduct_nav" class="nav-link" href="{% url 'code_of_conduct_list' %}">Code of Conduct</a>
                 </li>
               {% endif %}
+              {% if perms.events.add_event %}
+                <li class="nav-item {% if event_agreement_nav %}active{% endif %}">
+                  <a id="event_agreement_nav" class="nav-link" href="{% url 'event_agreement_list' %}">Event Agreements</a>
```
alistairewj: For consistency - rebase on dev to get the other changes to nav

```suggestion
                  <a id="nav_event_agreement" class="nav-link" href="{% url 'event_agreement_list' %}">Event Agreements</a>
```

physionet-django/console/test_views.py
```diff
@@ -614,3 +615,181 @@ def test_frontpage_button_delete(self):
...
+                'version': self.event_agreement_version_invalid,
+                'slug': self.event_agreement_slug_new_version,
+                'is_active': True,
+                'html_content': self.event_agreement_html_content,
+                'access_template': self.event_agreement_access_template
+            })
+
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, 'console/event_agreement_new_version.html')
+        self.assertContains(response,
```
alistairewj: I don't think asserting the response contains specific text is necessary. This just means that in the future if we want to update the text we have to also update the test. Asserting the server error is caught is sufficient.

physionet-django/events/forms.py
```diff
@@ -60,3 +60,35 @@ class EventApplicationResponseForm(forms.ModelForm):
...
+            'is_active': '* Only active agreements are usable in future events.',
+            'html_content': '* The agreement text displayed to the participant.',
+            'access_template': '* Instructions on accessing the dataset.'
+        }
+
+    def clean(self):
+        cleaned_data = super(EventAgreementForm, self).clean()
+        if EventAgreement.objects.filter(name=cleaned_data['name'], version=cleaned_data['version']).exists():
+            raise forms.ValidationError(
+                {"name": ["Agreement with this name and version already exists for this user."],
```
alistairewj: This is great, although because the model is no longer unique for the user, we should update the text to remove the mention of for this user:

```suggestion
                {"name": ["An agreement with this name and version already exists."],
```

physionet-django/events/forms.py
```diff
@@ -60,3 +60,35 @@ class EventApplicationResponseForm(forms.ModelForm):
...
+                 "version": ["Agreement with this name and version already exists for this user."]
```
alistairewj: Similar story here```suggestion
                 "version": ["An agreement with this name and version already exists."]
```

**Review by alistairewj — CHANGES_REQUESTED** [review]
Looking great - just a few minor text changes and I think it's good to go.

**Review by alistairewj — CHANGES_REQUESTED** [review]
One or two additional comments below!

File: physionet-django/console/templates/console/event_agreement_new_version.html [hunk 20]
```diff
@@ -0,0 +1,23 @@
...
+{% block title %}{{ SITE_NAME }} - Event Agreements{% endblock %}
+
+{% block content %}
+  <div class="card mb-3">
+    <div class="card-header">
+        {{ event_agreement.name }} - new version
+    </div>
+    <div class="card-body">
+        <form method="post" action="{% url 'event_agreement_new_version' event_agreement.pk %}">
+            {% include "inline_form_snippet.html" with form=event_agreement_form %}
```
alistairewj: Right forgot this - you need to change this to `content_inline_form_snippet.html` for the help text to display on this page

physionet-django/console/templates/console/event_agreement_detail.html
```diff
@@ -0,0 +1,23 @@
...
+{% block title %}{{ SITE_NAME }} - Event Agreements{% endblock %}
+
+{% block content %}
+  <div class="card mb-3">
+    <div class="card-header">
+        {{ event_agreement.name }} - edit
+    </div>
+    <div class="card-body">
+        <form method="post" action="{% url 'event_agreement_detail' event_agreement.pk %}">
+            {% include "inline_form_snippet.html" with form=event_agreement_form %}
```
alistairewj: Same as new version - you need to change this to `content_inline_form_snippet.html` for the help text to display on this page

superryeti: Oh yes, i missed this, and just did this on the listing page.
Fixing this right now

physionet-django/console/templates/console/event_agreement_new_version.html
superryeti: Oh yes, i missed this, and just did this on the listing page. 
Fixing this right now

physionet-django/console/test_views.py
superryeti: Updated this, Quick note in some cases because of how views are implemented the status code will be 200 and in most cases it will be 300.

So in terms of readability, it might look like we created a valid Event Agreement, and tried to create and invalid eventagreement or new invalid event agreement. but didnot do further verification except the status code and template(but the combination of status code and template used seems sufficient here)

physionet-django/console/templates/console/event_agreement_detail.html
```diff
@@ -0,0 +1,18 @@
...
+SITE_NAME }} - Event Agreements{% endblock %} {% block content %}
+<div class="card mb-3">
+  <div class="card-header">{{ event_agreement.name }} - edit</div>
+  <div class="card-body">
+    <form
+      method="post"
+      action="{% url 'event_agreement_detail' event_agreement.pk %}"
+    >
+      {% include "project/content_inline_form_snippet.html" with
+      form=event_agreement_form %}
```
alistairewj: The template tag here `{% %}` here is spread across two lines - which doesn't work: https://code.djangoproject.com/ticket/8652

physionet-django/console/templates/console/event_agreement_new_version.html
```diff
@@ -0,0 +1,18 @@
...
+SITE_NAME }} - Event Agreements{% endblock %} {% block content %}
+<div class="card mb-3">
+  <div class="card-header">{{ event_agreement.name }} - new version</div>
+  <div class="card-body">
+    <form
+      method="post"
+      action="{% url 'event_agreement_new_version' event_agreement.pk %}"
+    >
+      {% include "project/content_inline_form_snippet.html" with
+      form=event_agreement_form %}
```
alistairewj: Same issue as in the detail view: the template tag here `{% %}` here is spread across two lines - which doesn't work: https://code.djangoproject.com/ticket/8652

physionet-django/console/templates/console/event_agreement_new_version.html
```diff
@@ -0,0 +1,18 @@
+{% extends "console/base_console.html" %} {% load static %} {% block title %}{{
+SITE_NAME }} - Event Agreements{% endblock %} {% block content %}
+<div class="card mb-3">
+  <div class="card-header">{{ event_agreement.name }} - new version</div>
+  <div class="card-body">
+    <form
+      method="post"
+      action="{% url 'event_agreement_new_version' event_agreement.pk %}"
+    >
```
alistairewj: Not sure if we have a style guide but the Google HTML/CSS one recommends putting the last `>` on the end of the last line with attributes, e.g.

```
<form ...
more text here ...>
```

https://google.github.io/styleguide/htmlcssguide.html#HTML_Line-Wrapping

alistairewj (changes requested):
Two minor bugs introduced in the form due to multi-line template tags (invalid in Django).

physionet-django/console/templates/console/event_agreement_detail.html
superryeti: Thank you.  i missed that. i was on my personal laptop and the linter messedup the tag.

physionet-django/console/templates/console/event_agreement_new_version.html
superryeti: I agree. I always have the tag closed with the attribute. its my bad.

alistairewj (approved):
Looks great!

superryeti:
Working on merge conflicts

physionet-django/events/models.py
```diff
@@ -153,3 +155,33 @@ def withdraw(self, comment_to_applicant=None):
             comment_to_applicant (str): The comment to add to the participant about the status.
         """
         self._apply_decision(self.EventApplicationStatus.WITHDRAWN, comment_to_applicant)
+
+
+class EventAgreement(models.Model):
```
tompollard, line 160: Now that we have a couple of different types of agreement, I'm wondering whether we should have a generic "Agreement" model (and then either inherit from that model, or have a new agreement_type attribute. Any thoughts on this? Also q for @bemoody

superryeti:
Rebased and fixed merge conflicts. I also ran the tests and did a manual QA and both Event Agreement and Add dataset feature seems to be working.

I will go through the individual commits to see if everything is okay/to see if i can find issues

superryeti:
Looks good to me.

physionet-django/events/models.py
superryeti: @tompollard Thats a good point. I like the 2nd idea `new agreeement_type attribute` with text enums.

We could add the attribute to the DUA model, make the default='Project' or something like that, and slightly change `CreateProjectForm` and related views.
and can use it directly in Events.

we will probably be able to do it after the Datathon(but weirdly it will be a bad decision to merge the current PR if we decide to go with the `new_agreement_type` attribute as we would be adding a table to database, just to remove it next week)

physionet-django/console/templates/console/console_navbar.html
```diff
@@ -136,7 +136,7 @@
       {% endif %}
 
       <!-- legal -->
-      {% if perms.project.add_dua or perms.project.add_codeofconduct or perms.project.add_license %}
```
tompollard, line 139: This doesn't look right to me. Up until now, only people with admin permissions have been granted access to the console. 

__I think__ `perms.events.add_event` is a permission granted to event hosts? If so, we are granting permission to the admin console for all hosts. Is this intentional?

tompollard: e.g., should test user rgmark be able to access http://127.0.0.1:8000/console/? He can.

physionet-django/events/models.py
tompollard: Both Benjamin and Alistair pointed out the benefit of not introducing dependencies between apps, so let's ignore this idea for now!

bemoody:
I haven't looked at this code in any detail.  I think that EventAgreementSignature should probably have a foreign key to EventAgreement (either in addition to, or instead of, a foreign key to Event.)

DUASignature doesn't do that and I think that is probably a mistake.

bemoody:
If these EventAgreements are supposed to be reusable across many events... I think that we probably want to have different user permissions for "creating an EventAgreement" versus "creating an event".

physionet-django/console/templates/console/console_navbar.html
superryeti: Thats is true, Only admin should be able to create the agreement. At first i thought we wanted to let event hosts, create their own Event Agreement and that Event Agreements for one host will be different than the other.

But Alistair corrected me on this, 

Like Benjamin suggested on another Comment, it is best to use the permission `perms.events.add_event_agreement`. I will fix and push a commit.

superryeti:
Thanks @bemoody , Thats true. Working on the suggested changes.

superryeti:
Thanks @tompollard , @bemoody .

I fixed the permission  and updated the `EventAgreement` model and squashed the commits

tompollard:

PR #1877, opened 2023-02-15
Only show denied downloads screen if user has been granted access.

alistairewj:
The allow file downloads flag should gate file access only. In the current iteration, it also gates project access, and prevents the user from seeing messages related to needing credentialing.

e.g. for a project where the user does not have access, they will currently see:

image

... whereas it would be better to provide instructions on how to access the data.

Also took the opportunity to clean up the comments.

### DIFF
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -347,7 +347,6 @@
       <!-- /.sidebar -->
     </div>
     <h2 id="files">Files</h2>
-    {% if project.allow_file_downloads %}
     {% if project.deprecated_files %}
       <div class="alert alert-danger col-md-8" role="alert">
         {% if project.is_latest_version %}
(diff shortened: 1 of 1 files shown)


kshalot:
I think it will be worth it to split the published project template into partials based on the scenarios that are documented.
For example:
* partial list the things that the user has to undergo some checks before they can access the project
* partial with the files listed
* partial for the other ways to access the project
etc. etc.

From what I see the file is poorly formatted so certain if/else/endif blocks are misaligned and it's very hard to visually tell whether some line falls under a specific block. Not related to this PR, just a thought.

physionet-django/project/templates/project/published_project.html
```diff
@@ -377,6 +375,8 @@ <h5>Access the files</h5>
           </p>
         {% endif %}
 
+        {% if project.allow_file_downloads %}
```
kshalot, line 378: Will the ever be `False`? From what I see this falls under the `has_access` condition:
```
{% if has_access %}
  {% if project.allow_file_downloads %}
  {% endif %}
{% endif %}
```
But if `project.allow_file_downloads` is `False`, then `has_access` will be `False` as well: https://github.com/MIT-LCP/physionet-build/blob/291a60b57cf1847f4b8da96573ed361784c2b882/physionet-django/project/modelcomponents/publishedproject.py#L218-L219

We'll need to add a distinction between having access to the project and being able to download the files.

alistairewj: Ah, I did not expect `has_access` to check that.

> We'll need to add a distinction between having access to the project and being able to download the files.

Yeah, and building on our discussion from earlier this week, I vote we limit `has_access` to "has *project content* access" functionality. We can then separately manage the distribution mechanisms in the template (for now, until it's too unwieldy).

alistairewj:
Taken over by #1884

PR #1879, opened 2023-02-15
Update Help text for research_summary on Credentialing Application

superryeti:
Currently, during credentialing, we have a minimum limit of 20 words(but this is not obvious to the user as this information is not displayed to user). Users only see the message to provide the details about their research. So when trying to submit credentialing application with less than 20 words, they will keep getting served error by validators(but the error doesnot make it clear about the issue)

As pointed out by January, it will be a good idea if we could show the minimum word limit explicitly.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -162,3 +162,6 @@ ENABLE_CREDENTIALING_AUTO_REJECTION = false
 # total applications to reject every time the management command is run
 DEFAULT_NUMBER_OF_APPLICATIONS_TO_REJECT = 5
 #### Auto rejection of credentialing applications End ####
+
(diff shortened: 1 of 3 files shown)


superryeti:
Also,  should we make the word limit configurable through the environment variable? @alistairewj @tompollard 
January suggested that we might want to reduce this limit for HDN.

tompollard:
@amitupreti good catch. I think making the length configurable is a good idea!

alistairewj:
yep, another env variable? :)

superryeti:
Okay. I will add that. Thank you

tompollard:
Thanks Amit!

PR #1883, opened 2023-02-16
refactor boolean `is_instructor` to `can_change_event` for clarity

superryeti:
Context: 
@tompollard  pointed out on the PR https://github.com/MIT-LCP/physionet-build/pull/1874#issuecomment-1433656882 that the two variable  have slightly confusing terminology. 

`is_instructor` is someone who has permission to create events 
`is_host` is someone who is host of an existing event

On this PR, i changed the  boolean `is_instructor` to `can_change_event` so that its easier to understand what it is even without any context.


tompollard:

PR #1884, opened 2023-02-17
Feature: Change the logic of showing access denied information

rafalkosla101:
Changed the logic of the prompts to firstly show:
image
and then:
image

instead of the opposite way as it is right now. This change will help user get the permissions to actually view files and check the requirements that are needed to do so.

Things to discuss:
 - The names
 - Whether the use of the original has_access is valid, i.e. whether we should actually check for allow_file_downloads in all the places I have changed (the original has_access method was used there). For example, it's used in the sign_dua view - why would that logic depend on the allow_file_downloads flag? Is it a bug or intended behaviour?

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -1168,7 +1168,7 @@ def users_aws_access_list_json(request):
             dataset['name'] = project_name
             dataset['accounts'] = []
             for user in users_with_awsid:
-                if project.has_access(user):
+                if project.can_view_files(user):
                     dataset['accounts'].append(user.cloud_information.aws_id)
             datasets['datasets'].append(dataset)
 
--- a/physionet-django/project/modelcomponents/publishedproject.py
+++ b/physionet-django/project/modelcomponents/publishedproject.py
@@ -207,6 +207,9 @@ class PublishedProject(Metadata, SubmissionInfo):
             args=(self.slug, self.version, os.path.join(subdir, file)))
 
     def can_view_files(self, user):
+        """
+        Whether the user has access to this project and files. Extends 'is_authorized'
+        """
         return self.allow_file_downloads and self.is_authorized(user)
 
     def is_authorized(self, user):
--- a/physionet-django/project/modelcomponents/publishedproject.py
+++ b/physionet-django/project/modelcomponents/publishedproject.py
@@ -206,16 +206,16 @@ class PublishedProject(Metadata, SubmissionInfo):
         return reverse('display_published_project_file',
             args=(self.slug, self.version, os.path.join(subdir, file)))
 
-    def has_access(self, user):
+    def can_view_files(self, user):
+        return self.allow_file_downloads and self.is_authorized(user)
+
+    def is_authorized(self, user):
         """
         Whether the user has access to this project
         The logic should mirror PublishedProjectManager#accessible_by,
         but for a specific project.
         """
         if self.deprecated_files:
-            return False
-
-        if not self.allow_file_downloads:
             return False
 
         if self.access_policy == AccessPolicy.OPEN:
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -364,7 +364,7 @@
         The files in this project version are under embargo until: {{ project.embargo_end_date|date }}
       </div>
     {% else %}
-      {% if has_access %}
+      {% if can_view_files %}
         {#  refactored code goes here            #}
         <p>Total uncompressed size: {{ main_size }}.</p>
         {# ZIP START #}
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -347,7 +347,6 @@
       <!-- /.sidebar -->
     </div>
     <h2 id="files">Files</h2>
-    {% if project.allow_file_downloads %}
     {% if project.deprecated_files %}
       <div class="alert alert-danger col-md-8" role="alert">
         {% if project.is_latest_version %}
@@ -364,7 +363,8 @@
         The files in this project version are under embargo until: {{ project.embargo_end_date|date }}
       </div>
     {% else %}
-      {% if can_view_files %}
+      {% if is_authorized %}
+        {% if project.allow_file_downloads %}
         {#  refactored code goes here            #}
         <p>Total uncompressed size: {{ main_size }}.</p>
         {# ZIP START #}
@@ -438,12 +438,12 @@
           {% include "project/files_panel.html" %}
         </div>
       {% else %}
-        {% include "project/published_project_unauthorized.html" %}
+        {% include "project/published_project_denied_downloads.html" %}
       {% endif %}
 
-    {% endif %}
-    {% else %}
-    {% include "project/published_project_denied_downloads.html" %}
+      {% else %}
+        {% include "project/published_project_unauthorized.html" %}
+      {% endif %}
     {% endif %}
     <br>
     {% if accepted_access_requests %}
--- a/physionet-django/project/templatetags/project_templatetags.py
+++ b/physionet-django/project/templatetags/project_templatetags.py
@@ -202,7 +202,7 @@ def call_method(obj, method_name, *args):
         args: arguments to be passed to the method
 
     Example:
-        {% call_method published_project 'can_view_files' request.user as user_has_access %}
+        {% call_method published_project 'can_view_files' request.user as user_can_view_files %}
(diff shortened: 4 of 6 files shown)


physionet-django/console/views.py
```diff
@@ -1136,7 +1136,7 @@ def users_aws_access_list_json(request):
             dataset['name'] = project_name
             dataset['accounts'] = []
             for user in users_with_awsid:
-                if project.has_access(user):
+                if project.has_access_to_project_and_files(user):
```
alistairewj: this should be `has_access_to_project` // `is_authorized`

physionet-django/project/templates/project/published_project.html
```diff
@@ -347,7 +347,7 @@ <h5 class="card-header">Versions</h5>
       <!-- /.sidebar -->
     </div>
     <h2 id="files">Files</h2>
-    {% if project.allow_file_downloads %}
+    {% if has_access_to_project %}
     {% if project.deprecated_files %}
```
alistairewj, line 350: deprecation warning should be moved above `has_access_to_project` as we would like to inform the user why the files are not accessible. the logic of restricting access can be solely handled by `has_access_to_project`

alistairewj: same story for the embargo notification just below `project.deprecated_files`

physionet-django/project/views.py
```diff
@@ -1919,7 +1921,7 @@ def sign_dua(request, project_slug, version):
         project.deprecated_files
         or project.embargo_active()
         or project.access_policy not in {AccessPolicy.RESTRICTED, AccessPolicy.CREDENTIALED}
-        or project.has_access(user)
+        or project.has_access_to_project_and_files(user)
```
alistairewj: I agree this should be `has_access_to_project` // `is_authorized`

physionet-django/project/views.py
```diff
@@ -2204,7 +2206,7 @@ def published_project_request_access(request, project_slug, version, access_type
                                             platform=access_type)
 
     # Check if the person has access to the project.
-    if not project.has_access(request.user):
+    if not project.has_access_to_project_and_files(request.user):
```
alistairewj: this should be `has_access_to_project` // `is_authorized`

alistairewj (changes requested):
Adding some thoughts from our separate discussion - https://github.com/T-CAIREM/physionet-build/pull/11

I think we can adopt two new functions (as suggested by @kshalot):

1. Can the user access the project content? -> `publishedproject.is_authorized`
    * this is currently called `has_access_to_project` in the PR
3. Can the user view the files natively on the system? -> `publishedproject.can_view_files`
    * although, the clarity of `has_access_to_project_and_files` is growing on me... might not be 100% obvious from the name `can_view_files` requires `is_authorized`?

Then, we can take the approach to check `is_authorized` for displaying / allowing access to content. I'll mark where I think it should be changed.

A few changes to consider for future PRs:

* `deprecated_files` would be better called `deprecated_content`
    * sometimes we refer to the metadata as the "content", so maybe this is confusing and we need a word other than content
* `self.allow_file_downloads` could be wrapped into the integrations section so that downloading files directly from the server is just one of the options for data access

With this view, I agree that is_authorized should not check allow_file_downloads, seems inconsistent.

physionet-django/project/views.py
```diff
@@ -1919,7 +1921,7 @@ def sign_dua(request, project_slug, version):
         project.deprecated_files
         or project.embargo_active()
         or project.access_policy not in {AccessPolicy.RESTRICTED, AccessPolicy.CREDENTIALED}
-        or project.has_access(user)
+        or project.can_view_files(user)
```
alistairewj: suggestion: change to `is_authorized`

physionet-django/project/views.py
```diff
@@ -2204,7 +2206,7 @@ def published_project_request_access(request, project_slug, version, access_type
                                             platform=access_type)
 
     # Check if the person has access to the project.
-    if not project.has_access(request.user):
+    if not project.can_view_files(request.user):
```
alistairewj: @kshalot to verify: suggestion: change to `is_authorized`

kshalot: This is the access to e.g. BigQuery or AWS. Can probably be left as is.

tompollard:
@rafalkosla101 please could you respond to the inline comments to let us know how they are addressed? https://github.com/T-CAIREM/physionet-build/pull/12 sounds like you are expecting this PR to be merged in its current state, but it is unclear to me whether the feedback was dealt with or not.

rafalkosla101:
Sorry for such a late response @tompollard. Answering your question - the feedback was dealt with in the other PR(the feedback itself is about changing some logic inside physionet that was not closely related to the changes I have made but could be handled here since we are already changing things) but there is no need to merge this PR as it is now. In a spare minute I will merge the 2nd PR to this one and we will handle those as one. I just did not want to block merging this one before the datathon by making big logic changes. Is is/was good to have but nothing necessary PR that I can spend a little bit more time on now.

physionet-django/project/templates/project/published_project.html
rafalkosla101: done

alistairewj (changes requested):
@rafalkosla101 I rebased to try and get my head around this (hope that's OK) - there's a bunch of formatting changes still in the published_project.html template that I'd like to get rid of because they hide the real changes.

Looking at the real changes, I don't like that we now call `is_authorized` twice; once through `is_authorized` and then again through `can_view_files`. One option is to do `if can_view_files`, `elif is_authorized`, `else deny`... but that only somewhat removes the issue. Alternatively we could revert `can_view_files` to `project.allow_file_downloads`?

rafalkosla101:
I would probably stick to the alternative that you have proposed - reverting those changes will not be the best option but I assume that we could be refactoring those anyway since that topic was already covered by one of the issues(https://github.com/MIT-LCP/physionet-build/issues/1927) - We would need a general refactor of that part of the code anyway when covering above mentioned issue.

alistairewj (approved):

alistairewj:
@rafalkosla101 yeah makes sense, cheers

@tompollard just tidied up this PR so feel free to merge after you take a quick look. briefly we:

- rename `has_access` to `is_authorized`
- create a `can_view_files` method that checks both authorization & the allow file downloads project option
- updated the calls to `has_access` to either `is_authorized` or `can_view_files`, as appropriate. we do (correctly) change behaviour in a few places with these updates
- fix the originally described bug where the research environment denial ([published_project_denied_downloads.html](https://github.com/MIT-LCP/physionet-build/blob/83f292a1a4d382a4e109b2fa37d60dd4bbb69be4/physionet-django/project/templates/project/published_project_denied_downloads.html)) is shown to users before they have been granted dataset access

physionet-django/project/modelcomponents/publishedproject.py
```diff
@@ -206,7 +206,10 @@ def file_display_url(self, subdir, file):
         return reverse('display_published_project_file',
             args=(self.slug, self.version, os.path.join(subdir, file)))
 
-    def has_access(self, user):
+    def can_view_files(self, user):
```
tompollard, line 209: Could we add a docstring here that briefly explains the purpose of this method?

physionet-django/project/templatetags/project_templatetags.py
```diff
@@ -202,7 +202,7 @@ def call_method(obj, method_name, *args):
         args: arguments to be passed to the method
 
     Example:
-        {% call_method published_project 'has_access' request.user as user_has_access %}
+        {% call_method published_project 'can_view_files' request.user as user_has_access %}
```
tompollard: Not really an issue, but `user_has_access` is slightly off now that we have changed from `has_access` to `can_view_files`

physionet-django/search/templates/search/content_list.html
```diff
@@ -28,7 +28,7 @@
     <p class="pub-details">Published: {{ published_project.publish_datetime|date }}.
       Version: {{ published_project.version }}</p>
     {% if is_lightwave_supported %}
-      {% call_method published_project 'has_access' request.user as user_has_access %}
+      {% call_method published_project 'can_view_files' request.user as user_has_access %}
```
tompollard: Should `user_has_access` also be updated in this template? (e.g. to `user_can_view_files`)?

tompollard (approved):
Thanks for doing this. I've added a couple of small comments, but looks good to me!

tompollard:

PR #1886, opened 2023-02-18
use ckeditor on event description

superryeti:
Context: On Events app, currently description is a simple text field. This means we cannot add line breaks, or make the description bold, use bulletpoints etc. So its pretty annoying as  description appears jumbled up

The PR is simple and uses the `SafeHtmlField` that already exists in physionet-build project.

One thing to note is that before the PR, the edit event was done with a modal popup, but on the PR i changed that to a separate page.  The reason that i decided to make a separate page for event edit(compared to previous model) is because i was not able to get ckeditor to work with both modals(add event and edit event modal)  properly. I found out that when there was two modal(add event and edit event modal) ckeditor was working for the first modal only.

Before change

image

After Change

image


tompollard:
@amitupreti please could you rebase on dev and regenerate the migration?

superryeti:
> @amitupreti please could you rebase on dev and regenerate the migration?

Thanks @tompollard . working on this

physionet-django/events/tests_views.py
```diff
@@ -125,7 +125,9 @@ def test_edit_event_invalid(self):
                 'allowed_domains': ''
             })
 
-        self.assertEqual(response.status_code, 302)
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, 'events/event_edit.html')
+        self.assertContains(response, 'Event with this title already exists')
```
superryeti, line 130: @tompollard, Do you prefer to remove the text assertion(`assertContains`) from here? I remember Alistair mention on another PR that this might add overhead later as if someone changes the validation message on forms, they would have to fix the test as well.

tompollard: I'm happy either way! I can see there is some benefit in having this test, but it is also fairly strict. It's easy for someone to fix the test if they want to change the text, so it doesn't bother me (but I can understand why it might bother other people!).

superryeti: yup. sounds good to me. It also looks good to me. without the text assertion, in this particular test, its kind of hard to make sure if the `title` was the issue

i will let it be and refactor if @alistairewj prefers this refactored.

superryeti:
Thanks @tompollard . i rebased and recreated the migration file

tompollard:
@amitupreti looks like there are conflicts. Please could you resolve them?

superryeti:
Thank you. Working on it

superryeti:
I found an issue where the Event description is not respecting line breaks for cases where i typed the description manually, but in cases where i copy and paste description from another website, it is working as expected. 

I am checking the issue and making this PR draft.

superryeti:
Thanks @tompollard . I rebased, and fixed merge conflicts. 
There was a small bug i had introduced due to using wrong bootstrap class, details here https://github.com/MIT-LCP/physionet-build/pull/1886/commits/30dbf2d5bf56b09c3a70b14cf99981702292a8df

I will be more careful with html templates, bootstrap and spend time learning them to avoid similar issues in future.

bemoody:
When changing a text field to HTML, need to sanitize existing field values.

The easiest way is to call `full_clean`.  It should be possible to roll this into a single migration like so:

```
# Generated by Django 3.2.16 on 2023-02-24 14:35

from django.db import migrations
import project.modelcomponents.fields


def migrate_forward(apps, schema_editor):
    Event = apps.get_model('events', 'Event')
    for event in Event.objects.all():
        event.full_clean()
        event.save()


def migrate_reverse(apps, schema_editor):
    pass


    MIGRATE_AFTER_INSTALL = True

    dependencies = [
        ('events', '0006_eventagreement_eventagreementsignature'),
    ]

    operations = [
        migrations.AlterField(
            model_name='event',
            name='description',
            field=project.modelcomponents.fields.SafeHTMLField(blank=True, max_length=10000, null=True),
        ),
        migrations.RunPython(migrate_forward, reverse_code=migrate_reverse),
    ]
```

superryeti:
> full_clean

@bemoody good catch. I didn't think of the existing fields(they hadn't been sanitized before)

tompollard:
Looks good to me, except the CKEditor form in the "Create event" modal has a minor styling issue (items in the menu jump around). e.g. compare:

Screenshot 2023-03-02 at 4 03 39 PM

with: 

Screenshot 2023-03-02 at 4 03 54 PM

Perhaps the create event form should follow the same approach as the edit event form (i.e. move out of the modal?).

superryeti:
Sounds good. let me use the same approach for create event as well.

Thanks @tompollard

tompollard:
Looks good to me, thanks Amit. Please could you regenerate the migrations and deal with any conflicts?

superryeti:
Thank you, Doing this asap

superryeti:
I think i found a bug. Please hold off merge for few minutes

tompollard:
> I think i found a bug. Please hold off merge for few minutes

What was the bug? Is this ready now or are you still looking into it?

superryeti:
Thanks @tompollard . There was a repeated codeblock which was causing the `success` message to appear twice after editing the event.

I fixed it and squashed the comment

bemoody:
```
--- a/physionet-django/events/urls.py
+++ b/physionet-django/events/urls.py
@@ -6,6 +6,7 @@
 urlpatterns = [
     path('', views.event_home, name='event_home'),
     path('<slug:event_slug>/', views.event_detail, name='event_detail'),
+    path('create', views.create_event, name='create_event'),
```

There should be a slash after "create".

Does 'create' need to be listed first so it takes precedence?

superryeti:
> ```
> --- a/physionet-django/events/urls.py
> +++ b/physionet-django/events/urls.py
> @@ -6,6 +6,7 @@
>  urlpatterns = [
>      path('', views.event_home, name='event_home'),
>      path('<slug:event_slug>/', views.event_detail, name='event_detail'),
> +    path('create', views.create_event, name='create_event'),
> ```
> 
> There should be a slash after "create".
> 
> Does 'create' need to be listed first so it takes precedence?

Ah yes, about the /, 
About precedence, did you mean for case if someone tries to create an event with slug `create` or something like that in general? @bemoody

bemoody:
event_slug should be random (right?)  It should always be a random
12-character string, so "create" doesn't conflict.

But django doesn't know that; the path "create/" will match the
pattern "<slug:event_slug>/".

superryeti:
> event_slug should be random (right?)

---> Yes

> But django doesn't know that; the path "create/" will match the
> pattern "<slug:event_slug>/".

---> Yup,  I meant to ask that. I suppose it might not create bug in this specific case, but noted about the precedence. That makes a lot of sense, i had not thought of that before.

@bemoody  Thank you

I moved up the `create` and fixed the missing `/`

tompollard:

PR #1898, opened 2023-02-24
[WIP] code cleanup for events

superryeti:
Context:
Some features of event app were rushed because of Datathon deadline on 2023/02/23. This lead to a decreased code quality, unnecessary comments, vague docstrings etc. So i am starting this PR to do events code cleanup

### DIFF
--- a/physionet-django/events/models.py
+++ b/physionet-django/events/models.py
@@ -224,11 +224,8 @@ class EventDataset(models.Model):
 
     def has_access(self, user):
         """
-        checks if the user has access to the dataset for the Event
-        This method is independent of the PublishedProject.has_access method. it will only check if the user should
-        have access to the dataset for the event.
-        It is expected that the PublishedProject.has_access method will use this method to check if the user has access
-        in case the dataset access is made through an event.
+        Checks if the user has access to this EventDataset.
+        This does not check if the associated dataset(PublishedProject) is accessible
         """
         if not self.is_accessible():
             return False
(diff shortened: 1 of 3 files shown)


physionet-django/events/models.py
```diff
@@ -224,11 +224,8 @@ def is_accessible(self):
 
     def has_access(self, user):
         """
-        checks if the user has access to the dataset for the Event
-        This method is independent of the PublishedProject.has_access method. it will only check if the user should
-        have access to the dataset for the event.
-        It is expected that the PublishedProject.has_access method will use this method to check if the user has access
-        in case the dataset access is made through an event.
+        Checks if the user has access to this EventDataset.
```
superryeti, line 227: @tompollard Would you have any feedback for this doc string?

it is supposed to check if the attached dataset in the Event can be accessed by an user. 

Also for participant to access the dataset, the `PublishedProject` also needs to be accessible,(this method `EventDataset.has_access` only does event related checks but it doesn't guarantee that the user will be able to access the project as the `PublishedProject` has its own set of checks)

tompollard: I think we should rethink and refactor all of our code around managing access to files. It is a little confusing to have two independent "has_access" methods, but this is not going to be fixed by updating the doctring.

superryeti: I agree. It is confusing at this point. Let's regroup on Monday and discuss for this. 
Thank you

PR #1899, opened 2023-02-24
limit update event to the event host(Instructor) who created the event

superryeti:
Context:
On Events APP: Although, we don't allow other Instructor to edit someone else's event from the User Interface, There is a chance that someone can simulate a post request and edit someone else's event(As we were not verifying that the requester is the event host(Instructor) who created the event).

The change will do a check to make sure the current user making a request to edit an event is the right instructor.


tompollard:
thanks amit, looks good to me.

PR #1900, opened 2023-02-24
limit approval of participants to event host who created event

superryeti:
Context:
Similar change to https://github.com/MIT-LCP/physionet-build/pull/1899, 

For the view that allows event hosts to approve/disapprove participants, we were only checking if someone has the permission to create an event and allowing any user that has this permission to accept the participation request(on the `events.views.py`)(we were not verifying that the current user is the actual event host(the person who created the event).

Although its not a super security risk  as someone will need to guess the `EventApplication` id to submit a post request to exploit this, but it seems like a good idea to implement the validation

Also, this PR will later let us allow cohosts to handle participation request, So this PR sorts of acts as foundation for that.

### DIFF
--- a/physionet-django/events/views.py
+++ b/physionet-django/events/views.py
@@ -98,43 +98,42 @@ def event_home(request):
 
     form_error = False
 
-    if can_change_event:
-        # handle notifications to join an event
-        if request.method == 'POST' and 'participation_response' in request.POST.keys():
+    # handle notifications to join an event
+    if request.method == 'POST' and 'participation_response' in request.POST.keys():
 
-            formset = EventApplicationResponseFormSet(request.POST)
-            # only process the form that was submitted
-            for form in formset:
-                if form.instance.id == int(request.POST['participation_response']):
-                    if form.is_valid():
-                        event_application = form.save(commit=False)
-                        if event_application.status == EventApplication.EventApplicationStatus.APPROVED:
-                            event_application.accept(
-                                comment_to_applicant=form.cleaned_data.get('comment_to_applicant')
-                            )
-                            notification.notify_participant_event_decision(
-                                request=request,
-                                user=event_application.user,
-                                event=event_application.event,
-                                decision=EventApplication.EventApplicationStatus.APPROVED.label,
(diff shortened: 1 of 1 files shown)


superryeti:
I see the nesting issue(there 6 layers of nesting, that shouldn't be good). let me refactor the code.

superryeti:
@tompollard  i removed the nesting on the view here.

on second thought, do you think i should open a separate PR for removing nesting, get that PR merged and then rebase this PR?(i kind of like the 2nd idea better)

physionet-django/events/views.py
```diff
@@ -96,42 +96,50 @@ def event_home(request):
...
+            if not form.is_valid():
+                messages.error(request, form.errors)
+                return redirect(event_home)
+
+            event_application = form.save(commit=False)
+            event = event_application.event
+            if event.host != user:
+                messages.error(request, "You don't have permission to accept/reject this application")
+                return redirect(event_home)
+            elif event_application.status == EventApplication.EventApplicationStatus.APPROVED:
```
alistairewj, line 119:
```suggestion

            if event_application.status == EventApplication.EventApplicationStatus.APPROVED:
```

alistairewj (dismissed):

physionet-django/events/views.py
superryeti: Thank you @alistairewj . This is definitely more readable

tompollard:
@amitupreti please could you rebase and confirm this is ready for review?

superryeti:
@tompollard Thank you. This looks ready for review

physionet-django/events/views.py
```diff
@@ -98,43 +98,50 @@ def event_home(request):
...
+                event_application.accept(
+                    comment_to_applicant=form.cleaned_data.get('comment_to_applicant')
+                )
+                notification.notify_participant_event_decision(
```
tompollard, line 123: I feel like there is probably a way to structure this to avoid duplicated `notify_participant_event_decision` code but this is fine for now

superryeti: I agree, let me make a note and open a separate PR for that.

PR #1901, opened 2023-02-28
Remove "toggle_cohost_status"

bemoody:
This function is broken.  It's a security problem waiting to happen.  The only reason it's not a problem right now is that right now (as far as I can tell) the "cohost" flag is meaningless.

This functionality, if it's desired, needs to be completely re-implemented.  There's no way to make it secure without breaking compatibility, so there's no reason to keep the function at all.

### DIFF
--- a/physionet-django/events/templates/events/event_entries.html
+++ b/physionet-django/events/templates/events/event_entries.html
@@ -18,9 +18,9 @@
             <td>{{ participant.user.is_credentialed }}</td>
             <td>
                 {% if participant.is_cohost %}
-                    <input type="checkbox" name="toggle-cohost-status"  event-slug="{{ event.slug }}" value="{{ participant.id }}"  checked>
+                    <input type="checkbox" name="toggle-cohost-status"  event-slug="{{ event.slug }}" value="{{ participant.id }}" disabled checked>
                 {% else %}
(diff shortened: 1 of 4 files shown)


superryeti:
Thanks for the feedback, @bemoody! I learned a lot from your comments/commits, particularly the importance of design  , making requests  idempotent ,following basic HTTP standard and following the pattern on the existing codebase.
I'll be sure to read up on the HTTP standard and propose a solution.

I had a couple of questions, if you don't mind. Regarding the security concern you mentioned, were you referring specifically to the risk of someone injecting javascript code and making a participant a cohost? Or were there other potential security risks I should be aware of?

Also, in your first commit comment, you mentioned that there were other bad design choices made. I would love to hear more feedback if you have the time.

Thank you again for your help and expertise!

bemoody:
GET (and HEAD) requests must never modify any server-side state.  A GET request is a request to *get* a resource, nothing more.  Any request that changes server-side state must be a POST.

There is no reason not to just use a form for this.  But if there *were* some reason to desire a JSON API here, you want the client to be explicit about what it wants ("set the state to 1", not "set the state to the opposite of what it was before".)  Never assume that the client and the server have all the same information or that they agree about anything.

superryeti:
@bemoody  Thank you. Really appreciate your feedback. 👏

tompollard:
Thanks @bemoody!

PR #1902, opened 2023-02-28
bug fix: edit event agreement

superryeti:
Context:
On https://github.com/MIT-LCP/physionet-build/pull/1876, we introduced a bug on Event Agreement Edit. 
The bug is if you try to edit an Event Agreement for fields other than `title` and `version`, it will throw a validation error saying the "Name" and "Version" already exists.

The error comes from `EventAgreementForm`'s `clean` method. Here the if statement was throwing validation error if the name, version already exists as a row in database. The logic is wrong as it doesn't account for that fact that when you update an Event Agreement, the event agreement's name and version might be the same as before.
    
So in case of update/edit, we need to check if the existing row in the database is from the event agreement being edited; if yes, we shouldn't throw an error.

Changes
1. Form -  i updated the form to check if the current request is update, and if yes, dont throw the validation error
2. Tests - i added a new test that tries to edit just the html_content, so that it can test the edit functanality.

### DIFF
--- a/physionet-django/console/test_views.py
+++ b/physionet-django/console/test_views.py
@@ -632,6 +632,7 @@ class TestEventAgreements(TestMixin):
         self.event_agreement_slug_new_version = "a1b2c3d4e5"
         self.updated_event_agreement_name = "updated test event agreement"
         self.event_agreement_html_content = "<p>My test Event Agreement test content</p>"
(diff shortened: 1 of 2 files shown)


tompollard:
Thanks Amit!

PR #1904, opened 2023-03-01
Implement event agreement on event app

superryeti:
**Context:** 
On PR https://github.com/MIT-LCP/physionet-build/pull/1876, we added added models `EventAgreement` and `EventAgreementSignatures`. The goal of those models is to ultimately let Event Hosts attach `EventAgreement` to `Events`, and for the participants to read the event agreements and accept the event agreement before they are able to access the data.
This PR implements the Event Agreement on the Event APP for Hosts and Participants. Now Host is able to link and Event Agreement to an Event. The participants will then be required to accept the Event Agreement before they can access the data

**Workflow of the feature**
1. Platform Admin creates `Event Agreements` from Admin console, Thanks to https://github.com/MIT-LCP/physionet-build/pull/1876. These Agreements will then be used by Event Hosts
2. Event Hosts can choose from the list of existing `Event Agreements` when creating or editing the event.
3. If an event has an agreement, the participants will be required to read and accept the agreement before they can access the data
   - First the participant request to join the event
   - The Event Host approves the request
   - The Participant then sees the section on Event detail page, that will ask them to sign the agreement.
   - The participant signs the event agreement
   - The participant gets access to the data

**Questions**

1. Do we let event host change the Event Agreement from Event once someone has signed the agreement? If they change event agreement once people have signed the agreement then, things will get complicated(Project avoids problem by using immutable published project)
  - If the answer is yes, then i suppose we should be doing at least the following
    - Notify all the participants who had accepted the agreement, and ask them to agree to the new agreement.
    - Remove their access after a buffer time period or immediately?
  - If the answer is No, maybe we implement a feature that will prevent editing of Event after a certain date or maybe we can have a registration opens and closes date. So once registration opens, Event Host can't edit the event
  - or the third option, we trust the Event Host to not change the Event Agreement once someone has accepted the event(probably not a good design choice)
 

2. Should I open a separate PR for tests? I see on project we have a test that says `TestWorkflow` and tests all the Project stuff from start to finish? If we want to add a test like that, it makes more sense to open a separate PR. 
Also, I did not see the project having a test for just signing DUA, so thought it might be a good idea to follow the same pattern
		
3. Should Event Agreement be optional(when creating/editing event)?
		
4. Do we let people unsign an agreement?


superryeti:
@alistairewj @tompollard 

Could you please check the question section of the PR, when you have time?

I still have few small changes to make, so i marked this as draft, i will work on them tomorrow.

superryeti:
Screenshots to help review easily

1. Before participants registers to the event
image

2. After participants registers to the event
image (1)

3. After participants is approved to the event
image (2)

4. Sign Event Agreement page
image (3)

5. After the user signs the Event Agreement

image (4)

superryeti:
@tompollard @alistairewj , Ready for review.

There are few things that might need to be implemented pending your answers to the question on the PR(please check PR description)

bemoody:
> Do we let event host change the Event Agreement from Event once someone has signed the agreement?

I'm not sure whether you mean to say modifying an *agreement*, or switching an *event* to point to a different agreement.

The agreement is a fixed object and nobody should ever change it.  If changes are needed then we should create a new agreement, and point the event to the new version.

It seems unlikely that we would ever want to do this after an event is public.  If so, I think we'd want to say that yes, anybody who agreed to the first version of the agreement will now be required to click through the second version.

> Should I open a separate PR for tests?

No.  Please write test cases for a new feature along with the new feature.

> I see on project we have a test that says TestWorkflow and tests all the Project stuff from start to finish? If we want to add a test like that, it makes more sense to open a separate PR.

A single "comprehensive" test case tends to be a lot of work to write and also tends not to exercise all of the functionality.  Instead, try to write small and focused test cases.  Test a single request at a time if you can.  Use fixtures to define the initial state, make a request, and verify that it has the expected effect.

> Should Event Agreement be optional(when creating/editing event)?

Yes.

> Do we let people unsign an agreement?

No.

superryeti:
@bemoody 

Thank you for the feedback

> I'm not sure whether you mean to say modifying an agreement, or switching an event to point to a different agreement.

---> Here i meant to ask mainly about  switching. Noted about the expectation that admin's will not edit the existing Agreement.

I suppose my question was more of 
In case of switching of agreement linked to an event, should we trust the Host to not make switch or should we enforce it so that they are not able to make it even if they want.

If i understood correctly, your answer was yes let's enforce it, right?

bemoody:
Event *hosts* are not necessarily empowered to make decisions about who can access what data under what conditions.

Adding temporary database access to an event is something that in general has to be approved by the *data contributors*, and the data use agreement is (or might be) part of that.

So... event agreement should be tied to event dataset, I guess?

superryeti:
> Event _hosts_ are not necessarily empowered to make decisions about who can access what data under what conditions.
> 
> Adding temporary database access to an event is something that in general has to be approved by the _data contributors_, and the data use agreement is (or might be) part of that.
> 
> So... event agreement should be tied to event dataset, I guess?

@bemoody  I think that makes sense.

Currently an Event can have multiple Datasets attached to it. so i am having difficulty wrapping my mind around how to tie the Event Agreement to dataset(s) or that the implementation would probably be the same like what we have now[which is Event has dataset(s), and there will be 1 agreement created for that, and that agreement will have  the data use agreement  which covers all the dataset(s)]

As no matter how many number of dataset we have in the event, we would probably write all the agreement stuff for those dataset under that event in 1 single event agreement.


Or was your suggestion, that maybe we shouldn't let event host change the agreement or assign it at all? Maybe Platform Admin or whoever has the permission should be do it.

superryeti:
I added the tests for signing the event agreement, and updated the model to make event agreement optional.

For the part about requiring participants to resign the event agreement, I will open another PR after this one.

superryeti:
Currently paused while we discuss/figureout how we want to implement access to datasets/project

PR #1906, opened 2023-03-01
send reminder before auto rejecting credentialing applications

### LINKED ISSUE #1843: Send a reminder before auto rejecting credentialing applications.
Related to https://github.com/MIT-LCP/physionet-build/issues/1833

on [1833](https://github.com/MIT-LCP/physionet-build/issues/1833) , we are introducing change that will automatically delete credentialing application with `Awaiting Reference Response` pending.

So shouldn't we add something that sends a reminder or somekind of notification to the references or the user(although we could also expect the users to followup with their reference as i assume credentialing is important part of getting access to data)

on the other hand, it might be bunch of extra work for little gain

superryeti:
fixes https://github.com/MIT-LCP/physionet-build/issues/1843

**Context**: On PR https://github.com/MIT-LCP/physionet-build/issues/1833 , we introduced a change that automatically rejects credentialing application with Awaiting Reference Response pending after certain timeframe.

**Changes**
1. Adds a field `reminder_datetime` which is supposed to track the date n reminder was sent to the user about the reference. 
2. Adds a management function that loads the list of credentialing applications which are waiting for response from reference since x(configured on .env) number of days, and then sends reminder to the user.
3. Adds a test that tests the added management function
4. Adds a cronjob that runs the management command every hour and notifies 5 new users at a time.

**Questions:** 
1. I am wondering if it's the best idea to add a new field `reminder_datetime`. The other alternative that i could think of was creating a new model for `Logging`(for which i think the way to go forward is open a new PR, get that merged and come back to this). Maybe there is another easier way i am not seeing, i will leave it upto reviewers for suggestions.
2. Weird Edge Case(s): What if the platform administrator decides to turn on autorejection+reminder feature after a long time (say 6 months)
   - Problem: In that case, what will happen is , the system will start rejecting people and sending reminder rightaway. Suppose if there are users whose reference haven't responded for about 30 days, then the system will do both. It will probably be weird experience to the user as they will receive two emails where one says they have x number of days before autorejection, and another one that says the application was autorejected.
   - Solution: I suppose we could change the workflow in this way
				a. refactor autoreject code to only reject applications where we have sent reminder
				b. check the number of days after the reminder and only autoreject once the configured number of days have passed.


superryeti:
@tompollard  could you please check the questions on the PR description.

I think the solution on questions looks good to me. Please let me know what you think? 
If yes, i think we should be doing that on a separate PR after the current one is merged right?(otherwise there will be too much stuff on this PR)

physionet-django/user/management/commands/utility.py
```diff
@@ -0,0 +1,25 @@
+import logging
+
+from django.utils import timezone
+
+from user.models import CredentialApplication
+
+
+def get_pending__reference_response_credentialing_application(n, days, ignore_reminded_application=False):
```
tompollard: why the double underscore after pending? this is a loooooooong name! perhaps the fact that it is a management command makes length less of an issue, but something shorter might be nice? I don't have a good alternative unfortunately. Maybe `get_credential_pending_reference` or `get_credential_awaiting_reference`?

physionet-django/user/management/commands/utility.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    today = timezone.now()
+    limit = today - timezone.timedelta(days=days)
+
+    # get all applications that have decision pending
+    applications = CredentialApplication.objects.filter(status=0)
+    # get applications with Reference Response pending
+    response_applications = applications.filter(credential_review__status=30)
+    # ignore applications that have been reminded already
+    if ignore_reminded_application:
```
tompollard, line 16: Maybe the default should be set to True?

physionet-django/user/management/commands/utility.py
```diff
@@ -0,0 +1,25 @@
...
+
+def get_pending__reference_response_credentialing_application(n, days, ignore_reminded_application=False):
+    """
+    Get n CredentialApplication that has been pending reference response for more than the given days.
+    """
+    today = timezone.now()
+    limit = today - timezone.timedelta(days=days)
+
+    # get all applications that have decision pending
+    applications = CredentialApplication.objects.filter(status=0)
```
tompollard: These queries/filters look like they could be combined. Do you prefer to keep them separate?

physionet-django/notification/templates/notification/email/notify_credential_reference_response_reminder.html
```diff
@@ -0,0 +1,18 @@
+{% load i18n %}{% autoescape off %}{% filter wordwrap:70 %}
+Dear {{ applicant_name }},
+
+This is to remind you that your application for credentialed access to {{ SITE_NAME }} is still pending. We are currently waiting from your reference {{ application.reference_name }} to submit their response.
+
+We contacted your reference on {{ application.reference_contact_datetime }}
+
+Please follow up with your reference so that they can submit their response as soon as possible.
+
+Please note that your application will be automatically rejected if we do not receive a response from your reference within {{ auto_rejection_days }} days of the date of this email.
```
tompollard: "Rejected" doesn't sound very friendly, so I prefer to avoid saying this where we can. In this case, perhaps "closed" would be a good alternative? "`...will be closed if...`".

physionet-django/notification/templates/notification/email/notify_credential_reference_response_reminder.html
```diff
@@ -0,0 +1,18 @@
...
+
+It may take several weeks to process your request after we receive response from your reference. Thank you for your understanding and patience. You can follow the status of your application at: {{ url_prefix }}{% url 'credential_application' %}.
```
tompollard: Typically I think we would give approval shortly after receiving a response from the reference. Perhaps change this to: "`It may take several days to process ....`"

physionet-django/notification/templates/notification/email/notify_credential_reference_response_reminder.html
```diff
@@ -0,0 +1,18 @@
+{% load i18n %}{% autoescape off %}{% filter wordwrap:70 %}
+Dear {{ applicant_name }},
+
+This is to remind you that your application for credentialed access to {{ SITE_NAME }} is still pending. We are currently waiting from your reference {{ application.reference_name }} to submit their response.
+
+We contacted your reference on {{ application.reference_contact_datetime }}
```
tompollard: Add full stop (`.`) at the end of this sentence.

physionet-django/user/management/commands/utility.py
superryeti: Ah yes, weird that i put double underscore.

I really like your recommendation. The reason it turned out super long is because, i kind of wanted to make it very clear and ended up overdoing it

physionet-django/user/management/commands/utility.py
superryeti: Ah yes. i was trying to make it more readable.

But i think you are right, it doesnot make much of a difference, the queries are quite simple. let me refactor these

physionet-django/user/management/commands/utility.py
superryeti: the idea was,

- to add the parameter in a way that we wont need to touch the existing use of the function such as when autorejecting applicants. For that case we dont want to ignore the reminded application(we want to get all applications that are past the configured days)

So currently in `remind_pending_apps` we explicitly do `ignore_reminded_application=True` and make sure we dont resend the notification to the same person


may i ask why you like the default set to True?

physionet-django/notification/templates/notification/email/notify_credential_reference_response_reminder.html
superryeti: Ah yes, my bad, i remember you had pointed this out that the wording is negative, on of the previous PR.

superryeti:
Thanks @tompollard . I made changes according to the review. There is one last review remaining, to which i am waiting for your response before making the change.

physionet-django/user/management/commands/utility.py
tompollard: Thanks Amit, you have a good point. Let's leave the default as False. I'd suggested True because I felt that we would always want to avoid re-chasing people who had already been chased, but on reflection I think it is better to leave as False for the reason you suggest (i.e. more predictable behaviour).

tompollard:
Thanks @amitupreti, I'm happy with this. @bemoody please could you take a quick look before merging to check I haven't missed anything?

physionet-django/user/management/commands/utility.py
superryeti: Thank you. just saw some typo on my comment and fixed that.

physionet-django/notification/templates/notification/email/notify_credential_reference_response_reminder.html
```diff
@@ -0,0 +1,18 @@
+{% load i18n %}{% autoescape off %}{% filter wordwrap:70 %}
+Dear {{ applicant_name }},
+
+This is to remind you that your application for credentialed access to {{ SITE_NAME }} is still pending. We are currently waiting from your reference {{ application.reference_name }} to submit their response.
```
tompollard: `from` -> `for`

physionet-django/user/models.py
```diff
@@ -827,6 +827,7 @@ class AutoRejectionReason(models.TextChoices):
     responder_comments = models.CharField(max_length=500, default='',
                                           blank=True)
     revoked_datetime = models.DateTimeField(null=True)
+    reminder_datetime = models.DateTimeField(null=True)
```
tompollard: It isn't clear that this relates to the reference, rather the applicant.

deploy/production/etc/cron.d/physionet
```diff
@@ -16,3 +16,6 @@
 
 # auto reject pending credentialing applications in case the references don't respond
 0 */1 * * *  www-data  env DJANGO_SETTINGS_MODULE=physionet.settings.production /physionet/python-env/physionet/bin/python3 /physionet/physionet-build/physionet-django/manage.py reject_pending_credentialing_applications
+
+# auto remind users of pending credentialing applications in case the references don't respond(sent before the auto reject)
+0 */1 * * *  www-data  env DJANGO_SETTINGS_MODULE=physionet.settings.production /physionet/python-env/physionet/bin/python3 /physionet/physionet-build/physionet-django/manage.py remind_pending_credentialing_applications
```
tompollard: Instead of `remind_pending_credentialing_applications`, perhaps we should make it clear this relates to the reference? e.g. `remind_reference_identity_check`.

physionet-django/user/models.py
superryeti: Ah yes, thats true, Let me work on the changes

deploy/production/etc/cron.d/physionet
superryeti: Sounds good, i like the name `remind_reference_identity_check`. it much clearer.

physionet-django/notification/templates/notification/email/notify_credential_reference_response_reminder.html
superryeti: I just remembered the reason i didnot add full stop before.
It was because of the date, we got two fullstop, email sample attached for reference


```
From: localhost@localhost
To: amitupreti@fake_gmail.com
Date: Fri, 10 Mar 2023 17:22:24 -0000
Message-ID: <167846894445.58.2894547234442090599@1e454ba8beac>


Dear Amit Upreti,

This is to remind you that your application for credentialed access to
DataShare is still pending. We are currently waiting for your
reference Lorem Ipsum to submit their response.

We contacted your reference on Jan. 8, 2022, 1:49 p.m..

Please follow up with your reference so that they can submit their
response as soon as possible.

Please note that your application will be closed if we do not receive
a response from your reference within 15 days of the date of this
email.

It may take several days to process your request after we receive
response from your reference. Thank you for your understanding and
patience. You can follow the status of your application at:
http://localhost:8000/credential-application/.

Regards, The Development Team


```

superryeti:
Thanks @tompollard 

I renamed the field name to `reference_reminder_datetime` and also renamed the management command to `remind_reference_identity_check` and rebased, squashed the commits.

physionet-django/user/management/commands/remind_reference_identity_check.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # only perform auto reminder if the setting.ENABLE_CREDENTIALING_AUTO_REJECTION is True
+        if not settings.ENABLE_CREDENTIALING_AUTO_REJECTION:
+            LOGGER.info('Auto rejection of credentialing applications is disabled. Exiting.')
+            LOGGER.warning('If this was unintentional, please set '
+                           'ENABLE_CREDENTIALING_AUTO_REJECTION to True in .env file.')
+            return
+
+        total_applications_to_remind = options['number'] or settings.DEFAULT_NUMBER_OF_APPLICATIONS_TO_REMIND
+        auto_rejection_days = settings.MAX_REFERENCE_VERIFICATION_DAYS_BEFORE_AUTO_REJECTION - \
```
tompollard: I think PEP8 discourages `\` for multilines. Could you use `()` here instead?

superryeti: i missed this, making change right away. Thank you.

tompollard:
Thanks @amitupreti, looks good to me.

Issue #1910, opened 2023-03-02
Increase maximum length of event titles (currently 64 characters)


tompollard:
Currently the maximum length of the title for an event is 64 characters.

https://github.com/MIT-LCP/physionet-build/blob/2a8634f0cb778753d6dd939a3203a9fe02887b70/physionet-django/events/models.py#L12-L21

This limit has caused issues with test users (for example, @xborrat has an event whose title has been truncated to something like this: " prioritisation and organization of biomedical Information and Knowledg"

Can we increase the word limit to an arbitrarily bigger but not too big number (128 characters?).

superryeti:
Thanks Tom, i can work on this later today.

PR #1912, opened 2023-03-03
Event title length

superryeti:
Related https://github.com/MIT-LCP/physionet-build/issues/1910
**Context**
During testing of Event feature, it was discovered that in some cases the Event title length allotted was not sufficient. So as per @tompollard's suggestion, we increased the title length to 128 chars.


tompollard:
Thanks Amit, looks good to me! Closes https://github.com/MIT-LCP/physionet-build/issues/1910

Issue #1908, opened 2023-03-02
Status of Credential Application in Participants Tab for Events 


mscanlan-git:
To make the events functionality more useful it would be good to display the status of credential applications for host in the participants tab. If we display the status of credentialing applications directly to the event host, it saves time for us and the host because it isn't necessary to send emails.

This information could be displayed in the table below: 

![Screen Shot 2023-03-02 at 10 33 57 AM](https://user-images.githubusercontent.com/67070862/222475605-7467bc85-7d4a-466c-a44c-992786a168d3.png)

superryeti:
@tompollard  i took over this.

We would want to display  the credential status from the `CredentialReview`, one of the below

```
Not in review
ID check
Reference
Reference response
Final review
```

Right?

PR #1913, opened 2023-03-03
display details of credentialing status on view participants modal

superryeti:
Context:
As discussed on https://github.com/MIT-LCP/physionet-build/issues/1908, On Event Participation list, we just display True or False for credentialing status of a user, which is not super helpful.
As suggested in the issue, showing the actual status will be helpful.

On this PR, i added a method to user model that shows the status of the latest credentialing application for a user.

image

### DIFF
--- a/physionet-django/user/models.py
+++ b/physionet-django/user/models.py
@@ -1037,7 +1037,7 @@ class CredentialApplication(models.Model):
         elif self.credential_review.status <= 20:
             status = 'Awaiting review'
         elif self.credential_review.status == 30:
-            status = 'Awaiting a response from your reference'
+            status = 'Awaiting a response from reference'
         elif self.credential_review.status >= 40:
             status = 'Awaiting final approval'
 
--- a/physionet-django/user/models.py
+++ b/physionet-django/user/models.py
@@ -473,6 +473,23 @@ class User(AbstractBaseUser, PermissionsMixin):
 
         return users
 
+    def get_credentialing_status(self):
+        """
+        Returns the credentialing status of the user.
+        If the user is not credentialed, returns the status of the user's credentialing application.
+        """
+        if self.is_credentialed:
+            return 'Credentialed'
+        else:
+            application = self.credential_applications.last()
+            if not application:
+                return 'No application found'
+
+            if application.status == CredentialApplication.Status.PENDING:
+                return application.get_review_status()
+
+            return 'No application found'
+
 
 class UserLogin(models.Model):
     """Represent users' logins, one per record"""
--- a/physionet-django/events/templates/events/event_entries.html
+++ b/physionet-django/events/templates/events/event_entries.html
@@ -15,7 +15,7 @@
             <td>{{ participant.user.username }}</td>
             <td>{{ participant.user.get_full_name }}</td>
             <td>{{ participant.user.email }}</td>
-            <td>{{ participant.user.is_credentialed }}</td>
+            <td>{{ participant.user.get_credentialing_status }}</td>
             <td>
                 {% if participant.is_cohost %}
                     <input type="checkbox" name="toggle-cohost-status"  event-slug="{{ event.slug }}" value="{{ participant.id }}" disabled checked>


tompollard:
This pull request adds a dependency to the following chunk of code:

https://github.com/MIT-LCP/physionet-build/blob/76e134250c9e5f6d0c1cb3428c2416a9d0c6dbd1/physionet-django/user/models.py#L752-L758

Instead of adding a dependency on that chunk (which is kind of awkward, because the number representations are not human readable), is switching to a Django Enum an option?

physionet-django/user/models.py
```diff
@@ -464,6 +464,28 @@ def get_users_with_permission(app_label, permission_codename):
...
+        """
+        if self.is_credentialed:
+            return 'Credentialed'
+        else:
+            application = self.credential_applications.last()
+            if not application:
+                return 'No application found'
+            application_status = application.status
+            # pending applications should return the review status
+            if application_status == 0:
```
tompollard: I don't think that sharing the statuses below is appropriate (rejected. withdrawn, revoked). In these cases, it is enough to say that the user is not credentialed ("No application found" would be fine).

superryeti:
> dependency to the following

I agree, i had the same thought. The current solution felt very awkward when writing. Would you agree if i do the refactoring in another PR and come back to this PR?

physionet-django/user/models.py
superryeti: Noted. Let me do that.

tompollard:
> I agree, i had the same thought. The current solution felt very awkward when writing. Would you agree if i do the refactoring in another PR and come back to this PR?

This sounds good, thanks

superryeti:
Started refactoring here https://github.com/MIT-LCP/physionet-build/pull/1916

tompollard:
@amitupreti Now that https://github.com/MIT-LCP/physionet-build/pull/1916 is merged, please could you rebase/update?

superryeti:
> @amitupreti Now that #1916 is merged, please could you rebase/update?

Thank you @tompollard. Working on this ASAP

superryeti:
Thanks @tompollard. i updated the PR.

One note for a weird fix i did

For cases where the credential application has pending reference response, the status returned by `CredentialApplication.get_review_status` is 'Awaiting a response from your reference' which is a bit confusing to the Event Host to see as its someone else's application.
So i added a replace in the new method

```
return application.get_review_status().replace('your ', ' ')
```

This doesnot look like a good practice, and the only other option i could think of was to repeat the code of `CredentialApplication.get_review_status` which also might not be the best idea as then we need to update two functions

superryeti:
Current view participant model for reference

view participant model

tompollard:
> For cases where the credential application has pending reference response, the status returned by CredentialApplication.get_review_status is 'Awaiting a response from your reference' which is a bit confusing to the Event Host to see as its someone else's application.

Good catch.

> So i added a replace in the new method: `return application.get_review_status().replace('your ', ' ')`

I agree this feels hacky! It's likely to break at some point, probably silently. Why not tweak the language in the `get_review_status()` method? I think you could remove "your" in the source string without any complaints.

superryeti:
> > For cases where the credential application has pending reference response, the status returned by CredentialApplication.get_review_status is 'Awaiting a response from your reference' which is a bit confusing to the Event Host to see as its someone else's application.
> 
> Good catch.
> 
> > So i added a replace in the new method: `return application.get_review_status().replace('your ', ' ')`
> 
> I agree this feels hacky! It's likely to break at some point, probably silently. Why not tweak the language in the `get_review_status()` method? I think you could remove "your" in the source string without any complaints.

Thank you. let me update the PR

tompollard:
Looks good to me, thanks @amitupreti.

@bemoody are you okay with this change?

superryeti:
@tompollard  rebased this on top of latest change.

physionet-django/user/models.py
```diff
@@ -473,6 +473,24 @@ def get_users_with_permission(app_label, permission_codename):
...
+        Returns the credentialing status of the user.
+        If the user is not credentialed, returns the status of the user's credentialing application.
+        """
+        if self.is_credentialed:
+            return 'Credentialed'
+        else:
+            application = self.credential_applications.last()
+            if not application:
+                return 'No application found'
+            application_status = application.status
```
tompollard: Sorry Amit, I don't think this line is necessary (and for me it makes the code less readable). 

Before I merge, please could you drop the `application_status` variable and instead use `application.status`? 

(unless I'm missing a good reason for assigning `application_status`!

superryeti: Ah yes, i agree, earlier i think there were multiple if else statement so i had stored this in a variable.

Let me change and rebase quickly

superryeti:
Thanks @tompollard , i updated the code and squashed the commits

tompollard:
Looks good to me, thanks Amit!

PR #1916, opened 2023-03-03
refactor credentialapplication status to use enum

superryeti:
On PR https://github.com/MIT-LCP/physionet-build/pull/1913, we discussed the current system is not human readable because of the numbers.

For this, i added a Django integer Enum to the `CredentialApplication` `status`

Based on my understanding this should add a constraint check in database for the status field, but everything else should be the same. All the existing code should function as it is.
Since the purpose of this PR was for readability, i went ahead and refactored the code to use the enum in views, and forms.

I used the following strategy to refactor code
1. Search on all project files by using the following search(`status=`, `status =` and `status`) and refactor where appropriate

For testing, 
1. I ran the automated tests
2. I also tested the Credentialing workflow. I did the following tests
  - Create a new account
  - Submit Credentialing
  -- Login as Admin and  approve initially
  -- send verification request to reference
  --- reference approves
  ----Admin Approves credentialing
  ----Admin Rejects Credentialing
  --- reference rejects
 -Admin rejects credentialing directly
 -Admin approves credentialing directly (  Full approve)
 - Admin revokes Credentialing(previously approved)


bemoody:
`CredentialApplication.CredentialApplicationStatus` is super wordy.  Could it just be `CredentialApplication.Status`?  Or else could `CredentialApplicationStatus` be a global class in user.models?

superryeti:
> `CredentialApplication.CredentialApplicationStatus` is super wordy. Could it just be `CredentialApplication.Status`? Or else could `CredentialApplicationStatus` be a global class in user.models?

I agree, i like the `CredentialApplication.Status` as it kid of keeps everything within the `CredentialApplication` and is nearly same length as `CredentialApplicationStatus`
It was super weird to refactor the code because how long the name turned out.

I update the PR with changes

Thank you

superryeti:
Updated the PR as per suggestion and rebased

physionet-django/console/forms.py
```diff
@@ -406,23 +406,24 @@ class Meta:
...
         self.responder = responder
-        self.fields['status'].choices = CredentialApplication.REJECT_ACCEPT_WITHDRAW[:3]
+        self.fields['status'].choices = CredentialApplication.Status.choices_process_application
 
     def clean(self):
         if self.errors:
             return
 
-        if self.cleaned_data['status'] == 1 and not self.cleaned_data['responder_comments']:
+        if self.cleaned_data['status'] == CredentialApplication.Status.REJECTED\
```
tompollard: There should be a space before the `\` I think, but PEP8 discourages `\` for wrapping long lines, suggesting parentheses instead: https://peps.python.org/pep-0008/#multiline-if-statements

physionet-django/console/forms.py
```diff
@@ -449,23 +450,24 @@ class Meta:
...
```
tompollard: As above: There should be a space before the `\` I think, but PEP8 discourages `\` for wrapping long lines, suggesting parentheses instead: https://peps.python.org/pep-0008/#multiline-if-statements

physionet-django/user/forms.py
```diff
@@ -617,7 +617,8 @@ def clean(self):
         if state_required and not data['state_province']:
             raise forms.ValidationError("Please add your state or province.")
 
-        if not self.instance and CredentialApplication.objects.filter(user=self.user, status=0):
+        if not self.instance and CredentialApplication.objects\
```
tompollard: As above: There should be a space before the `\` I think, but PEP8 discourages `\` for wrapping long lines, suggesting parentheses instead: https://peps.python.org/pep-0008/#multiline-if-statements

physionet-django/console/forms.py
superryeti: Nice. i like this. Thank you for sharing. let me make the changes with parentheses.

superryeti:
Thanks @tompollard . i refactored the code to use parenthesis instead of `\` in the recent changes.

physionet-django/user/models.py
```diff
@@ -749,13 +749,18 @@ class CredentialApplication(models.Model):
...
     )
 
-    REJECT_ACCEPT_WITHDRAW = (
-        ('', '-----------'),
-        (1, 'Reject'),
-        (2, 'Accept'),
-        (3, 'Withdrawn'),
-        (4, 'Revoked')
-    )
+    class Status(models.IntegerChoices):
```
tompollard, line 761: Is there a reason to use Integers here? Why not TextChoices?

superryeti: Originally it was because the code before this was using integer, so my idea was if we use integerchoices we wont need to worry about the existing contents(as it was already integers and same values like right now)

In summary, mostly so that we dont make changes to the existing data.


PS: i learned about the migrate_forward,backward stuff few days ago, so if you prefer the textchoices, i can refactor the code.

tompollard: Thanks Amit, that's not a bad reason! I think it would be an improvement to switch to `TextChoices` if you think it wouldn't be too much extra effort? If we are going to make the change, now would be a good time. 

I notice `application.status` is referencing the numerical values in a few additional places (e.g. https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/user/management/commands/test_reject_pending_credentialing_applications.py and https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/user/templates/user/application_display_table.html so these would also need to be updated).

superryeti: For sure, let me switch to TextChoices.

Thank you i seem to missed the test.

Regarding the templates, i skipped them thinking that we will have to add the `Status` in the `context` returned by view.

Let me update the templates as well.

tompollard: @bemoody you mentioned that you thought keeping as integer would be easier/better? if so, please could you post your suggestion here before @amitupreti moves everything?

superryeti: @bemoody I hope you don’t mind me nudging you in case you missed @tompollard’s question.

bemoody:
Changing the type from integer to text is theoretically possible by renaming the field.  It would take some careful thinking.  I think it would be a lot more work than it's worth.

tompollard:
> Changing the type from integer to text is theoretically possible by renaming the field. It would take some careful thinking. I think it would be a lot more work than it's worth.

Okay, @amitupreti let's keep as it is!

physionet-django/user/models.py
tompollard: decision is to leave as integer type

superryeti:
> > Changing the type from integer to text is theoretically possible by renaming the field. It would take some careful thinking. I think it would be a lot more work than it's worth.
> 
> Okay, @amitupreti let's keep as it is!

Thanks @tompollard . let me make changes to the templates and few places that i had missed earlier.

tompollard:
@amitupreti please could you regenerate the migration?

superryeti:
Thanks @tompollard . I rebased and regenerated the migration file

tompollard:
Thanks @amitupreti, looks good to me!

PR #1919, opened 2023-03-07
Fix: Archiving files throws "No such file or directory" error

rafalkosla101:
Changing those parts will lead to reactive implementation that can be used both by local files system and remote one by setting proper ENV variables.


tompollard:
Looks good to me, thanks Rafal!

PR #1920, opened 2023-03-07
Fix projects not being copyeditable by their author

kshalot:
We could also do:
```python
if self.author_editable() and self.authors.filter(is_submitting=True, user=user).exists():
    return True
if self.copyeditable() and user == self.editor:
    return True
return False
```
but the version proposed in the PR makes it clear what the logic behind a project being editable is without forcing the reader to analyze the branches. Python evaluates boolean conditions lazily so at worst this version is adding an additional call to `self.copyeditable()` which is very light.


tompollard:
Looks good, thanks @kshalot (+ like this implementation!)

PR #1922, opened 2023-03-07
Improve TestURLs coverage

bemoody:
To improve test coverage and to more easily examine the HTML generated by the site, we added "URL testing" a few months ago.  See pull #1695 for an overview.

Currently, TestURLs covers URLs in the following apps:
- events
- export
- notification
- physionet
- search

This pull request adds coverage for:
- lightwave
- user

Still to be done:
- console
- project

In addition, this makes the test requirements slightly stricter - treat redirection to the login page as an error.  We want to see what these pages look like when viewed by a logged-in user, so authentication failure is an error.  A few existing tests had to be fixed as a consequence.

### DIFF
--- a/physionet-django/events/urls.py
+++ b/physionet-django/events/urls.py
@@ -11,4 +11,8 @@ urlpatterns = [
 ]
 
 # Parameters for testing URLs (see physionet/test_urls.py)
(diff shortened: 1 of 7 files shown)


tompollard:
Thanks Benjamin, looks good to me.

PR #1923, opened 2023-03-08
Events: show inactive participants to host and Admin

superryeti:
Context:
Currently on Events Homepage, and on Event management page(admin console), we can only see the list of people approved to the event. We cant see who has been rejected to the event or if someone has withdrawn access to the event.
It will be helpful to hosts if they can have access to the unsuccessful participants and the reason that they were rejected.

This PR adds another table on the existing modal which lets us switch tabs between active, inactive participants(the wording is a bit wrong here, maybe this should be `inactive applicants/unsuccessful aplications`)

Pictures for reference

Active tab

InActive Tab


superryeti:
Tab variation option(this one doesnot have the dark background and black text)
image

tompollard:
Listing application history for users is a good idea, but the changes in this PR need a little more thought. Possibly just a case of tweaking the language (e.g. "Inactive Participant" is an odd name for someone who is not a participant; "inactive applications" is an odd way to describe an application that was withdrawn or closed).

superryeti:
> Listing application history for users is a good idea, but the changes in this PR need a little more thought. Possibly just a case of tweaking the language (e.g. "Inactive Participant" is an odd name for someone who is not a participant; "inactive applications" is an odd way to describe an application that was withdrawn or closed).

Could we create 1 tab per type, for example
1. `Approved`
2. `Not Approved`
3. `Withdrawn`
4. `Revoked`

superryeti:
> > Listing application history for users is a good idea, but the changes in this PR need a little more thought. Possibly just a case of tweaking the language (e.g. "Inactive Participant" is an odd name for someone who is not a participant; "inactive applications" is an odd way to describe an application that was withdrawn or closed).
> 
> Could we create 1 tab per type, for example
> 
> 1. `Approved`
> 2. `Not Approved`
> 3. `Withdrawn`
> 4. `Revoked`

Let me think a bit about the names and get back. The multiple tab option is also not looking very good

superryeti:
Hello @tompollard, thank you for your feedback. I agree that the language in the menu name and the application status needs to be more precise and accurate. I have the following suggestions:

1. Use `Application History` instead of `Inactive Application` as the menu name and then include the `status` on a column
2. Or use different tabs for different types of applications, such as: 
   a. `Not Approved Applications `
   b. `Withdrawn Applications `
   c. `Revoked Applications` 
   d. `Closed Applications`

2nd option would allow Host to filter  applications by status

What do you think of these suggestions? Do you approve them or do you have any other feedback? Please let me know.

kshalot:
Hey, @tompollard! (@alistairewj I'm also roping you in. In case you have any comments)

  I’m doing a small review of the old events PRs to see what can be pushed forward and what needs details. I resolved the conflicts in this branch so we can see the changes.

Amit proposed two ways (https://github.com/MIT-LCP/physionet-build/pull/1923#issuecomment-1464183049) to get rid of the “inactive” wording before he left.

He mentioned that the multiple tab option is not ideal, but functionally it’s closer to other existing flows (for example the training view, pictured below) and separates the different statuses in a cleaner way (it’s going to be way easier to answer more specific questions like “How many people have withdrawn from the event? Who was it?” without introducing some filtering capabilities).

Trainings

Showing those tabs in the “total participants” view might be confusing though, since they are not counted towards the total (the UI is visible in a screenshot below). There’s also no indication that anything apart from accepted participants will be visible when clicking the button. This could be fixed with a simple rewording (total participants -> accepted participants, view participants -> view applications) although it still wouldn't be ideal:

ViewApplications

We can also try making it granular and show it in the event details directly:

Granular

Similar changes would be made to the Events page (outside of the admin panel).

I think just keeping it simple for now and having a single button would be the better option because I feel that the UI is bad either way. Showing the participants in a modal feels more like trying to simplify the implementation of the events pages than a conscious design decision, but I feel that if we want to rework the page, it should be a separate issue/PR. I'm curious to hear your thoughts.

physionet-django/events/templates/events/event_pending_applications.html
```diff
@@ -0,0 +1,31 @@
...
+                  <td>{{ application.comment_to_applicant }}</td>
+                  <td>{{ application.decision_datetime | date }}</td>
+                </tr>
+              {% endfor %}
+            {% endwith %}
+            </tbody>
+        </table>
+    </div>
+  </div>
+</div>
```
kshalot: I'd end all the files with newlines. Just for consistency.

physionet-django/events/templatetags/participation_status.py
```diff
@@ -23,3 +23,34 @@ def is_on_waiting_list(user, event):
 @register.filter(name='has_access_to_event_dataset')
 def has_access_to_event_dataset(user, dataset):
     return has_access_to_event_dataset_func(user, dataset)
+
+
+@register.filter(name='get_inactive_applications')
+def get_inactive_applications(event):
```
kshalot: I don't see this function used anywhere.

physionet-django/events/templatetags/participation_status.py
```diff
@@ -23,3 +23,34 @@ def is_on_waiting_list(user, event):
...
+    return event.applications.filter(
+        status__in=[EventApplication.EventApplicationStatus.WITHDRAWN]
+    )
+
+
+@register.filter(name='get_rejected_applications')
+def get_rejected_applications(event):
+    return event.applications.filter(
+        status__in=[EventApplication.EventApplicationStatus.NOT_APPROVED]
+    )
```
kshalot: This looks like you are trying to place logic inside the template. It's not the right way in my opinion. Instead of sending all the applications and then filtering inside the view (`{% with rejected_applications=event|get_rejected_applications %}`) you could send `rejected_applications`, `withdrawn_applications` etc. This way the business logic is not embedded inside the template, which in this case could trigger an SQL query.

physionet-django/console/templates/console/event_management.html
```diff
@@ -76,4 +116,59 @@ <h5 class="modal-title">Participants</h5>
...
+          <h5 class="modal-title">Withdrawn Applications</h5>
+          <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+            <span aria-hidden="true">&times;</span>
+          </button>
+        </div>
+        {% include 'events/event_withdrawn_applications.html' %}
+      </div>
+    </div>
+  </div>
+
```
kshalot: I see a lot of duplication in this snippet. Maybe introducing a variable like:
```python
[
  {
    "id": "withdrawn_applications",
    "title": "Withdrawn Applications",
    "template": "event_withdrawn_applications.html"
  },
  ...
]
```
and rendering that dynamically within a for-loop would be better.

physionet-django/console/templates/console/event_management.html
```diff
@@ -46,11 +46,51 @@ <h1>{{ event.title }}</h1>
...
+                      data-target="#participants">View</button>
             </div>
           </div>
         </div>
       </div>
+      <div class="row mb-1">
+        <div class="col-md-3">Pending Applications:</div>
+        <div class="col-md-9">
+          <div class="row mb-1">
+            <div class="col-md-1">{{ event.participants.count }}</div>
```
kshalot: `even.participants.count` could be memoized using the `with` templatetag.

physionet-django/events/templates/events/event_home.html
```diff
@@ -114,7 +117,46 @@ <h5 class="modal-title">Participants</h5>
...
+                  <div class="modal-header">
+                    <h5 class="modal-title">Withdrawn Application</h5>
+                    <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                      <span aria-hidden="true">&times;</span>
+                    </button>
+                  </div>
+                    {% include 'events/event_withdrawn_applications.html' %}
+                  </div>
+                </div>
+              </div>
```
kshalot: Similarly to the modals in the admin panel, this could be dynamically rendered using a list of dictionaries. (the buttons should also reference the IDs. The buttons could be dynamically rendered as well to maintain the id<->id consistency.

physionet-django/events/templates/events/event_pending_applications.html
```diff
@@ -0,0 +1,31 @@
+{% load participation_status %}
+<div class="tab-content" id="nav-tabContent">
+  <div class="tab-pane fade show active" id="active-participant" role="tabpanel" aria-labelledby="active-participant-tab">
```
kshalot: In both those places the id will not be unique (since you are rendering all 3 templates, you'll have 3 divs with the "active-participant" id for example).

Also - this should also be a single template and the only configurable part would be the list of participants. I see that all the templates use the same table body:
```html
<td>{{ application.user.username }}</td>
<td>{{ application.user.get_full_name }}</td>
<td>{{ application.user.email }}</td>
<td>{{ application.comment_to_applicant }}</td>
<td>{{ application.decision_datetime | date }}</td>
```

Is there any other difference? Either way be can make it configurable with arguments and have one template that renders the list of applications.

kshalot (changes requested):

physionet-django/events/templates/events/event_pending_applications.html
Rutvikrj26: Done.

physionet-django/console/templates/console/event_management.html
Rutvikrj26: I've updated the view to include the count alongwith the actual applications as suggested in another review snippet.

physionet-django/console/templates/console/event_management.html
Rutvikrj26: Updated the template and the view accordingly.

physionet-django/events/templatetags/participation_status.py
Rutvikrj26: The Logic after update is no longer embedded in the templatetags. 
I've updated the view to pass the event details to the frontend.

physionet-django/events/views.py
```diff
@@ -146,6 +147,43 @@ def event_home(request):
         else:
             form_error = True
 
+    event_details = {}
+    for selected_event in events_all:
+        participants = selected_event.participants.all()
+        pending_applications = selected_event.applications.filter(
+            status__in=[EventApplication.EventApplicationStatus.WAITLISTED])
```
kshalot: Won't this work?
```suggestion
            status=EventApplication.EventApplicationStatus.WAITLISTED)
```

physionet-django/events/views.py
```diff
@@ -146,6 +147,43 @@ def event_home(request):
         else:
             form_error = True
 
+    event_details = {}
+    for selected_event in events_all:
+        participants = selected_event.participants.all()
```
kshalot: This is the N+1 problem. The participants should be prefetched for the events.

physionet-django/events/views.py
```diff
@@ -146,6 +147,43 @@ def event_home(request):
...
+        rejected_applications = selected_event.applications.filter(
+            status__in=[EventApplication.EventApplicationStatus.NOT_APPROVED])
+        withdrawn_applications = selected_event.applications.filter(
+            status__in=[EventApplication.EventApplicationStatus.WITHDRAWN])
+
+        event_details[selected_event.id] = [
+            {
+                'id': 'participants',
+                'title': 'Total participants:',
+                'count': participants.count(),
```
kshalot: I'm just thinking that maybe this would be faster? `len` doesn't require an additional query and the results are going to be small here.
```suggestion
                'count': len(participants),
```

physionet-django/console/views.py
```diff
@@ -2984,14 +2985,54 @@ def event_management(request, event_slug):
...
         'console/event_management.html',
         {
             'event': selected_event,
             'event_dataset_form': event_dataset_form,
             'event_datasets': event_datasets,
+            'status_options': status_options,
+            'participants': participants,
+            'pending_applications': pending_applications,
+            'rejected_applications': rejected_applications,
+            'withdrawn_applications': withdrawn_applications,
```
kshalot: Is this used in the template?

physionet-django/console/views.py
```diff
@@ -2984,14 +2985,54 @@ def event_management(request, event_slug):
...
+    participants = selected_event.participants.all()
+    pending_applications = selected_event.applications.filter(
+        status__in=[EventApplication.EventApplicationStatus.WAITLISTED])
+    rejected_applications = selected_event.applications.filter(
+        status__in=[EventApplication.EventApplicationStatus.NOT_APPROVED])
+    withdrawn_applications = selected_event.applications.filter(
+        status__in=[EventApplication.EventApplicationStatus.WITHDRAWN])
+
     event_datasets = selected_event.datasets.filter(is_active=True)
+    status_options = [
```
kshalot: I don't like the name `option` which also is later used in the templates like `option.objects` - this is unclear imo. Maybe something like `participant_info` or something?

kshalot (changes requested):

Rutvikrj26:
@kshalot Thanks for the comments. I hadn't considered the efficiency and number of queries. 
I've updated the codebase and resolved the conflicts with the dev. 
Please check and let me know if this looks good.

physionet-django/console/views.py
```diff
@@ -5,6 +5,7 @@
 from collections import OrderedDict
 from datetime import datetime
 from itertools import chain
+import stat
```
kshalot: I think you added this by accident.

physionet-django/console/views.py
```diff
@@ -2985,41 +2986,91 @@ def event_management(request, event_slug):
...
-                        access_type=event_dataset_form.cleaned_data['access_type'],
-                        is_active=True).count() == 0:
+                if (
+                    selected_event.datasets.filter(
+                        dataset=event_dataset_form.cleaned_data["dataset"],
+                        access_type=event_dataset_form.cleaned_data["access_type"],
+                        is_active=True,
+                    ).count()
+                    == 0
+                ):
```
kshalot: I'd refrain from huge logic like this inside the if statement. I'd rather see a meaningful variable name here, like `active_datasets` or something.

physionet-django/events/views.py
```diff
@@ -114,53 +116,123 @@ def event_home(request):
...
                     decision=EventApplication.EventApplicationStatus.NOT_APPROVED.label,
-                    comment_to_applicant=form.cleaned_data.get('comment_to_applicant')
+                    comment_to_applicant=form.cleaned_data.get("comment_to_applicant"),
                 )
 
             return redirect(event_home)
         else:
             form_error = True
 
+    events_all = Event.objects.all().prefetch_related(
```
kshalot: Why not just `events`?

physionet-django/events/views.py
```diff
@@ -114,53 +116,123 @@ def event_home(request):
...
+        "participants",
+        "applications",
+    )
+
+    event_details = {}
+    for selected_event in events_all:
```
kshalot:
```suggestion
    for event in events:
```

Isn't that clearer?

physionet-django/events/views.py
```diff
@@ -114,53 +116,123 @@ def event_home(request):
...
+        all_applications = selected_event.applications.all()
+        pending_applications = [
+            app
```
kshalot: In my head "app" means "program" or "software application". I think using "application" here would be better, no need to use shorthands.

physionet-django/events/views.py
```diff
@@ -114,53 +116,123 @@ def event_home(request):
...
+        rejected_applications = [
+            app
+            for app in all_applications
+            if app.status == EventApplication.EventApplicationStatus.NOT_APPROVED
+        ]
+        withdrawn_applications = [
+            app
+            for app in all_applications
+            if app.status == EventApplication.EventApplicationStatus.WITHDRAWN
+        ]
```
kshalot: Technically you could do this in one sweep instead of four, but for readability's sake I'd sake it's fine the way it is

physionet-django/events/views.py
```diff
@@ -3,7 +3,7 @@
 from django.http import JsonResponse
 from django.shortcuts import render, get_object_or_404, redirect
 from django.contrib import messages
-from django.db.models import Q
+from django.db.models import Q, Prefetch
```
kshalot: Is this used? I mean `Prefetch`

kshalot (changes requested):

physionet-django/events/views.py
Rutvikrj26: Events_all is a reference to all the events rather than just the active or archived events.

physionet-django/events/views.py
Rutvikrj26: event is used above in the function by amit for some other logic and hence tried to avoid using event.

physionet-django/console/views.py
```diff
@@ -2985,41 +2985,88 @@ def event_management(request, event_slug):
...
 
+    participants = selected_event.participants.all()
+    pending_applications = selected_event.applications.filter(
+        status__in=[EventApplication.EventApplicationStatus.WAITLISTED]
+    )
+    rejected_applications = selected_event.applications.filter(
+        status__in=[EventApplication.EventApplicationStatus.NOT_APPROVED]
+    )
+    withdrawn_applications = selected_event.applications.filter(
+        status__in=[EventApplication.EventApplicationStatus.WITHDRAWN]
```
kshalot: Everywhere you are doing `status__in`, woudln't `status=` suffice?

Rutvikrj26:
@tompollard @kshalot requesting further review and/or merge. 
PR #1925 partially builds on top of this PR, and would like to get this one complete & merged.
Thanks!

physionet-django/console/views.py
```diff
@@ -2985,41 +2985,88 @@ def event_management(request, event_slug):
...
             if event_dataset_form.is_valid():
-                if selected_event.datasets.filter(
-                        dataset=event_dataset_form.cleaned_data['dataset'],
-                        access_type=event_dataset_form.cleaned_data['access_type'],
-                        is_active=True).count() == 0:
+                active_datasets = selected_event.datasets.filter(
+                    dataset=event_dataset_form.cleaned_data["dataset"],
+                    access_type=event_dataset_form.cleaned_data["access_type"],
+                    is_active=True)
+                if len(active_datasets) == 0:
```
kshalot: Here it actually makes sense to use `.count()` rather than `len`. `len` makes sense when you already have the data loaded from the database and in your program's memory since it just iterates over the Python iterable to count the items. `.count()` fetches the length of the collection from the database, so if you already have the data on hand it's simpler to just count it locally rather than sending a request to the database.

In this case it's better to go for:

```suggestion
                if active_datasets.count() == 0:
```

physionet-django/events/views.py
```diff
@@ -114,53 +116,123 @@ def event_home(request):
...
             form_error = True
 
+    events = Event.objects.all().prefetch_related(
+        "participants",
+        "applications",
+    )
+
+    event_details = {}
+    for event in events:
+        all_applications = event.applications.all()
```
kshalot: Same logic as with `events_all` would apply here - wouldn't `applications` or `event_applications` be enough? It's just a thought though, I'm not a stickler for naming.

kshalot (commented):
A few tiny thoughts.

physionet-django/console/views.py
Rutvikrj26: yeah - sorry - that must have gotten carried over while I was trying to fix the black formatter issue.

physionet-django/console/views.py
Rutvikrj26: Thanks - I had the notion that the line about defining the filter leads to a query, but it actually does not trigger a query, and the query is only triggered when the variable is evaluated.

physionet-django/events/views.py
Rutvikrj26: Definitely!
I think what Tom mentioned the other day in the meeting makes sense - it can always be looked up, and even though the naming looks more clear in the context of this PR, in terms of the whole codebase, it can look confusing. 
I've updated the naming here as well.

kshalot (approved):

tompollard:
Looks good, thanks!

PR #1924, opened 2023-03-08
Show permissions groups on Admin Console

### LINKED ISSUE #1548: Display permissions groups in the admin console
In https://github.com/MIT-LCP/physionet-build/pull/1487 we added new functionality for management of permission groups. We should update the admin console so that permissions groups can be viewed for a user.

e.g. In addition to displaying the list of admin users at: http://localhost:8000/console/users/admin/, we should indicate which permission groups have been assigned to the user.

superryeti:
Closes https://github.com/MIT-LCP/physionet-build/issues/1548

**Context:**
On https://github.com/MIT-LCP/physionet-build/pull/1487, we added a new feature to manage permission groups, and on https://github.com/MIT-LCP/physionet-build/pull/1719/files, we refactored the existing codebase to use remove the use of `is_admin` and use the permission groups.

With those two changes, the Admin(people who are in Permission Group `Admin`) list page that exists right now  http://localhost:8000/console/users/admin/ is a bit outdated. because now we should have an option to see all the existing Permission groups and detail of each permission group.

**This PR does the following**
1. Adds a Permission group list page that lists the permission group and shows basic details.
2. Adds a Permission group detail page, that shows details of permission group
3. updates the user management page to link to permission group detail page

**Screenshots for reference**

Permission group list page http://localhost:8000/console/users/groups/

image

Permission group detail page http://localhost:8000/console/users/groups/Admin/

image

**TODO after this**
1. Remove the admin url https://localhost:8000/console/users/admin/ as it will be less relevant after this PR

### DIFF
--- a/physionet-django/console/templates/console/console_navbar.html
+++ b/physionet-django/console/templates/console/console_navbar.html
@@ -238,6 +238,9 @@
             <li class="nav-item {% if group == 'all' %}active{% endif %}">
               <a id="nav_all_users" class="nav-link" href="{% url 'users' 'all' %}">All Users</a>
             </li>
+            <li class="nav-item {% if user_groups_nav%}active{% endif %}">
+              <a id="nav_user_groups" class="nav-link" href="{% url 'user_groups' %}">User Groups</a>
+            </li>
             <li class="nav-item {% if group == 'admin' %}active{% endif %}">
               <a id="nav_all_users" class="nav-link" href="{% url 'users' 'admin' %}"> {{admin}}Administrators</a>
             </li>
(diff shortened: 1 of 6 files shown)


superryeti:
Question:

For displaying permission would using pills/badge make more sense(instead of table)? https://getbootstrap.com/docs/4.0/components/badge/

Something like
image

tompollard:
Useful, thanks! 

> For displaying permission would using pills/badge make more sense(instead of table)? https://getbootstrap.com/docs/4.0/components/badge/

I prefer the table. Clearer than badges.

PR #1925, opened 2023-03-08
Event: Allow host to add cohosts

superryeti:
Related to  https://github.com/MIT-LCP/physionet-build/pull/1901

**Context**

On https://github.com/MIT-LCP/physionet-build/pull/1901, we reverted the add cohost feature because of security issue.
On this PR we reimplement the feature by using forms

Cohosts are supposed to be helper for hosts and can manage the participants(approving, rejecting participants)
To be cohost, one has to be first approved to join the event.

**Workflow**

The event host clicks the `Make Cohost` or the `Remove Cohost` button, and the following will happen

1. we will immediately disable the button(so that host cant spam the button), it also acts like a feedback to host that the task is processing.
2. Then we send Ajax request to our view
3. Depending on the response from view,
 a. If successful, we will either change the button from `Make Cohost` to `Remove Cohost` or vice versa(depending on whether the host was trying to make cohost or remove cohost)
 b. If unsuccessful, we will enable the button again to allow host to try again.

**Screen Record for quick review**

[Feature Demo](https://user-images.githubusercontent.com/24412619/223860722-1260598d-fff6-496c-9436-64e9eff2e51c.webm)


bemoody:
What is a "cohost"?  I feel like that question has not been answered.

Is "cohost" status a meaningless gold star?  Or is it intended to confer some permissions on the person?

Why would we want to have "cohosts" rather than having multiple "hosts"?

superryeti:
> What is a "cohost"? I feel like that question has not been answered.
> 
> Is "cohost" status a meaningless gold star? Or is it intended to confer some permissions on the person?
> 
> Why would we want to have "cohosts" rather than having multiple "hosts"?

My bad, let me explain
Cohost is a participant who will have some extra permission like approving/disapproving other participants, viewing the list of applications.

> Why would we want to have "cohosts" rather than having multiple "hosts"?

The idea for multiple hosts never came up before during the discussion. Tagging @tompollard  and @alistairewj  for feedback on multiple hosts.

alistairewj:
I think we'd want to avoid multiple hosts to avoid having multiple users with permission to update event content (similar to the role of the corresponding author in projects).

bemoody:
Okay, yeah, I see this was mentioned briefly in issue #1839.

This needs further discussion, I'm not really comfortable with being able to grant permissions over data access control so casually.

bemoody:
@alistairewj: this is true, though only allowing a single person to edit is often *annoying* and it'd be better if we could avoid that.

superryeti:
@alistairewj @bemoody @tompollard i remember we wanted to meetup do discuss about Events anyway, I was trying to setup the meeting, but i was unable to because i couldn't get everyone's availability. So brining this up to setup meeting as looks like we need to discuss/plan bunch of things :) 


Also sorry i couldn't find all 3 of your on Slack so brought this up here

superryeti:
Email templates for reference

For Make Cohost

```
physionet-build-dev-1   | [10/Mar/2023 14:28:38] "GET /events/ HTTP/1.1" 200 31822
physionet-build-dev-1   | Content-Type: text/plain; charset="utf-8"
physionet-build-dev-1   | MIME-Version: 1.0
physionet-build-dev-1   | Content-Transfer-Encoding: 7bit
physionet-build-dev-1   | Subject: DataShare Event Cohost Status Change
physionet-build-dev-1   | From: localhost@localhost
physionet-build-dev-1   | To: amitupreti@fake_gmail.com
physionet-build-dev-1   | Date: Fri, 10 Mar 2023 19:28:41 -0000
physionet-build-dev-1   | Message-ID: <167847652126.73.15411785239375743028@f0b72aed4da1>
physionet-build-dev-1   | 
physionet-build-dev-1   | 
physionet-build-dev-1   | Dear Amit Upreti,
physionet-build-dev-1   | 
physionet-build-dev-1   | You have been added as Cohost to the Event : Test Event physionet.
physionet-build-dev-1   | 
physionet-build-dev-1   | You can now manage the event participants from your events dashboard.
physionet-build-dev-1   | 
physionet-build-dev-1   | You can view further information about the event using the following
physionet-build-dev-1   | link:
physionet-build-dev-1   | http://localhost:8000/events/PsenZXNhPluI/
physionet-build-dev-1   | 
physionet-build-dev-1   | Regards
physionet-build-dev-1   | The DataShare Team
physionet-build-dev-1   | 
physionet-build-dev-1   | 
physionet-build-dev-1   | -------------------------------------------------------------------------------
physionet-build-dev-1   | Content-Type: text/plain; charset="utf-8"
physionet-build-dev-1   | MIME-Version: 1.0
physionet-build-dev-1   | Content-Transfer-Encoding: 7bit
physionet-build-dev-1   | Subject: DataShare Event Cohost Status Change
physionet-build-dev-1   | From: localhost@localhost
physionet-build-dev-1   | To: amitupreti@fake_gmail.com
physionet-build-dev-1   | Date: Fri, 10 Mar 2023 19:28:41 -0000
physionet-build-dev-1   | Message-ID: <167847652126.73.5870397640578469008@f0b72aed4da1>
physionet-build-dev-1   | 
physionet-build-dev-1   | 
physionet-build-dev-1   | Dear Roger Greenwood Mark,
physionet-build-dev-1   | 
physionet-build-dev-1   | You have provided Amit Upreti Cohost access to the Event : Test Event
physionet-build-dev-1   | physionet.
physionet-build-dev-1   | 
physionet-build-dev-1   | If this was done by mistake, you can change the cohost status by
physionet-build-dev-1   | visiting the events dashboard.
physionet-build-dev-1   | 
physionet-build-dev-1   | http://localhost:8000/events/
physionet-build-dev-1   | 
physionet-build-dev-1   | Regards
physionet-build-dev-1   | The DataShare Team
physionet-build-dev-1   | 
physionet-build-dev-1   | 
```

For Remove Cohost

```
physionet-build-dev-1   | -------------------------------------------------------------------------------
physionet-build-dev-1   | [10/Mar/2023 14:28:41] "POST /events/manage_co_hosts/ HTTP/1.1" 200 40
physionet-build-dev-1   | Content-Type: text/plain; charset="utf-8"
physionet-build-dev-1   | MIME-Version: 1.0
physionet-build-dev-1   | Content-Transfer-Encoding: 7bit
physionet-build-dev-1   | Subject: DataShare Event Cohost Status Change
physionet-build-dev-1   | From: localhost@localhost
physionet-build-dev-1   | To: amitupreti@fake_gmail.com
physionet-build-dev-1   | Date: Fri, 10 Mar 2023 19:28:45 -0000
physionet-build-dev-1   | Message-ID: <167847652579.73.11830662540034523928@f0b72aed4da1>
physionet-build-dev-1   | 
physionet-build-dev-1   | 
physionet-build-dev-1   | Dear Amit Upreti,
physionet-build-dev-1   | 
physionet-build-dev-1   | Your Cohost access has been removed from the Event : Test Event
physionet-build-dev-1   | physionet.
physionet-build-dev-1   | 
physionet-build-dev-1   | You can view further information about the event using the following
physionet-build-dev-1   | link:
physionet-build-dev-1   | http://localhost:8000/events/PsenZXNhPluI/
physionet-build-dev-1   | 
physionet-build-dev-1   | Regards
physionet-build-dev-1   | The DataShare Team
physionet-build-dev-1   | 
physionet-build-dev-1   | 
physionet-build-dev-1   | -------------------------------------------------------------------------------
physionet-build-dev-1   | Content-Type: text/plain; charset="utf-8"
physionet-build-dev-1   | MIME-Version: 1.0
physionet-build-dev-1   | Content-Transfer-Encoding: 7bit
physionet-build-dev-1   | Subject: DataShare Event Cohost Status Change
physionet-build-dev-1   | From: localhost@localhost
physionet-build-dev-1   | To: amitupreti@fake_gmail.com
physionet-build-dev-1   | Date: Fri, 10 Mar 2023 19:28:45 -0000
physionet-build-dev-1   | Message-ID: <167847652580.73.11619867650555598756@f0b72aed4da1>
physionet-build-dev-1   | 
physionet-build-dev-1   | 
physionet-build-dev-1   | Dear Roger Greenwood Mark,
physionet-build-dev-1   | 
physionet-build-dev-1   | You have removed Amit Upreti's Cohost access from the Event : Test
physionet-build-dev-1   | Event physionet.
physionet-build-dev-1   | 
physionet-build-dev-1   | If this was done by mistake, you can change the cohost status by
physionet-build-dev-1   | visiting the events dashboard.
physionet-build-dev-1   | 
physionet-build-dev-1   | http://localhost:8000/events/
physionet-build-dev-1   | 
physionet-build-dev-1   | Regards
physionet-build-dev-1   | The DataShare Team
physionet-build-dev-1   | 
physionet-build-dev-1   | 
physionet-build-dev-1   | -------------------------------------------------------------------------------
```

physionet-django/notification/utility.py
```diff
@@ -1005,6 +1005,45 @@ def notify_participant_event_decision(request, user, event, decision, comment_to
     send_mail(subject, body, settings.DEFAULT_FROM_EMAIL, [user.email], fail_silently=False)
 
 
+def notify_event_cohost_cohost_status_change(request, cohost, event, status='Make cohost'):
```
kshalot: I don't see a reason why the status would have a default value here.

physionet-django/events/views.py
```diff
@@ -235,3 +235,58 @@ def event_detail(request, event_slug):
...
+            return JsonResponse({'success': 'Cohost removed successfully'})
+        elif 'Make cohost' in request.POST.get('submit'):
+            if participant.is_cohost:
+                return JsonResponse({'error': 'User is already a cohost of this event'}, status=403)
+            participant.is_cohost = True
+            participant.save()
+            # notify the cohost that they have been added as a cohost for the event
+            notification.notify_event_cohost_cohost_status_change(request=request, cohost=participant.user,
+                                                                  event=event, status='Make cohost')
+            # notify the host that they have added a cohost for the event
```
kshalot: I don't see a point in adding comments that just explain code that should be self-documenting. I'd remove them.

physionet-django/events/templates/events/email/event_cohost_cohost_status_change.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n %}{% autoescape off %}{% filter wordwrap:70 %}
```
kshalot, line 1: I don't like the names of both of the mail templates:
`event_cohost_cohost_status_change.html` and `event_host_cohost_status_change.html`

They are very confusing and, in my opinion, do not convey the information in the template.

physionet-django/events/templates/events/email/event_cohost_cohost_status_change.html
```diff
@@ -0,0 +1,15 @@
...
+Dear {{ name }},
+{% if status == 'Make cohost' %}
+You have been added as Cohost to the Event : {{ event_title }}.
+
+You can now manage the event participants from your events dashboard.
+{% elif status == 'Remove cohost' %}
+Your Cohost access has been removed from the Event : {{ event_title }}.
+{% endif %}
+You can view further information about the event using the following link:
+{{ url_prefix }}{{ event_url }}
```
kshalot: Why is it set to this instead of the `event_url` that is available here anyway? This looks like trying to reimplement `reverse`.

physionet-django/events/templates/events/email/event_host_cohost_status_change.html
```diff
@@ -0,0 +1,14 @@
+{% load i18n %}{% autoescape off %}{% filter wordwrap:70 %}
+Dear {{ host_name }},
+{% if status == 'Make cohost' %}
+You have provided {{ cohost_name }} Cohost access to the Event : {{ event_title }}.
+{% elif status == 'Remove cohost' %}
+You have removed {{ cohost_name }}'s Cohost access from the Event : {{ event_title }}.
+{% endif %}
+If this was done by mistake, you can change the cohost status by visiting the events dashboard.
+
+{{ url_prefix }}/events/
```
kshalot: The way the URL is built here is inconsistent with the other template. We need to align this (should be sent from the controller).

physionet-django/events/templates/events/event_entries.html
```diff
@@ -17,11 +17,23 @@
...
+                    <input type="submit" value="Remove cohost" class="btn btn-sm btn-danger">
+                </form>
+              {% else %}
+                  <form class="manage-cohost" action="{% url 'manage_co_hosts' %}">
+                    {% csrf_token %}
+                    <input type="text" name="event" value="{{ event.slug }}" hidden>
+                    <input type="text" name="participant" value="{{ participant.id }}" hidden>
+                    <input type="submit" value="Make cohost" class="btn btn-sm btn-success">
+                </form>
+              {% endif %}
```
kshalot: From what I see only the submit button is different here. Maybe the if-statement could be moved deeper to not duplicate the code.

kshalot (changes requested):
I did an initial review. There is a lot more coming up, but I wanted to give you some immediate pointers @Rutvikrj26

physionet-django/notification/utility.py
Rutvikrj26: yup, agreed.
The call is only made twice, and both times provides the value for the attribute `status`

physionet-django/events/views.py
Rutvikrj26: Removed the comments.

physionet-django/events/templates/events/email/event_cohost_cohost_status_change.html
Rutvikrj26: @kshalot This follows a pattern set up by Amit for the rest of the notifications. 
In the context of the pattern, the names make sense. 

That being said, out of context, it does look confusing. I don't have any specific suggestions for the new names.

physionet-django/events/templates/events/email/event_cohost_cohost_status_change.html
Rutvikrj26: Addressed - updated the event_url to be the absolute URL using reverse.

physionet-django/events/templates/events/email/event_host_cohost_status_change.html
Rutvikrj26: I'm changing this one as well to the same event_url to keep it consistent across all the templates.

physionet-django/events/templates/events/event_entries.html
Rutvikrj26: I've made the changes here, but the entire template will have to be changed, as a lot of things have been changed in the inactive_participants, and will need to be changed accordingly in this template.

Rutvikrj26: I've made the changes and moved the if statement further in.

physionet-django/notification/utility.py
```diff
@@ -3,6 +3,7 @@
 """
 from email.utils import formataddr
 from functools import cache
+import re
```
kshalot: It's not something to be addressed in this PR but rather in the entire codebase later, but usually package imports are placed above imports of specific modules `import` vs `from ... import`

physionet-django/events/templates/events/email/event_cohost_cohost_status_change.html
kshalot: We can stick to it for now then to not break the naming pattern.

physionet-django/events/templates/events/email/event_cohost_cohost_status_change.html
```diff
@@ -0,0 +1,15 @@
...
+Dear {{ name }},
+{% if status == 'Make cohost' %}
+You have been added as Cohost to the Event : {{ event_title }}.
+
+You can now manage the event participants from your events dashboard.
+{% elif status == 'Remove cohost' %}
+Your Cohost access has been removed from the Event : {{ event_title }}.
+{% endif %}
+You can view further information about the event using the following link:
+{{ event_url }}
```
kshalot: I'm not a fan of checking the status as strings to decide the logic of the template. It would be better to either have separate templates for each status or check the Enum rather than some specific string.

physionet-django/events/urls.py
```diff
@@ -6,6 +6,7 @@
 urlpatterns = [
     path('', views.event_home, name='event_home'),
     path('create/', views.create_event, name='create_event'),
+    path('manage_co_hosts/', views.manage_co_hosts, name='manage_co_hosts'),
     path('<slug:event_slug>/', views.event_detail, name='event_detail'),
     path('<slug:event_slug>/edit_event/', views.update_event, name='update_event'),
     path('<slug:event_slug>/details/', views.get_event_details, name='get_event_details'),
```
kshalot: The endpoint is a bit out of place in my opinion. I'd rather see some nesting here to make it more RESTful, maybe `<slug:event_slug>/cohosts` rather than passing the slug as a post parameter. It makes it cleaner. That's why is there a `<slug:event_slug>/edit_event/` endpoint rather than `manage_events`

physionet-django/events/views.py
```diff
@@ -238,3 +238,54 @@ def event_detail(request, event_slug):
...
+
+
+@login_required
+def manage_co_hosts(request):
+    """
+    Manage co-hosts of an event
+    """
+    user = request.user
+
+    if request.method == 'POST' and request.META.get('HTTP_X_REQUESTED_WITH') == 'XMLHttpRequest':
```
kshalot: Do we need to check for `XMLHttpRequest`? What's the difference from our standpoint?

physionet-django/events/views.py
```diff
@@ -238,3 +238,54 @@ def event_detail(request, event_slug):
...
+                return JsonResponse({'error': 'User is not a cohost of this event'}, status=403)
+            participant.is_cohost = False
+            participant.save()
+            notification.notify_event_cohost_cohost_status_change(request=request, cohost=participant.user,
+                                                                  event=event, status='Remove cohost')
+            notification.notify_event_host_cohost_status_change(request=request, cohost=participant.user, event=event,
+                                                                status='Remove cohost')
+
+            return JsonResponse({'success': 'Cohost removed successfully'})
+        elif 'Make cohost' in request.POST.get('submit'):
```
kshalot: This logic seems wonky - why would some substring of some parameter decide on the logic? Can't we have something like an "action" attribute that would contain a value like `remove`/`add` that we can deserialize and throw an error if it's wrong?

physionet-django/events/views.py
```diff
@@ -238,3 +238,54 @@ def event_detail(request, event_slug):
...
+        event = get_object_or_404(Event, slug=event_slug)
+
+        if not event.host == user:
+            return JsonResponse({'error': 'You are not the host of this event'}, status=403)
+
+        if event.end_date < datetime.now().date():
+            return JsonResponse({'error': 'You cannot manage co-hosts of an event that has ended'}, status=403)
+
+        if not event.participants.filter(id=participant_id).exists():
+            return JsonResponse({'error': 'User is not a participant of this event'}, status=403)
```
kshalot: The only difference here is the error message. The validation could set some variable like:
```python
if not event.host == user:
  error_message = 'You are not the host of this event'
...

if error_message:
  return JsonResponse({'error': error_message}, status=403)
```

physionet-django/events/views.py
```diff
@@ -238,3 +238,54 @@ def event_detail(request, event_slug):
...
+
+        if 'Remove cohost' in request.POST.get('submit'):
+            if not participant.is_cohost:
+                return JsonResponse({'error': 'User is not a cohost of this event'}, status=403)
+            participant.is_cohost = False
+            participant.save()
+            notification.notify_event_cohost_cohost_status_change(request=request, cohost=participant.user,
+                                                                  event=event, status='Remove cohost')
+            notification.notify_event_host_cohost_status_change(request=request, cohost=participant.user, event=event,
+                                                                status='Remove cohost')
```
kshalot: Passing the status here as a raw string looks very loose - we need some enum value to decide which action we are going to take. Right now if you make a typo anywhere in the codebase, you introduce a bug.

kshalot (changes requested):
Sorry for a late review - I'm leaving some comments. If you need more context then let me know Rutvik, we can even jump on a call.

physionet-django/notification/utility.py
Rutvikrj26: Thanks!
Maybe I can open up a new PR after this one, just cleaning up similar imports.

physionet-django/events/views.py
Rutvikrj26: Amit implemented an AJAX-based request and hence was checking from the backend for the same. Even after removing it from the backend, the page reloads still won't happen - probably due to the request still being sent the same way.

so basically no difference in the functionality.

physionet-django/events/views.py
Rutvikrj26: @kshalot I've currently replaced the strings with the is_participant attribute from the EventParticipant model for addressing one of the earlier comments. 

In case that does not seem like a good idea, we can go with the option you've suggested here. Also, it would be helpful if you could add a bit more detail to the second option.

physionet-django/events/views.py
Rutvikrj26: Implemented the return statement at the end, if the error message exists else the redirect.

physionet-django/events/views.py
Rutvikrj26: Agreed, I've changed it to the actual participant cohost_status.

physionet-django/events/views.py
Rutvikrj26: Oh okay, got you - you're mentioning the "Remove Cohost" string parameter in the request form - which decides whether it is a remove or add button.

kshalot:
@Rutvikrj26 This requires a backmerge/rebase.

physionet-django/events/models.py
```diff
@@ -10,6 +11,11 @@
 from project.validators import validate_version, validate_slug
 
 
+class cohostStatus(IntEnum):
```
kshalot: Conventionally, classes in Python are PascalCase.

```suggestion
class CohostStatus(IntEnum):
```

physionet-django/events/templates/events/event_home.html
```diff
@@ -11,193 +11,255 @@
...
+      e.preventDefault();
+      let $form = $(this);
+      $form.find('input[type="submit"]').prop('disabled', true);
+
+      let participant_id = e.target.participant.value;
+      let event_slug = e.target.event.value;
+      let submit_value = $form.find('input[type="submit"]').val();
+      let action = $form.find('input[type="submit"]').attr('data-action');
+      let csrftoken = e.target.csrfmiddlewaretoken.value;
+      let url = e.target.action;
```
kshalot: We could always prefer `const` over `let`. It's always easier to reason about the code when you cannot change certain values.

physionet-django/events/templates/events/event_home.html
```diff
@@ -11,193 +11,255 @@
...
-{% endblock %}
+<script src="{% static 'custom/js/resize-ck.js' %}"></script>
+<script>
+  $(document).ready(function () {
+    $('.manage-cohost').submit(function (e) {
+      e.preventDefault();
+      let $form = $(this);
+      $form.find('input[type="submit"]').prop('disabled', true);
+
+      let participant_id = e.target.participant.value;
```
kshalot: This is just convention, but the JavaScript community prefers camelCase for JavaScript variables.

physionet-django/events/templates/events/event_home.html
```diff
@@ -11,193 +11,255 @@
...
+          'participant_id': participant_id,
+          'event_slug': event_slug,
+          'csrfmiddlewaretoken': csrftoken,
+          'action': action,
+          'submit': submit_value
+        },
+        success: function (data) {
+          $form.find('input[type="submit"]').prop('disabled', false);
+
+          console.log(data);
```
kshalot: Leftover.

physionet-django/console/templates/console/event_management.html
```diff
@@ -2,90 +2,138 @@
...
+        },
+        success: function (data) {
+
+          $form.find('input[type="submit"]').prop('disabled', false);
+
+          current_action = $form.find('input[type="submit"]').attr('data-action');
+          make_cohost_action = '{{ cohostStatus.MAKE_COHOST }}';
+          remove_cohost_action = '{{ cohostStatus.REMOVE_COHOST }}';
+
+          console.log(current_action, make_cohost_action, remove_cohost_action);
```
kshalot: Leftover.

physionet-django/events/templates/events/event_home.html
```diff
@@ -11,193 +11,255 @@
...
+            $form.find('input[type="submit"]').attr('data-action', remove_cohost_action);
+            $form.find('input[type="submit"]').val('Remove cohost');
+            $form.find('input[type="submit"]').removeClass('btn-primary');
+            $form.find('input[type="submit"]').addClass('btn-danger');
+          } else if (current_action == remove_cohost_action) {
+            $form.find('input[type="submit"]').attr('data-action', make_cohost_action);
+            $form.find('input[type="submit"]').val('Make cohost');
+            $form.find('input[type="submit"]').removeClass('btn-danger');
+            $form.find('input[type="submit"]').addClass('btn-success');
+          }
```
kshalot: Maybe we should split this into multiple functions. Right now it's a big chunk of code, handling success, error and all the independent logic could be function calls that are properly named. This would make the code more self-documenting

physionet-django/events/views.py
```diff
@@ -1,4 +1,6 @@
+from asyncio import events
```
kshalot: I think this is here by mistake since the PhysioNet codebase doesn't use asyncio. I guess the IDE autocompleted this by mistake when you were trying to import the events module.

physionet-django/events/views.py
```diff
@@ -310,3 +314,60 @@ def event_detail(request, event_slug):
          'is_waitlisted': is_waitlisted,
          'event_datasets': event_datasets,
          })
+
+
+@login_required
+def manage_co_hosts(request, event_slug):
```
kshalot: There's a little bit too much logic in this view. I don't think the codebase has a notion of a "service" at this point, but ideally the view should just be parsing the request, dispatching some logic via a service and then preparing the response.

Here, it's parsing the request, handling errors, sending notifications, removing cohosts, adding cohosts etc. It's very hard to reason about this code. Maybe we should think about a separate module where this would be placed.

physionet-django/events/views.py
```diff
@@ -1,4 +1,6 @@
+from asyncio import events
 from datetime import datetime
+import re
```
kshalot: By the way, this is not used anywhere I think.

physionet-django/notification/utility.py
kshalot: Is this import used?

kshalot (changes requested):

physionet-django/events/views.py
Rutvikrj26: Thanks @kshalot  for catching. I'll keep in mind to go through the changes once before requesting review.

physionet-django/events/templates/events/event_home.html
Rutvikrj26: @kshalot - good point. It took me a while to understand the full functionality when I was trying to fix the function. I'm breaking this function down the entire function into smaller ones, each having a single responsibility scope. Please check and let me know if that looks good. 
It should also be self-documenting through names and easy to follow.

physionet-django/events/views.py
Rutvikrj26: As Discussed during the meeting - this needs to be addressed in a new Issue more generic on refactoring of codebase to reduce fat views and have service modules. I will document the discussion briefly and open up a new issue.
Resolving the comment here @kshalot

physionet-django/events/templates/events/event_home.html
```diff
@@ -11,191 +11,281 @@
...
-
 {% block local_js_bottom %}
-  <script>
-  </script>
-  <script src="{% static 'custom/js/resize-ck.js' %}"></script>
-{% endblock %}
+<script src="{% static 'custom/js/resize-ck.js' %}"></script>
+<script>
+  $(document).ready(function () {
+    $('.manage-cohost').submit(function (e) {
```
kshalot: I don't love the fact that this is duplicated. Couldn't we extract the participants list to a separate template so the logic is only in one place?

physionet-django/events/urls.py
```diff
@@ -6,6 +6,7 @@
 urlpatterns = [
     path('', views.event_home, name='event_home'),
     path('create/', views.create_event, name='create_event'),
+    path('<slug:event_slug>/manage_co_hosts/', views.manage_co_hosts, name='manage_co_hosts'),
```
kshalot: In terms of API design, wouldn't it be cleaner to have:
1. Separate endpoints for adding and removing.
or
2. Distinguishing the action based on the method (single endpoint with `PUT` and `DELETE`) handlers.


Then there is no need to define the `CohostStatus` enum at all. Having separate endpoints also makes the view logic more comprehensible since it would handle each case separately (no branching).

kshalot (changes requested):

physionet-django/events/urls.py
Rutvikrj26: I feel the PUT & DELETE will be more cleaner than having a separate endpoint for Add & one for Remove. 
Furthermore, the implementation by Amit is in a way that it's more like a backend view than a REST Endpoint.
I have updated the functionality to use PUT & DELETE for the changes.

Rutvikrj26: Django has inherent support for get & post requests, but not the same for put & delete, and it seems that only these requests are supported by URL testing and hence the tests for put & delete are failing.

physionet-django/events/templates/events/event_home.html
Rutvikrj26: > extract the participants list to a separate template

Not sure what you mean - as the part you've highlighted is the Ajax call for one participant and doesn't have anything to do with a list of participants.

physionet-django/events/templates/events/event_applications.html
```diff
@@ -1,27 +1,139 @@
...
+        data: JSON.stringify(formData),
+        beforeSend: function(xhr) {
+          xhr.setRequestHeader('X-CSRFToken', formData.csrfmiddlewaretoken);
+        },
+        success: function (data) {
+          handleSuccess($form, data);
+        },
+        error: function (data) {
+          handleError($form, data);
+          console.log(data);
```
kshalot:
```suggestion
```

I think that's a leftover.

kshalot (changes requested):

physionet-django/events/views.py
```diff
@@ -310,3 +314,46 @@ def event_detail(request, event_slug):
...
+
+@require_http_methods(["PUT", "DELETE", "GET", "POST"])
+def manage_co_hosts(request, event_slug):
+    """
+    Manage co-hosts of an event
+    """
+    user = request.user
+    error_message = None
+
+    body = json.loads(request.body)
```
kshalot, line 341: @Rutvikrj26 here's the line that parses the body. It parses it regardless of the HTTP method, even though the views specified the following: `["PUT", "DELETE", "GET", "POST"]`.

physionet-django/events/views.py
```diff
@@ -310,3 +314,46 @@ def event_detail(request, event_slug):
...
+    participant_id = body.get('participant_id')
```
kshalot, line 342: Now that I think of it, a `DELETE` call shouldn't contain a body either (it's semantics are undefined). Maybe this endpoint should be redesigned after all.

kshalot (changes requested):

physionet-django/events/views.py
Rutvikrj26: Yup, Agreed.
But, redesigning the endpoint - as in completely re-designing or breaking the current into two different endpoints - add & remove?

Rutvikrj26:
@kshalot @tompollard 
I've implemented single endpoints for Add & Remove co-hosts. Please review and let me know your thoughts.

tompollard:
Sorry to say this so late in the day, but I feel like an "invite co-host" button somewhere might be cleaner (both for user and implementation-wise).

Rutvikrj26:
As discussed during the regular Standup, this feature implementation needs to be redone for security issues.
I'll be opening a new PR and implement an invite cohost button on the main event page.

PR #1928, opened 2023-03-10
Handle bucket root correctly when copying GCS objects

kshalot:
Opening this PR for @rafalkosla101.

Wojtek found that creating a new version of a project will eventually fail when using the GCS backend, because the `LICENSE.txt` file is not ignored properly. This is because we try to ignore `/LICENSE.txt` while the file is called `LICENSE.txt` (note the leading slash).

This is the same edge case inside the `GCSObject` integration as before and before. "Folders" inside the buckets must contain a trailing slash, e.g. "test/" is a folder while "test" is a file. The bucket root is an empty string, so marking it as a folder ends up with "/" which then cannot be used as a prefix for objects because prepending it creates a leading slash in the path which is never correct.

Changing this now risks breaking something more critical, so we just extracted the fix logic to a property and added it to the `_cp_dir` method.

### DIFF
--- a/physionet-django/physionet/gcs.py
+++ b/physionet-django/physionet/gcs.py
@@ -66,6 +66,11 @@ class GCSObject:
         """Return the name"""
         return self._object_name
 
+    @property
+    def local_name(self):
+        """Return the relative name inside the bucket"""
+        return '' if self.name == '/' else self.name
+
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks Karol, looks good to me.

PR #1929, opened 2023-03-10
Event: Hide dataset section unless event has dataset

### LINKED ISSUE #1926: Remove "Datasets" section from the Event Details page if no datasets are selected
Right now, there is a section on the Event Detail page (e.g. https://staging.physionet.org/events/lEIQjniyUEoU/) titled "Datasets". This section is intended to display datasets that have been made available to participants in the event.

Screenshot 2023-03-09 at 11 34 41 AM

In general we do not want to display the "Datasets" section. It isn't relevant to most people. We should hide the section unless a dataset has been added to the event.

superryeti:
Closes https://github.com/MIT-LCP/physionet-build/issues/1926

**Context:**
As discussed on https://github.com/MIT-LCP/physionet-build/issues/1926, currently we show the dataset section on Event detail page even when there is no dataset attached to the event.
The Event app can be used by instructor in other ways where they might not need to use any dataset, so as suggested by @tompollard we shouldn't display the dataset section unless datasets are used on Event.

### DIFF
--- a/physionet-django/events/templates/events/event_detail.html
+++ b/physionet-django/events/templates/events/event_detail.html
@@ -53,8 +53,9 @@
 
     {% include  "events/event_participation.html" %}
      <hr>
-    {% include  "events/event_datasets.html" %}
-
+    {% if event_datasets|length != 0 %}
+      {% include  "events/event_datasets.html" %}
+    {% endif %}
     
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks Amit!

PR #1932, opened 2023-03-13
refactor events app code to follow pep guidelines

superryeti:
Context:
Previously on some of the PRs, the code was using backslash in multiple places for events app. It looks like PEP recommends to use parenthesis over backslash. 

https://peps.python.org/pep-0008/#multiline-if-statements
https://doc.casthighlight.com/alt_multilineinstruction-backslash-line-continuation-makes-code-less-readable/#:~:text=This%20is%20helpful%20in%20some,use%20parentheses%20around%20your%20elements.

Thanks to @tompollard  for pointing this out.


tompollard:
thanks amit!

PR #1935, opened 2023-03-14
add token/secret to credential_reference URLs

### LINKED ISSUE #1933: credential_reference URLs should be secret
When a credentialing application is submitted, the reviewer has the option to send an automated message to the applicant's reference (see `notification.utility.contact_reference`).

This message instructs the recipient to visit a URL such as

http://localhost:8000/credential-reference/3niq5ZuHnhhdtSX2qHlU/

This URL (see `user.views.credential_reference`) does nothing to check who is actually submitting the form.  Anyone who knows this URL can claim to be the applicant's reference and verify (or deny) the application.

This isn't as bad as it sounds, because there isn't any way I can see for the applicant or other unauthorized people to know the slug of their application.  Still, these slugs are visible to site administrators, and it'd be quite easy to make the mistake of making the slug visible to the applicant somewhere (in general we don't use the word "slug" to refer to a secret token.)

Instead, the CredentialApplication should have a secret "verification_token" distinct from the "slug".  In the email message (`notification/email/contact_reference.html`), the URL should include this verification token instead of (or in addition to) the slug.  Then, either:

(a) define a new function similar to `user.views.credential_reference` which requires a verification token; after a month, remove the old function

(b) modify `user.views.credential_reference` to optionally accept a verification token; after a month, make the token mandatory

Note that if the reference has a PhysioNet account then there are better ways to handle this.  But we cannot assume that and need to address the general case.

superryeti:
closes https://github.com/MIT-LCP/physionet-build/issues/1933

**Context:** 
As discussed in https://github.com/MIT-LCP/physionet-build/issues/1933, currently there is a chance that site administrators might accidentally make the credential app slug visible to the applicant which can let them(applicant or someone who is not the reference) submit the reference verification themselves. To solve this, we had discussed to add a token to the credential reference response url, so that even somehow someone knows the slug, they(participant or someone else) wont be able to submit the reference response.

**Changes:**
Here i added a new field for verification token to `CredentialApplication`. This token is generated when a user submits the credential application. For existing applications(pending applications, where reference has not responded), the token is generated with a `migrate_forward` method when migration is applied to database.
For the credential verification i added a new view `user.views.credential_reference_verification`  which is similar to the old view `user.views.credential_reference`.

**TODO**
After a month, we can remove the existing view  `user.views.credential_reference` that does the  credential reference verification for the cases where reference request has already been sent, The assumption is that in 1 month the reference will have plenty of time to respond. Currently i added a new method and didn't modify the existing view `user.views.credential_reference` because its kind of easier to later just remove the old `view`


bemoody:
Thanks!

In physionet-django/user/forms.py:
```
        token = get_random_string(32)
        while CredentialApplication.objects.filter(reference_verification_token=token):
            token = get_random_string(32)
        credential_application.reference_verification_token = token
```
Checking for duplicate tokens might be slow, and I'm not sure we care about duplicates since the credential_reference_verification URL includes both the slug and token.  So we could probably get rid of the while loop here, as well as in the migration.

If we do want to ensure uniqueness then the field should be defined with `unique=True` (which also makes the `filter` operation faster.)

superryeti:
> nsure uniqueness then

Thank you @bemoody . Thats a very good point about not needing to make the token unique, since we have slug + token. 
And since the token is 32 chars long(made from upper and lower string chars, and 0-9), we will most likely never see same token appearing twice.

My vote is to not use the unique=True on token because 
1. Like you mentioned , we really don't care it its unique
2. In a weird case if the token gets duplicated, the user will see an error when submitting the credential application with no fault of their own.

Please let me know what you think.

bemoody:
This looks great.

One small issue is that this doesn't add verification tokens for credential applications in demo-user.json.  And updating the demo-*.json files is a real mess, which we ought to sort out.  But for a quick and dirty solution, we could just run the following command:
```
sed '/"reference_response_text"/a\    "reference_verification_token": "rJ2i7vlzh6AgZ1Wwtcz8zCoI5BqxH0kU",' -i.orig user/fixtures/demo-user.json
```

superryeti:
> This looks great.
> 
> One small issue is that this doesn't add verification tokens for credential applications in demo-user.json. And updating the demo-*.json files is a real mess, which we ought to sort out. But for a quick and dirty solution, we could just run the following command:
> 
> ```
> sed '/"reference_response_text"/a\    "reference_verification_token": "rJ2i7vlzh6AgZ1Wwtcz8zCoI5BqxH0kU",' -i.orig user/fixtures/demo-user.json
> ```

Thank you @bemoody. i updated the PR

Note to self

for Mac, here is the replace code

```
sed -i .orig '/"reference_response_text"/a\
    "reference_verification_token": "rJ2i7vlzh6AgZ1Wwtcz8zCoI5BqxH0kU",
' user/fixtures/demo-user.json

```

bemoody:
Looks great.  Thanks!

Issue #1937, opened 2023-03-20
CITI API client workflow 


lamawmouk:
As discussed, we want to obtain information from the CITI API to be able to automate training report checks. This issue describes the workflow for the API call, which will enable the extraction of specific data from the XML response:

1) Making a SOAP request to a web service, parsing the XML response, and extracting specific information from the response. It will involve sending a SOAP request to a web service at a specified URL, passing in specific user credentials and email information. The response will be parsed using the XML parsing library, e.g. `ElementTree`. The codebase will extract the email and member ID from the parsed XML response. Later, it will set the URL to the endpoint of the SOAP web service that is going to be called: url = "https://webservices.citiprogram.org/SOAP/CITISOAPService.asmx".
It will then set up the XML payload that will be sent to the web service as part of the SOAP request. This payload will include the user's credentials and the email information. The request will be made using the requests library, with the POST method being used to send the SOAP request to the specified URL, where the headers and payload will be passed in as arguments. The code will then extract the email and member ID (primary key) from the parsed XML response to link to other report information.

2) Making a SOAP request to a web service to retrieve information on CITI program courses for a specific member ID (primary key). The XML payload will be generated and will include the member ID variable. The payload will request data about a member's courses based on their ID. The XML response will then be parsed, and specific data will be extracted by finding certain elements. The extracted data will include course names, scores, expiration dates, and course IDs.

3) Setting up the configuration of the CITI API client to integrate it on PhysioNet.

PR #1938, opened 2023-03-21
CITISOAPSERVICE configuration

lamawmouk:
As discussed with @bemoody and given the CITI API workflow in #1937, this PR will create the configuration to set up the credentials of the CITI SOAP SERVICE on the PhysioNet platform.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -167,3 +167,10 @@ DEFAULT_NUMBER_OF_APPLICATIONS_TO_REMIND = 5
 
 # minimum number of word needed for research_summary field for Credentialing Model.
 MIN_WORDS_RESEARCH_SUMMARY_CREDENTIALING = 20
+
+# CITISOAPService API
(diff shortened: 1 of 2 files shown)


bemoody:
Please add more explanation in `.env.example`.

The comment in .env.example should say something about *what* this username/password are used for and *why*.

The comment should also say something about how to obtain such credentials (link to public documentation from CITI if possible).

Most importantly, the comment should explain how to change or revoke the credentials in case of a security breach.

All of the secrets that are currently stored in .env *should* be changed regularly.  (None of them currently *are* being changed regularly AFAIK, which is a problem.  But at a bare minimum we need to document the process for doing so.)

bemoody:
Also, leave these values blank in .env.example so that the feature is not enabled by default in devlelopment configurations.

lamawmouk:
Thank you @bemoody! I have updated the changes.

alistairewj:
Merged in #1964

Issue #1921, opened 2023-03-07
Avoid in-place pip upgrade


bemoody:
Currently packages are installed and upgraded in `/physionet/python-env/physionet` by invoking `/physionet/python-env/physionet/bin/pip3 install ...`.  (See `deploy/update`.)

Some of the many problems with this:
- packages removed from `requirements.txt` don't get removed
- no guarantee in general that upgrading will give reproducible results
- no reliable way to roll back to a previous version
- pip is implemented in python and therefore installed packages can screw up pip
- installation can fail for a million reasons, in which case we'd prefer the django server to continue using the old working libraries rather than the new partially-installed ones

Instead it would be preferable to install packages from scratch every time (the way we do every time we run the test suite.)

Some potential obstacles to doing that:
- if requirements.txt isn't changed, we don't want to waste time installing stuff every time we deploy to the server
- if pypi.org is down, we still want to be able to deploy changes (assuming those changes don't change requirements.txt)
- a virtualenv prefix can never be moved or renamed

A possible solution:
- Generate a hash of the requirements.txt file, e.g. `45454a9fdba6595291c93265e1b6c2cc`.
- If `/physionet/python-env/45454a9fdba6595291c93265e1b6c2cc` already exists, then simply create a symlink `/physionet/python-env/physionet -> 45454a9fdba6595291c93265e1b6c2cc`.
- Otherwise, create a new virtualenv prefix there and install the packages from requirements.txt, then create the symlink.
- (Some magical process to automatically delete prefixes that are no longer used.)

bemoody:
> If /physionet/python-env/45454a9fdba6595291c93265e1b6c2cc already exists, then simply create a symlink /physionet/python-env/physionet -> 45454a9fdba6595291c93265e1b6c2cc.

might work but has some odd effects, just as renaming does.

Basically, in this case, scripts installed by virtualenv or pip (such as `pip` itself, `django-admin`, or the `activate` file) will hardcode the original path (`/physionet/python-env/45454a9fdba6595291c93265e1b6c2cc`).

But if python3 is invoked as `/physionet/python-env/physionet/bin/python3`, then `/physionet/python-env/physionet` is what gets stuck into `sys.path`.  I think thhe same is true of the python interpreter invoked by uwsgi (see /etc/uwsgi/vassals/physionet_uwsgi.ini)

[...]

PR #1939, opened 2023-03-21
[WIP] Re-install python libraries rather than upgrading

bemoody:
Change the git deployment process to install pip packages by *re-installing* into a new directory each time, rather than trying to *upgrade* in place.

See issue #1921.  There are some potential problems with this (virtualenv is *ugly*, yo) and it needs thorough testing.

TODO:

- `install-pn-test-server` won't set up the initial prefix the same way `update` does, so the first push will always force a reinstall.  Should it?

- push options don't actually work with `update`?!  (also, push options are disabled by default)

### DIFF
--- a/deploy/test-server/install-pn-test-server
+++ b/deploy/test-server/install-pn-test-server
@@ -114,8 +114,9 @@ su pn -c '
    echo staging > deploy-branch
    mkdir -p physionet-build python-env deploy
 
-   virtualenv --no-download -ppython3 /physionet/python-env/physionet
-   . /physionet/python-env/physionet/bin/activate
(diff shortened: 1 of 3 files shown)


bemoody:
This writes a somewhat amazing amount of junk to /data/log/pn/update.log (about 10 MB each time.)  Maybe better to log the pip stuff elsewhere.

Also, why does it recompile psycopg every time?  I mean, yeah, that is kind of what I asked for, but I thought pip had a cache specifically to avoid that kind of thing.  What is pip's cache for if not that?

tompollard:
@bemoody is this still a WIP or ready for review?

PR #1940, opened 2023-03-21
CITI class to store account information

lamawmouk:
This PR will create the CITI class (a Django model) that creates a database table to store CITI account information.

The CITI class has five fields: user, member_id, course_id, course_name, course_expiration, and report_score. The user field is a one-to-one relationship with Django's built-in User model and is used to associate each CITI account with a specific user. The remaining fields correspond to the information stored in a CITI account, including the member ID, course ID, course name, course expiration date and time, and report score. These fields are defined as CharField and DateTimeField with appropriate lengths and validators to ensure that the data stored in the fields adhere to the expected format.


alistairewj:
Merged in #1964

Issue #1915, opened 2023-03-03
cryptography / pyopenssl / pip incompatibility    


bemoody:
The combination of these packages breaks pip (see pull #1911):
```
$ rm -rf ve1
$ virtualenv ve1
$ ./ve1/bin/pip install --no-deps pyopenssl==19.1.0 \
                                  cffi==1.15.1 \
                                  cryptography==39.0.1
$ ./ve1/bin/pip freeze
[...]
AttributeError: module 'lib' has no attribute 'X509_V_FLAG_CB_ISSUER_CHECK'
```
Omit any one of the three and it works.

tompollard:
Thanks for getting to the bottom of this! First time I've seen something like this happen/

tompollard:
@bemoody this remains an issue, but not for us. Can we close or is there a benefit in keeping it open?

bemoody:
Yep, this was addressed by #1953.

PR #1947, opened 2023-03-23
Sanity check pip in github workflow

bemoody:
Installing packages using pip can break pip for weird reasons (see issue #1915).

Add a sanity check to catch seriously broken cases before they're merged.  This may be unnecessary if we stop doing pip upgrades (see issue #1921 / pull #1939) but isn't a bad idea anyway.


tompollard:
thanks, looks good to me!

PR #1948, opened 2023-03-23
Add Django redirects app to allow URLs to be redirected. Fixes #1946

### LINKED ISSUE #1946: Add Django redirects app to allow URLs to be redirected
physionet.org has existed for a long time! We've tried to persist URLs, but we often find that papers, github repositories, websites, etc include broken links (sometimes links to pages that were deleted at some point in the past 20 years, sometimes simply links with typos). 

It would be helpful to have an approach for creating redirects when we discover a broken link. @bemoody pointed out that Django has a built in app for this: https://docs.djangoproject.com/en/4.1/ref/contrib/redirects/

> The app creates a django_redirect table in your database. This is a lookup table with site_id, old_path and new_path fields. The [RedirectFallbackMiddleware](https://docs.djangoproject.com/en/4.1/ref/contrib/redirects/#django.contrib.redirects.middleware.RedirectFallbackMiddleware) does all of the work. Each time any Django application raises a 404 error, this middleware checks the redirects database for the requested URL as a last resort. Specifically, it checks for a redirect with the given old_path with a site ID that corresponds to the [SITE_ID](https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-SITE_ID) setting.

Redirects can be added in two ways:

> - If you’ve activated the automatic Django admin interface, you should see a “Redirects” section on the admin index page. Edit redirects as you edit any other object in the system.
> - via the shell: 
```python
>>> from django.conf import settings
>>> from django.contrib.redirects.models import Redirect
>>> # Add a new redirect.
>>> redirect = Redirect.objects.create(
...     site_id=1,
...     old_path='/contact-us/',
...     new_path='/contact/',
... )
> - >>> # Change a redirect.
>>> redirect.new_path = '/contact-details/'
>>> redirect.save()
>>> redirect
<Redirect: /contact-us/ ---> /contact-details/>
>>> # Delete a redirect.
>>> Redirect.objects.filter(site_id=1, old_path='/contact-us/').delete()
(1, {'redirects.Redirect': 1})
```

tompollard:
We often find the external websites, articles, etc, include links to non-existent URLs on the physionet.org domain. This pull request adds the Django Redirects App, allowing us to set up URL redirects (addressing #1946). 

The changes are:

1) Add the Redirects App, following instructions at: https://github.com/MIT-LCP/physionet-build/issues/1946
2) Add a test to confirm that setting up a redirect results in URLs being redirected.
3) Display a new page in the console that lists redirects: http://localhost:8000/console/redirects/. The page is listed under the "Pages" item in the console menu. Not an ideal place, but the best I could think of. The whole menu needs a clean up.
4) Grant the admin user the permission necessary to view the redirects page in the console (`redirects.redirect.view_redirect`).

Couple of notes:

- We will need to add the new permission (`redirects.redirect.view_redirect`) to relevant permissions groups if/once the change is merged.
- For now, redirects will need to be added via the shell using the syntax below:

```
from django.contrib.redirects.models import Redirect
# Add a new redirect.
redirect = Redirect.objects.create(site_id=settings.SITE_ID,
                                   old_path='/contact-us/',
                                   new_path='/contact/')
```

### DIFF
--- a/physionet-django/console/templates/console/console_navbar.html
+++ b/physionet-django/console/templates/console/console_navbar.html
@@ -314,7 +314,7 @@
       {% endif %}
 
       <!-- Pages -->
-      {% if perms.user.change_staticpage or perms.user.change_frontpagebutton %}
+      {% if perms.user.change_staticpage or perms.user.change_frontpagebutton or perms.redirect.view_redirect %}
       <li class="nav-item" data-toggle="tooltip" data-placement="right">
         {% if frontpage_buttons_nav or static_pages_nav %}
           <a id="nav_pages_dropdown" class="nav-link nav-link-collapse drop" data-toggle="collapse" href="#pagesComponent" data-parent="#sideAccordion" aria-expanded="true">
@@ -325,16 +325,19 @@
             <span class="nav-link-text">Pages</span>
           </a>
           <!-- submenu -->
-        {% if frontpage_buttons_nav or static_pages_nav %}
+        {% if frontpage_buttons_nav or static_pages_nav or redirects_nav %}
           <ul class="sidenav-second-level collapse show" id="pagesComponent" style="">
         {% else %}
           <ul class="sidenav-second-level collapse" id="pagesComponent">
         {% endif %}
             <li class="nav-item {% if static_pages_nav %}active{% endif %}">
-              <a id="nav_console_news" class="nav-link" href="{% url 'static_pages' %}">Static Pages</a>
+              <a id="nav_pages_dropdown" class="nav-link" href="{% url 'static_pages' %}">Static Pages</a>
             </li>
             <li class="nav-item {% if frontpage_buttons_nav %}active{% endif %}">
-              <a id="nav_console_news" class="nav-link" href="{% url 'frontpage_buttons' %}">Frontpage Buttons</a>
+              <a id="nav_pages_dropdown" class="nav-link" href="{% url 'frontpage_buttons' %}">Frontpage Buttons</a>
+            </li>
+            <li class="nav-item {% if redirects_nav %}active{% endif %}">
+              <a id="nav_pages_dropdown" class="nav-link" href="{% url 'redirects' %}">Redirects</a>
             </li>
           </ul>
       </li>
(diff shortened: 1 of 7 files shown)


bemoody:
The Site stuff is weird.  IIUC, this just ignores the Host (because SITE_ID is set)?  (So following your example it would return the same response for `https://physionet.org/contact-us/`, `https://alpha.physionet.org/contact-us/`, etc.?)

tompollard:
I can look into it, but setting the `site_id` seemed necessary.

If you update `site_id` in physionet-django/physionet/test_urls.py to something other than `settings.SITE_ID`, then the redirect will not work. 

e.g. if `site_id` is set to 1 (alpha.physionet.org) in the following chunk, the second `response.status_code` will return a 404:

```
    def test_redirect(self):
        """
        Creates and tests a redirect.
        """
        site_id = settings.SITE_ID
        old_path = '/redirect/me/'
        new_path = '/news/'

        response = self.client.get(old_path)
        self.assertEqual(response.status_code, 404)

        self._add_redirect(site_id=site_id, old_path=old_path, new_path=new_path)

        response = self.client.get(old_path)
        self.assertEqual(response.status_code, 301)
```

tompollard:
(@bemoody Let me know if I missed your point...re-reading, I wonder if I did?)

bemoody:
Looks good!

In the test case:
```
+    def test_redirect(self):
+        """
+        Creates and tests a redirect.
+        """
+
+        ...
+
+        redirect = Redirect.objects.get(old_path=old_path)
+        redirect.delete()
```
There is no need to delete objects here; Django's TestCase rolls back all database changes after running each test case.  This class also doesn't need to inherit from TestMixin, just TestCase.  (We ought to rename TestMixin...)

tompollard:
Thanks @bemoody. I have updated the PR based on your comments above:

> There is no need to delete objects here; Django's TestCase rolls back all database changes after running each test case.

Removed the lines that deleted the objects.

> This class also doesn't need to inherit from TestMixin, just TestCase

Now inheriting from TestCase only.

PR #1949, opened 2023-03-27
Feature: Add privacy policy consent into registration form

rafalkosla101:
Added privacy policy consent for registration form and provided a link to those resources configurable via env variables. The env variable is needed since we are creating privacy policy as a backend static page that may change its name etc which would occur in wrong data.

image


rafalkosla101:
@tompollard Should we move the whole privacy policy customisation to the env variable?

tompollard:
> @tompollard Should we move the whole privacy policy customisation to the env variable?

Thanks Rafal. Hmm, yes, I think this would be good if you don't mind making the change.

.env.example
```diff
@@ -102,6 +102,7 @@ EMAIL_SIGNATURE="Regards, The Development Team"
 FOOTER_MANAGED_BY="DataShare Lab"
 FOOTER_SUPPORTED_BY="people"
 FOOTER_ACCESSIBILITY_PAGE=
+PRIVACY_POLICY_MESSAGE="I consent to the collection and use of my personal information. Without your consent, we can not create an account"
```
alistairewj: Add comment above providing example of using HTML tag

physionet-django/user/forms.py
```diff
@@ -298,6 +299,7 @@ class RegistrationForm(forms.ModelForm):
                     widget=forms.TextInput(attrs={'class': 'form-control'}),
                     validators=[validate_name])
 
+    privacy_policy = forms.BooleanField(required=True, label=mark_safe(settings.PRIVACY_POLICY_HTML))
```
tompollard: Could we wrap this in a `if privacy_policy` clause, so that platforms have the ability to switch the feature off?

tompollard (commented):
Thanks @rafalkosla101, looks good to me. I think it would be nice if there was the option to switch this feature off (suggested changes inline).

physionet-django/physionet/settings/base.py
```diff
@@ -31,6 +31,7 @@
 ENABLE_LIGHTWAVE = config('ENABLE_LIGHTWAVE', default=True, cast=bool)
 SSO_REMOTE_USER_HEADER = config('SSO_REMOTE_USER_HEADER', default='HTTP_REMOTE_USER')
 SSO_LOGIN_BUTTON_TEXT = config('SSO_LOGIN_BUTTON_TEXT', default='Login')
+PRIVACY_POLICY_HTML = config('PRIVACY_POLICY_HTML')
```
tompollard: Could we add some a default=None (or similar) here, and then omit the checkbox on the registration page in this case?

tompollard:

PR #1950, opened 2023-03-28
create training basics  --> Add create training from admin console by uploading json

superryeti:
## What?
Here i have added a feature that will allow admin to create training from admin console by uploading json file. This also lays foundation to implement take training feature and resume training feature for user later.

## Why?
A training feature was desired by HDN so that we can build/customize a onplatform training for users which will ultimately be used to control access to projects.
The new onplatform training will work just like the existing Trainings and work with projects without any additional changes,  Currently users take training outside the platform and then upload the certificate to the platform.

## How?
We created a separate app called Training. This app will be used to create/manage training courses, and allow users to take onplatform training.

We created two major types of Models in  training.models

**1. Platform Training (`training.models.OnPlatformTraining`)**

The idea here is that a new Training Course will be defined with `user.models.TrainingType`. This is the ultimate(top) model for a training course. 

The training content for `TrainingType` model will be implemented by the new training app. On Training app, `OnPlatformTraining` instance can be created for each TrainingType. For different versions of the same training course, we can create as many `OnPlatformTraining` models as we want as long as the version is different.

A training is divided into modules. Each module has a description and a list of contents and quizzes. Modules are like chapters in a book. Each module has a list of contents and quizzes. Contents are like paragraphs in a chapter. Quizzes are like questions in a chapter. Contents and quizes can be organized in any order which is controlled by `order` field. For ordering the modules, contents and quizzes, we have used `order` field. The ordering is unique for each instance of the parent model. 

For example : Under Module 1, we have content 1(order=1), content 2(order=2), content 3(order=4) and quiz 1(order=3). Then the ordering of the contents and quizzes will be content 1, content 2, quiz 1, content 3.

It is expected that the order will start from 1 and will be incremented by 1 with no gaps.

**2. Tracking User Progress during training** (`training.models.OnPlatformTrainingProgress`)

When a user starts a training, a `OnPlatformTrainingProgress` instance should be created for that user and the version of the training type that they are taking. This model tracks the progress of the user during the training. Similarly, when a user starts a module, a `ModuleProgress` model should be created for each module in the training. For quiz, content progress, we should create a instance of `CompletedContent` or `CompletedQuiz` model when the user completes a content or quiz.

I don't think we need to track when someone started a content or quiz as these are expected to complete in few minutes.

## Testing?

Next PR https://github.com/MIT-LCP/physionet-build/pull/1951

### How to create a training from admin console?

1. Login as admin
2. Navigate to `Training` tab in admin console
3. Click on `Create/Update OP Training` button
4. Select `Create New Training` from the dropdown
5. Upload the example json from `training/fixtures/example-training-create.json`

### How to edit a training from admin console?

1. Follow steps 1-3 from above
2. Select Existing Training from the dropdown
3. Upload the example json from `training/fixtures/example-training-update.json`

Notes:
This is the first part of Training PR, there is another PR https://github.com/MIT-LCP/physionet-build/pull/1951 which is rebased on this PR.


physionet-django/console/templates/console/training_type/index.html
```diff
@@ -0,0 +1,74 @@
...
+
+{% block title %}Training Types{% endblock %}
+
+{% block content %}
+ <div class="card mb-3">
+  <div class="card-header">
+    Training Types <span class="badge badge-pill badge-info">{{ training_types|length }}</span>
+  </div>
+  <div class="card-body">
+    <div><button type="button" class="btn btn-sm btn-danger" data-toggle="modal" data-target="#op-training">Create/Update On Platform Training</button></div>
```
tompollard: `btn-danger` doesn't seem right! just a standard green button is fine.

physionet-django/console/templates/console/training_type/index.html
```diff
@@ -0,0 +1,74 @@
...
+    <div class="table-responsive">
+      <table class="table table-bordered">
+        <thead>
+          <tr>
+            <th>Name</th>
+            <th>Valid Duration</th>
```
tompollard, line 33: Perhaps I'll understand this after seeing the whole pull request, but the meaning of "valid duration" isn't clear to me here. e.g. what does it mean for a training course to be valid for 1 week?

physionet-django/console/templates/console/training_type/index.html
```diff
@@ -0,0 +1,74 @@
...
+            <th>On Platform Training</th>
+          </tr>
+        </thead>
+        <tbody>
+        {% for training in training_types %}
+            <tr>
+                <td>{{ training.name|title }}</td>
+                <td>{{ training.valid_duration }}</td>
```
tompollard: For me this is rendered as `1095 days, 0:00:00`. Should we filter to just "X days"?

physionet-django/console/templates/console/training_type/index.html
```diff
@@ -0,0 +1,74 @@
...
+  </div>
+  <div class="card-body">
+    <div><button type="button" class="btn btn-sm btn-danger" data-toggle="modal" data-target="#op-training">Create/Update On Platform Training</button></div>
+    <div class="table-responsive">
+      <table class="table table-bordered">
+        <thead>
+          <tr>
+            <th>Name</th>
+            <th>Valid Duration</th>
+            <th>On Platform Training</th>
```
tompollard: What is the purpose of the "On Platform Training" column? Can you give an example of the kind of non-on-platform training that would be listed on this page?

physionet-django/console/templates/console/training_type/index.html
```diff
@@ -0,0 +1,74 @@
...
+                        <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
+                    </div>
+            </div>
+            </form>
+        </div>
+    </div>
+    
+  </div>
+</div>
+{% endblock %}
```
tompollard: Missing newline at end of file.

physionet-django/notification/utility.py
```diff
@@ -1025,3 +1025,19 @@ def notify_event_participant_application(request, user, registered_user, event):
     body = loader.render_to_string('events/email/event_registration.html', context)
     # Not resend the email if there was an integrity error
     send_mail(subject, body, settings.DEFAULT_FROM_EMAIL, [user.email], fail_silently=False)
+
+
+def notify_users_of_training_expiry(user, training, expiry):
```
tompollard: I see, so the idea is that the training expires after a fixed period, and then it needs to be retaken?

physionet-django/static/sample/create.json
```diff
@@ -0,0 +1,23 @@
...
+        "choices": [
+          {
+            "body": "string",
+            "is_correct": "boolean"
+          }
+        ]
+      }
+    ]
+  }
+}
```
tompollard: Missing newline at end of file.

physionet-django/console/templates/console/training_type/index.html
tompollard: I'm not keen on phrase "On Platform Training". It's awkward to read and write, and I'm not sure it's necessary. 

I think probably just "training courses" is fine. We can discuss if we need to create a clear distinction between courses hosted on PhysioNet and external courses.

physionet-django/console/templates/console/training_type/index.html
```diff
@@ -0,0 +1,74 @@
...
+                <div class="modal-content">
+                    <div class="modal-body">
+                        <p>Download Sample Json File by <a href="{% static 'sample/create.json' %}" download>Clicking Here</a></p>
+                        <div>
+                            {% csrf_token %}
+                            <div class="form-group">
+                                <label>Training Type: </label>
+                                <select name="training_id" class="form-control" required>
+                                    <option disabled>Choose...</option>
+                                    <option value="-1">Create new training</option>
```
tompollard: In places like this, I think "course" is clearer than "training". e.g. "Create new course" or "Create new training course". There's something about "training" being uncountable (and countable?) that can make it a confusing word to use.

physionet-django/static/sample/create.json
```diff
@@ -0,0 +1,23 @@
+{
+  "op_trainings": {
+    "version": "float",
+    "contents": [
+      {
+        "body": "string",
+        "order": "integer"
+      }
+    ],
+    "quizzes": [
```
tompollard: For me, "quizzes" doesn't seem quite right here. Perhaps "Questions"?

physionet-django/static/sample/create.json
```diff
@@ -0,0 +1,23 @@
...
+      {
+        "question": "string",
```
tompollard: Should we support different types of questions, e.g. multiple_choice, short_answer, etc? Maybe not necessary?

physionet-django/training/fixtures/example-training-create.json
```diff
@@ -0,0 +1,47 @@
...
+                "body": "I am a choice4",
+                "is_correct": false
+              }
+            ]
+          }
+        ]
+      }
+    ]
+  }
+}
```
tompollard: newline

physionet-django/training/models.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+class QuizChoice(models.Model):
+    quiz = models.ForeignKey('training.Quiz',
+                             on_delete=models.CASCADE, related_name='choices')
+    body = models.TextField()
+    is_correct = models.BooleanField('Correct Choice?', default=False)
+
+
+class OnPlatformTrainingProgress(models.Model):
```
tompollard: As above, I don't think we want to be using the term "On Platform". Probably just `TrainingProgress` is fine here.

physionet-django/training/models.py
```diff
@@ -0,0 +1,142 @@
...
+    class Status(models.TextChoices):
+        IN_PROGRESS = 'IP', 'In Progress'
+        COMPLETED = 'C', 'Completed'
+
+    user = models.ForeignKey('user.User', on_delete=models.CASCADE)
+    training = models.ForeignKey('training.OnPlatformTraining', on_delete=models.CASCADE)
+    status = models.CharField(max_length=2, choices=Status.choices, default=Status.IN_PROGRESS)
+    started_datetime = models.DateTimeField(auto_now_add=True)
```
tompollard: In the comment, I think you mentioned that we don't want to capture the start time, or did I misunderstand? Anyway, I do think we should capture the start time, because it is interesting/helpful to know how long it takes people to complete the course.

tompollard (commented):
I've taken a quick look and added some comments. I didn't get all of the way through the changes, so will need to take another look later. It would be good to chat more about the approach, perhaps at our Tuesday meeting.

physionet-django/console/templates/console/training_type/index.html
superryeti: From what i understand, the valid duration was a part of the `user.models.TrainingType`, so i think it might have been Harry's idea that some trainings might  need to be retaken after certain timeframe.

physionet-django/console/templates/console/training_type/index.html
superryeti: I agree, we should be removing this. Because the tab on console says "OP Training", i cant think of a case where, we would have a non OP training here

physionet-django/notification/utility.py
superryeti: Yes, the training would expire and users would be unable to access the projects/data and would need to retake the training.

physionet-django/static/sample/create.json
superryeti: Good point, We can support multiple choice questions right now if we want(because same model should work for it). 

tagging @alistairewj to discuss about support for short answers, and if something else also?

physionet-django/training/models.py
superryeti: Sorry for the confusion, We do capture the start time for a course and each module inside the course, but we donot track start time for a quiz(question) and a content block. my idea was that the quiz, content block would probably just take few minutes to complete, so i was not sure if its a good idea to track the start time.

tompollard: Got it, thanks, makes sense.

physionet-django/console/templates/console/training_type/index.html
tompollard: Thanks for explaining, makes sense.

physionet-django/static/sample/create.json
alistairewj: Not sure how we would grade them if it's short answer. Maybe a better option is to allow for multiple_choice or checkbox (where the user can check many answers rather than just select a single one). This would give you a way to allow for different types of answering questions without introducing a short answer type that we don't know how to handle.

superryeti: Oh yes, my bad earlier i meant to say we can support the checkbox one where there can be more than 1 right answer.

Good point about short answers, we would need to worry about grading them.

alistairewj:
Could we rename it from Training to Course for clarity/mental separation from other parts of the platform?

superryeti:
Thank you @tompollard @alistairewj  working on the feedback

physionet-django/static/sample/create.json
superryeti: I agree @tompollard  however i think the reason Harry choose the Quizzes instead of Questions is because we already have `Question` model on `user` which has many to many relation to `TrainingType`. So it might get a confusing if we have two questions

physionet-django/training/models.py
```diff
@@ -0,0 +1,141 @@
...
+        IN_PROGRESS = 'IP', 'In Progress'
+        COMPLETED = 'C', 'Completed'
+
+    course_progress = models.ForeignKey('training.CourseProgress', on_delete=models.CASCADE,
+                                        related_name='module_progresses')
+    module = models.ForeignKey('training.Module', on_delete=models.CASCADE)
+    status = models.CharField(max_length=2, choices=Status.choices, default=Status.IN_PROGRESS)
+    last_completed_order = models.PositiveIntegerField(null=True, default=0)
+    started_datetime = models.DateTimeField(null=True, blank=True)
+    updated_datetime = models.DateTimeField(auto_now=True)
```
alistairewj:
```suggestion
    updated_at = models.DateTimeField(auto_now=True)
```

physionet-django/training/models.py
```diff
@@ -0,0 +1,141 @@
...
+    class Status(models.TextChoices):
+        IN_PROGRESS = 'IP', 'In Progress'
+        COMPLETED = 'C', 'Completed'
+
+    course_progress = models.ForeignKey('training.CourseProgress', on_delete=models.CASCADE,
+                                        related_name='module_progresses')
+    module = models.ForeignKey('training.Module', on_delete=models.CASCADE)
+    status = models.CharField(max_length=2, choices=Status.choices, default=Status.IN_PROGRESS)
+    last_completed_order = models.PositiveIntegerField(null=True, default=0)
+    started_datetime = models.DateTimeField(null=True, blank=True)
```
alistairewj:
```suggestion
    started_at = models.DateTimeField(null=True, blank=True)
```

physionet-django/training/models.py
```diff
@@ -0,0 +1,141 @@
...
+class CourseProgress(models.Model):
+    class Status(models.TextChoices):
+        IN_PROGRESS = 'IP', 'In Progress'
+        COMPLETED = 'C', 'Completed'
+
+    user = models.ForeignKey('user.User', on_delete=models.CASCADE)
+    course = models.ForeignKey('training.Course', on_delete=models.CASCADE)
+    status = models.CharField(max_length=2, choices=Status.choices, default=Status.IN_PROGRESS)
+    started_datetime = models.DateTimeField(auto_now_add=True)
+    completed_datetime = models.DateTimeField(auto_now=True)
```
alistairewj:
```suggestion
    completed_at = models.DateTimeField(auto_now=True)
```

physionet-django/training/models.py
alistairewj:
```suggestion
    started_at = models.DateTimeField(auto_now_add=True)
```

physionet-django/notification/utility.py
```diff
@@ -1025,3 +1025,19 @@ def notify_event_participant_application(request, user, registered_user, event):
...
+    """
+    Send the training expiry email.
+    """
+
+    subject = f"{settings.SITE_NAME} Training Validation Notification"
```
alistairewj: Wouldn't "Training Expiry Notification" be more appropriate?

physionet-django/notification/utility.py
```diff
@@ -1025,3 +1025,19 @@ def notify_event_participant_application(request, user, registered_user, event):
...
+    context = {
+        'name': user.get_full_name(),
+        'SITE_NAME': settings.SITE_NAME,
+        'training': training,
+        'expiry': expiry
+    }
+    body = loader.render_to_string('training/email/training_notification.html', context)
```
alistairewj: for clarity I would rename the e-mail to training_expiry_notification.html

physionet-django/static/sample/create.json
```diff
@@ -0,0 +1,23 @@
+{
+  "courses": {
```
alistairewj: I think it would be better to have actual example data here. Otherwise the sample is fairly hard to follow. After downloading the JSON I have difficulty knowing exactly what to put where.

The typing info is useful though, so it makes me think we should also have a guideline page on the structure of the JSON. There is already a tab in the admin console for that (Guidelines), so we'd just need to add a tab under that ("Course Creation").

physionet-django/static/sample/create.json
alistairewj: I think Quizzes is correct because each one of the elements is itself a quiz (made up of multiple questions).

If we want a more generic term, we could do "Assessments", or something similar. But quizzes seems simple and good enough.

Again, I think having example data here (e.g. multiple quizzes, and multiple questions within each quiz) would make it easier to understand this JSON format.

physionet-django/physionet/urls.py
```diff
@@ -25,6 +25,8 @@
     path('console/', include('console.urls')),
     # user app
     path('', include('user.urls')),
+    # training app
+    path('', include('training.urls')),
```
alistairewj, line 28: For events, we have:

- the event content and URLs under `events/`
- views in the console page displaying the events which are located in the console app

Conversely, for the training app, we have:

- no content under `training/`
- views for the console page within the training app

Seems inconsistent (and at the least the URL redirect from `console/` to `training/` in the admin console seems wrong).

physionet-django/static/sample/create.json
alistairewj: You can have the questions be something like: "The answer is A". Just to make it easier to edit, and skip filler content.

physionet-django/console/templates/console/training_type/index.html
```diff
@@ -0,0 +1,75 @@
...
+            <th>Name</th>
+            <th>Valid Duration</th>
+            <th>Version</th>
+          </tr>
+        </thead>
+        <tbody>
+        {% for training in training_types %}
+
+            <tr>
+                <td>{{ training.name|title }}</td>
```
alistairewj: I think it would be useful to make the training content available by clicking on the title. The simplest way I can imagine doing that is an icon next to it or something similar which lets you download the JSON. The reason I suggest putting an icon next to the title (rather than making the title a hyperlink) is it makes it clearer that clicking the link will download a file.

physionet-django/console/templates/console/training_type/index.html
```diff
@@ -0,0 +1,75 @@
...
+                                    {% for training in training_types %}
+                                    <option value="{{ training.id }}">{{ training.name|title }}</option>
+                                    {% endfor %}
+                                </select>
+                            </div>
+                            <div class="form-group">
+                                <label>File: </label>
+                                <input type="file" name="json_file" id="json_file" required="True" class="form-control">
+                            </div>
+{#                            <div class="form-group">#}
```
alistairewj: Delete I think

physionet-django/physionet/urls.py
alistairewj: After discussion, decision was to import the views into the console app so that the urls are rendered via the console

physionet-django/training/views.py
```diff
@@ -0,0 +1,56 @@
...
+
+        if request.POST.get('training_id') != "-1":
+            training_type = get_object_or_404(TrainingType, pk=request.POST.get('training_id'))
+        else:
+            training_type = None
+
+        json_file = request.FILES.get("json_file", "")
+
+        if not json_file.name.endswith('.json'):
+            messages.error(request, 'File is not of JSON type')
```
alistairewj: Perhaps a warning is better than an error here? The parser would raise an error if it fails to deserialize.

alistairewj (commented):
Have some brief organizational comments. Haven't reviewed the models yet but I plan to!

physionet-django/static/sample/create.json
superryeti: Hi, Yes, we have the example for create and update on the same folder. This one is like a blue print that tells about expected data types, structure etc.

superryeti: oh sorry, just understood what you meant. Yeah i agree your recommendation makes sense. Thank you @alistairewj

superryeti:
Thanks @alistairewj , i addressed the comments.

Brief summary of updates, 
1. Move Course page inside console app
2. Add an option to download the course data as json(which users can use to migrate the content or make some changes and update the course). Here i added options to allow admins to select the version of course they want to download also.
3. Add a guideline page on how to create course

image

image


image

physionet-django/console/templates/console/console_navbar.html
```diff
@@ -101,14 +101,31 @@
...
+        {% else %}
+          <a id="nav_pages_dropdown" class="nav-link nav-link-collapse drop collapsed" data-toggle="collapse" href="#trainingComponent" data-parent="#sideAccordion" aria-expanded="false">
+        {% endif %}
+            <i class="fa fa-fw fa-window-maximize"></i>
+            <span class="nav-link-text">Training</span>
           </a>
-        </li>
+          <!-- submenu -->
+        {% if training_type_nav or training_nav %}
+          <ul class="sidenav-second-level collapse show" id="trainingComponent" style="">
```
alistairewj: remove empty ruleset for style

```suggestion
          <ul class="sidenav-second-level collapse show" id="trainingComponent">
```

physionet-django/console/templates/console/training_type/index.html
```diff
@@ -0,0 +1,71 @@
+{% extends "console/base_console.html" %}
+
+{% load static %}
+
+{% block title %}Training Types{% endblock %}
```
alistairewj: Not sure Training Types is correct anymore

```suggestion
{% block title %}On Platform Courses{% endblock %}
```

or perhaps include the variable for the platform name here

physionet-django/console/views.py
```diff
@@ -1902,6 +1902,15 @@ def guidelines_review(request):
                   {'guidelines_review_nav': True})
 
 
+@permission_required('training.can_view_course_guidelines', raise_exception=True)
+def guidelines_course(request):
+    """
+    Guidelines for course creators.
+    """
+    return render(request, 'console/guidelines_course.html',
```
alistairewj, line 2033: I don't like that the guidelines for the course (what I would consider training app content) is here in the console app, but the template extends the base console template.

@tompollard - any suggestions on a better organization?

physionet-django/training/views.py
```diff
@@ -0,0 +1,71 @@
...
+
+        json_file = request.FILES.get("json_file", "")
+
+        if not json_file.name.endswith('.json'):
+            messages.warning(request, 'File is not of JSON type')
+            return redirect("courses")
+        file_data = JSONParser().parse(json_file.file)
+        serializer = TrainingTypeSerializer(training_type, data=file_data, partial=True)
+        if serializer.is_valid(raise_exception=False):
+            serializer.save()
```
alistairewj, line 54: This will fail if you violate a constraint on the model, e.g. if you try to reupload a course with the same training_type_id and version. We should have a check to verify the the course doesn't already exist. If it does, notifying the user that the upload failed probably due to an already used version is important.

We'd also want to verify here that the version number is increasing and semantically correct.

physionet-django/console/templates/console/guidelines_course.html
```diff
@@ -0,0 +1,161 @@
+{% extends "console/base_console.html" %}
+{% load static %}
+{% block title %}Guidelines for Courses{% endblock %}
+
+{% block content %}
+  <div class="card mb-3">
+    <div class="card-header">
+      Guidelines for creating and updating courses courses
```
alistairewj: typo

```suggestion
      Guidelines for creating and updating courses
```

physionet-django/console/templates/console/guidelines_course.html
```diff
@@ -0,0 +1,161 @@
...
+    </div>
+    <div class="card-body">
+      <div class="table-responsive">
+        <p>To create a new course or update a course, you will need to organize all the course content in a json file
+          and upload via
```
alistairewj:
```suggestion
          and upload via the
```

physionet-django/console/templates/console/guidelines_course.html
```diff
@@ -0,0 +1,161 @@
...
+              href="{% static 'sample/example-course-update.json' %}">update a course</a></p>
+
+        <h4>General Schema Explanation</h4>
+        <pre>
+{
+  "name": "string",
+  "description": "string",
+  "valid_duration": "string",
+  "courses": [{
+    "version": "float",
```
alistairewj: not a float, right? it could be 1.0.1. even if we follow the instructions below (Major.Minor), we still wouldn't want a number as that would mean `1.1` == `1.10`.

physionet-django/console/templates/console/guidelines_course.html
```diff
@@ -0,0 +1,161 @@
...
+            above).
+            Please use the Semantic Versioning system (Major.Minor) for version numbers.
+            If you update a course's major version (e.g., from 1.5 to 2.0), all existing training certificates provided
+            to
+            users will expire after a month. To regain access to resources, users must retake the new version of the
+            training.</small>
+
+        </ol>
+        <ol start="2">
+          <li><strong>Modules:</strong> A course may have one or more modules that contain the contents,quizzes for each
```
alistairewj: add space
```suggestion
          <li><strong>Modules:</strong> A course may have one or more modules. Each module contains the course content and quizzes.
          Modules must include:
```

physionet-django/console/templates/console/guidelines_course.html
```diff
@@ -0,0 +1,161 @@
...
+          <ul>
+            <li>A name for the module</li>
+            <li>A description of the module in html</li>
+            <li>An order for the module</li>
+            <li>One or more content items in the module</li>
+            <li>One or more quizzes in the module</li>
+          </ul>
+        </ol>
+        <ol start="3">
+          <li><strong>Content:</strong> A module may have one or more content. Each content should include:</li>
```
alistairewj:
```suggestion
          <li><strong>Content:</strong> A module may have one or more content blocks. Each content block should include:</li>
```

physionet-django/console/templates/console/guidelines_course.html
```diff
@@ -0,0 +1,161 @@
...
+            <li>One or more choices for the quiz, including the correct answer</li>
+            <small><strong>Order:</strong> Note that the order is used to determine which content or quiz comes first,
+              so please
+              make sure you order the content and quiz properly. <br> For example, here the user will see the content
+              and quiz in the following order
+              content1, quiz 1, content 2, quiz 2.
+              <pre><code>
+"contents": [
+    {
+      "body": "content 1",
```
alistairewj: I think it will be useful for the text in this example to give the reader an idea of how the content will flow. So my following suggestions aim towards having the text describe when/where the content will appear and how the course will actually function.


```suggestion
      "body": "This content will display first for the given module.",
```

physionet-django/console/templates/console/guidelines_course.html
```diff
@@ -0,0 +1,161 @@
...
+      "order": 1
+    },
+    {
+      "body": "content 2",
```
alistairewj:
```suggestion
      "body": "This content will appear third (order = 3), after the quiz below.",
```

physionet-django/console/templates/console/guidelines_course.html
```diff
@@ -0,0 +1,161 @@
...
+      "order": 3
+    }
+],
+"quizzes": [
+  {
+    "question": "quiz 1",
```
alistairewj:
```suggestion
    "question": "This quiz will appear after the first content block (order = 2).",
```

physionet-django/console/templates/console/guidelines_course.html
```diff
@@ -0,0 +1,161 @@
...
+        "is_correct": true
+      },
+      {
+        "body": "choice 2",
+        "is_correct": false
+      }
+    ]
+  },
+  {
+    "question": "quiz 2",
```
alistairewj:
```suggestion
    "question": "This quiz will appear last, after the final content block (order = 4).",
```

physionet-django/console/templates/console/guidelines_course.html
```diff
@@ -0,0 +1,161 @@
...
+      "order": 3
+    }
+],
+"quizzes": [
+  {
+    "question": "quiz 1",
+    "order": 2,
+    "choices": [
+      {
+        "body": "choice 1",
```
alistairewj:
```suggestion
        "body": "This answer is correct. The user will proceed to the next block if they answer it.",
```

physionet-django/console/templates/console/guidelines_course.html
```diff
@@ -0,0 +1,161 @@
...
+        "is_correct": true
+      },
+      {
+        "body": "choice 2",
```
alistairewj:
```suggestion
        "body": "This choice is incorrect. The user will return to the beginning of the module if they answer it.",
```

physionet-django/notification/utility.py
alistairewj: Yes, and just to add to this, a new training version should trigger a notification of expiry for users with the older training.

Since this would be potentially a large number of emails, and because it might be confusing for someone to know their training expires a year from now and then suddenly receive an e-mail saying their training is expiring in 30 days, it might be worth having a `notify_users_of_training_update`. This would batch e-mails users about the new training version using [`send_mass_mail`](https://docs.djangoproject.com/en/4.2/topics/email/#send-mass-mail). That being said, I don't know the real-world impact in efficiency of only opening one connection to the mail server for the e-mails.

physionet-django/training/models.py
```diff
@@ -0,0 +1,144 @@
+from django.db import models
+
+from project.modelcomponents.fields import SafeHTMLField
+
+
+class Course(models.Model):
+    training_type = models.ForeignKey('user.TrainingType',
+                                      on_delete=models.CASCADE, related_name='courses')
+    version = models.FloatField(default=1.0)
```
alistairewj: see also my note on the guidelines -> this should be a string.

the version in the projects is a CharField length 15
https://github.com/MIT-LCP/physionet-build/blob/0af16dcfbefc54c81e59a4c1b123b959780b254e/physionet-django/project/modelcomponents/metadata.py#L60

we could just be consistent with that. note also it provides a validator

physionet-django/training/models.py
```diff
@@ -0,0 +1,144 @@
...
+
+    class Meta:
+        default_permissions = ('change',)
+        unique_together = ('training_type', 'version')
```
alistairewj: Docs say this may be deprecated soon - https://docs.djangoproject.com/en/4.2/ref/models/options/#unique-together

You can switch to UniqueConstraint


```suggestion
    constraints = [
        models.UniqueConstraint(fields=['training_type', 'version'], name='unique_course')
    ]
```

physionet-django/training/models.py
```diff
@@ -0,0 +1,144 @@
...
+class CourseProgress(models.Model):
+    class Status(models.TextChoices):
+        IN_PROGRESS = 'IP', 'In Progress'
+        COMPLETED = 'C', 'Completed'
+
+    user = models.ForeignKey('user.User', on_delete=models.CASCADE)
+    course = models.ForeignKey('training.Course', on_delete=models.CASCADE)
+    status = models.CharField(max_length=2, choices=Status.choices, default=Status.IN_PROGRESS)
+    started_at = models.DateTimeField(auto_now_add=True)
+    completed_at = models.DateTimeField(auto_now=True)
```
alistairewj:
```suggestion
    completed_at = models.DateTimeField(null=True, blank=True)
```

... otherwise this will be updated every time the object is saved

physionet-django/training/models.py
```diff
@@ -0,0 +1,144 @@
...
+
+    class Meta:
+        unique_together = ('user', 'course')
```
alistairewj: similar to the above, can switch syntax to:


```suggestion
        constraints = [
            models.UniqueConstraint(fields=['user', 'course'], name='unique_course_progress')
        ]
```

physionet-django/training/serializers.py
```diff
@@ -0,0 +1,169 @@
...
+
+class TrainingTypeSerializer(serializers.ModelSerializer):
+    courses = CourseSerializer(many=True)
+
+    class Meta:
+        model = TrainingType
+        fields = ['name', 'description', 'valid_duration', 'courses']
+        read_only_fields = ['id']
+
+    def update_course_for_major_version_change(self, instance):
```
alistairewj: I wouldn't consider this serialization. This is updating the database and informing the user. This should probably be refactored into the model and called in the view.

physionet-django/training/serializers.py
```diff
@@ -0,0 +1,169 @@
...
+        _ = trainings.update(
+            process_datetime=(
+                timezone.now() - (instance.valid_duration - timezone.timedelta(
+                    days=NUMBER_OF_DAYS_SET_TO_EXPIRE))))
+
+        for training in trainings:
+            notify_users_of_training_expiry(
+                training.user, instance.name, NUMBER_OF_DAYS_SET_TO_EXPIRE)
+
+    def update(self, instance, validated_data):
```
alistairewj, line 89: A lot of the functionality in `update` and `create` overlaps. You could refactor out most of the code into a function which you call in `update` and `create`.

physionet-django/training/serializers.py
```diff
@@ -0,0 +1,169 @@
...
+
+        with transaction.atomic():
+            course = validated_data.pop('courses')[0]
+            modules = course.pop('modules')
+
+            course['training_type'] = instance
+
+            course_instance = Course.objects.create(**course)
+
+            for module in modules:
```
alistairewj: There is a lot of duplication in the `update`/`create` codes. One possible refactor would be to create functions:

```python
def create_quizzes(module_instance, quizzes_data):
    ...
    QuizChoice.objects.bulk_create(choice_bulk)
    return

def create_contentblocks(module_instance, content_data):
    ...
    ContentBlock.objects.bulk_create(content_bulk)
    return
```

then wrap that in a `create_modules`:

```python
def create_modules(course_instance, modules_data):
    for module_data in modules_data:
        quizzes = module_data.pop('quizzes')
        contents = module_data.pop('contents')

        module_data['course'] = course_instance
        module_instance = Module.objects.create(**module)

        create_quizzes(module_instance, quizzes)
        create_contentblocks(module_instance, contents)
    return
```

so that you can do something like:


```python
with transaction.atomic():
    course = validated_data.pop('courses')[0]
    modules = course.pop('modules')

    course['training_type'] = instance

    course_instance = Course.objects.create(**course)
    create_modules(course_instance, modules)
```

physionet-django/training/models.py
```diff
@@ -0,0 +1,144 @@
...
+    module = models.ForeignKey('training.Module', on_delete=models.CASCADE)
+    status = models.CharField(max_length=2, choices=Status.choices, default=Status.IN_PROGRESS)
+    last_completed_order = models.PositiveIntegerField(null=True, default=0)
+    started_at = models.DateTimeField(null=True, blank=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    def __str__(self):
+        return f'{self.course_progress.user.username} - {self.module}'
+
+    def get_next_content_or_quiz(self):
```
alistairewj: Maybe it's worth removing this and the next function (`update_last_completed_order`) and only including them in the next PR? They are hard to reason about without seeing the front-end that actually uses the JSON.

physionet-django/static/sample/create-course-schema.json
```diff
@@ -0,0 +1,32 @@
+{
```
alistairewj, line 1: We have these samples, but we actually don't seem to use them? Only the hard-coded JSON field is used in the guidelines page. I do think it's nicer to have these as static assets in this folder which we then call an include to show on the guidelines page.

alistairewj (changes requested):
This is really great! I've made some comments on refactoring / rewriting some bits.

superryeti:
> This is really great! I've made some comments on refactoring / rewriting some bits.

Thanks @alistairewj , just saw this. Weirdly it was in my spam(Gmail thought there was some dangerous link in the email, 😆 )

image

alistairewj:
probably the links to physionet-build ;)

physionet-django/notification/utility.py
```diff
@@ -1025,3 +1025,19 @@ def notify_event_participant_application(request, user, registered_user, event):
...
+
+    subject = f"{settings.SITE_NAME} Training Expiry Notification"
+    context = {
+        'name': user.get_full_name(),
+        'SITE_NAME': settings.SITE_NAME,
+        'training': training,
+        'expiry': expiry
+    }
+    body = loader.render_to_string('training/email/training_expiry_notification.html', context)
+    send_mail(subject, body, settings.DEFAULT_FROM_EMAIL, [user, ], fail_silently=False)
```
alistairewj: let's change this to a user-specific banner, at the beginning just on the training page

physionet-django/console/templates/console/training_type/index.html
Rutvikrj26: Added the platform name variable

physionet-django/training/views.py
Rutvikrj26: Added the following checks :
1. If the Content of the submitted JSON file is a valid JSON and can be parsed by the JSON Parser
2. If the course exists & the version number is semantically incorrect
3. If the course exists & the new version number is either equal to or smaller than the last version number

If none of these are present, then the new course gets added.

physionet-django/training/serializers.py
Rutvikrj26: Refactored the function to model and added a check while final saving. If the first digit of the version number is not the same as the first digit of the latest version, then the major_update method is called on the existing course with the training_type as the instance.

Final review by @alistairewj pending.

physionet-django/training/serializers.py
Rutvikrj26: Refactored the Update & Create Functions in a similar pattern as mentioned in the comment.

Confirming the fucntionality: Create function is currently returning `training_type` at the end.

physionet-django/training/models.py
Rutvikrj26: Updated the branch connected to PR #1951 with these functions and removed them from the current branch.

physionet-django/console/views.py
tompollard: Not really, other than when I look at the two course PRs I keep thinking this new functionality would be better as a standalone app.

physionet-django/console/templates/console/guidelines_course.html
Rutvikrj26: Converting it to a string.

physionet-django/training/models.py
Rutvikrj26: Updated the version to a charfield similar to the project version, using the version validator from projects.

physionet-django/notification/utility.py
Rutvikrj26: @alistairewj Need some more clarification on 
user-specific banner, at the beginning just on the training page.

Is this comment hinting at changes in a template or may be I'm missing something? some clarification would help.

tompollard:
@Rutvikrj26 Try running `makemigrations`. I think tests are failing because the PR is missing a migration.

tompollard:
Tests just failing on a style issue in the migration now (E501 line too long):

```
physionet-django/training/migrations/0002_alter_course_unique_together_and_more.py:25:120: E501 line too long (124 > 119 characters)

             model_name='course',
             name='version',
             field=models.CharField(blank=True, default='', max_length=15, validators=[project.validators.validate_version]),
                                                                                                                        ^

physionet-django/training/models.py:23:17: E1[26](https://github.com/MIT-LCP/physionet-build/actions/runs/5731564942/job/15532734091?pr=1950#step:15:27) continuation line over-indented for hanging indent

         ]
         permissions = [
                 ('can_view_course_guidelines', 'Can view course guidelines'),
                 ^

physionet-django/training/models.py:24:13: E121 continuation line under-indented for hanging indent

         permissions = [
                 ('can_view_course_guidelines', 'Can view course guidelines'),
             ]
             ^
```

physionet-django/notification/utility.py
Rutvikrj26: Added a conditional message displayed on the training page when training expires in the next 30 days.

tompollard:
@Rutvikrj26 I'll keep an eye out for your updates to the settings console (adding a "Courses" tab and moving some of the content to it from the Training section). Let me know if it is helpful to chat before you make the pull request.

Rutvikrj26:
Heyy @tompollard, I've fixed the errors and rebased the branch on dev. Please check and let me know if any further changes are required.

physionet-django/training/templates/training/email/training_expiry_notification.html
```diff
@@ -0,0 +1,9 @@
+{% load i18n %}{% autoescape off %}{% filter wordwrap:70 %}
+Dear {{ name }},
+
+Your training {{ training }} on {{ domain }} will be expiring in {{ expiry }} days. To retain the access it provides, kindly login to your account to retake it.
```
tompollard: This email could be improved I think, e.g. by giving guidance on how to retake training and ideally giving more detail about which resources require the training. Perhaps for now we could update to something like: 

> Your {{ training }} training on {{ domain }} will be expiring in {{ expiry }} days. After the expiry date, you may lose access to certain resources.  To retain access, please submit a new training certificate via the [training page](LINK) in your user profile.

physionet-django/training/models.py
```diff
@@ -0,0 +1,188 @@
...
+from django.utils import timezone
+
+from project.modelcomponents.fields import SafeHTMLField
+from user.models import Training
+from project.validators import validate_version
+
+NUMBER_OF_DAYS_SET_TO_EXPIRE = 30
+
+
+class Course(models.Model):
```
tompollard, line 9: All of the classes here would benefit from Docstrings. It is not at all clear what is meant by "Course", "Module", "Quiz", "QuizChoice", "CourseProgress", "CompletedContent", etc. The code needs to be clear about how all these components fit together.

physionet-django/training/views.py
```diff
@@ -0,0 +1,98 @@
...
+
+from user.models import Training, TrainingType, TrainingQuestion, RequiredField
+from user.enums import TrainingStatus
+
+from training.models import Course, Quiz, QuizChoice, ContentBlock
+from training.models import CourseProgress, ModuleProgress, CompletedContent, CompletedQuiz
+from training.serializers import TrainingTypeSerializer
+
+
+@permission_required('training.change_course', raise_exception=True)
```
tompollard, line 29: As above, all of these functions should have docstrings to at minimum explain their purpose.

physionet-django/training/views.py
```diff
@@ -0,0 +1,98 @@
...
+        existing_course = Course.objects.filter(training_type=training_type)
+        if existing_course.exists():
+            # checking if the version number is valid
+            if not all(map(lambda x: x.isdigit() or x == '.', str(file_data['courses'][0]['version']))):
+                messages.error(request, 'Version number is not valid.')
+            # checking if the version number is greater than the existing version
+            elif float(file_data['courses'][0]['version']
+                       ) <= float(existing_course.order_by('-version').first().version):
+                messages.error(request, 'Version number should be greater than the existing version.')
+            else:  # Checks passed and moving to saving the course
```
tompollard: I think PEP8 discourages inline comments ("Checks passed and moving to saving the course"). Also not sure this comment is necessary.

physionet-django/training/views.py
```diff
@@ -0,0 +1,98 @@
...
+        try:
+            file_data = JSONParser().parse(json_file.file)
+        except json.decoder.JSONDecodeError:
+            messages.error(request, 'JSON file is not properly formatted.')
+            return redirect("courses")
+
+        # Checking if the Training type with the same version already exists
+        existing_course = Course.objects.filter(training_type=training_type)
+        if existing_course.exists():
+            # checking if the version number is valid
```
tompollard: Shouldn't validation happen in a validators module? e.g.:

https://github.com/MIT-LCP/physionet-build/blob/46ef14feb4198faf6386c22cde48b33310ed188d/physionet-django/project/validators.py#L85-L95

physionet-django/training/views.py
```diff
@@ -0,0 +1,98 @@
...
+                messages.error(request, 'Version number is not valid.')
+            # checking if the version number is greater than the existing version
+            elif float(file_data['courses'][0]['version']
+                       ) <= float(existing_course.order_by('-version').first().version):
+                messages.error(request, 'Version number should be greater than the existing version.')
+            else:  # Checks passed and moving to saving the course
+                serializer = TrainingTypeSerializer(training_type, data=file_data, partial=True)
+                if serializer.is_valid(raise_exception=False):
+                    # A Major Version change is detected : The First digit of the version number is changed
+                    if int(str(existing_course.order_by('-version').first().version).split('.')[0]) != int(str(
```
tompollard: Some weird things happening with types here. Is it necessary to convert to string and then to int? I feel like we probably want to pull version handling out into separate, reusable functions.

physionet-django/training/models.py
```diff
@@ -0,0 +1,188 @@
...
+        constraints = [
+            models.UniqueConstraint(
+                fields=["training_type", "version"], name="unique_course"
+            )
+        ]
+        permissions = [
+            ("can_view_course_guidelines", "Can view course guidelines"),
+        ]
+
+    def update_course_for_major_version_change(self, instance):
```
tompollard: Right now, the behaviour of `update_course_for_major_version_change` is unintuitive. If I want to bump the major version of a course, I should be able to bump the major version. If I want to expire a course, I should be able to expire a course. These don't need to be coupled.

physionet-django/training/models.py
```diff
@@ -0,0 +1,188 @@
+from django.db import models
+from django.utils import timezone
+
+from project.modelcomponents.fields import SafeHTMLField
+from user.models import Training
+from project.validators import validate_version
+
+NUMBER_OF_DAYS_SET_TO_EXPIRE = 30
```
tompollard: Does `NUMBER_OF_DAYS_SET_TO_EXPIRE` needs to be hardcoded? Should we allow administrators to explicitly set the expiry period instead?

physionet-django/training/views.py
```diff
@@ -0,0 +1,98 @@
...
+            return redirect("courses")
+
+        # Checking if the Training type with the same version already exists
+        existing_course = Course.objects.filter(training_type=training_type)
+        if existing_course.exists():
+            # checking if the version number is valid
+            if not all(map(lambda x: x.isdigit() or x == '.', str(file_data['courses'][0]['version']))):
+                messages.error(request, 'Version number is not valid.')
+            # checking if the version number is greater than the existing version
+            elif float(file_data['courses'][0]['version']
```
tompollard: I think @bemoody recently coded up a function to check for versioning (for projects). We might be able to reuse his implementation here?

physionet-django/training/views.py
```diff
@@ -0,0 +1,98 @@
...
+        {
+            'training_types': training_types,
+            'training_type_nav': True,
+        })
+
+
+@permission_required('training.change_course', raise_exception=True)
+def download_course(request, pk, version):
+    training_type = get_object_or_404(TrainingType, pk=pk)
+    version = float(version)
```
tompollard: I think we should review how versioning is handled (perhaps referring to Benjamin's implementation for projects). It would be best if we could avoid these type changes being scattered about the place.

tompollard (commented):
@Rutvikrj26 I have added some thoughts. Sorry for the delay! The functionality works well for me, but I think we'll want to do a second round of code clean up after these initial points are addressed.

physionet-django/training/templates/training/email/training_expiry_notification.html
Rutvikrj26: Agreed, some more guidance would be useful.
Adding the particular resource they are credentialed for using the training, which is expiring could be a good idea?
I'm adding the current update, while we can discuss more.

physionet-django/training/models.py
Rutvikrj26: I've added the docstrings to the classes based on my understanding of the code, though they don't include the reasoning behind this specific implementation & naming as they were written by Amit.

physionet-django/training/views.py
Rutvikrj26: Thanks for bringing this to my notice. I have implemented this function to check the validity of the version number.

physionet-django/training/views.py
```diff
@@ -0,0 +1,98 @@
...
+        # Checking if the Training type with the same version already exists
+        existing_course = Course.objects.filter(training_type=training_type)
+        if existing_course.exists():
+            # checking if the version number is valid
+            if not all(map(lambda x: x.isdigit() or x == '.', str(file_data['courses'][0]['version']))):
+                messages.error(request, 'Version number is not valid.')
+            # checking if the version number is greater than the existing version
+            elif float(file_data['courses'][0]['version']
+                       ) <= float(existing_course.order_by('-version').first().version):
+                messages.error(request, 'Version number should be greater than the existing version.')
```
Rutvikrj26: To ensure that the new version number is greater than the previous version, I am using a float-based comparision, which does not work if the version has multiple decimal counters. Any suggestions on how we can compare that the new version is greater than the existing one?
@tompollard @bemoody @kshalot

tompollard: As mentioned above, there should be existing code that implements versioning for projects. I would suggest taking a look at that.

physionet-django/training/views.py
Rutvikrj26: @tompollard agreed, I'll take a look at how the project versioning is handled.

physionet-django/training/views.py
tompollard: See: https://github.com/MIT-LCP/physionet-build/pull/2068

Rutvikrj26: I've implemented the sematic package versioning used by python for version check.

physionet-django/training/views.py
Rutvikrj26: I just identified a major flaw in the download logic. It does not give you the version you try to download, as the version kind of no-longer exists in the system. It'll take more time to fix this issue, than I initially estimated. 

As it turns out, in this function - the primary key passed down along with the request is the same irrespective of the version it gets, and hence the only thing that changes is the name of the file that is downloaded.

Rutvikrj26: The function was broken and was using TrainingType serializer, rather than the Course serializer. 
Fixed the functionality - now working.

Rutvikrj26:
Implemented a new UI and an is_active flag in the model of the course. 
image

tompollard:
@Rutvikrj26 thanks for chatting today. Quick summary of the main points we discussed:
 
- If we find existing training code confusing, we should fix it first.
- We should clarify need for both `TrainingType` and `Course` models. Are both needed? Should the current `TrainingType` be renamed to `Course`?
- The `Create/Update` button should be renamed `Create` and it should focus on creating new courses. Updating courses should happen on the page that lists all active/archived versions of a single course.

bemoody:
> Should the current TrainingType be renamed to Course?

No, I don't think that makes sense.

@Rutvikrj26, do not merge dev into a feature branch.  Please hard reset to 3e75ab970 and rebase instead.

Rutvikrj26:
> * The `Create/Update` button should be renamed `Create` and it should focus on creating new courses. Updating courses should happen on the page that lists all active/archived versions of a single course.

@tompollard This has been implemented and is ready for testing. 
Furthermore, @bemoody it would be great if you could review the PR and let me know your thoughts.

tompollard:
I took a look this morning and have added a few thoughts below:

#### Minor formatting things: 

  - The menu item could be cleaned up (capitalise first letter for consistency, use a unique icon)
  - Some files are missing newlines at the end.

#### Usability: 

- The page that displays Active/Archived courses is really not very clear from a user perspective. For example: What happens to a course when I set it to "expire"? The course is moved to "archived", but the expiry date disappears. How can I tell whether a course is expired or not?

If you haven't already, I'd suggest running through the design of this new feature with January. I think the page would benefit from some tweaks to layout and content. Perhaps also add some short explanatory sentences to the page to help guide the user?

#### TypeError when trying to update a course:

When attempting to update a course using the following JSON (!) (taken from the suggested example), I receive a TypeError:

```
{
  "name": "Course 2",
  "description": "<p>Test content description</div>",
  "valid_duration": "1095 00:00:00",
  "courses": [
    {
      "version": 2.0,
      "modules": [
        {
          "name": "Module 1",
          "description": "<p>Module description</div>",
          "order": 0,
          "contents": [
            {
              "body": "<p>Hello This is a test</p><div><p>Test content1</p></div>",
              "order": 0
            }
          ],
          "quizzes": [
            {
              "question": "What is the correct answer(choice1)?",
              "order": 1,
              "choices": [
                {
                  "body": "I am a choice1",
                  "is_correct": true
                },
                {
                  "body": "I am a choice2",
                  "is_correct": false
                },
                {
                  "body": "I am a choice3",
                  "is_correct": false
                },
                {
                  "body": "I am a choice4",
                  "is_correct": false
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

```
Request Method: POST
Request URL: http://localhost:8000/console/courses/3/

Django Version: 4.1.10
Python Version: 3.10.13
Installed Applications:

  File "/Users/tompollard/projects/physionet-build/env/lib/python3.10/site-packages/django/core/handlers/exception.py", line 56, in inner
  File "/Users/tompollard/projects/physionet-build/env/lib/python3.10/site-packages/django/core/handlers/base.py", line 197, in _get_response
  File "/Users/tompollard/projects/physionet-build/env/lib/python3.10/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapped_view
  File "/Users/tompollard/projects/physionet-build/env/lib/python3.10/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapped_view
  File "/Users/tompollard/projects/physionet-build/physionet-django/training/views.py", line 104, in course_details
    if validate_version(new_course_version) is not None:
  File "/Users/tompollard/projects/physionet-build/physionet-django/project/validators.py", line 90, in validate_version
    if not re.fullmatch(r'[0-9]+(\.[0-9]+)+', value) or '..' in value:
  File "/opt/homebrew/Cellar/python@3.10/3.10.13/Frameworks/Python.framework/Versions/3.10/lib/python3.10/re.py", line 195, in fullmatch
    return _compile(pattern, flags).fullmatch(string)

Exception Type: TypeError at /console/courses/3/
Exception Value: expected string or bytes-like object
```

bemoody:
Some things that jump out at me:


* Courses should be identified by slug plus version (rather than integer ID plus version.)  Either we want a `slug` field in the TrainingType, or in the Course.


* I think the course metadata should be part of the versioned object.  That is, the course title (what you've called `name`, I would call `title`) and `description`, and possibly also `valid_duration`.


* It is very strange to have `"courses"` in the JSON file.  If I understand correctly, one JSON file is one Course (one version.)  So why is this an array?


* The function of the `is_active` flag is not totally clear to me.  Let's break down the expected life-cycle of a course:

1. Version 1.0 is published.  It's the only version.  People requesting access are directed to take this course.  People who have completed it are treated as "certified" for five years (or 100 years, or whatever) after the completion date.

2. Version 1.1 is published.  People requesting access are directed to take version 1.1.  But people who already completed version 1.0 are still treated as "certified" for five years after completion.

3. Version 2.0 is published.  People requesting access are directed to take version 2.0.  People who already completed version 1.0 are now treated as "certified" for at most 30 days after the date that version 2.0 was published.

How would these three states be represented in terms of the Course model?

(Note that `update_course_for_version_change` appears to be modifying the stored *completion date* in order to enforce what is actually an *altered validity period*.  That seems like a bad idea to me.)

Rutvikrj26:
Please checkup the following workflow for new training addition and an old training expiry. Let me know if they sound okay. 

Admin adds a new training-type & a user wants to get credentialed

- User tries to take the course(belongs to a training-type) and is given the latest course for the training-type( to be ensured in course taking - second PR)
- User completes the quizzes
- User gets the certification / training attached to his account for the training type

Admin adds another course (a new version) for an existing training-type

- The admin sees both the training type as active in the admin panel
- Admin chooses to expire the previous version of the training
- Admin goes into admin panel, puts in the number of days from now should any training/certification accredited for the "to-be-expired" version should be valid.
- The validity duration reduces( the logic can be modified to accommodate a safe change - change in validity duration / adding an `expiry_date` field / somethingelse - any suggestions?)
- The user gets notified to re-take the course for the training type. 
- user takes the new course and gets credentialed again.

Rutvikrj26:
> When attempting to update a course using the following javascript (taken from the suggested example JSON), I receive a TypeError:

The type for version has changed to `string` to match the versioning of the projects. The schema examples have become outdated. Will be updating them in the next push.

Rutvikrj26:
@tompollard @bemoody 

As per our discussion, I have most of the changes implemented except two :

> `update_course_for_version_change` appears to be modifying the stored _completion date_ in order to enforce what is actually an _altered validity period_. That seems like a bad idea to me.

This is the only way as the trainings does not have any other field that can be updated to implement the change in validity.
Any suggestions - if we do want to gear away, we might need to add another field to the `trainings` model. Please let me know if there is some other way to implement this that I have missed. 

> * Either we want a `slug` field in the TrainingType, or in the Course.

The current navigation uses trainingtype id  and currently, `TrainingType` model does not have a slug field. Do we want to allow null for the slug value or is there a better strategy for back filling the slug that is preferred.  

Please provide your thoughts on these two.

Rutvikrj26:
> The current navigation uses trainingtype id and currently, `TrainingType` model does not have a slug field. Do we want to allow null for the slug value or is there a better strategy for back filling the slug that is preferred.

As discussed during the tuesday standup, a good idea will be use django-slugify for bacfilling the slugs in training type. 
I've implemented the slug field in training type, added a migration to handle the backfilling and updated the routes & views to work with slug rather thank primary key.

physionet-django/console/urls.py
```diff
@@ -92,6 +93,8 @@
 
     # guidelines
     path('guidelines/review/', views.guidelines_review, name='guidelines_review'),
+    path('guidelines/course/', views.guidelines_course, name='guidelines_course'),
```
tompollard, line 96: This page is not currently listed in the admin console (not sure if this is intentional).

tompollard (approved):

PR #1951, opened 2023-03-28
create training basics  --> implement take training feature and resume training feature for user

superryeti:
## What?
Here i have added a feature that will allow users to take the training from settings. User can choose from a list of available 
trainings and take the onplatform training. Users can also see the progress of the training and can resume the training from where they left off.

## Why?
This is a follow up PR for https://github.com/MIT-LCP/physionet-build/pull/1950

## How?
Here we create 3 new views to handle the training.

**1. `training.views.take_training`**

This view will take care of training homepage. It will show progress of the training if the user has already started the training, show them details of the training and allow them to start/resume and review the training.

**2.  `training.views.take_module_training`**

This view will take care of letting users take individual modules of the training. It will allow users to take the module from where they left off.
Users will need to finish module1 before they can start module2 and so on.

Here we have used a considerable amount of javascript to make the training interactive. How it works is when the user loads the module page,
we initially load all the content, quizzes and order them based on the `order` field, and initially we hide all the content and quizzes.

Then we show either the first content,quiz or the next one after a content, quiz that they had completed last time. and we keep doing this until the user has completed the module.

**3. `training.views.update_module_progress`**

    This view uses a simple post request and expects an ajax request from the client. It will update the progress of the module and the progress of the training.

## Testing? [WIP]

Following tests will be added
1. Check access to the training homepage
2. Check access to the training module page
3. Check valid and invalid submission of the training module
4. Check valid and invalid submission of the training

## Screenshots (optional)

image

image

image

image

### How to take a training as a user?

1. Login as user
2. Navigate to `Settings`
3. Select the training from the dropdown
4. Start the training


superryeti:
Quick Note: There is a weird bug i just noticed

Currently OnPlatform training is treated unique based on version_number and training_type. So if a user had taken a training with version 1.1, and then next month there is a minor update with version 1.2, their previous training would still be valid.

But if they go and they try to review the training again, they will be retaking the new version 1.2 of training.(because by default we always provide them with the latest version of training).

Which brings to the weird bug which is that if the user hadn't finished the training 1.1 and the version 1.2 was lunched, then when they try to resume the training, it will be like taking the training from scratch for them

tompollard:
I merged this PR on top of https://github.com/MIT-LCP/physionet-build/pull/1950 to do some testing. There were a few issues I came across while testing:

1. Duplicate training:

Minor thing, but it is currently possible to upload duplicate training courses at: 
http://localhost:8000/console/courses/

2. Getting stuck after hitting "back"

- Sign in and start a training course at: http://localhost:8000/settings/training/
- Click "Start" to start the training
- Click "Next" after reading the explanatory content on "The Americas"
- Select a response to the first question, then click "Back"
- "Next" is greyed out and cannot be clicked. You are stuck!

3. Reloading the page after following point 2 above

Reloading the page after following the points above triggers the following error. Returning to http://localhost:8000/settings/platform-training/3/ and clicking "Continue" triggers the same error.

```
Request Method: GET
Request URL: http://localhost:8000/settings/platform-training/3/module/3/

  File "/Users/tompollard/projects/physionet-build/env/lib/python3.10/site-packages/django/core/handlers/exception.py", line 56, in inner
  File "/Users/tompollard/projects/physionet-build/env/lib/python3.10/site-packages/django/core/handlers/base.py", line 197, in _get_response
  File "/Users/tompollard/projects/physionet-build/env/lib/python3.10/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapped_view
  File "/Users/tompollard/projects/physionet-build/physionet-django/training/views.py", line 147, in take_module_training
    resume_content_or_quiz_object = resume_content_or_quiz_module.get_next_content_or_quiz()

Exception Type: AttributeError at /settings/platform-training/3/module/3/
Exception Value: 'ModuleProgress' object has no attribute 'get_next_content_or_quiz'
```

4. Reviewing a completed course

After completing a course, returning to http://localhost:8000/settings/platform-training/3/ and clicking "Review" triggers the same `Exception Value: 'ModuleProgress' object has no attribute 'get_next_content_or_quiz'` error.

Rutvikrj26:
Addressing the comments by @tompollard 

1. Added a Check in course creation workflow, ensuring that two course with same name can not be added. Check for appropriate versions have already been added in #1950. 
2. Added the `get_next_content_or_quiz` method in module progress. It was removed while addressing refactor comments from 1950, which might have caused the error. 

The course addition & usage functionality should be working. Please check and let me know.

tompollard:
Thanks for running through this today @Rutvikrj26. As discussed, please could you remove all javascript elements as far as possible? If dynamic features are needed, consider HTMX.

Rutvikrj26:
@bemoody I've updated the URL for the heading and the PR is ready for further review / merge.

bemoody:
A couple of small things:

In training/test_views.py, we don't want to duplicate the TestMixin class.  I don't think this is needed here at all - we can just remove lines 16-169 and have TestPlatformTraining inherit from TestCase instead of TestMixin.

In physionet-django/user/templates/user/edit_training.html:

> You can either complete a training on a third-party platform and update the training or take the training on this platform if available.

The phrasing here is confusing, and particulary doesn't make sense if the platform doesn't have any training.

I would like to redesign this page, and improve all the wording - but that doesn't have to be done in this pull request.  Instead, perhaps just remove the above sentence (or only display this text if `take_course_form` is set.)

Rutvikrj26:
@bemoody I've implemented the updates. 

I agree with you on the fact that the whole page would need updating. currently, there is hardcoded behavior to show a particular template info when CITI training shows up, which is not ideal. We can create a new issue, and list out the desirable changed and what we want it to look like.

bemoody:
The URLs are a little messy:
```
    path('settings/platform-training/<training_slug>/', views.take_training, name='platform_training'),
    path('settings/platform-training/<training_slug>/module/<int:module_id>-<int:order>/', views.current_module_block,
         name='platform_training_module_block'),
    path('settings/platform-training/<training_slug>/next/<int:module_id>-<int:order>/', views.next_module_block,
         name='next_module_block'),
    path('settings/platform-training/<training_slug>/previous/<int:module_id>-<int:order>/',
         views.previous_module_block, name='previous_module_block'),
```
Using the `module_id` (primary key) in the URL is not great.  I would instead use the module's `order` (so you'd have `<int:module_order>-<int:content_order>`).

In the views, there are some issues with the queries.  When you do:
```
    course = Course.objects.filter(training_type__slug=training_slug).last()
```
There's no `order_by`, so if there are multiple versions this isn't guaranteed to give the one you want.  You're also not handling courses that don't exist.

I'm not sure how we *want* to handle multiple versions, if the version should be in the URL or if we should always display the latest version.

If what you want is to fetch the latest active version, you could do something like:
```
    course = Course.objects.filter(
        training_type__slug=training_slug,
        is_active=True,
    ).order_by('-version').first()
    if course is None:
        raise Http404()
```
(or `order_by('version').last()` if you like.)

Similarly, when querying the module:
```
    module = course.modules.filter(pk=module_id).first()
```
Instead, use:
```
    module = get_object_or_404(course.modules, order=module_order)
```

When creating objects, this is okay:
```
    course_progress = CourseProgress.objects.filter(user=request.user, course__id=course.id).last()
    if not course_progress:
        course_progress = CourseProgress.objects.create(user=request.user, course_id=course.id)
```
But this is simpler (note the `_`):
```
    course_progress, _ = CourseProgress.objects.get_or_create(
        user=request.user,
        course=course,
    )
```
and similarly for `ModuleProgress`.

I also notice that `ModuleProgress` doesn't have a uniqueness constraint.  I think this class should have:
```
    class Meta:
        unique_together = [('course_progress', 'module')]
```

Rutvikrj26:
Heyy @bemoody Thanks for the detailed review and I really appreciate your thoughtful comments! I'll keep these points in mind while writing queries now onwards.

bemoody:
* When possible, please use the same name for the function in views.py as the "name" used for the URL in urls.py.  It is confusing to see `redirect('platform_training_module_block', ...)` which corresponds to the function `current_module_block`.

* The `take_training` function is used to do two different things ("start_training" and "platform_training"), and there isn't any real reason for that.

Notice that if you visit the URL `http://localhost:8000/settings/platform-training/`, it will instantly crash.  We don't want to have URLs on PhysioNet that instantly crash.

You could split this into two separate view functions, one that corresponds to "start_training" and one that corresponds to "platform_training".  Or you could move the "start_training" logic into `user.views.edit_training`.

* Notice also that if you visit the URL `http://localhost:8000/settings/platform-training/asdfgjkl/`, it will crash.  Use `get_object_or_404` here.

* The use of "last_completed_order" is confusing.  It also doesn't seem to work properly.

If I start the course, go to `http://localhost:8000/settings/platform-training/world-101-introduction-to-continents-and-countries/module/1-1/`, and click the Next button, the link takes me to `.../next/1-1/`, which redirects to `.../module/1-2/`, which redirects back to `.../module/1-1/`.

Previously the default value for `last_completed_order` was zero.  Here you're setting it to 1 by default - that seems wrong.  It seems to me that, if the field is called "last completed order", then it should only be set to 1 when the trainee has *completed* block 1 (and answered the quiz if any.)

You could make things easier by saying that the block order has to begin with 1 and increase by 1 for each block.  Then, if the last completed block is `N`, the next block to be completed is `N + 1`.

I would try to fold `current_module_block` and `next_module_block` into one function.

The "back" button could just be a link to the previous block (use `reverse` to get the appropriate URL) - `previous_module_block` isn't really needed.

The "submit" or "next" button is a form submit button.  When submitted, this checks if the responses were correct, then updates the trainee's progress if necessary, and redirects to the next block.

* I don't think the following is needed:
```
        training_questions = []
        for question in course.training_type.questions.all():
            training_questions.append(TrainingQuestion(training=training, question=question))

        TrainingQuestion.objects.bulk_create(training_questions)
```
(TrainingQuestion refers to the administrator's review checklist - there is no administrator review here.)

Rutvikrj26:
@bemoody Thanks for your review!
I did see there were a few loose ends in the way views and routing was managed. Furthermore as you pointed out, there were logical issues as well in the views. I have fixed it now, and tested it on my end to run through everything. 
Furthermore, since the current_module_block was becoming too big ( the original reason I did not want both in one) I broke it down into utility functions to make it more readable and manageable. But those utility functions are being used just once, and hence am not really sure if it is a good idea to have them out there. 

Let me know your thoughts on it. @tompollard I'd also appreciate your thoughts on the matter. 

Apart from that, the logical flow of the view is now as benjamin suggested, with next & submit being form actions, that verify, update the database and redirect to the next module block.

tompollard:
Thanks Rutvik, this is definitely getting there! I've added a few thoughts below:

### 1. Location of the "update" button in the console

Minor thing, but I think the "Update" button [on the courses page](http://localhost:8000/console/courses/world-101-introduction-to-continents-and-countries/) would look much cleaner if it sat in the table next to the relevant course (to the left or right of the "expire" button):

Screenshot 2024-07-15 at 2 49 32 PM

### 2. "View" training on Certification page

When I go to the [Certification page](http://localhost:8000/settings/certification/) in the user settings after completing the training,  I see a button to "View" my World 101 training in the Active panel.

When I click view, I see text saying "World 101: Introduction to Continents and Countries Training
Submitted: July 15, 2024". I already completed the course, so why does it say "Submitted?". Shouldn't it say "Active"?

### 3. Add to "In Progress" tab to the Certification page?

Perhaps there should be an "In progress" Tab on the the [Certification page](http://localhost:8000/settings/certification/) to list courses that are in progress?

### 4. Clicking the "Continue" button to continue a course raises a `NoReverseMatch` error

- Go to http://localhost:8000/settings/platform-training/world-101-introduction-to-continents-and-countries/
- Click "Start"
- Go to http://localhost:8000/settings/platform-training/world-101-introduction-to-continents-and-countries/
- Click "Continue"
Raises:

```
NoReverseMatch at /settings/platform-training/world-101-introduction-to-continents-and-countries/module/1-0/
Reverse for 'current_module_block' with keyword arguments '{'training_slug': 'world-101-introduction-to-continents-and-countries', 'module_order': 1, 'order': ''}' not found. 1 pattern(s) tried: ['settings/platform\\-training/(?P<training_slug>[^/]+)/module/(?P<module_order>[0-9]+)\\-(?P<order>[0-9]+)/\\Z']
```

### 5. What is the expected behaviour of the "Review" button for completed courses

When I click Review, I am asked to redo the course (i.e. I need to provide answers to each question to continue. Is this the intended behaviour?

### 6. Behaviour for "expiring" a course is unclear

The effect of expiring a course should be clear to the user.

As far as I can see, when I go to the [management page for a course in the console](http://localhost:8000/console/courses/world-101-introduction-to-continents-and-countries/) and click "Expire" to expire a course:

- The course is moved to "Archived" (the expiry date that I set does not appear to be listed anywhere).
- The course is immediately unavailable to users
- Anyone in the middle of the course gets a 404 when trying to complete it.

Some improvements would be:

- Display the expiry date to the admin user and explain impact of the course expiring.
- Display the expiry date in the user settings and explain impact of the course expiring.
- Ideally do something better than returning a 404 to someone in the middle of the course (though perhaps save for later PR).
- 
### 7. It is possible to start training that is complete

If I go to http://localhost:8000/settings/training/, I have the option to start the World 101 training, even though I already completed this training. 

It then says "Are you sure you want to start the training?". When I say yes, I am taken to a page that indicates the training is complete. Probably we should make it clearer to the user that the training is already complete and they cannot restart it.

Rutvikrj26:
> Location of the "update" button in the console

Versions of a course are immutable by design (a course version that has been published and taken by users are certified for that particular version of the course - and hence once published on the platform, it cannot be - updated)
What the update button does is, it created a new version. 

May be I can rename the button to `upload new version` which might make it more clear.

> 2. "View" training on Certification page

This was abstracted out to a different PR #2223 as it dealt with updating the certifications page and had an updated in the queryset logic. It is a followup PR on this, and will be addressed in it. 

> ### Add to "In Progress" tab to the Certification page?

This sounds like a good idea - I'll update the #2223 with this as a requirement.

> Clicking the "Continue" button to continue a course raises a `NoReverseMatch` error

I haven't been able to replicate the error.

> What is the expected behaviour of the "Review" button for completed courses

Yes, the review is just to go through the text once again to brush up anything if the user wants. Given that these are going to be basic mandatory courses on it security or something similar, it doesn't have any real purpose.



> * Display the expiry date to the admin user and explain impact of the course expiring.
> * Display the expiry date in the user settings and explain impact of the course expiring.
> * Ideally do something better than returning a 404 to someone in the middle of the course (though perhaps save for later PR).

The Expiry date is the date the admin chooses on the date field on the side of the expire button. I can add a new column of date for the archived courses and then add a line at the top of the archived table detailing that the trainings allotted to the people who took that version of the course will no longer be valid after the expiry date listed. And yes, a new issue/pr sounds better for that, as this one is long as it is. 
(edit) as I see, that note is already there. 

> Are you sure you want to start the training?

Agreed, this looks a bit misleading. It should rather say open the training page, where you can then choose to start/continue/review the training modules.

Rutvikrj26:
@tompollard I've made the changes you requested on switching from expiring to archiving and simplifying the logic. However, there is a weird issue I am facing in the tests. The test keeps giving 404 error - which it technically shouldn't.

The two url endpoints failing in the tests are:
`/settings/platform-training/world-101-introduction-to-continents-and-countries/`
`/settings/platform-training/world-101-introduction-to-continents-and-countries/module/1-1/`

But these work perfectly fine when I open them in my desktop. I tried figuring out what's going on, but couldn't figure out the error. One reason of something like this happening is the training element not getting loaded up - but the loaddemo command loads up the demo training when I test locally.
@tompollard @bemoody any ideas what could be the issue here.

tompollard:
@Rutvikrj26 odd, it's not immediately clear to me why the tests are failing. i added a breakpoint at:

https://github.com/MIT-LCP/physionet-build/blob/75b7f911378b5e36422c5745025bfd533c18f688/physionet-django/physionet/test_urls.py#L191-L194

...to catch the error. The url and parameters appear to be set correctly (`'/settings/platform-training/world-101-introduction-to-continents-and-countries/'`, `<User: tompollard>`) but as you say, the page returns a 404.

`Course.objects.all()` returns: `<QuerySet [<Course: World 101: Introduction to Continents and Countries v1.0>]>` so it looks like the data is loaded.

tompollard:
@Rutvikrj26 It looks like the issue is that the course is not active. If you drop a breakpoint() around `self.assertLess(response.status_code, 400) `:

```
try:
    self.assertLess(response.status_code, 400) 
except:
    breakpoint()
```

Then run:

```
from training.models import Course
course = Course.objects.all()[0]
```

You'll see that `course.is_active` returns `False`. Perhaps the course gets archived in an earlier test? (Fix is to reverse the archive action that happens previously, or add another course to the fixtures).

physionet-django/training/templates/training/course_block.html
```diff
@@ -0,0 +1,71 @@
...
+            {% endif %}
+            <br>
+            <div id="alert" class="alert" role="alert"></div>
+        </div>
+    </div>
+{% endblock %}
+
+{% block local_js_bottom %}
+    <script src="{% static 'custom/js/resize-ck.js' %}"></script>
+{% endblock %}
```
tompollard: missing newline

physionet-django/training/views.py
```diff
@@ -1,31 +1,87 @@
...
+def get_min_max_order_in_module(module):
+    # calculating the max and min order of the module
+    module_blocks = ContentBlock.objects.filter(module=module).values_list('order', flat=True)
+    quiz_blocks = Quiz.objects.filter(module=module).values_list('order', flat=True)
+    module_max_order = max(max(module_blocks, default=0), max(quiz_blocks, default=0))
+    module_min_order = min(min(module_blocks, default=0), min(quiz_blocks, default=0))
+    return module_min_order, module_max_order
+
+
+def handle_course_completion(course, course_progress):
```
tompollard, line 83: Please could you add docstrings for these functions. Their purpose isn't always clear (e.g. "handle course completion" does not give much information about the purpose of this function).

physionet-django/training/views.py
```diff
@@ -1,31 +1,87 @@
...
+    course_progress.status = CourseProgress.Status.COMPLETED
+    course_progress.save()
```
tompollard: Should the `course_progress` and `training` transactions be atomic (in `with transaction.atomic():`?)

physionet-django/console/urls.py
```diff
@@ -166,7 +166,7 @@
     path('courses/<training_slug>/download/<str:version>',
          training_views.download_course, name='download_course_version'),
     path('courses/<training_slug>/expire/<str:version>',
```
tompollard: Should `expire` be updated to `/archive/` in the URL?

physionet-django/static/custom/js/quiz.js
```diff
@@ -0,0 +1,138 @@
+
+
+function changeSlide(n) {
```
tompollard: Is all of this javascript still necessary?

physionet-django/training/models.py
```diff
@@ -38,14 +36,12 @@ class Meta:
             ("can_view_course_guidelines", "Can view course guidelines"),
         ]
 
-    def expire_course_version(self, instance, number_of_days):
+    def archive_course_version(self):
         """
         This method expires the course by setting the is_active field to False and expires all
```
tompollard: Change expires to archives in the docstring? 

I think "expires" should just be used when referring to training that has reached the expiry date (i.e. if someone takes a course that is valid for one year, the course expires when a year has elapsed).

Rutvikrj26:
@tompollard I've addressed all the comments and updated the fixtures to have a fallback version of the course so that even after archiving tests works fine.

Rutvikrj26:
@tompollard I need this PR to be merged to complete the work on certifications page #2223. Please merge it if everything looks okay.

physionet-django/training/views.py
```diff
@@ -175,3 +249,103 @@ def download_course(request, training_slug, version):
     response = JsonResponse(response_data, safe=False, json_dumps_params={'indent': 2})
     response['Content-Disposition'] = f'attachment; filename={training_type.name}--version-{version}.json'
     return response
+
+
+@login_required
+def take_training(request, training_slug):
```
tompollard, line 255: This function should have a docstring, at minimum explaining its purpose.

physionet-django/training/views.py
```diff
@@ -175,3 +249,103 @@ def download_course(request, training_slug, version):
...
+    return render(request, 'training/course.html', {
+        'course': course,
+        'modules': modules,
+        'ModuleStatus': ModuleProgress.Status,
+    })
+
+
+@login_required
+def current_module_block(request, training_slug, module_order, order):
+
```
tompollard, line 325: This function should have a docstring, at minimum explaining its purpose. I think we also want to try to work towards simpler functions at some point, perhaps moving logic to a separate module.

Rutvikrj26:
@tompollard @bemoody There is a weird error in the Testing where it fails while running Docker commands saying docker-compose is command is not available. Any recent changes that might have triggered this?

tompollard:
> There is a weird error in the Testing where it fails while running Docker commands saying docker-compose is command is not available. Any recent changes that might have triggered this?

@Rutvikrj26 Should be fixed by https://github.com/MIT-LCP/physionet-build/pull/2269. Please could you rebase on the dev branch?

Rutvikrj26:
@tompollard Looks good to merge.

tompollard:
@Rutvikrj26 it looks like the user migrations are stale ([0061_cloudinformation_aws_userid_and_more.py](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/user/migrations/0061_cloudinformation_aws_userid_and_more.py) collides with [0061_training_course.py](https://github.com/MIT-LCP/physionet-build/pull/1951/files#diff-af31e53ab7e97b19ed0f2f67332df5890000e5ca1914c06cc1524b8e3d564227)). 

Please could you regenerate the migrations so we can go ahead and merge? This is also an opportunity to rebase and clean up commit history if you'd like to do that.

Rutvikrj26:
@bemoody @tompollard I've squashed the commits and converted it to one - plus added migrations separately.

A quick question - The `python manage.py resetdb` command fails to run in the GitHub action workflow. I tried re-running with debugging logs but can't get to the bottom of why it is occurring. Locally, I've been able to test and run it - so can't reproduce for further debugging as well.

tompollard:
@Rutvikrj26 in case you didn't receive the notification, tests are failing.

Rutvikrj26:
> @Rutvikrj26 in case you didn't receive the notification, tests are failing.

@tompollard I'm not looking to get this one merged yet. I saw the tests are failing. The comment was to ask if you have an idea of what could be causing this, as I am not able to replicate the command failure locally.

tompollard:
> @tompollard I'm not looking to get this one merged yet. I saw the tests are failing. The comment was to ask if you have an idea of what could be causing this, as I am not able to replicate the command failure locally.

Oh sorry, I'd missed the earlier question. I'll try to take a look shortly/

Rutvikrj26:
> > @tompollard I'm not looking to get this one merged yet. I saw the tests are failing. The comment was to ask if you have an idea of what could be causing this, as I am not able to replicate the command failure locally.
> 
> Oh sorry, I'd missed the earlier question. I'll try to take a look shortly/

I think I just fixed it. I had generated migrations but for some reason, it did not get added to tracking files. This should fix the issue.

tompollard:
> I think I just fixed it. I had generated migrations but for some reason, it did not get added to tracking files. This should fix the issue.

Easy to miss new, untracked files!

Rutvikrj26:
@tompollard The PR is now finally ready to be merged!

Issue #1930, opened 2023-03-13
Add asterisk to reference requirement on credential applications


mscanlan-git:
As discussed in #1804, we've made changes to the reference requirements to ask everyone to provide a reference, regardless of their application category (student, researcher, etc.).

What was left out from this, however, is providing an asterisk for the new reference requirement as seen below: 

![image](https://user-images.githubusercontent.com/67070862/224759080-2e8416a4-bde5-4a2b-b913-2a8fa5be54e0.png)

If we're going to require everyone to provide a reference, it should be shown with an asterisk so applicants understand that this is a requirement, not an optional input.

PR #1960, opened 2023-03-30
adds reference requirement asterisk

lamawmouk:
This PR addresses the issue in #1930. 

It updates the `CredentialApplication` model where if a user submits a form that does not include the reference field, then the form won't be processed and will raise  a validation error. This PR removes the blank=True to indicate that the field is required to be filled and thus will render the asterisks for reference required fields in the reference form.

The reference form  found under  the `Credential_application.html` includes another template `descriptive_inline_form_snippet.html`.  The reference form is rendered using the `descriptive_inline_form_snippet.html` partial template.  This template renders the form object `reference_form` and has a conditional statement that checks if a form field is required or not, and adds a red asterisk symbol (*) next to the field label if the field is required. The pull request will render the asterisks for reference required fields in the reference form.


tompollard:
(it looks like you are working on an old branch. you should be able to fix the tests by deleting the migration, rebasing on dev, and then regenerating the migration).

tompollard:
Thanks Lama, looks good to me.

Issue #1927, opened 2023-03-09
Integrate local file access and research environments with the existing `DataAccess` model


kshalot:
(I used orgmode to generate this so now it looks more serious than it actually is)

# Table of Contents

1.  [Overview](#orge93e257)
    1.  [Use the existing `DataAccess` abstraction to get rid of implicit project access](#orgcd1c1c7)
        1.  [File downloads](#org43a40d1)
        2.  [Research environments](#org2ab5ff6)
        3.  [Events](#org71dc39c)
        4.  [Defaults](#org4f08a29)
        5.  [Cleanup](#orga2235f5)
    2.  [Refactor](#orgbaffb4a)

<a id="orge93e257"></a>

# Overview

<a id="orgcd1c1c7"></a>

## Use the existing `DataAccess` abstraction to get rid of implicit project access

Before refactoring the `DataAccess` functionality itself, the mechanism could be made generic (i.e. local file access and research environment access should be controlled by the `DataAccess` model). This will make any access via any source fall under the same mechanism so it&rsquo;s easier to reason about a potential refactor in the future.

<a id="org43a40d1"></a>

### File downloads

-   `DataAccess` already contains a `local` platform which is unused.
-   A project being accessible via file download would be denoted as `DataAccess(platform=0, location=None)`.
-   Steps:
    1.  Move the files panel into the [published\_project\_data\_access](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/templates/project/published_project_data_access.html) template.
    2.  Control displaying of the file panel via `if item.platform == 0`.
    3.  Check whether the `local` access is enabled before serving files/zip (technically, the Google Cloud Storage `ProjectFiles` backend should also support `wget`. This can be aligned later when needed).
    4.  Create a data migration/script (apply on PhysioNet only) that will create a local `DataAccess` for each PublishedProject.

<a id="org2ab5ff6"></a>

### Research environments

-   A project being accessible via a research environment would be denoted as `DataAccess(platform=5, location="GCP group controlling access to the dataset")`.
-   Steps:
    1.  Add a `research_environment` platform to `DataAccess`.
    2.  (Optionally) The platform should be hidden unless `ENABLE_CLOUD_RESEARCH_ENVIRONMENTS` is set to `True`.
    3.  Add a new branch to [published\_project\_data\_access](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/project/templates/project/published_project_data_access.html):
[...]

PR #1961, opened 2023-03-31
refactor authorization to separate module

superryeti:
Context:
This PR is a small part of https://github.com/MIT-LCP/physionet-build/issues/1927#issuecomment-1468633078
On this PR, i tried to move all the code related to authorization into a separate module, so that in future, we would only need to worry about making changes here(like maybe once `DataSource` is refactored` we could just change the implementation inside the authorization module).

Here the summary of change is that i only copied the code from `PublishedProject.has_access` and `PublishedProject.objects.accessible_by`(used by healthdatanexus) into `projects.authorization.access. can_access_project` and `projects.authorization.access.get_accessible_projects`

To take this even further and make everything super simply, we might be able to put all the authorization code into just a single method `projects.authorization.access.get_accessible_projects`  and use this instead of `project.has_access`

For example:
Currently, to check if a user has access to a project
```
has_access = project.has_access(user)
```

we could do the following using just the single new method `get_accessible_projects`

```
>>> from project.authorization.access import get_accessible_projects
>>> from user.models import User
>>> user = User(username='amitupreti')
>>> queryset = get_accessible_projects(user)
>>> has_access = queryset.contains(project) #boolean result
```

### DIFF
--- a/physionet-django/project/authorization/access.py
+++ b/physionet-django/project/authorization/access.py
@@ -86,7 +86,13 @@ def get_accessible_projects(user):
 
 
 def can_access_project(project, user):
-    """Checks if the project is accessible by the user"""
+    """
+    Checks if the project is accessible by the user
+    Users may access a project through different ways, for example, thorough direct download on the physionet website,
+    or through s3 bucket links, or gcs storage.
+    This function only checks access to the project in general, users might still not be able to access the files
+    even if they can access the project.
+    """
     if project.deprecated_files:
         return False
 
@@ -122,3 +128,11 @@ def can_access_project(project, user):
             == project.required_trainings.count()
         )
     return False
+
+
+def can_view_project_files(project, user):
+    """
+    Checks if the project files are  directly accessible by the user
+    Currently used to allow direct file downloads and to show project files on the platform
+    """
+    return can_access_project(project, user) and project.allow_file_downloads
--- a/physionet-django/project/authorization/access.py
+++ b/physionet-django/project/authorization/access.py
@@ -1,0 +1,61 @@
+from datetime import datetime
+
+from django.db.models import Q
+
+from events.models import Event, EventDataset
+from project.models import AccessPolicy, DUASignature, DataAccessRequest, PublishedProject
+from user.models import Training, TrainingType
+
+
+def get_accessible_projects(user):
+    """
+    Returns query filter for published projects accessible by a specified user
+    """
+    query = Q(access_policy=AccessPolicy.OPEN) & Q(deprecated_files=False)
+
+    dua_signatures = DUASignature.objects.filter(user=user)
+
+    if user.is_authenticated:
+        query |= Q(access_policy=AccessPolicy.RESTRICTED) & Q(
+            duasignature__in=dua_signatures
+        )
+
+    if user.is_credentialed:
+        completed_training = (
+            Training.objects.get_valid()
+            .filter(user=user)
+            .values_list("training_type")
+        )
+        not_completed_training = TrainingType.objects.exclude(
+            pk__in=completed_training
+        )
+        required_training_complete = ~Q(
+            required_trainings__in=not_completed_training
+        )
+
+        accepted_data_access_requests = DataAccessRequest.objects.filter(
+            requester=user, status=DataAccessRequest.ACCEPT_REQUEST_VALUE
+        )
+        contributor_review_with_access = Q(
+            access_policy=AccessPolicy.CONTRIBUTOR_REVIEW
+        ) & Q(data_access_requests__in=accepted_data_access_requests)
+
+        credentialed_with_dua_signed = Q(
+            access_policy=AccessPolicy.CREDENTIALED
+        ) & Q(duasignature__in=dua_signatures)
+
+        query |= required_training_complete & (
+            contributor_review_with_access | credentialed_with_dua_signed
+        )
+
+    # add projects that are accessible through events
+    events_all = Event.objects.filter(Q(host=user) | Q(participants__user=user))
+    active_events = set(events_all.filter(end_date__gte=datetime.now()))
+    accessible_datasets = EventDataset.objects.filter(event__in=active_events, is_active=True)
+    accessible_projects_ids = []
+    for event_dataset in accessible_datasets:
+        if event_dataset.has_access(user):
+            accessible_projects_ids.append(event_dataset.dataset.id)
+    query |= Q(id__in=accessible_projects_ids)
+
+    return PublishedProject.objects.filter(query).distinct()
--- a/physionet-django/project/authorization/access.py
+++ b/physionet-django/project/authorization/access.py
@@ -7,56 +7,80 @@ from project.models import AccessPolicy, DUASignature, DataAccessRequest, Publis
 from user.models import Training, TrainingType
 
 
-def get_accessible_projects(user):
-    """
-    Returns query filter for published projects accessible by a specified user
-    """
-    query = Q(access_policy=AccessPolicy.OPEN) & Q(deprecated_files=False)
+def get_public_projects_query():
+    """Returns query filter for public published projects"""
+    return Q(access_policy=AccessPolicy.OPEN)
 
+
+def get_restricted_projects_query(user):
+    """Returns query filter for restricted published projects accessible by a specified user"""
+    dua_signatures = DUASignature.objects.filter(user=user)
+    query = Q(access_policy=AccessPolicy.RESTRICTED) & Q(duasignature__in=dua_signatures)
+    return query
+
+
+def get_credentialed_projects_query(user):
+    """Returns query filter for credentialed published projects accessible by a specified user"""
     dua_signatures = DUASignature.objects.filter(user=user)
 
-    if user.is_authenticated:
-        query |= Q(access_policy=AccessPolicy.RESTRICTED) & Q(
-            duasignature__in=dua_signatures
-        )
+    completed_training = (
+        Training.objects.get_valid()
+        .filter(user=user)
+        .values_list("training_type")
+    )
+    not_completed_training = TrainingType.objects.exclude(pk__in=completed_training)
+    required_training_complete = ~Q(required_trainings__in=not_completed_training)
 
-    if user.is_credentialed:
-        completed_training = (
-            Training.objects.get_valid()
-            .filter(user=user)
-            .values_list("training_type")
-        )
-        not_completed_training = TrainingType.objects.exclude(
-            pk__in=completed_training
-        )
-        required_training_complete = ~Q(
-            required_trainings__in=not_completed_training
-        )
+    accepted_data_access_requests = DataAccessRequest.objects.filter(
+        requester=user, status=DataAccessRequest.ACCEPT_REQUEST_VALUE
+    )
 
-        accepted_data_access_requests = DataAccessRequest.objects.filter(
-            requester=user, status=DataAccessRequest.ACCEPT_REQUEST_VALUE
-        )
-        contributor_review_with_access = Q(
-            access_policy=AccessPolicy.CONTRIBUTOR_REVIEW
-        ) & Q(data_access_requests__in=accepted_data_access_requests)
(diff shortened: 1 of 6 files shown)


physionet-django/project/authorization/access.py
```diff
@@ -0,0 +1,100 @@
...
+    elif project.access_policy == AccessPolicy.RESTRICTED:
+        return user.is_authenticated and DUASignature.objects.filter(project=project, user=user).exists()
+    elif project.access_policy == AccessPolicy.CREDENTIALED:
+        return (
+            user.is_authenticated
+            and user.is_credentialed
+            and DUASignature.objects.filter(project=project, user=user).exists()
+            and Training.objects.get_valid()
+            .filter(training_type__in=project.required_trainings.all(), user=user)
+            .count()
```
tompollard, line 113: seems odd to be comparing a count here?

physionet-django/project/authorization/access.py
```diff
@@ -0,0 +1,100 @@
...
+            and user.is_credentialed
+            and DataAccessRequest.objects.get_active(
+                project=project,
+                requester=user,
+                status=DataAccessRequest.ACCEPT_REQUEST_VALUE
+            ).exists()
+            and Training.objects.get_valid()
+            .filter(training_type__in=project.required_trainings.all(), user=user)
+            .count()
+            == project.required_trainings.count()
```
tompollard, line 128: as above, odd to be comparing count here

physionet-django/project/authorization/access.py
```diff
@@ -0,0 +1,100 @@
...
+def get_accessible_projects(user):
+    """
+    Returns query filter for published projects accessible by a specified user
+    """
+    query = Q(access_policy=AccessPolicy.OPEN) & Q(deprecated_files=False)
+
+    dua_signatures = DUASignature.objects.filter(user=user)
+
+    if user.is_authenticated:
+        query |= Q(access_policy=AccessPolicy.RESTRICTED) & Q(
```
tompollard: We are not excluding deprecated files here, so does this mean that the filter will return deprecated files (despite them being excluded in the earlier step?). I might be misunderstanding how query construction works, so just checking...

physionet-django/project/authorization/access.py
```diff
@@ -0,0 +1,100 @@
...
+        credentialed_with_dua_signed = Q(
+            access_policy=AccessPolicy.CREDENTIALED
+        ) & Q(duasignature__in=dua_signatures)
+
+        query |= required_training_complete & (
+            contributor_review_with_access | credentialed_with_dua_signed
+        )
+
+    # add projects that are accessible through events
+    events_all = Event.objects.filter(Q(host=user) | Q(participants__user=user))
```
tompollard, line 54: It feels messy for events to appear here, but perhaps the refactoring is just making the problems in our current approach more apparent (and so its a good thing that this seems awkward, because it is a step towards us fixing it!).

physionet-django/project/authorization/access.py
```diff
@@ -0,0 +1,100 @@
+from datetime import datetime
+
+from django.db.models import Q
+
+from events.models import Event, EventDataset
+from project.models import AccessPolicy, DUASignature, DataAccessRequest, PublishedProject
+from user.models import Training, TrainingType
+
+
+def get_accessible_projects(user):
```
tompollard: The logic in this function is difficult to follow. I wonder whether splitting it into some subfunctions might help with readability?

tompollard (commented):
Thanks @amitupreti, I think moving to a new module is a good idea. I've added a few thoughts, and hopefully @kshalot will have a chance to look too.

physionet-django/project/authorization/access.py
alistairewj: should raise an issue about this since it has to do with handling expired trainings

superryeti:
Opened a new issue here for the discussion on the existing code https://github.com/MIT-LCP/physionet-build/issues/1979

physionet-django/project/authorization/access.py
superryeti: Thanks @tompollard  working on this. Also looks like another change by Rafal was merged which kinda of make the new logic a bit difficult to rebase quickly, i will edit the last commit on this PR and make changes.

superryeti:
Hi @tompollard Ready for review, Quick summary
1.  i broke down the function into smaller modules. 
2. Updated code to work with Rafal's changes

tompollard:
@amitupreti this basically looks good to me, except that there are several new functions (e.g. `get_accessible_projects`) that don't appear to be used anywhere. Please could you provide a bit more detail on these unused functions? Are they intended to be included in this PR?

superryeti:
> get_accessible_projects

Ah sorry, yes. let me explain here, and also update the PR description. 
Yes about the functions being intended to be included in this PR.

Currently, the PublishedProject#accessible_by is only used by HDN, to load a list of project for a user.
This get_accessible_projects is a the new version of PublishedProject#accessible_by which is supposed to be used by the hdn-research-environment package.

The other smaller functions are just breakdown of `get_accessible_projects` to make it more readable

tompollard:
Thanks Amit, and one other quick question: Should the original functions/methods (e.g. the one below) be removed in this PR, or is there a reason to keep them in place?

```
    def can_view_files(self, user):
        """
        Whether the user has access to this project and files. Extends 'is_authorized'
        """
        return self.allow_file_downloads and self.is_authorized(user)
```

superryeti:
> Thanks Amit, and one other quick question: Should the original functions/methods (e.g. the one below) be removed in this PR, or is there a reason to keep them in place?
> 
> ```
>     def can_view_files(self, user):
>         """
>         Whether the user has access to this project and files. Extends 'is_authorized'
>         """
>         return self.allow_file_downloads and self.is_authorized(user)
> ```

The reason i let them be was because the `hdn-research-environment` is still using `PublishedProject#accessible_by` and will probably use it for few days at least. So my idea was to remove the old functions at once in another PR once this is merged and `hdn-research-environment` is also updated.

Now that i think about it we should remove the  `is_authorized` and `can_view_files`. Please let me know if you think that makes sense or if we should remove it later

superryeti:
i found one place(content_list.html) where i had missed to change the code, please wait before merging, i am working on it

superryeti:
Thanks @tompollard , Updated the code in template and removed the two methods

`PublishedProject.is_authorized` and `PublishedProject.can_view_files`

physionet-django/project/authorization/access.py
```diff
@@ -0,0 +1,123 @@
...
+    return Q(access_policy=AccessPolicy.OPEN)
+
+
+def get_restricted_projects_query(user):
+    dua_signatures = DUASignature.objects.filter(user=user)
+    query = Q(access_policy=AccessPolicy.RESTRICTED) & Q(duasignature__in=dua_signatures)
+    return query
+
+
+def get_credentialed_projects_query(user):
```
tompollard, line 22: Please could you add docstrings for all of the new functions?

physionet-django/project/authorization/access.py
```diff
@@ -0,0 +1,123 @@
...
+        if event_dataset.has_access(user):
+            accessible_projects_ids.append(event_dataset.dataset.id)
+    query = Q(id__in=accessible_projects_ids)
+    return query
+
+
+def get_accessible_projects(user):
+    """
+    Returns query filter for published projects accessible by a specified user
+    """
```
tompollard, line 72: Please could you try to use consistent formatting? This function has lots of white space. The previous function (`get_projects_accessible_through_events`) is dense!

physionet-django/project/authorization/access.py
```diff
@@ -0,0 +1,123 @@
...
+            ).exists()
+            and Training.objects.get_valid()
+            .filter(training_type__in=project.required_trainings.all(), user=user)
+            .count()
+            == project.required_trainings.count()
+        )
+    return False
+
+
+def can_view_project_files(project, user):
```
tompollard, line 133: I think in an earlier version of the PR, this was just `can_view_files`? There is now inconsistent language. Sometimes we use `can_view_project_files` and sometimes we use `can_view_files`, I think for the same concept?

physionet-django/project/authorization/access.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    Checks if the project files are accessible by the user
```
tompollard: It would be good to use the docstrings as an opportunity to clarify the difference between `can_view_project_files` and `can_access_project`. Currently I don't think we make the distinction clear:

- The docstring for `can_access_project` is: "Checks if the project is accessible by the user"; 
- The docstring for `can_view_project_files` is: "Checks if the project files are accessible by the user".

physionet-django/project/authorization/access.py
```diff
@@ -0,0 +1,123 @@
...
+
+    if user.is_credentialed:
+        query |= get_credentialed_projects_query(user)
+
+    query |= get_projects_accessible_through_events(user)
+
+    return PublishedProject.objects.filter(query).distinct()
+
+
+def can_access_project(project, user):
```
tompollard, line 88: Thinking about the function names, why does `can_view_project_files` include "files", but `can_access_project` does not? Does it relate to the purpose of these functions?:

1) e.g. is `can_access_project` determining whether or not someone can access the project description and files (e.g. for an unpublished project) while
2) `can_view_project_files` specifically relates to the files?

If so, makes sense! Just checking...

physionet-django/project/authorization/access.py
superryeti: My bad. let me work on them. i will remember in future. Thank you

physionet-django/project/authorization/access.py
superryeti: > I think in an earlier version of the PR, this was just can_view_files

yes, sorry about this, i was trying to make it consistent(trying to make`can_access_project` and `can_view_project_files ` similar) but ended up with this looking weird.

Do you think it would be consistent with everything `can_view_project_files` . Right now the reason we have `can_view_files` in the template tag is because we need to import `can_view_project_files`  on template tag so i didnot make the name of tag `can_view_project_files`

Now that i think about it, we could have also imported the access.py as a whole and use `access.can_view_project_files` in the template tag

physionet-django/project/authorization/access.py
superryeti: From what i saw in the code before refactor we had

`is_authorized` ---> Checks if a user has access to project (user might still not have access to the files download/view)
`can_view_files`--->  checks `is_authorized`+ if the allow_file_downloads boolean is turned on for the project

So here also, my idea was has `can_access_project` checks if they have access to the project (files might not be accessible if `allow_file_downloads` is off) -->also  on the User interface, i think if this boolean is off, we dont display the files to the user on project page

superryeti: Sorry, if i made things further confusing or misunderstood. Happy to jump on a call to clarify things quickly

superryeti: I updated the doc string, to try to explain this better.

physionet-django/project/views.py
```diff
@@ -1802,8 +1803,8 @@ def published_project(request, project_slug, version, subdir=''):
     an_url = request.get_signed_cookie('anonymousaccess', None, max_age=60 * 60)
     has_passphrase = project.get_anonymous_url() == an_url
 
-    can_view_files = project.can_view_files(user) or has_passphrase
-    is_authorized = project.is_authorized(user) or has_passphrase
+    can_view_project_files = can_view_project_files(project, user) or has_passphrase
```
tompollard: Are you okay with overwriting the function with a boolean? I feel like it would be better to avoid this.

superryeti: sorry, happened with autorefactor, made the change already and force pushed

physionet-django/project/authorization/access.py
superryeti: refactored all function and template tag to can_view_project_files

tompollard:
Thanks Amit! Looks good to me.

Issue #1962, opened 2023-03-31
New layout for CITI Course Instructions Page


mscanlan-git:
The CITI course instructions page is a bit jarring to view as a new user who isn't familiar with the platform as the layout is more complicated than it needs to be. We can simplify this layout by taking the instructions and integrating them into the screenshots so that users don't need to refer to figures every time they are trying to follow said instructions. 

This is currently what the layout looks like: 

<img width="834" alt="Screenshot 2023-03-31 at 12 08 35 PM" src="https://user-images.githubusercontent.com/67070862/229173389-cc8e9f56-7e78-46d0-ac67-63926b39e743.png">

My proposed new layout would look like this:

<img width="825" alt="Screenshot 2023-03-31 at 12 05 03 PM" src="https://user-images.githubusercontent.com/67070862/229173552-0e9023b1-d668-4605-82af-01ad05f92432.png">

The actual content (images and text) will be updated to better reflect what the new CITI course website layout looks like, plus, simplifying the language and instructions by making the navigation easier for new users. The images and text will be updated in a future issue/pull request, but for now, I wanted to demonstrate what the new layout would like with the text paired with the images that are associated with it.

I am looking for feedback on this new layout and would like any criticism that could better improve it. Thanks.

_Again, please ignore the actual images and text, this will be updated._

tompollard:
@mscanlan-git I agree that the [current layout](https://physionet.org/about/citi-course/) could be improved. In your proposed layout, I think the sequence of steps gets a little lost.  Instead, could we display the text prompts and screenshots in order, similar to the example below?: 

> 1. If your institution is affiliated with the CITI program, please log in to the CITI site using your institutional portal. Otherwise, visit the CITI program site: https://www.citiprogram.org/, create an account if you don’t already have one, and log in.

> IMAGE 1

> 2. Once you are logged in to the CITI site, open the tab “Click here to affiliate with another institution” and click on the link (fig 1)). Select your organizational affiliation as “Massachusetts Institute of Technology Affiliates”.

> IMAGE 2

> 3. After affiliating with MIT, follow the links to add a Massachusetts Institute of Technology Affiliates course (fig 2). In the Human Subjects training category, select the “Data or Specimens Only Research” course (fig 3). Answer the other questions as appropriate and submit the form.

> IMAGE 3

> etc...

mscanlan-git:
[...]

PR #1963, opened 2023-04-03
CITI Course Instructions Page Layout

mscanlan-git:
Re-organized the CITI Course instructions page on how to become a credentialed user. Moved the text from the block paragraph into the images and made images descent vertically, making navigation easier for new users who are working on their credential application.

All these changes were discussed originally in #1962


tompollard:
Thanks @mscanlan-git, I think this is improvement on the previous content! 

Before merging, it would be good if we could:

1) Remove the old screenshots, which are now unused.
2) Clean up the filenames (no need to have `...3_edit.png` or `..._v2.png`)

I think the styling could be improved (e.g. perhaps by adding a border to the screenshots and maybe putting them inside the list items so they are indented), but this can always be addressed in a later pull request.

tompollard:
You could add something like this to `physionet-django/static/custom/css/physionet.css`:

```
.screenshot {
  width: 600px;
  border: 1px solid rgba(160, 160, 160, 0.15);
  margin-top: 1rem;
  margin-bottom: 1rem;
}
```

Then replace `style="width:600px"` with `class="screenshot"` in `physionet-django/templates/about/citi_course.html`

tompollard:
Thanks Michael, looks good to me!

PR #1964, opened 2023-04-04
Python CITI module for PhysioNet

lamawmouk:
As discussed with @bemoody, we would like to create a python module for the CITI training to host the function needed to use the CITI API calls to enable the integration with PhysioNet.
This PR adds the python script that creates the `get_citiprogram_completion` function that address points one and two described in this issue #1937. The output of the function returns a list of dictionaries, where each dictionary represents the completion report information for the Human Research: Data or Specimens Only Research course. The dictionary will contain the completion report, course group, course ID, user's score, date of course completion, and expiration date for the course.


bemoody:
Thanks, Lama.  This is a good start.

### Code style things

I would move this into the `user` app, not the `physionet` app, though it's not a big deal either way.

I would break this into two functions: one function to search for a member by email (takes email as argument and returns a member_id), and one function to list the courses for a given member_id.

It would be wise to define a helper function that takes the XML payload as a string, retrieves the response, parses it, and returns an ElementTree.

I would define the request URL as a settings variable (settings.CITI_SOAP_URL) alongside CITI_USERNAME and CITI_PASSWORD.

### XML templates

Instead of using `f"""..."""`, it would be safer to use Django templates.

For example, you could create a file `physionet-django/user/templates/user/citi/get_inst_member_by_email.xml` containing:
```
<?xml version="1.0" encoding="utf-8"?>
<soap12:Envelope xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                 xmlns:xsd="http://www.w3.org/2001/XMLSchema"
                 xmlns:soap12="http://www.w3.org/2003/05/soap-envelope">
  <soap12:Body>
    <GetInstMemberByEmail xmlns="https://webservices.citiprogram.org/">
      <usr>{{ username }}</usr>
      <pwd>{{ password }}</pwd>
      <strEmail>{{ email }}</strEmail>
    </GetInstMemberByEmail>
  </soap12:Body>
</soap12:Envelope>
```
and then:
```
from django.template import loader

def get_citiprogram_completion(email):

    ...

    payload = loader.render_to_string(
        'user/citi/get_inst_member_by_email.xml', {
            'username': settings.CITI_USERNAME,
            'password': settings.CITI_PASSWORD,
            'email': email,
        }
    )
```

### HTTP headers

Looking at the examples here:

https://webservices.citiprogram.org/SOAP/CITISOAPService.asmx?op=GetInstMemberByEmail

This seems to indicate that for SOAP 1.2, the Content-Type in the request should be "application/soap+xml; charset=utf-8".  It might not matter but I would adhere to the headers they recommend.

lamawmouk:
Thanks @bemoody for the recommendations, I have implemented the above suggestions. I think we could perhaps merge this PR and I could add the test cases in a separate PR afterwards.

physionet-django/user/citi_training_module.py
```diff
@@ -0,0 +1,111 @@
+# Import the necessary libraries
```
tompollard: No need for this comment.

physionet-django/user/citi_training_module.py
```diff
@@ -0,0 +1,111 @@
...
+import requests
```
tompollard, line 1: Minor thing but perhaps tweak order follow PEP8 style for imports: https://peps.python.org/pep-0008/#imports

> Imports should be grouped in the following order:
> 
> Standard library imports.
> Related third party imports.
> Local application/library specific imports.

physionet-django/user/citi_training_module.py
```diff
@@ -0,0 +1,111 @@
...
+import xml.etree.ElementTree as ET
+from datetime import datetime
+from django.conf import settings
+from django.template import loader
+
+
+def convert_data_time(str_date_time):
```
tompollard: convert "datetime" or "data_time"?

physionet-django/user/citi_training_module.py
```diff
@@ -0,0 +1,111 @@
...
+    # Convert the string date time to a datetime object
```
tompollard: Please could you add doctrings to your functions? This comment could go in the docstring.

physionet-django/user/citi_training_module.py
```diff
@@ -0,0 +1,111 @@
...
+    datetime_object = datetime.strptime(str_date_time.text, '%Y-%m-%dT%H:%M:%S.%f%z')
+    return datetime_object
+
+
+def send_request(xml_payload):
+    # Define the headers for the request
```
tompollard: We can see the header is being defined, so no need for this comment. Same for the "# URL for the request" and "# Send the POST request and get the response" comments below. In general, just add a comment if the purpose/behaviour of the code is unclear.

physionet-django/user/templates/user/citi/get_courses_by_member.xml
```diff
@@ -0,0 +1,11 @@
...
+    <soap12:Envelope xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:soap12="http://www.w3.org/2003/05/soap-envelope">
+      <soap12:Body>
+        <GetMemberCoursesbyID xmlns="https://webservices.citiprogram.org/">
+        <usr>{{username}}</usr>
+        <pwd>{{password}}</pwd>
+        <intMemberID>{{memberid}}</intMemberID>
+        </GetMemberCoursesbyID>
+      </soap12:Body>
+    </soap12:Envelope>
```
tompollard: This file and the next file are missing a newline at the end.

tompollard:
Thanks @lamawmouk I haven't tested yet, but this looks pretty good to me. I have added a few very minor suggestions, mostly related to style. Please could you take a look and update the pull request. I'll try giving this a go in the meantime!

lamawmouk:
>  I have added a few very minor suggestions, mostly related to style.

Sure, where could I find them?

tompollard:
> Sure, where could I find them?

If you click the following link, you'll see the comments in the pull request: https://github.com/MIT-LCP/physionet-build/pull/1964. e.g.: 

Screenshot 2023-05-04 at 12 23 52 PM

Ideally you should: 
1. look over the comments
2. update the pull request as needed
3. add a short reply to each comment to add your thoughts (e.g. I made this change; I would prefer to make this change later because ... so have raised a new issue; I don't think this change is necessary because ....

lamawmouk:
Thanks @tompollard, I have updated the styling change.

tompollard:
@lamawmouk it looks like you've made a single change relating to one of the comments about "datatime -> "datetime" (plus an unrelated capitalisation change).

Please review all of the feedback and respond to each point. If it's easier, you can "Files changed" tab on this page and review the comments inline.

lamawmouk:
I am unable to see the other comments inline under the "Files changed" tab.
image

tompollard:
> I am unable to see the other comments inline under the "Files changed" tab.

I'm not in the lab today, but perhaps you can ask for help from someone who is there. Perhaps @briangow or @Chrystinne

tompollard:
@danamouk my bad, possibly I hadn't submitted them! Can you take another look?

lamawmouk:
No problem, yes I can see them now!

physionet-django/user/citi_training_module.py
lamawmouk: I have made this change.

physionet-django/user/citi_training_module.py
lamawmouk: Done!

physionet-django/user/citi_training_module.py
lamawmouk: Done!

physionet-django/user/citi_training_module.py
lamawmouk: I have added docstrings for the 3 functions.

physionet-django/user/citi_training_module.py
lamawmouk: Removed all unnecessary comments

physionet-django/user/templates/user/citi/get_courses_by_member.xml
lamawmouk: Done!

lamawmouk:
Thanks @tompollard, I have added the suggested changes above.

physionet-django/user/citi_training_module.py
tompollard: Thanks Lama, sorry to be fussy - please could you follow this guideline too: "You should put a blank line between each group of imports.".

physionet-django/user/citi_training_module.py
```diff
@@ -0,0 +1,155 @@
+import requests
+from datetime import datetime
+import xml.etree.ElementTree as ET
+from django.conf import settings
+from django.template import loader
+
+
+def convert_date_time(str_date_time):
```
tompollard, line 10: There is something odd about the input argument here. If `str_date_time` is expected to be a string, `str_date_time.text` would raise an error. 

Adding type hints across the PR might be helpful. e.g. this function would look something like:

```
def convert_date_time(str_date_time: str) -> datetime:
```

Perhaps:
1. Change `str_date_time.text` to `str_date_time` here.
2. Change the input arguments (below) to `dteExpiration.text` and `dtePassed.text`

physionet-django/user/citi_training_module.py
```diff
@@ -0,0 +1,155 @@
...
+            'memberEmail': memberEmail.text,
+            'strCompletionReport': strCompletionReport.text,
+            'intGroupID': intGroupID.text,
+            'strGroup': strGroup.text,
+            'intStageID': intStageID.text,
+            'intStageNumber': intStageNumber.text,
+            'strStage': strStage.text,
+            'intCompletionReportID': intCompletionReportID.text,
+            'intMemberStageID': intMemberStageID.text,
+            'dtePassed': convert_date_time(dtePassed),
```
tompollard: you should probably pass `dtePassed.text` here for consistency with the other fields.

tompollard: As above, you should probably pass `dteExpiration.text` here for consistency with the other fields, and so that your `convert_date_time` actually parses a string, instead of an object.

tompollard (commented):
Thanks Lama, sorry I picked up a minor thing or two when looking back at this. Once this is resolved (and after a bit of testing against the API), I think we should be good to merge, assuming @bemoody is also happy.

physionet-django/user/citi_training_module.py
lamawmouk: Sounds good, I have added a blank link between the standard imports, related third party imports, and local app (e.g. django) imports.

physionet-django/user/citi_training_module.py
lamawmouk: Thanks Tom for catching that, I have updated the code for consistency.

physionet-django/user/citi_training_module.py
lamawmouk: yes changed it to .text

lamawmouk: updated it to .text

tompollard:

Issue #1870, opened 2023-02-14
Managing `poetry.lock` and `requirements.txt`


kshalot:
@alistairewj @bemoody @tompollard  Like we saw, there seems to be an inconsistency in the way we manage the dependencies (`requirements.txt` in particular).

The way we approached this in Health Data Nexus is:
1. Add the dependency to `pyproject.toml`
2. Generate a new `poetry.lock` via for example `poetry lock --no-update`
3. Generate a new `requirements.txt` via `poetry export -f requirements.txt --output requirements.txt --with dev`

Step 3 seems to embellish the file with extra markers (see https://github.com/MIT-LCP/physionet-build/pull/1867#discussion_r1106019930). This is more of a discussion starter than an issue report. We should probably decide on a specific process to get this done to now get surprising results.

bemoody:
```
$ diff -u <(git show 8b5980b2260339a4eebb4d23f7c1937d063e0184:./requirements.txt | grep == | sed 's| \\||' | tr A-Z a-z | sort) <(git show ad4757d35d1e8e4180aca31c200d6334fcce1c20:./requirements.txt | grep == | sed 's| ; python_version >= "3.8" and python_version < "4[.0]*" \\$||' | sed 's| ; python_version < "4[.0]*" and python_version >= "3.8" \\||' | tr A-Z a-z | sort)
--- /dev/fd/63  2023-02-14 12:22:03.488557908 -0500
+++ /dev/fd/62  2023-02-14 12:22:03.488557908 -0500
@@ -1,52 +1,56 @@
 asgiref==3.5.2
-asn1crypto==0.24.0
+backports-zoneinfo==0.2.1 ; python_version >= "3.8" and python_version < "3.9" \
 bleach==3.3.0
-cachetools==3.1.1
+cachetools==4.2.2
 certifi==2022.12.7
 cffi==1.15.1
 chardet==3.0.4
 charset-normalizer==2.0.12
-coverage==4.4.2
+coverage==4.5.4
 cryptography==36.0.1
 django-autocomplete-light==3.9.4
 django-background-tasks-updated==1.2.7
 django-ckeditor==6.5.1
-django-compat==1.0.15
 django-debug-toolbar==3.2.4
 django-js-asset==2.0.0
 django-sass==1.1.0
-django-storages==1.12.3
+django-storages[google]==1.12.3
 django==4.1.6
 djangorestframework==3.14.0
-google-api-core==1.30.0
[...]

PR #1965, opened 2023-04-05
Bump python from 3.8 to 3.9 and update dependencies. Ref #1870.

tompollard:
This pull request bumps Python from >=3.8 to >=3.9 in `pyproject.toml` and updates dependencies with the following steps:

1. Generate a new poetry.lock file with: `poetry lock --no-update`
2. Generate a new requirements.txt with: `poetry export -f requirements.txt --output requirements.txt --with dev`

The change is a response to the following comment by @bemoody in https://github.com/MIT-LCP/physionet-build/issues/1870: 

>  I would be fine with increasing the minimum python version to 3.9 (in pyproject.toml) which might help remove some of the cruft *(in the requirement file).*

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -16,35 +16,6 @@ files = [
 tests = ["mypy (>=0.800)", "pytest", "pytest-asyncio"]
 
 [[package]]
-name = "backports.zoneinfo"
-version = "0.2.1"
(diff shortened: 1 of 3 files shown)


bemoody:
Thanks!

PR #1967, opened 2023-04-10
[WIP] refactor DataAccess to DataSource

superryeti:
Context:
As the first step to solve https://github.com/MIT-LCP/physionet-build/issues/1927, we want to refactor the current code to centralize all the project data access mechanism though DataSource. The goal of this PR is to refactor the current project to use DataSource without messing the access logic. 

List of expected changes
1. Create DataSource model
2. Refactor code to use DataSource and control all access mechanism through DataSource
2.1 Direct Access(Files stored on server)
2.2 Research Environment Access(All other access disabled, users can only create the Research Environment)
2.3 GCP, AWS(WIP)

3. Create migration script to migrate the access location from dataaccess to datasource for all existing projects for (Pending)
    a. direct access
    b. RE
    c. aws, gcp


superryeti:
@kshalot  opened for early feedback.

Currently, i did a rough refactor to control access to project files though direct access(files on server) and research environment(gcs). I wanted to see if i am on the right track or totally off-course :D 

Please check commitwise serially otherwise it might be confusing to review

physionet-django/project/managers/publishedproject.py
```diff
@@ -61,4 +61,9 @@ def accessible_by(self, user):
                 accessible_projects_ids.append(event_dataset.dataset.id)
         query |= Q(id__in=accessible_projects_ids)
 
+        # for Research Environment,the project should have a DataSource with access_mechanism = 'research-environment'
+        projects_with_research_environment = DataSource.objects.filter(
+            access_mechanism=DataSource.AccessMechanism.RESEARCH_ENVIRONMENT,
+            project=self).values_list('project_id', flat=True)
```
kshalot: I don't think this is the way we want to go - let's not make any assumptions regarding which data source is expected. Right now we still operate with the assumption that being authorized to view a project's files means that you can access all of its data sources. Therefore, the `accessible_by` and `is_authorized` methods should simply return the list of accessible projects (or `True` in case of `is_authorized`). We can then use the `DataSource` relation to only choose the relevant datasets.

So research environments would get the list of all accessible projects using a generic method and simply filter out the irrelevant ones. We could also create more methods, e.g.

```python
def can_access_directly(): ...
def can_create_research_environment(): ...
```

This will be even more apparent if we decide to refactor the authorization to, for example, the access grant model (see https://github.com/MIT-LCP/physionet-build/issues/1927#issuecomment-1477021739).

physionet-django/physionet/settings/base.py
```diff
@@ -612,3 +612,5 @@ class StorageTypes:
 
 # minimum number of word needed for research_summary field for Credentialing Model.
 MIN_WORDS_RESEARCH_SUMMARY_CREDENTIALING = config('MIN_WORDS_RESEARCH_SUMMARY_CREDENTIALING', cast=int, default=20)
+
+DEFAULT_PROJECT_ACCESS_MECHANISM = config('DEFAULT_PROJECT_ACCESS_MECHANISM', default='direct')
```
kshalot: This could be validated + the default value could be some enumerated value instead of a bare string.

physionet-django/project/modelcomponents/access.py
```diff
@@ -167,6 +168,35 @@ class Meta:
         default_permissions = ()
 
 
+class DataSource(models.Model):
+    """
+    Controls all access to project data.
+    """
+    class DataLocation(models.TextChoices):
+        DIRECT = 'DI', _('Direct')
```
kshalot: Do we need this to be translatable? Almost all the options are proper nouns that are not translatable (e.g. AWS S3, AWS Open Data). Also, nothing else in the console is translatable.

physionet-django/project/projectfiles/gcs.py
```diff
@@ -117,6 +118,14 @@ def publish_initial(self, active_project, published_project):
 
     def publish_complete(self, active_project, published_project):
         self.rm_dir(active_project.file_root())
+        if settings.DEFAULT_PROJECT_ACCESS_MECHANISM == 'research_environment':
```
kshalot: In other places you use 
```
DataSource.AccessMechanism.RESEARCH_ENVIRONMENT
```

physionet-django/project/modelcomponents/access.py
```diff
@@ -167,6 +168,35 @@ class Meta:
         default_permissions = ()
 
 
+class DataSource(models.Model):
```
kshalot, line 174: I think we'll need some proper validations on this model (require URI for certain mechanisms, require email for certain mechanisms etc.)

physionet-django/project/modelcomponents/access.py
```diff
@@ -167,6 +168,35 @@ class Meta:
...
+
+    class AccessMechanism(models.TextChoices):
+        GOOGLE_GROUP_EMAIL = 'google-group-email', _('Google Group Email')
+        S3 = 's3', _('S3')
+        RESEARCH_ENVIRONMENT = 'research-environment', _('Research Environment')
+
+    project = models.ForeignKey('project.PublishedProject',
+                                related_name='data_sources', db_index=True, on_delete=models.CASCADE)
+    files_available = models.BooleanField(default=False)
+    data_location = models.CharField(max_length=3, choices=DataLocation.choices, default=DataLocation.DIRECT)
```
kshalot: Aren't the default values here a bit misleading?

kshalot (changes requested):
A few initial comments

superryeti:
> A few initial comments

Thank you @kshalot , Super appreciate the feedback. Let me get on it right away

physionet-django/console/forms.py
```diff
@@ -689,6 +690,26 @@ def save(self):
...
+        model = DataSource
+        fields = ('data_location', 'access_mechanism', 'files_available', 'email', 'uri' )
+
+    def __init__(self, project, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.project = project
+
+        if not settings.ENABLE_CLOUD_RESEARCH_ENVIRONMENTS:
+            self.fields['access_mechanism'].choices = [
+                choice for choice in self.fields['access_mechanism'].choices if choice[0] != 'research-environment']
```
kshalot, line 704: I'm wondering whether we want to care about this being enabled or not... It makes sense and it would be my first instinct to do so, but right now it seems that research environments are singled out as something that can be present or not.

The HDN deployment is not integrated with any other data source other than local/research environment, so in that case it makes no sense showing anything other than that. Research environments are the only ones that have an explicit setting to disable them, but still.

I guess it doesn't hurt to have this, it just got me thinking!

physionet-django/project/managers/publishedproject.py
kshalot: Which case is this covering? Why do we need to check whether a source exists to check whether a project is accessible by a user? This just makes this method less generic, since it only makes sense in case of HDN.

I think my previous point still stands - we should only check access mechanisms here (event, DUA, credentialing etc.) and not sources.

physionet-django/project/modelcomponents/access.py
```diff
@@ -167,6 +168,35 @@ class Meta:
...
+        AWS_S3 = 'AS3', _('AWS S3')
+
+    class AccessMechanism(models.TextChoices):
+        GOOGLE_GROUP_EMAIL = 'google-group-email', _('Google Group Email')
+        S3 = 's3', _('S3')
+        RESEARCH_ENVIRONMENT = 'research-environment', _('Research Environment')
+
+    project = models.ForeignKey('project.PublishedProject',
+                                related_name='data_sources', db_index=True, on_delete=models.CASCADE)
+    files_available = models.BooleanField(default=False)
```
kshalot, line 192: I'm going to go back to this to give it some more thought - the `files_available` boolean is dubious now, seeing it in practice. It's a bit confusing what it's supposed to achieve unless you were part of the previous conversations 🤔 

I think it's going to be more clear once we see it actually used in practice, i.e. replace the current `DataAccess` with `DataSource`. Originally the idea of this boolean was to denote the source that gives us access to the file browser. I think at this point we could not care about this and assume only the `direct` source gives us direct file download access. We could easily go back to this and make it more generic in the future.

The point is that _in theory_ the GCS source could also give us access to files, since we are using buckets in HDN to store the projects, with the exact same UX as PhysioNet has for direct upload. But right now the mechanisms are tightly coupled with their specific use cases, so maybe we could just not care about it for now. This would mean we wouldn't initially need the `files_available` bool.

physionet-django/physionet/settings/base.py
kshalot: Shouldn't the default be  `DataLocation.DIRECT` or something similar?

physionet-django/project/projectfiles/gcs.py
kshalot: I'd extract the logic of creating a default `DataSource` to a separate class that would be configured during the app's initialization with the default source. Then, after the project is published, some generic `#create_data_source` method would be called from the publishing code. This way the logic is nicely separated and not embedded everywhere.

Also - I don't think this is the place to create a `DataSource`. This should probably happen inside the `publish` method in a project (or similar). This part of the code is simply about publishing files. It shouldn't care that an abstraction like `DataSource` exists or not.

physionet-django/project/projectfiles/local.py
```diff
@@ -120,7 +121,13 @@ def publish_rollback(self, active_project, published_project):
...
 
     def publish_complete(self, active_project, published_project):
-        pass
+        if settings.DEFAULT_PROJECT_ACCESS_MECHANISM == 'direct':
+            DataSource.objects.create(
+                project=published_project,
+                files_available=True,
+                data_location=DataSource.DataLocation.DIRECT,
+            )
+
```
kshalot: Same comment as with the GCS files. We should have a simple hook after the project that would call into a class/module to do this. We want this to be as abstract as possible.

physionet-django/project/managers/publishedproject.py
kshalot: Looking at the `can_create_research_environment` part of the code, this looks like a leftover from some old logic

physionet-django/project/modelcomponents/publishedproject.py
```diff
@@ -210,7 +210,19 @@ def can_view_files(self, user):
...
+            data_location=DataSource.DataLocation.DIRECT).exists()
+        return self.is_authorized(user) and is_accessible
+
+    def can_create_research_environment(self, user):
+        """
+        Whether the user has access to create a research environment for this project.
+        """
+        is_accessible = DataSource.objects.filter(
+            project=self,
+            access_mechanism=DataSource.AccessMechanism.RESEARCH_ENVIRONMENT).exists()
```
kshalot: `is_accessible` sounds a bit misleading when compared to the manager's `accessible_by` method - one is about authorization, the other is about configuration.

Maybe something like `has_research_environment_source` would be more explicit?

kshalot (changes requested):
@amitupreti I left a few comments. The base looks good, but what I think would be very helpful here is if we would see how this new `DataSource` would play into the existing codebase, i.e. trying to use this model to control whether a project's files are visible etc (instead of the way it is done, which uses the `allow_file_downloads` boolean.

My comments might be a bit all over the place - if you encounter any trouble planning the next steps here then let me know. We can jump on a call and figure this out together. Whatever works for you!

superryeti:
> @amitupreti I left a few comments. The base looks good, but what I think would be very helpful here is if we would see how this new `DataSource` would play into the existing codebase, i.e. trying to use this model to control whether a project's files are visible etc (instead of the way it is done, which uses the `allow_file_downloads` boolean.
> 
> My comments might be a bit all over the place - if you encounter any trouble planning the next steps here then let me know. We can jump on a call and figure this out together. Whatever works for you!

Thank you. This is super helpful

physionet-django/project/modelcomponents/access.py
superryeti: I agree, we  don't need this to be translatable. To be honest, i was just following what i had seen in Django docs, and had not looked at the import. This is a good lesson for me to understand the code more.

physionet-django/console/forms.py
superryeti: I can see what you mean.  It makes sense about for HDN like deployment, we might just want to have the local/research environment and disable other stuff.

From what i understand, we might want to allow the deployments to have any combination of access mechanisms. 
so what if we have a configurable env like
ACCESS_MECHANISMS = 'research-environment,direct, awss3' 
and use that.


or maybe since this is something only the Project Editor or console admin can do, we might be okay with showing them all the option and trust them to use the right ones.

physionet-django/project/managers/publishedproject.py
superryeti: Yes, i agree, this method should be generic and we should be creating separate methods which use the `accessible_by` and `is_authorized` to check access via specific data access mechanism.

I see your point, the accessible_by method should only give us project that the user has access to. This method should not care what access mechanisms are set for this project. Also, appreciate your question, it is super helpful to think about the desig, i will remember to ask myself while working on other functions.

superryeti:
Hi @kshalot , i abstracted the creation of datasource on project publish to a separate class, which is called after the project is published.

Regarding the `files_available`, i let it be and it is currently used with `direct access`(files stored and served from the server). So with `direct access` and `files_available` =True, the assumption is that user can view the files on browser and also download it with wget, which i think works only for `direct access/local` . Reason for letting it be is because as you said, later we might want to let users see/access files from gcs bucket through physionet.

physionet-django/project/modelcomponents/access.py
```diff
@@ -167,6 +169,67 @@ class Meta:
         default_permissions = ()
 
 
+class DataSource(models.Model):
+    """
+    Controls all access to project data.
+    """
+    class DataLocation(models.TextChoices):
+        DIRECT = 'DI', 'Direct'
```
kshalot, line 179: Is there any advantage to using abbreviations like `DI` or `AS3` over `DIRECT`, `AWS_S3`, `AWS_OPEN_DATA` etc? I guess it will save a teeny-tine amount of space but makes things a bit incomprehensible when looking at it from the data's perspective.

physionet-django/project/modelcomponents/access.py
```diff
@@ -167,6 +169,67 @@ class Meta:
...
+                raise ValidationError('AWS S3 data sources must use the S3 access mechanism.')
+            if not self.uri:
+                raise ValidationError('AWS S3 data sources must have a URI.')
+        elif self.data_location == self.DataLocation.DIRECT:
+            if self.email:
+                raise ValidationError('Direct data sources must not have an email address.')
+            if self.uri:
+                raise ValidationError('Direct data sources must not have a URI.')
+        else:
+            raise ValidationError('Invalid data location.')
```
kshalot, line 232: This is very hard to follow. You could approach it differently - define what exactly is required for each data source and have a single piece of logic that validates it.

```python
required_fields = {
  self.DataLocation.DIRECT: [],
  self.DataLocation.AWS_S3: ["uri"],
  ...
}

forbidden_fields = {
  self.DataLocation.DIRECT: ["uri", "email"],
  ...
}

for required_field in required_fields[self.data_location]:
  if not getattr(self, required_field): # Or something like that
    raise ValidationError("...")

for forbidden_field in forbidden_fields[self.data_location]:
  if getattr(self, forbidden_field): # Or something like that
    raise ValidationError("...")
```

kshalot: Same goes for data access:
```python
data_location_access_mechanisms = {
  self.DataLocation.DIRECT: None
  self.DataLocation.AWS_S3: self.AccessMechanism.AWS_S3,
  ...
}
```

physionet-django/project/utility.py
```diff
@@ -677,3 +679,42 @@ def convert_url(self, attr_name, url):
...
         return url
+
+
+class DataSourceCreator:
+    def __init__(self, **kwargs):
+        self.data_location = kwargs.get('data_location', None)
+        self.files_available = kwargs.get('files_available', None)
+        self.email = kwargs.get('email', None)
+        self.uri = kwargs.get('uri', None)
+        self.access_mechanism = kwargs.get('access_mechanism', None)
```
kshalot: Hmm, why did you opt for keyword arguments here? Why not:
```suggestion
    def __init__(self, data_location=None, files_available=None, email=None, uri=None, access_mechanism=None):
        self.data_location = data_location
        self.files_available = files_available
        self.email = email
        self.uri = uri
        self.access_mechanism = access_mechanism
```

physionet-django/project/utility.py
```diff
@@ -677,3 +679,42 @@ def convert_url(self, attr_name, url):
...
+        DataSource.objects.create(
+            project=project,
+            files_available=self.files_available,
+            data_location=self.data_location,
+            access_mechanism=self.access_mechanism,
+            email=self.email,
+            uri=self.uri,
+        )
+
+    @staticmethod
```
kshalot: I know `self` is not needed here, but this is called on an instance everywhere (see your uses of this method - you first instantiate the class). Maybe `@staticmethod` is not necessary here.

physionet-django/project/utility.py
```diff
@@ -677,3 +679,42 @@ def convert_url(self, attr_name, url):
...
+    def create_default(project):
+        if (settings.DEFAULT_PROJECT_DATA_LOCATION == DataSource.DataLocation.DIRECT
+                and settings.STORAGE_TYPE == StorageTypes.LOCAL):
+            DataSource.objects.create(
+                project=project,
+                files_available=True,
+                data_location=DataSource.DataLocation.DIRECT,
+            )
+        elif (settings.DEFAULT_PROJECT_ACCESS_MECHANISM == DataSource.DataLocation.RESEARCH_ENVIRONMENT
```
kshalot: That's too many conditions IMO, it should be extracted somewhere. It's also not clear why the condition is there in the first place. I.e. why do we need all 3 settings to be set to a certain value.

If the settings are that coupled, then maybe we should opt for a different setup. Have one environment variable, e.g. `RESEARCH_ENVIRONMENT_DEPLOYMENT` that will set all three (`DEFAULT_PROJECT_ACCESS_MECHANISM`, `DEFAULT_PROJECT_DATA_LOCATION` and `STORAGE_TYPE`) if set to `True` and if set to `False` it will set the two from the condition above.

kshalot (changes requested):
@amitupreti I left a few comments. I think we are generally going in the right direction.

Correct me if I'm wrong though - this is still not integrated with the project page, right? I.e. it does not control whether the file tab is shown, whether the S3 message is shown etc?

I think you should at least integrate the direct access with the file viewer so we can see it in action. This is the most important part, because the file viewer is not part of the current `DataAccess` model (the one which you are refactoring into `DataSource`.

superryeti:
> Correct me if I'm wrong though - this is still not integrated with the project page, right? I.e. it does not control whether the file tab is shown, whether the S3 message is shown etc?

I think i had added that part of integration. i.e using datasource to control access in project on `Direct Download refactor` commits. Let me check your comments and make things clear and write how everything comes to play in details.

Its kind of chaotic right now. I will work on it and update the code.

Thanks again

PR #1969, opened 2023-04-14
news:increase fontsize of Readmore section

### LINKED ISSUE #1958: "Read more: <URL>" text on news items should be larger
We have received feedback to say that the "Read more: " text on news items is too small to read easily. Looking at the website, I agree!

For example, see the screenshot below (reproduced from https://physionet.org/news/post/413):

Screenshot 2023-03-30 at 3 44 32 PM

The template that needs updating is:

https://github.com/MIT-LCP/physionet-build/blob/0f53f540832693a79af9bf8af56b8c772e2e7f60/physionet-django/notification/templates/notification/news_content.html#L6-L10

superryeti:
Closes https://github.com/MIT-LCP/physionet-build/issues/1958

As reported on the issue #1958 , the font size can be made more readable.

Before

image

After

image

### DIFF
--- a/physionet-django/notification/templates/notification/news_content.html
+++ b/physionet-django/notification/templates/notification/news_content.html
@@ -6,5 +6,5 @@
   {{ news.content|safe }}
 </div>
 {% if news.url %}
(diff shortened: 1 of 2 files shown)


tompollard:

Issue #1971, opened 2023-04-14
Improve search functionality for projects


tompollard:
Currently we display a search tool at https://physionet.org/content/ (and also in the right of the site header). The tool is basically unusable right now, because (1) it does exact matching on single words and (2) the results are not weighted, meaning top matches get lost in noise.

e.g. try searching for "mimic database" and the results are not what you expect: https://physionet.org/content/?topic=mimic+database (first hit is "Lobachevsky University Electrocardiography Database") 

We should consider swapping out the current search backend to something a little more sophisticated (e.g. supports fuzzy matching, takes into account relevance/match density in order of results).

tompollard:
@amitupreti @Rutvikrj26 I wonder whether it makes sense for the search to sit on top of the new project API (https://github.com/MIT-LCP/physionet-build/pull/1697)? 

This might help us to move towards a cross-site search later (e.g. a search tool that returns results from both PhysioNet and Health Data Nexus)?

Side note, but when speaking to someone recently about API design, they strongly recommended following the OpenAPI spec (https://www.openapis.org/).

superryeti:
Yeah , totally, this sounds like a good idea to me. I will catchup with @Rutvikrj26  on how to move ahead with this, as i remember he is already working on api here https://github.com/MIT-LCP/physionet-build/issues/1885

bemoody:
See also: #349

Rutvikrj26:
I discussed this with @amitupreti, and we've decided to add an API endpoint that talks in the Post request, calls the search function, and returns the results, which will then be serialized to show in the output. 

This will enable integration of the Cross-Site search functionality via API, as @tompollard mentioned.

Issue #349, opened 2019-04-08
Site search backend


bemoody:
(I tried to post this as a reply to pull request #348, but it
looks like it didn't go through, and it belongs as a separate
issue anyway.)

I'm sorry to jump in after Lucas has already done a lot of work
on this, but for the backend implementation of the site search,
have you considered using an existing search engine such as
Xapian?

Xapian has Python bindings (although I don't know how easy they
are to use), and it does a lot of useful things (parsing complex
queries, stemming, weighting keywords, phrase searching) rather
than us having to reinvent the wheel.

tompollard:
I don't know Xapian, but if you think it has potential to improve search then it is definitely worth exploring. A major part of Lucas's work has been in improving page structure and usability, so any backend improvements would complement his efforts.  I've tracked the issue at https://github.com/MIT-LCP/physionet-build/projects/1 (search seems pretty okay for now, so left migration of existing content as the current task).

PR #1973, opened 2023-04-20
Improve search functionality for projects

superryeti:
**Context**
Related to https://github.com/MIT-LCP/physionet-build/issues/1971

The current search mechanism does
1. Takes the search term from user and breaks it into words
2. Finds projects that has the words on `title`, `abstract` and `description`
3. Calculate relevance score based on how many times the word appear on the above fields of the projects
4. calculate has_keys by checking on which field the words are found for the selected projects, and providing highest preference to the matches found in title
5. Sort results by relevance and has_keys and display to users

This PR implements full text search with Django Postgres for the project model. It adds the same logic as the normal search with Django Postgres full text search,

This PR improves the relevance/results by stemming the search terms which comes with `SearchRank`, which means that it will match different forms of the same word. For example, if someone searches for "ponies", "pony" or "ponying", they will still get results that have the word "pony" in them. This makes the search more flexible and accurate.

However, this PR doesn't do anything about the performance issues  which might happen when the project database gets very large. The documentation says to improve performance we should index the dataset

Indexing: https://docs.djangoproject.com/en/4.1/ref/contrib/postgres/search/#performance

SearchRank: https://docs.djangoproject.com/en/4.1/ref/contrib/postgres/search/#searchrank

### DIFF
--- a/physionet-django/search/views.py
+++ b/physionet-django/search/views.py
@@ -81,9 +81,9 @@ def get_content(resource_type, orderby, direction, search_term):
     # Relevance
     for t in search_term:
         published_projects = (published_projects.annotate(has_keys=Case(
-                When(topics__description__iregex=r'{0}{1}{0}'.format(wb, t),
-                     then=Value(3)),
                 When(title__iregex=r'{0}{1}{0}'.format(wb, t),
+                     then=Value(3)),
+                When(topics__description__iregex=r'{0}{1}{0}'.format(wb, t),
                      then=Value(2)),
                 When(abstract__iregex=r'{0}{1}{0}'.format(wb, t),
                      then=Value(1)),
--- a/physionet-django/search/views.py
+++ b/physionet-django/search/views.py
@@ -55,6 +55,19 @@ def get_content(resource_type, orderby, direction, search_term):
     Helper function to get content shown on a resource listing page
     """
 
+    if 'postgresql' in settings.DATABASES['default']['ENGINE']:
+        published_projects = get_content_postgres_full_text_search(resource_type, orderby, direction, search_term)
+    else:
+        published_projects = get_content_normal_search(resource_type, orderby, direction, search_term)
+
+    return published_projects
+
+
+def get_content_postgres_full_text_search(resource_type, orderby, direction, search_term):
+    pass
+
+
+def get_content_normal_search(resource_type, orderby, direction, search_term):
     # Word boundary for different database engines
     wb = r'\b'
     if 'postgresql' in settings.DATABASES['default']['ENGINE']:
--- a/physionet-django/search/views.py
+++ b/physionet-django/search/views.py
@@ -79,9 +79,11 @@ def get_content_postgres_full_text_search(resource_type, orderby, direction, sea
               + SearchVector('topics__description', weight='C'))
 
     # Filter projects by latest version and annotate relevance field
-    published_projects = (PublishedProject.objects.annotate(search=vector).
-                          filter(query, is_latest_version=True).
-                          annotate(relevance=SearchRank(vector, search_query)))
+    published_projects = PublishedProject.objects.annotate(search=vector).filter(query, is_latest_version=True)
+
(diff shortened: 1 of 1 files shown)


bemoody:
Thanks for looking at this!  So, my thought (not having tried it) is that these changes are likely to be a great improvement, but probably still very limited by the capabilities of Postgres.

I think a better long term strategy would be to use Xapian (or some other package designed for the purpose, though Xapian is the one I'm familiar with.)  It's possible we could implement Xapian-like features using Django and Postgres, but I think it's likely to be a huge amount of work and ultimately less useful.

Of course I'm not allowed to complain because I posted issue #349 four years ago and haven't done anything about it since.  But that would be my gentle suggestion: if you want to spend time on improving the search feature, switching it to a real search engine would be a good start. :)

superryeti:
> Thanks for looking at this! So, my thought (not having tried it) is that these changes are likely to be a great improvement, but probably still very limited by the capabilities of Postgres.
> 
> I think a better long term strategy would be to use Xapian (or some other package designed for the purpose, though Xapian is the one I'm familiar with.) It's possible we could implement Xapian-like features using Django and Postgres, but I think it's likely to be a huge amount of work and ultimately less useful.
> 
> Of course I'm not allowed to complain because I posted issue #349 four years ago and haven't done anything about it since. But that would be my gentle suggestion: if you want to spend time on improving the search feature, switching it to a real search engine would be a good start. :)

Thank you. I will check the Xapian and the old issues

tompollard:
@amitupreti I don't remember if I already posted a link to it, but Django Haystack seems popular: https://django-haystack.readthedocs.io/en/master/index.html and it integrates with "Solr, Elasticsearch, Whoosh, Xapian, etc".

tompollard:
@amitupreti this looks like it is duplicating results in the search. e.g. compare the following URL between the dev branch and the feature branch: http://localhost:8000/content/?topic=eicu

One the dev branch, we see a single result. The screenshot below shows the results on the feature branch:

Screenshot 2023-04-25 at 11 33 44 PM

superryeti:
Thats weird, i dont remember seeing that. let me do some test by creating lots of project and test with that.

Thanks @tompollard

superryeti:
This seems to be coming from  topics__description, Looks like if the publishedproject.topics.all().count() is 3, we will see the published project duplicated 3 times.(for some reason with `related_model` in search, we get duplicate results)
For example
if we search for "eicu" it will appear appears 3 times
if we search for "Demo ECG Signal Toolbox" it will appear 4 times

So far, the solution i found is to 
1. use subquery
2. we cant seem to use distinct because postgres requires us to have the first field in `order_by` to be same  as the field in `distinct`
I am still not sure why it returns duplicate in the first place, i will ask Karol for help.

superryeti:
Thanks @tompollard , i added a subquery to remove the duplicate

related https://forum.djangoproject.com/t/searchquery-and-duplicate-in-queryset/13152

tompollard:

tompollard:
This is definitely an improvement on the previous search. e.g. searching for MIMIC now returns "[MIMIC-III Clinical Database CareVue subset](https://physionet.org/content/mimic3-carevue/1.4/)" and "[MIMIC-IV](https://physionet.org/content/mimiciv/2.2/)" in the top 3 hits.

Previously the top hits for MIMIC were [Chest ImaGenome Dataset](https://physionet.org/content/chest-imagenome/1.0.0/), [Nosocomial Risk Datasets from MIMIC-III](https://physionet.org/content/nosocomialriskdata/1.0/), and [Synthetic Acute Hypotension and Sepsis Datasets Based on MIMIC-III and Published as Part of the Health Gym Project](https://physionet.org/content/synthetic-mimic-iii-health-gym/1.0.0/).

I notice that there are still occasional duplicates being returned, but it is not a huge issue (e.g. take a look at https://physionet.org/content/?topic=eicu). Perhaps better to focus in the new search than worrying about fixing these duplicates

PR #1980, opened 2023-04-26
allow curly apostrophe in research summary

### LINKED ISSUE #1972: Should we allow curly apostrophe “” on Credentialing form-> Research Summary?
**Context:** 
We were just having someone test the https://healthdatanexus.ai/ platform, and they were going though the process of applying for Credentialing. They were unable to submit the form because the keyboard somehow added a curly apostrophe instead of straight apostrophe.

Something like this will not go through

> this is a test “lorem ipsum” This is a test description lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem 

@january-adams also found that this will happen if a user write the description in google doc or microsoft docs, and try to copy and paste it to the form

Maybe instead of adding new allowed special characters, we could add some kind of processor to clean and convert the special characters to normal ones(like from curly to straight apostrophe)

superryeti:
Closes https://github.com/MIT-LCP/physionet-build/issues/1972

Context: During testing, a user had reported that they couldn't use curly apostrophe on the summary. After discussion, we decided to call it a day and allow curly apostrophe. See https://github.com/MIT-LCP/physionet-build/issues/1972

Changes:
Here i added the opening and closing single and double curly apostrophe.

### DIFF
--- a/physionet-django/user/validators.py
+++ b/physionet-django/user/validators.py
@@ -204,7 +204,7 @@ def validate_research_summary(value):
     spaces, underscores, apostrophes and the following special
     characters: !@#$%&*()[]~+={};:"<>?,./`-
     """
-    if not re.fullmatch(r'[a-zA-Z][\w\'\[\]\n\r~!@#$%&*()+={};:"<>?,./` -]+', value):
+    if not re.fullmatch(r'[a-zA-Z][\w\'\[\]\n\r~!@#$%&*()+={};:‘’“”"<>?,./` -]+', value):
         raise ValidationError('Letters, numbers, spaces, apostrophes and [!@#$%&*()[]~_+-={ };:"<>?,./`] characters only. Must begin with a letter.')
 
 


tompollard:
Thanks Amit, I'm happy to merge this, but I think the decision was a bit more than simply allowing apostrophes. We would like to be much more permissive about allowed characters (i.e. unless a character poses a security risk, it should be allowed).

superryeti:
Oh sorry, i misunderstood. let me open another PR. @tompollard 

Should i check other validators to determine places where were are too restrictive?

tompollard:
> Should i check other validators to determine places where were are too restrictive?

Yes, please, if it isn't too much extra work!

superryeti:
Yup, Sounds good. let me do that., i also was thinking the same, I see Alistair also opened an issue issue because we don't let people use ampersand in bunch of places.

PR #1981, opened 2023-04-27
Remove hardcoded redirects to physiobank and physiotools

tompollard:
In https://github.com/MIT-LCP/physionet-build/pull/1948 we added functionality for managing URL redirects. Our current code has some hardcoded redirects that could be replaced with this new functionality.

This pull request removes the following hardcoded redirects:
- https://physionet.org/physiobank/ -> https://physionet.org/data/
- https://physionet.org/physiobank/database/ -> https://physionet.org/data/
- https://physionet.org/physiotools/ -> https://physionet.org/software/

I have added these redirects to our database on production, so removing this code should have no effect on the live platform.

### DIFF
--- a/physionet-django/search/urls.py
+++ b/physionet-django/search/urls.py
@@ -79,7 +79,6 @@ urlpatterns = [
     path('charts/', views.charts, name='charts'),
 
     # Redirect from legacy
-    path('physiotools/', views.physiotools),
     path('physiobank/database/wfdbcal', views.wfdbcal),
     re_path('^physiobank/database/(?P<project_slug>[\w\-]+)/$', views.redirect_project),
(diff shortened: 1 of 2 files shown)


bemoody:
Note that /physiotools/ and /physiobank/database/ are hard-coded in /etc/nginx/snippets/physionet-common.conf .  They're not handled by django (that's why they're currently 301s rather than 302s.)

tompollard:
> Note that /physiotools/ and /physiobank/database/ are hard-coded in /etc/nginx/snippets/physionet-common.conf . They're not handled by django (that's why they're currently 301s rather than 302s.)

Thanks, I should have picked that up. That's not a bad place for the custom redirects, so I guess those can stay for now!

PR #1984, opened 2023-04-27
Search all user emails in console, not just primary emails. Fixes #1978.

### LINKED ISSUE #1978: User search in admin console only considers primary email addresses
There is a tool for searching users by email address at https://physionet.org/console/users/active/. This tool only searches for primary email addresses (as far as I can see). It should search *all email addresses* (or at least all email active addresses).

tompollard:
This is a quick fix for the issue described in https://github.com/MIT-LCP/physionet-build/issues/1978. Currently, our search tool in the user console will only search for primary emails (not for other linked emails).

This pull request updates the search query so that associated emails are also searched. To test the change: 

1) On a fresh database on the dev branch, try searching for `admin3@mit.edu` in the user console at http://localhost:8000/console/users/active/. No results are returned.
2) Now switch to the feature branch and run the same search. You'll see that the admin user is returned (because `admin3@mit.edu` is a secondary email address for this user).

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -1126,7 +1126,9 @@ def users_search(request, group):
 
         users = User.objects.filter(Q(username__icontains=search_field)
                                     | Q(profile__first_names__icontains=search_field)
-                                    | Q(email__icontains=search_field))
+                                    | Q(email__icontains=search_field)
+                                    | Q(associated_emails__email__icontains=search_field)
+                                    ).distinct()
 
         if 'inactive' in group:
             users = users.filter(is_active=False)


bemoody:
email__icontains should be redundant at this point, isn't it?

Funny that this searches first_names and not last_name, too...

tompollard:
> `email__icontains` should be redundant at this point, isn't it?

In theory, it should be redundant (because the email should also be captured as an AssociatedEmail), but I think it is safest to retain it here. I do now realise that we were duplicating users returned by the query, so added a `distinct()`.

> Funny that this searches first_names and not last_name, too...

Good point! Shall I update this PR to search those, or open a new PR? My preference is new PR.

tompollard:
> `email__icontains` should be redundant at this point, isn't it?

Might be wrong (and it doesn't change the fact that I think it's safest to keep `email__icontains`) but I think that `AssociatedEmail` is only created on account activation, so dropping `email__icontains` would break the search for inactive accounts.

bemoody:
> In theory, it should be redundant (because the email should also be captured as an AssociatedEmail), but I think it is safest to retain it here.

OK.  I think in the long run the User email field probably should go away.

> Good point! Shall I update this PR to search those, or open a new PR? My preference is new PR.

Sounds good, see issue #1990.

PR #1987, opened 2023-04-29
JSON export of AWS user accounts (restricted to MIMIC-IV 2.2)

Chrystinne:
This PR provides a JSON file of AWS user accounts that can be downloaded by administrators. It includes only the latest version of mimic-iv that is currently mirrored on AWS (i.e., version 2.2).


tompollard:
Thanks Chrystinne, looks good as a temporary solution!

PR #1986, opened 2023-04-29
customizable theming

superryeti:
## Context:

Healthdatanexus theme is a  bit different from physionet. HDN has a bluish color and different gradient color and also other smaller css changes and background images.

Currently, this is achieved by modifying the forked physionet code directly, which requires us to manually edit the css and copy the css files during deployment. (Check https://github.com/T-CAIREM/physionet-build/pull/3/files to see what changes are needed on HDN )

This PR adds environment variable, and makes the theme customizable through them.

Changes
1. update scss files to use the environment variables. by default the theme is set to normal physionet colors, gradient
2. make background image customizable through environment variable
3. Demo change to get feedback on making the existing css(such as home.css) customizable(as hdn tweaks color on some other css also)

So if this [commit](https://github.com/MIT-LCP/physionet-build/commit/ce458e4656bf5ce6620a9d53c5feed68a502d4d5)  looks acceptable, i can go ahead  and make the rest of relevant  customizable like this(Is it a good idea to replace use of all css with scss or maybe we just modify the ones we need?)

## To test(to get the healthdatanexus tints), please add the following to your .env

```
# SCSS Theme Variables
# These variables are used to set the theme colors for the site
DARK='#1E3765'
PRIMARY='#337ab7'
SECONDARY='#6D247A'
SUCCESS='#8DBF2E'
INFO='#6FC7EA'
WARNING='#F1C500'
DANGER='#DC4633'
LIGHT='#007FA3'

## Gradient colors on homepage

GRADIENT_60 = 'rgba(30, 55, 101, 0.6)'
GRADIENT_85 = 'rgba(30, 55, 101, 0.85)'

## physionet background image(local path or direct download url)
BACKGROUND_IMAGE='https://storage.googleapis.com/healthdatanexus-prod-hdn-static/images/background.jpg'
```

Many thanks to @kshalot for the help

PR #1991, opened 2023-05-02
customizable theming followup

superryeti:
**Context:**
On https://github.com/MIT-LCP/physionet-build/pull/1986, we merged things in a rush of excitement and missed few things

This PR cleans up on the #1986.

**Changes**

Here we just untracked the home.css files as it is now generated with compilestatic.
Also we updated the default background image to a nicer image(Thanks to @tompollard )


.env.example
```diff
@@ -157,7 +157,7 @@ GRADIENT_60 = 'rgba(42, 47, 52, 0.6)'
 GRADIENT_85 = 'rgba(42, 47, 52, 0.85)'
 
 ## physionet background image(direct download url)
-BACKGROUND_IMAGE='https://physionet.org/static/images/background.jpg'
+BACKGROUND_IMAGE='https://cdn.vox-cdn.com/thumbor/WbjC4KbwavUienc3q7DEIn1xMsg=/0x0:1440x900/1200x628/filters:focal(697x524:698x525)/cdn.vox-cdn.com/uploads/chorus_asset/file/22310830/NmJgg.jpg'
```
tompollard: Much as I like this, I'm not sure it is a good idea to suggest downloading a random file from the internet! Perhaps we could add a demo image to this repository and link to this instead?

.gitignore
```diff
@@ -148,4 +148,5 @@ physionet-django/static/bootstrap/scss/theme.scss
 physionet-django/static/bootstrap/css/bootstrap-grid.css
 physionet-django/static/bootstrap/css/bootstrap-reboot.css
 physionet-django/static/bootstrap/css/bootstrap.css
-physionet-django/static/bootstrap/css/theme.css
\ No newline at end of file
+physionet-django/static/bootstrap/css/theme.css
```
tompollard, line 151: Missing newline

tompollard (commented):
Thanks! Looks good to me, other than a couple of minor things that I've commented on.

.env.example
superryeti: ah i agree, it's not a good idea to trust a third party link, let me do that. Good idea about demo image on our own repo

tompollard: If there is an existing background, we can use that. Otherwise you can find a open licensed image instead (possibly the one in your previous link?).

superryeti: Yup, i agree, i was also worried about the usage rights of the image above and was about to ask you on slack.

superryeti: or because we kind of really need to merge this PR, i can remove the commit about image upload(and for now use the physionet background from physionet.org) and add image in another PR.

tompollard: > or because we kind of really need to merge this PR, i can remove the commit about image upload(and for now use the physionet background from physionet.org) and add image in another PR.

Sounds good. Looks to me like there might be some copyright issues with Nyan Cat that it would be best to avoid: https://en.wikipedia.org/wiki/Nyan_Cat

tompollard:
looks good to me, thanks amit

PR #1992, opened 2023-05-02
Make required text more visible for state/province on Credentialing form

superryeti:
**Context:**

During onboarding users to healthdatanexus.ai platform, January noticed that many users were confused with the State/Province field(on Credentialing form) as with a quick glance it looks like an optional field(but is actually required for US/Canada residents). This caused them to have to resubmit the form.

It seems like a good idea to have the Required message on the input label, as the eye seems to see that first.(see screenshots below, on the 2nd screenshot its very noticeable) 

Before
image

After
image

PS: Because this was such a small change, i opened a PR instead of an issue(as we close this if this doesn"t seem necessary or quickly review it if it makes sense)

Thanks to @january-adams for reporting it and solution.

### DIFF
--- a/physionet-django/user/forms.py
+++ b/physionet-django/user/forms.py
@@ -439,7 +439,7 @@ class PersonalCAF(forms.ModelForm):
             'job_title': """Your job title or position (e.g., student) within
                 your institution or organization.""",
             'city': "The city where you live.",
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks Amit and January, I agree this is clearer.

PR #1995, opened 2023-05-04
No downloading stuff in compilestatic

bemoody:
No, you may not download files from physionet.org as part of the compilestatic command.  That is madness.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -156,10 +156,6 @@ LIGHT='#007FA3'
 GRADIENT_60 = 'rgba(42, 47, 52, 0.6)'
 GRADIENT_85 = 'rgba(42, 47, 52, 0.85)'
 
-## physionet background image(direct download url)
-BACKGROUND_IMAGE='https://physionet.org/static/images/background.jpg'
-
-
 # Users settings
 # maximum number of emails that can be associated to a user model
 MAX_EMAILS_PER_USER = 10
--- a/physionet-django/user/management/commands/compilestatic.py
+++ b/physionet-django/user/management/commands/compilestatic.py
@@ -1,7 +1,5 @@
 # Management Command to compile static Sass files
 import os
-import urllib.request
-import urllib.error
 from django.core.management.base import BaseCommand, CommandError
 from decouple import config
 from django.core.management import call_command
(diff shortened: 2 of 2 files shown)


tompollard:
Is it bad enough to require undoing? Isn't the only risk that you do something stupid with your .env file, or is it worse than that?  @amitupreti, perhaps you could modify the approach to only allow local files to be added?

bemoody:
I'm sorry for not explaining.

Downloading stuff as part of the installation is not a good idea.  The installation process needs to be reproducible - installing the server twice with the same code must produce the same results.  Additionally for offline development and for a lot of testing, it's important to be able to run the server without external network access.  So no downloads!

(This particular case is especially bizarre because you've literally made that URL a prerequisite for itself.  We need to be able to set up a physionet server *when physionet.org is inaccessible*.)

There are countless ways of making the stylesheets locally customizable; I don't really care how that is done.  But you can't overwrite files tracked in the git repository and you can't require network access during installation.

tompollard:
Thanks Benjamin. 

> The installation process needs to be reproducible - installing the server twice with the same code must produce the same results. 

Hmm, I'd assumed that the .env file also supported local paths, but now I see this isn't the case (e.g. `BACKGROUND_IMAGE='./physionet-build/physionet-django/static
/images/background.jpg'`) fails with an `unknown url` error. Again, should have checked the PR more carefully. Hopefully @amitupreti can provide an option for local images files.

tompollard:
@amitupreti Okay if I merge this PR for now and we add back the custom image later? Will this cause any issues for Health Data Nexus?

alistairewj:
@amitupreti is off today but feel free to merge, it's nothing we can't manage.

superryeti:
> s it bad enough to require undoing? Isn't the only risk that you do something stupid with your .env file, or is it worse than that? @amitupreti, perhaps you could modify the approach to only allow local files to be added?

For sure, i remember initially i had added the option for local or download, but i removed the local option for some reason i dont remember(stupid)
Let me open another pr that lets us use local files.

Regarding merging, it should be fine as we haven't deployed changes to the HDN, so HDN should be good.

Issue #1990, opened 2023-05-01
User search should include last name


bemoody:
The search tool at http://localhost:8000/console/users/active/ searches based on the person's first name(s), username, and primary email address, but not their *last name*.

(This may have gone unnoticed because so often people have their last name as part of their email address.)

tompollard:
@mscanlan-git if you have time, do you want to give this a go? I can help out. It is a similar change to the one at: https://github.com/MIT-LCP/physionet-build/pull/1984/files (i.e. you are editing the same search query, but this time to search the lastname).

mscanlan-git:
@tompollard Sure, can we talk a bit about this on Tuesday? Have some questions about testing and making sure I'm using a proper approach.

tompollard:
Fixed in https://github.com/MIT-LCP/physionet-build/pull/1999

PR #1999, opened 2023-05-10
Update user search to include last names

mscanlan-git:
This PR addresses #1990 to allow searching for a user by their last name.


tompollard (approved):
Very cool, thanks @mscanlan-git. Looks good to me! Fixes #1990

Issue #1982, opened 2023-04-27
Reduce strictness of text validators


alistairewj:
Building on #1972 - we just noted that individuals can't include an ampersand (`&`) in many fields of their credentialing application, which raises an issue as quite a few organizations are described using an ampersand (e.g. Jack & Jill Foundation). @amitupreti fairly sure you are already working on this, so just something for you to explicitly keep an eye out for.

PR #2001, opened 2023-05-11
Relax validators to allow non risk characters

superryeti:
## Context:
As discussed on [Issue#1982](https://github.com/MIT-LCP/physionet-build/issues/1982), we want to make the validators less restrictive unless there is a security concern. This is a summary of the changes that will be made to the validators(mostly validators for longer text).

### Unit tests (Added unit tests for validators Pilot)

When writing code i often seem to make mistakes, so i wanted to experiment with TDD. I tried to write unit tests for all the validators first and then the code to make the tests pass.
It seemed to make the coding process easier and i was able to catch some bugs that i would have missed otherwise.
So i wanted to see what others thought of adding unit test for all the small methods/functions as well(Normally we seem to be doing this for just views) in the future or maybe if it looks like an overkill, i can remove the tests.

### validate_reference_response

Before the change, the reference response is very restrictive, users can only selected special characters are permitted.

After the change, we allow everything as long as the first character is a letter or a number.

`re.fullmatch(r'[^\W_0-9]([\w\',. -])+', value) (OLD)`
`re.fullmatch(r'\w.*', value, re.DOTALL) (NEW)`

### validate_research_summary

Before the change, the research summary is very restrictive, users can only use from a list of authorized special characters.

After the change, we allow everything as long as the first character is a letter or a number.

`re.fullmatch(r'[a-zA-Z][\w\'\[\]\n\r~!@#$%&*()+={};:‘’“”"<>?,./` -]+', value) (OLD)`
`re.fullmatch(r'\w.*', value, re.DOTALL) (NEW)`

### validate_organization

Previously we didn't allow special characters in org names, which fails for cases like Johnson & Johnson or 21st century fox.
Currently, we only validate that the first character is not a special character and there is no new line(i dont think Organization names have new lines)

`r'[^\W_0-9]([\w\',. -])+' (OLD)`
`r'^[a-zA-Z0-9].*' (NEW)`

I also checked all existing `validator_*` functions and they seemed fine to me.

### DIFF
--- a/physionet-django/user/test_validators.py
+++ b/physionet-django/user/test_validators.py
@@ -1,0 +1,20 @@
+from unittest import TestCase
+
+from django.core.exceptions import ValidationError
+
+from user.validators import validate_organization, validate_reference_response, validate_research_summary
+
+
+class TestValidators(TestCase):
+
+    def test_organization_with_newline_is_invalid(self):
+        self.assertRaises(ValidationError, validate_organization, 'Johnson\nJohnson')
+
+    def test_organization_with_special_character_is_valid(self):
+        self.assertIsNone(validate_organization('Johnson & Johnson'))
+
+    def test_organization_with_number_as_first_character_is_valid(self):
+        self.assertIsNone(validate_organization('21st century fox'))
+
+    def test_organization_with_special_character_as_first_character_is_invalid(self):
+        self.assertRaises(ValidationError, validate_organization, '&Johnson')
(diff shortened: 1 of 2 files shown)


tompollard:
Thanks Amit, looks good to me! Not convinced that we need both `validate_reference_response` and `validate_research_summary` (why not merge?), but we can worry about that another time.

superryeti:
> Thanks Amit, looks good to me! Not convinced that we need both `validate_reference_response` and `validate_research_summary` (why not merge?), but we can worry about that another time.

I agree, since both of them collect similar type of data(bunch of paragraphs from user)

I will open another PR with merged functions, just in case we might forget to do this later.

PR #2003, opened 2023-05-12
OAuth Provider Implementation 

Rutvikrj26:
This PR adds an [OAuth 2.0](https://oauth.net/2/) Provider. The Provider can be tested following these [procedures](https://www.toolsqa.com/postman/oauth-2-0-authorization-with-postman/).

Testing and Usage Instructions :

 1. Create an Application from the admin panel to get the client-id & client-secrets
image
(Also put in the redirect-uri for Postman : https://www.getpostman.com/oauth2/callback)

2. Fill in the details in the postman
image

For a detailed version - check out the procedures mentioned [here](https://www.toolsqa.com/postman/oauth-2-0-authorization-with-postman/).

**Usage Scenarios** :

This enables the Platform to be a provider and provide API-based resource authorization using the [OAuth](https://oauth.net/2/) workflow. This can enable Physionet-based authentication for third-party users. 
For Eg: Someone who is not registered with HDN but is registered with Physionet - can use this functionality(once HDN is added as a client) to access a pre-determined resource. This functionality needs to be extended for the implementation of these scenarios and is a first step towards it.


tompollard:
Thanks @Rutvikrj26, please could you provide some detail on how you anticipate this new functionality being used?

Rutvikrj26:
> Thanks @Rutvikrj26, please could you provide some detail on how you anticipate this new functionality being used?

Added it in the main section. Please let me know if you want me to add more details to it.

tompollard:
@Rutvikrj26 please could you rebase on dev?

tompollard:
@bemoody please could you take a look at this when you have a chance?

bemoody:
```
+CORS_ORIGIN_ALLOW_ALL = True
```
Sorry - this is **definitely not** something we can allow.

(What this does, apparently: adds `Vary: Authorization` and `Vary: Origin` and adds `Access-Control-Allow-Origin: *` for any request that specifies `Origin`.

It's totally cool to set `Access-Control-Allow-Origin: *` for any *public* resources.  For any resources that require authentication or a cookie of any kind, the answer is probably *no*.  It is complicated and messy and requires careful consideration.)

From what I understand, though, this should not be related to OAuth (but maybe has to do with how you're testing it.)  You might consider making this an option that you can turn on temporarily for development, with the understanding that this isn't the way it's going to work in production.

Rutvikrj26:
Hello Benjamin, 

Thanks for catching this oversight! 
As I commented on top of it, I had put it in while I was debugging a particular error related to access and missed out while removing the code. I am really sorry this came into the PR. I'll remove the code right away.

physionet-django/physionet/settings/base.py
```diff
@@ -82,13 +85,16 @@
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     # RedirectFallbackMiddleware should go at end of list, according
     # to the docs: https://docs.djangoproject.com/en/4.1/ref/contrib/redirects/
-    'django.contrib.redirects.middleware.RedirectFallbackMiddleware'
+    'django.contrib.redirects.middleware.RedirectFallbackMiddleware',
```
tompollard, line 90: Note the comment above about `RedirectFallbackMiddleware` being placed at the end of the list.

physionet-django/physionet/settings/base.py
```diff
@@ -82,13 +85,16 @@
...
+    'corsheaders.middleware.CorsMiddleware',
```
tompollard: According to the corsheaders documentation (https://pypi.org/project/django-cors-headers/):

> CorsMiddleware should be placed as high as possible, especially before any middleware that can generate responses such as Django’s CommonMiddleware or Whitenoise’s WhiteNoiseMiddleware. If it is not before, it will not be able to add the CORS headers to these responses.

tompollard:
@Rutvikrj26 please could you resolve the conflicts and double check the middleware order in base.py? I think we can then merge.

tompollard (approved):

physionet-django/physionet/settings/base.py
Rutvikrj26: Thanks for pointing out!
I read more about it and found it is essential to keep it as above as possible but below the security middleware. Furthermore, I'm keeping it above CSRF and auth middleware as they can further modify the response headers. I have also read onto the OAuth middleware and adjusted it's position to ensure it is below the session middleware to access the session details.

physionet-django/physionet/settings/base.py
Rutvikrj26: Done.

tompollard:
Thanks Rutvik, looks good to me!

tompollard:
Side note, I tested this following the steps at: https://django-oauth-toolkit.readthedocs.io/en/latest/getting_started.html#authorization-code

PR #2012, opened 2023-05-23
Remove pop up on email settings page, and clean up text and formatting

tompollard:
This is a small clean up of the content at: https://physionet.org/settings/emails/. The changes are:

1. Remove a "Please add an institutional email address" pop up that was causing issues for some users (blocking them from accessing the page)
2. Take this opportunity to clean up some of the language on the page.

### DIFF
--- a/physionet-django/user/templates/user/edit_emails.html
+++ b/physionet-django/user/templates/user/edit_emails.html
@@ -91,7 +91,7 @@
 
 <hr>
 <h4>Add Email</h4>
-<p>Add a new email address(Max emails: {{ max_associated_emails_allowed }}).</p>
+<p>Add a new email address (maximum {{ max_associated_emails_allowed }}).</p>
 <form action="{% url 'edit_emails' %}" method="post" class="form-signin row no-pd" name="add_email">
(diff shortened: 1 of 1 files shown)


bemoody:
```
-<h4>Primary Email</h4>
-<p>The email address used for authentication, and email notifications. We recom
mend using an institutional/academic email address where possible.</p>
+<h4>Primary Email (for login and notifications)</h4>
+<p>Please use an institutional email address where possible.</p>
```

I don't think this is clear.  It should be explained that this is the address we will use to contact you, and can be used to reset your password.

We shouldn't be telling people they need to set their institutional address as their primary address.  If we're relying on that, for credentialing or project submission or whatever, that's wrong and should be fixed.

tompollard:
I basically agree with both points, but my focus is removing the pop up while making a couple of tweaks to clean up typos and improve readability. If you have specific suggestion for new text, please could you post it here so I can add it? Otherwise perhaps we can raise an issue and then update later?

tompollard:
(undid changes to the text)

bemoody:
Thanks!

PR #2013, opened 2023-05-25
Display cloud credentials for users in the user management console

### LINKED ISSUE #2010: Display cloud credentials for users in the user management console
We should update the user management page in the console (https://physionet.org/console/user/manage/USERNAME/) so that it displays the users cloud credentials, e.g. 

- What is the user's Amazon ID?
- What is the user's Google Cloud ID?

mscanlan-git:
This PR resolves #2010 by creating two new lines in the user console to display a user's GCP email as well as AWS ID number.

### DIFF
--- a/physionet-django/console/templates/console/user_management.html
+++ b/physionet-django/console/templates/console/user_management.html
@@ -35,6 +35,30 @@
       {{ subject.is_credentialed }}
     </div>
   </div>
+  <div class="row mb-1">
+    <div class="col-md-3">
+      Amazon ID
+    </div>
+    <div class="col-md-9">
+      {% if aws_info %}
+        {{ aws_info }}
+      {% else %}
+        Not provided.
+      {% endif %}
+    </div>
+  </div>
+  <div class="row mb-1">
+    <div class="col-md-3">
+      Google Cloud ID:
+    </div>
+    <div class="col-md-9">
+      {% if gcp_info %}
+        {{ gcp_info }}
+      {% else %} 
+        Not provided.
+      {% endif %}
+    </div>
+  </div>
   
     <div class="row mb-1">
       <div class="col-md-3">
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -69,7 +69,8 @@ from user.models import (
     Training,
     TrainingType,
     TrainingQuestion,
-    CodeOfConduct
+    CodeOfConduct,
+    CloudInformation
 )
 from physionet.enums import LogCategory
 from console import forms, utility, services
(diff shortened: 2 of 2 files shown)


physionet-django/console/views.py
```diff
@@ -1072,6 +1073,14 @@ def user_management(request, username):
     Admin page for managing an individual user account.
     """
     user = get_object_or_404(User, username__iexact=username)
+    try:
+        aws_info = CloudInformation.objects.filter(user=user).values_list('aws_id', flat=True).get(pk=1)
```
tompollard: A simpler query is: 

`aws_info = CloudInformation.objects.get(user=user).aws_id`

physionet-django/console/views.py
```diff
@@ -1072,6 +1073,14 @@ def user_management(request, username):
...
+    except CloudInformation.DoesNotExist:
+        aws_info = None
+    try:
+        gcp_info = CloudInformation.objects.filter(user=user).values_list('gcp_email', flat=True).get(pk=1)
```
tompollard: `gcp_info = CloudInformation.objects.get(user=user).gcp_email`

physionet-django/console/templates/console/user_management.html
```diff
@@ -35,6 +35,30 @@ <h3>Profile</h3>
       {{ subject.is_credentialed }}
     </div>
   </div>
+  <div class="row mb-1">
```
tompollard, line 38: Later we might want to move the cloud details into a separate subsection, but this is fine for now.

physionet-django/console/views.py
```diff
@@ -1109,7 +1118,9 @@ def user_management(request, username):
                                                             'emails': emails,
                                                             'projects': projects,
                                                             'training_list': training,
-                                                            'credentialing_app': credentialing_app})
+                                                            'credentialing_app': credentialing_app,
+                                                            'aws_info': aws_info,
```
tompollard, line 1122: Instead of getting separate variables for each cloud ID (`aws_info`, `gcp_info`), we could just pass the `CloudInformation` information object and then access it in the template with `cloud_object.aws_id`, `cloud_object.gcp_email` etc, but I think your approach is fine for now.

tompollard (commented):
Thanks @mscanlan-git, this is great. 

Before we merge, please could you update the queries as suggested (e.g. user `aws_info = CloudInformation.objects.get(user=user).aws_id` instead of `aws_info = CloudInformation.objects.filter(user=user).values_list('aws_id', flat=True).get(pk=1)`).

I have made a couple of other comments about things we might want to think about in future (e.g. displaying the cloud content in a separate section on the page) but these can wait.

tompollard:
Looks good to me, thanks!

PR #2017, opened 2023-05-28
Rs/search api

Rutvikrj26:
This PR adds search functionality to the API-based access. 

**Usage :** 

API Endpoint : `http://127.0.0.1:8000/api/v1/project/published/search`

Type of requests : only `GET`

A direct GET request will give list of all the projects, same as the published endpoint.

For simple search of a keyword, you can send a get request to the endpoint with param `search_term` key and the actual keyword as the `value`. This includes all resources types : databases, software projects, challenges, and models.

Further granular access is possible via adding `resource_type` to params : 
for each resource_type, you can add `&resource_type=<chosen-resource-type>`
image

### DIFF
--- a/physionet-django/export/urls.py
+++ b/physionet-django/export/urls.py
@@ -6,12 +6,12 @@ from export import views
 urlpatterns = [
     # API V1 Routes
     path('v1/project/published/', views.PublishedProjectList.as_view(), name='Published_project_list'),
+    path('v1/project/published/search/', views.PublishedProjectSearch.as_view(), 
+         name='Published_project_search'),    
     path('v1/project/published/<str:project_slug>/', views.ProjectVersionList.as_view(),
          name='Published_Project_versions'),
     path('v1/project/published/<str:project_slug>/<str:version>/', views.PublishedProjectDetail.as_view(),
          name='Published_project_detail'),
-    path('v1/project/published/search/', views.PublishedProjectSearch.as_view(), 
-         name='Published_project_search'),
 ]
 
 # Parameters for testing URLs (see physionet/test_urls.py)
--- a/physionet-django/export/urls.py
+++ b/physionet-django/export/urls.py
@@ -5,13 +5,13 @@ from export import views
 
 urlpatterns = [
     # API V1 Routes
-    path('v1/project/published/', views.PublishedProjectList.as_view(), name='Published_project_list'),
-    path('v1/project/published/search/', views.PublishedProjectSearch.as_view(), 
-         name='Published_project_search'),    
+    path('v1/project/published/', views.PublishedProjectList.as_view(), name='published_project_list'),
+    path('v1/project/published/search/', views.PublishedProjectSearch.as_view(),
+         name='published_project_search/'),
     path('v1/project/published/<str:project_slug>/', views.ProjectVersionList.as_view(),
-         name='Published_Project_versions'),
+         name='published_project_versions'),
     path('v1/project/published/<str:project_slug>/<str:version>/', views.PublishedProjectDetail.as_view(),
-         name='Published_project_detail'),
+         name='published_project_detail'),
 ]
 
 # Parameters for testing URLs (see physionet/test_urls.py)
--- a/physionet-django/export/urls.py
+++ b/physionet-django/export/urls.py
@@ -10,6 +10,8 @@ urlpatterns = [
          name='Published_Project_versions'),
     path('v1/project/published/<str:project_slug>/<str:version>/', views.PublishedProjectDetail.as_view(),
          name='Published_project_detail'),
+    path('v1/project/published/search/', views.PublishedProjectSearch.as_view(), 
+         name='Published_project_search'),
 ]
 
 # Parameters for testing URLs (see physionet/test_urls.py)
--- a/physionet-django/export/views.py
+++ b/physionet-django/export/views.py
@@ -7,9 +7,9 @@ from rest_framework import generics
 from rest_framework.authentication import SessionAuthentication, BasicAuthentication
 from rest_framework import mixins
 from rest_framework.response import Response
-
-# Importing the get_content function from Search Module's views.py
 from search.views import get_content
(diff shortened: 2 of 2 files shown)


physionet-django/export/views.py
```diff
@@ -61,3 +64,45 @@ def get(self, request, project_slug, version, *args, **kwargs):
...
+
+class PublishedProjectSearch(mixins.ListModelMixin, generics.GenericAPIView):
+    """
+    Search for a Published Project using the get_content function inside Search Module's views.py
+    """
+
+    serializer_class = PublishedProjectSerializer
+
+    def get_queryset(self):
+        resource_type = self.request.POST.get('resource_type', '0,1,2,3') or self.request.GET.get('resource_type',
```
alistairewj: change to string and keep any integer logic internal

physionet-django/export/views.py
```diff
@@ -61,3 +64,45 @@ def get(self, request, project_slug, version, *args, **kwargs):
...
+                                                                                                  '0,1,2,3')
+        resource_type = [int(x) for x in resource_type.split(',')]
+        orderby = self.request.POST.get('orderby', 'relevance-desc') or self.request.GET.get('orderby',
```
alistairewj: change to `GET`

Rutvikrj26: Only GET Requests are allowed in the updated code.

physionet-django/export/views.py
Rutvikrj26: This has been updated to take in the string values of the resource types, namely database, software, challenge, model, or all.

physionet-django/export/urls.py
```diff
@@ -6,6 +6,8 @@
 urlpatterns = [
     # API V1 Routes
     path('v1/project/published/', views.PublishedProjectList.as_view(), name='Published_project_list'),
+    path('v1/project/published/search/', views.PublishedProjectSearch.as_view(),
+         name='Published_project_search/'),
```
tompollard: These names should be lower case, I think (e.g. `published_project_search`). 

Can we fix the others at the same time?

physionet-django/export/views.py
```diff
@@ -8,6 +8,9 @@
 from rest_framework import mixins
 from rest_framework.response import Response
 
+# Importing the get_content function from Search Module's views.py
```
tompollard: I don't think we need this comment (it is clear from the import what is happening)

physionet-django/export/views.py
```diff
@@ -61,3 +64,49 @@ def get(self, request, project_slug, version, *args, **kwargs):
...
+    Search for a Published Project using the get_content function inside Search Module's views.py
+    """
+
+    serializer_class = PublishedProjectSerializer
+
+    def get_queryset(self):
+        resource_type = self.request.GET.getlist('resource_type', 'all')
+        search_term = self.request.GET.get('search_term', ' ')
+
+        print(search_term, resource_type)
```
tompollard: Could you briefly explain the purpose of the print statements?

physionet-django/export/views.py
```diff
@@ -61,3 +64,49 @@ def get(self, request, project_slug, version, *args, **kwargs):
...
+
+        # Define a dictionary to map resource types to their corresponding integer values
```
tompollard: Could this be imported instead of hardcoded? It looks like we capture the types here:

https://github.com/MIT-LCP/physionet-build/blob/07c02703ca533ba379059d38dc4534fe99a1caa9/physionet-django/project/modelcomponents/coreproject.py#L57-L69

physionet-django/export/views.py
```diff
@@ -61,3 +64,49 @@ def get(self, request, project_slug, version, *args, **kwargs):
...
+            'software': 1,
+            'challenge': 2,
+            'model': 3,
+        }
+
+        # Ensure that resource_type is a list of valid options
+        resource_type_list = []
+        if resource_type != 'all':
+            for r_type in resource_type:
+                print("Checking resource_type: " + r_type)
```
tompollard: Is this print statement needed?

physionet-django/export/views.py
```diff
@@ -61,3 +64,49 @@ def get(self, request, project_slug, version, *args, **kwargs):
...
+            resource_type_list = [0, 1, 2, 3]
+
+        print(search_term, resource_type_list)
+
+        # Default to relevance descending order
+        queryset = get_content(resource_type_list, 'relevance', 'desc', search_term)
+
+        return queryset
+
+    def get(self, request, *args, **kwargs):
```
tompollard, line 105: Perhaps add short docstring?

physionet-django/export/views.py
```diff
@@ -61,3 +64,49 @@ def get(self, request, project_slug, version, *args, **kwargs):
...
+
+
+class PublishedProjectSearch(mixins.ListModelMixin, generics.GenericAPIView):
+    """
+    Search for a Published Project using the get_content function inside Search Module's views.py
+    """
+
+    serializer_class = PublishedProjectSerializer
+
+    def get_queryset(self):
```
tompollard, line 83: Perhaps add short docstring?

physionet-django/export/urls.py
Rutvikrj26: I have a bad habit of making the first characters upper case while typing. I've fixed the names to lower case and will keep it in mind from now on.

physionet-django/export/views.py
Rutvikrj26: I use these comments as a guide for writing the code. 
I will ensure to remove simple ones.

physionet-django/export/views.py
Rutvikrj26: The print statements were used while I was debugging - no need for them in production.
Removed them.

physionet-django/export/views.py
tompollard: Now getting these from `ProjectType`

tompollard:
Thanks Rutvik, looks good to me!

Issue #2016, opened 2023-05-27
Move events access control into authorization module


alistairewj:
In #1961 @amitupreti refactored the access checks into a separate module in the project app, so that all project authorization is done in one place, which has made project access in the code a lot easier to reason about. There are effectively two functions in the authorization module:

- get_accessible_projects(user) - list out all projects a user can access
- can_access_project(user, project) - can the user access the specified project?

(Side note: in my opinion, the reason to have the latter function with separate functionality is it is a hot path, and it might be worth optimization. Otherwise, it is trivially derived from `get_accessible_projects(user).filter(project=project)`).

Currently the events/models.py app skip the authorization module entirely:

https://github.com/MIT-LCP/physionet-build/blob/851735ff071b7d356e5b1c65061c02cae00be01a/physionet-django/events/models.py#L225-L241

This code should be moved into the authorization module. To keep it simple, creating a new `has_event_access(user, project)` function to the authorization module makes sense. Removing redundancy between has_event_access and has_access can be worked out in the future, and similarly improving the mechanism through which datasets are provisioned via events can equally be improved.

PR #2019, opened 2023-06-08
Move Event authorization code to authorization module

superryeti:
As discussed on https://github.com/MIT-LCP/physionet-build/issues/2016,  we want to move all authorization related logic to projects.authorization module to centralize authorization logic.

Note: here i created an events.py file inside authorization, because if we try to add this func inside the projects.authorization.access.py, we will get circular import. Also, this kind of organizes the authorization for event in a single file(instead of mixing up on access.py)

In future, it might make sense to move authorization as a separate app.

For testing, i created a new event, added dataset to it, joined the event waitlist, and approved the event waitlist(The events workflow)

### DIFF
--- a/physionet-django/events/templatetags/participation_status.py
+++ b/physionet-django/events/templatetags/participation_status.py
@@ -1,6 +1,7 @@
 from django import template
 
 from events.models import EventApplication
+from project.authorization.events import has_access_to_event_dataset as has_access_to_event_dataset_func
 
 register = template.Library()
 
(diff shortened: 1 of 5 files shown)


physionet-django/events/templatetags/participation_status.py
```diff
@@ -21,4 +22,4 @@ def is_on_waiting_list(user, event):
 
 @register.filter(name='has_access_to_event_dataset')
 def has_access_to_event_dataset(user, dataset):
-    return dataset.has_access(user)
+    return has_access_to_event_dataset_func(user, dataset)
```
tompollard, line 25: Do we need to rename the function here? Not sure it is necessary, and it is inconsistent with the approach used below (when importing to `authorization/access.py`).

tompollard (approved):

physionet-django/events/templatetags/participation_status.py
tompollard: I think Amit has moved on from the project, so I'll leave this to be addressed later.

PR #2020, opened 2023-06-13
Tests on Debian 12

bemoody:
Debian 12 "bookworm" is hot off the presses.  There's no rush, but I would like to upgrade in the near future - probably mid to late July.

(While Health Data Nexus certainly doesn't have to upgrade at the same time PhysioNet does, it'll be easier if we don't have a long transitional period.)

This pull request should (I think - I still don't fully understand this GitHub stuff) run the test suites in parallel in both Debian 11 and Debian 12 containers (both running on an Ubuntu host VM.)

*For the time being, please continue to use Debian 11 and Python 3.9 for regular development* (though if you feel adventurous, go ahead and set up both environments to test stuff in parallel.)

### DIFF
--- a/.github/workflows/physionet-build-test.yml
+++ b/.github/workflows/physionet-build-test.yml
@@ -14,7 +14,7 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
-    container: debian:11
+    container: ${{ matrix.container }}
     env:
       DJANGO_SETTINGS_MODULE: physionet.settings.settings
       TEST_GCS_INTEGRATION: false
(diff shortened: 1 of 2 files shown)


bemoody:
```
*** blocked system call 0x06 arch=0xc000003e rdi=0x1 rsi=0x7f0b8a1e7dd5 rdx=0x7ffd26565180 r10=0x000 r8=0x7f0b8a224c60 r9=0x7f0b8a224da0 [newfstatat]
```
oh, glibc, what did you do this time

(also, "newfstatat", seriously?)

(Actually it's pretty cool that this error message worked as intended! So that's a lightwave bug, probably an easy fix.)

bemoody:
> probably an easy fix

I should know better than to make comments like that

tompollard:
Looks good to me.

PR #2021, opened 2023-06-15
Fetch credentialed project example instead of hardcoding MIMIC-III

kshalot:
MIMIC-III is a dataset specific to PhysioNet. Right now the emails link to a page that does not exist for deployments other than PhysioNet.

This is just an example solution to this problem that we deployed to Health Data Nexus because we have an influx of credentialing applications at the moment. If you need to link a specific project (like MIMIC-III) and not the first one that the query finds, we could add a debranding envvar (e.g. `CREDENTIALING_EXAMPLE_PROJECT_SLUG`) which will specify which project should be linked or even combine the two approaches and default to the first one if the envvar is not present.

Also, using `example_project.title` will result in showing `...can access MIMIC-III Clinical Database
 by following the steps...` which is probably missing "the" before the title in this specific case. Other titles might not require the "the" article though

### DIFF
--- a/physionet-django/notification/templates/notification/email/process_training_complete.html
+++ b/physionet-django/notification/templates/notification/email/process_training_complete.html
@@ -15,14 +15,14 @@ Dear {{ applicant_name }},
 If you are able to address the issue(s), please resubmit your training at: {{ url_prefix }}{% url 'edit_training' %}.
 
 {% elif training.status.name == "ACCEPTED" %}We are pleased to say that your "{{ training.training_type.name }}" training was approved.
-{% if training.user.is_credentialed %}
-You are now able to access protected databases upon agreeing to the terms of usage. For example, you can access MIMIC-III by following the steps below:
+{% if training.user.is_credentialed and example_project %}
+You are now able to access protected databases upon agreeing to the terms of usage. For example, you can access {{ example_project.title }} by following the steps below:
 
-- Go to the project page at {{ url_prefix }}{% url 'published_project_latest' 'mimiciii' %}
+- Go to the project page at {{ url_prefix }}{% url 'published_project_latest' example_project.slug %}
 - Find the “Files” section in the project description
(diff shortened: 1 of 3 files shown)


physionet-django/notification/templates/notification/email/process_training_complete.html
```diff
@@ -15,19 +15,19 @@
...
+{% if training.user.is_credentialed and example_project %}
+You are now able to access protected databases upon agreeing to the terms of usage. For example, you can access {{ example_project.title }} by following the steps below:
 
-- Go to the project page at {{ url_prefix }}{% url 'published_project_latest' 'mimiciii' %}
+- Go to the project page at {{ url_prefix }}{% url 'published_project_latest' example_project.slug %}
 - Find the “Files” section in the project description
 - Click “Sign the data use agreement” to agree to the terms of usage for this dataset
 {% else %}
-Before accessing protected databases such as MIMIC-III, you will also need to complete the credentialing process.  
+Before accessing protected databases such as MIMIC-III, you will also need to complete the credentialing process.
```
tompollard, line 25: This sentence also contains "MIMIC-III"

tompollard (approved):
The following chunk returns MIMIC-III for us. I'd be happy to merge as is, and then add the `CREDENTIALING_EXAMPLE_PROJECT_SLUG` later:

```
PublishedProject.objects.filter(
        access_policy=AccessPolicy.CREDENTIALED, is_latest_version=True).first()
```

> Also, using example_project.title will result in showing ...can access MIMIC-III Clinical Database by following the steps... which is probably missing "the" before the title in this specific case. Other titles might not require the "the" article though

Not a big deal I don't think. I wouldn't worry about this for now.

physionet-django/notification/templates/notification/email/process_training_complete.html
tompollard: Added as an issue: https://github.com/MIT-LCP/physionet-build/issues/2022

PR #2018, opened 2023-06-05
Display project creation date in submission timeline.

tompollard:
The "submission timeline" pages display a timeline of events for a project (for example, "The project was submitted for review"; "Waiting for editor to be assigned.", etc...). 

There are times when it may be helpful to view the date of project creation.  This is a minor change that adds the date of project creation.

Screenshot 2023-06-01 at 10 43 42 PM

Issue #612, opened 2019-07-23
Handling of rejected/withdrawn projects


alistairewj:
When an editor rejects a project, it moves into the closed submission page for the user, and they can only view the submission history. This raises two issues:

1. Users will have no way of retaining any metadata - so potentially they lose their work with no warning.
2. Administrators have no way of managing the project - all you can see is the submission history and a record of the creation/rejection. Potentially these projects will cause bloat (e.g. a project can take up 1 TB of space and there's no record of that on the admin page).

PR #2023, opened 2023-06-16
Show history of all archived projects, not just those that were rejected.

tompollard:
## Motivation for the change

There are different reasons for projects to be "archived". One reason is that users decide to close a project themselves. Another reason is that a project is rejected after submission.

As far as I'm aware, the only archived projects that are viewable in the system are ones that were rejected. These can be viewed in the console at: https://physionet.org/console/rejected-submissions/

Currently, there are a large number of archived submissions which are invisible to administrators. We should fix this, to make sure all archived projects can be found.

## What is the change

This pull request follows https://github.com/MIT-LCP/physionet-build/pull/2018. We now display all archived projects in the console, not just those that were rejected.

## Future work

In a later pull request, I intend to make additional improvements, for example displaying the archiving reason in the submission history and in the project list at: https://physionet.org/console/rejected-submissions/

https://github.com/MIT-LCP/physionet-build/issues/612 is a related issue. Once this pull request is (I hope) merged, then I will follow up with changes that allow more metadata to be retained.


physionet-django/console/templates/console/archived_submissions.html
alistairewj, line 1: Line 5, 14, and 27 look like they need an update to "Archived"

alistairewj (changes requested):
One quick change, otherwise LGTM

physionet-django/console/templates/console/archived_submissions.html
tompollard: Good catch, thanks. I think this page will need a bit work later (preferably in a new PR), so that it displays more meaningful information in the list (e.g. reason for archiving).

alistairewj (approved):

PR #2025, opened 2023-06-20
Add libcap-dev as a build dependency

bemoody:
The libcap-dev package is required in order to build the sandboxed-lightwave server to run in "unprivileged mode" (without using setuid).  

This pull request doesn't actually change anything about how we're installing lightwave; it merely makes it possible to compile the master branch of lightwave.


tompollard:
Thanks Benjamin

Issue #2024, opened 2023-06-16
Display personal website for users in the user management console


tompollard:
In https://github.com/MIT-LCP/physionet-build/pull/2013 we added cloud information to the users page in the user management console. We should also display the user's personal website URL.

The URL is stored on the `User` model here:

https://github.com/MIT-LCP/physionet-build/blob/a96e768d2a879233371257f04ef3dff7b310fa3e/physionet-django/user/models.py#L684

We could also think about displaying the user's profile photo:

https://github.com/MIT-LCP/physionet-build/blob/a96e768d2a879233371257f04ef3dff7b310fa3e/physionet-django/user/models.py#L685C4-L687

...though it might be best to save this for a different pull request.

PR #2027, opened 2023-06-21
Display personal website for users in the user management console

mscanlan-git:
This PR address #2024 by adding a way to display the user's personal website in the user management console.


tompollard:
Thanks Michael! I now see that we were displaying the personal website when it was provided, but the whole row (i.e. both label and value) was skipped in cases where the website was not provided. I think your approach of always displaying the label ("Personal website: ") is an improvement.

PR #2028, opened 2023-06-21
Django 4.1.9 file upload compatibility, and tests

bemoody:
Between Django 4.1.7 and 4.1.9 there have been some changes to how multiple-file form fields are handled.  Here's what the announcement says (https://www.djangoproject.com/weblog/2023/may/03/security-releases/):

> CVE-2023-31047: Potential bypass of validation when uploading multiple files using one form field
>
> Uploading multiple files using one form field has never been supported by forms.FileField or forms.ImageField as only the last uploaded file was validated. Unfortunately, Uploading multiple files topic suggested otherwise.
>
> In order to avoid the vulnerability, ClearableFileInput and FileInput form widgets now raise ValueError when the multiple HTML attribute is set on them. To prevent the exception and keep the old behavior, set allow_multiple_selected to True.
>
> For more details on using the new attribute and handling of multiple files through a single field, see Uploading multiple files.
>
> Thanks Moataz Al-Sharida and nawaik for reports.

And this I guess is the relevant commit: https://github.com/django/django/commit/e7c3a2ccc3a562328600be05068ed9149e12ce64

Actually, "setting allow_multiple_selected to True" doesn't exactly keep the old behavior - there's also something funky going on (something to do with `value_from_datadict`, I guess) which means that in addition to subclassing `FileInput` we also have to subclass `FileField`.

The code here is lifted from the current Django docs (https://docs.djangoproject.com/en/4.2/topics/http/file-uploads/#uploading-multiple-files), but also explicitly sets the 'multiple' attribute for backward compatibility with Django 4.1.7.

This *should* work, but please test it before merging.

### DIFF
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -101,13 +101,42 @@ class ActiveProjectFilesForm(forms.Form):
         return data
 
 
+class MultipleFileInput(forms.ClearableFileInput):
+    """
+    Variant of ClearableFileInput that allows uploading multiple
+    files in a single form field.
+    """
+    allow_multiple_selected = True
+
+    def __init__(self, attrs={}):
+        super().__init__(attrs={'multiple': True, **attrs})
+
+
+class MultipleFileField(forms.FileField):
+    """
+    Variant of FileField that allows uploading multiple files in a
+    single form field.
+    """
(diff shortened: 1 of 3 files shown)


tompollard:
Looks great Benjamin, thanks for doing this.

Issue #715, opened 2019-09-03
URLs should only be displayed in public profiles of credentialed users


tompollard:
Allowing any user to create a profile with a link to an external site is a bit risky. If we are going to support user-generated URLs in public profiles, then we need to do some kind of review of the user first. My suggestion is that URLs can only be displayed in public profiles if the user is credentialed.

elfeto:
There are several credentialed people that have random external websites in their profile. It might be their company, but it still the home page for the website, and there are even at least one that cannot be reached. 

The same goes for un-credentialed people that most of them are the same, but one that is a link to a laptop buying site. 

Maybe doing what @alistairewj  loves, and create a dashboard like admin page, which lists all changes for the current week? 
For example:
 - UserX added a email address
 - UserY signed DUA for O database
 - UserA added a picture
 - UserL added a website to his profile

Then we can add things like Project X has an outstanding storage request there also, and more things?

PR #2029, opened 2023-06-22
Display personal website for credentialed users only on public profile

mscanlan-git:
This PR addresses #715 by only displaying a user's website link if they are a credentialed user and if they have a link available to display. If both these conditions are not met, there will be no link listed on the public profile.

### DIFF
--- a/physionet-django/user/templates/user/public_profile.html
+++ b/physionet-django/user/templates/user/public_profile.html
@@ -68,14 +68,14 @@ Profile for {{ public_user.username }}
       {% endif %}
 
       {% if public_user.is_credentialed and profile.website %}
-    <div class="row mb-1">
-      <div class="col-md-2">
-        Website:
-      </div>
-      <div class="col-md-10">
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks Michael, I think this is an improvement on existing behavior. 

Instead of displaying ("not provided") for non-credentialed users, perhaps we should completely omit the website field (both the "Website:" and the URL if provided)?

It could be confusing for people otherwise (i.e. why is it telling me the website has not been provided, when it has been?).

mscanlan-git:
> Thanks Michael, I think this is an improvement on existing behavior.
> 
> Instead of displaying ("not provided") for non-credentialed users, perhaps we should completely omit the website field (both the "Website:" and the URL if provided)?
> 
> It could be confusing for people otherwise (i.e. why is it telling me the website has not been provided, when it has been?).

Good idea. The website url will now only show if the user is a credentialed user and has a website url to include on their public profile.

tompollard:
Thanks!

PR #2038, opened 2023-07-03
Add styling to SSO login button

kshalot:
This just adds an icon and some styling to the SSO button. This does not impact the regular login at all.


tompollard:
Thanks Karol

Issue #2034, opened 2023-06-30
TypeError (argument of type 'NoneType' is not iterable): if '@' in username


tompollard:
We are seeing multiple versions of the following error, across multiple projects. The error seems related to the recent update in https://github.com/MIT-LCP/physionet-build/pull/2003 to introduce OAuth support.

The `HTTP_USER_AGENT` looks suspicious! In all errors the agent is either `Microsoft Office Excel 2014`,`Microsoft Office PowerPoint 2014`; or `Microsoft Office Word 2014`.

```
Internal Server Error: /files/SLUG/1.0/FOLDER/


TypeError at /files/SLUG/1.0/FOLDER/
argument of type 'NoneType' is not iterable

Request Method: HEAD
Request URL: 
Django Version: 4.1.9
Python Executable: /home/physionet/python-env/physionet/bin/uwsgi
Python Version: 3.9.2
Python Path: ['.', '', '/usr/lib/python39.zip', '/usr/lib/python3.9', '/usr/lib/python3.9/lib-dynload', '/physionet/python-env/physionet/lib/python3.9/site-packages']

Traceback:

File "/physionet/python-env/physionet/lib/python3.9/site-packages/django/core/handlers/exception.py" in inner
 56.                 response = get_response(request)

File "/physionet/python-env/physionet/lib/python3.9/site-packages/oauth2_provider/middleware.py" in __call__
 32.                 user = authenticate(request=request)

File "/physionet/python-env/physionet/lib/python3.9/site-packages/django/views/decorators/debug.py" in sensitive_variables_wrapper
 42.             return func(*func_args, **func_kwargs)

File "/physionet/python-env/physionet/lib/python3.9/site-packages/django/contrib/auth/__init__.py" in authenticate
 77.             user = backend.authenticate(request, **credentials)

File "./user/backends.py" in authenticate
 15.         if '@' in username:

Exception Type: TypeError at /files/SLUG/1.0/FOLDER/
Exception Value: argument of type 'NoneType' is not iterable
Request information:
USER: [unable to retrieve the current user]

GET: No GET data
POST: No POST data
FILES: No FILES data
COOKIES: No cookie data

META:
HTTP_USER_AGENT = 'Microsoft Office Excel 2014'
REMOTE_ADDR = 'XXXXXX'
```

Rutvikrj26:
I tried recreating the issue using the Demo version but was unsuccessful. 
[...]

PR #2039, opened 2023-07-05
Fix server error if username is none

alistairewj:
We're receiving a lot of `TypeError` due to, presumably, crawlers trying to login using token based auth (e.g. user agent Microsoft Excel 2014). This is a quick fix to prevent the server errors so we can investigate further.

Ref: #2034

### DIFF
--- a/physionet-django/user/backends.py
+++ b/physionet-django/user/backends.py
@@ -12,6 +12,10 @@ class DualAuthModelBackend(ModelBackend):
     This is a ModelBacked that allows authentication with either a username or an email address.
     """
     def authenticate(self, request, username=None, password=None):
+        # After implementation of OAuth2, (presumably) many crawlers started to attempt
+        # authentication with tokens, causing server errors due to the comparison below.
+        if username is None:
+            return None
         if '@' in username:
             kwargs = {'email': username.lower()}
         else:


tompollard:
@bemoody pointed out that we can reproduce the error with:

```
curl -H user-agent:'Microsoft Office Word 2014' -H accept-auth:'badger,Wlid1.1,Bearer,Basic,NTLM,Digest,Kerberos,Negotiate,Nego2' -H authorization:'Bearer' localhost:8000
```

He suggested that:

1. Possibly our custom authorization (`DualAuthModelBackend`) is not behaving in the way that Django expects it to behave. We should look at the current authorization module and compare to our implementation.
2. It's possible that the fix in this PR will break OAuth.
3. We need to better understand the OAuth module and work out what it is doing. We could consider temporarily removing the OAuth middleware in the meantime.

alistairewj:
> 2\. It's possible that the fix in this PR will break OAuth.

I think it absolutely will, but no one is using OAuth right now AFAIK?

tompollard:
> I think it absolutely will, but no one is using OAuth right now AFAIK?

Right now no, but we will still need to fix this. I raised an issue at: https://github.com/MIT-LCP/physionet-build/issues/2040

Issue #1129, opened 2020-09-08
Mismatch between maximum length of affiliation in the Affiliation and Contact models


tompollard:
When a project is published, an error is raised in cases when a user has added an affiliation of over 150 characters and he/she is the contact author (or an error should be raised...SQLLite seems to overlook max lengths, so the issue is not reproducible on a local sqlite setup).

The problem is a mismatch between the following two models in `/project/models.py` (no idea why 202 was chosen for affiliations):

```py
class Affiliation(models.Model):
    """
    Affiliations belonging to an author
    """
    name = models.CharField(max_length=202, validators=[validate_affiliation])
    author = models.ForeignKey('project.Author', related_name='affiliations',
        on_delete=models.CASCADE)

...

class Contact(models.Model):
    """
    Contact for a PublishedProject
    """
    name = models.CharField(max_length=120)
    affiliations = models.CharField(max_length=150)
    email = models.EmailField(max_length=255)
    project = models.OneToOneField('project.PublishedProject',
        related_name='contact', on_delete=models.CASCADE)
```

When a project is published, the `Affiliation.name` is copied to the `Contact.affiliations` which errors if the length is >150. A fix would be to bump up the max_length of the `Contact.affiliations` field.

bemoody:
Good catch.  The name field is wrong, too; the maximum length of a
name would be 100 + 1 + 50 (based on the equally arbitrary limits in
user.models.Profile).

Perhaps the Contact fields simply shouldn't have maximum lengths at
all.  Or perhaps they should refer to constants (e.g.
MAX_AFFILIATION_LENGTH), similar to the MAX_FILENAME_LENGTH.

I think the Contact fields probably *should* have validators
(user.validators.validate_name and
user.validators.validate_affiliation).

tompollard:
> Perhaps the Contact fields simply shouldn't have maximum lengths at all.

I thought about this, but it seems like it is probably a good idea to keep some kind of constraint. 

> Or perhaps they should refer to constants (e.g. MAX_AFFILIATION_LENGTH), similar to the MAX_FILENAME_LENGTH.

I like this idea. I'll try to update https://github.com/MIT-LCP/physionet-build/pull/1130 later today.

bemoody:
Actually, this is more complicated than I thought.

An Author can have multiple affiliations (how many?) and each of them
can be up to 202 characters.  "affiliations" lists all of the
affiliations, separated by "; " (ActiveProject.publish, line 1745.)

[...]

PR #2043, opened 2023-07-07
Increase maximum character length for affiliations 

lamawmouk:
This PR addresses the issue #1129. The problem arises when a project is published, and an error is raised if a user includes an affiliation of over 150 characters while being the contact author.

In this pull request, I have modified the codebase to increase the maximum character length for affiliations in the `Contact` class model. This update ensures that longer affiliations can be accommodated without triggering any errors.

Please review and merge the change at your convenience.


tompollard:
Thanks for looking at this Lama.

Increasing the length to 300 characters is definitely an improvement. It doesn't solve the underlying issue though (that three affiliations, each of 150 characters, get combined into a single field). 

Instead of bumping to 300 characters, how about following Benjamin's suggestion below by removing the length constraint?

> Perhaps the Contact fields simply shouldn't have maximum lengths at all. 

This should prevent us getting more errors when publishing projects. We can then think about improving the validation on the fields later.

bemoody:
> Instead of bumping to 300 characters, how about following Benjamin's suggestion below by removing the length constraint?

This would mean either:

- changing the `models.CharField` to a `models.TextField` and removing `max_length`

- keeping `CharField` and setting `max_length=None` (supported since Django 4.2, if I understand correctly?) which potentially has portability issues

I think I would slightly prefer having an explicit limit (MAX_AFFILIATIONS * (MAX_AFFILIATION_LENGTH + 2)) but removing the limit should also work.

tompollard:
> I think I would slightly prefer having an explicit limit (MAX_AFFILIATIONS * (MAX_AFFILIATION_LENGTH + 2)) but removing the limit should also work.

This also sounds fine to me!

tompollard:
Thanks Lama, looks good! Before we merge, please could you:

- Remove the trailing whitespace in physionet-django/project/forms.py (line 670)
- Squash the commits? (perhaps one commit with the code change, and one commit for the migration)

lamawmouk:
Thank you Benjamin and Tom! I have added the changes.

tompollard:
Looks great, thanks @lamawmouk.

I see the migration was generated using an older version of Django (`# Generated by Django 3.1.14 on 2023-07-11 20:51`). Not an issue here, but we're now using Django 4 so you might want to update it at some point.

PR #2045, opened 2023-07-14
Allow an associated publication to be added on the management page for published projects

tompollard:
## Background

On the landing page for published projects, we display three suggested citations:

1) The citation for the data/software/etc publication itself. This is required to meet the FAIR criteria. People should be able to directly cite the published object.
2) [Optional] A citation to an associated paper, such as a data paper in NPJ Scientific Data. This is partly to support the legacy of proxy papers being needed to gain credit for data sharing.
3) A citation to a PhysioNet paper.

## Problem

After publication of a dataset, it is fairly common for us to receive a request to add a citation for point 2. Right now, adding the paper involves either (a) creating a new version of the dataset (burdensome) (b) or manually adding the paper via the shell (risky).

I think option (a) is still desirable, because it is best if the paper is also added as a formal reference. It is burdensome to release new versions though, and so we often end up doing (b).

## What is the change introduced in this pull request?

This pull request adds a section to the project management form at http://localhost:8000/console/published-projects/SLUG/VERSION/ (e.g. http://localhost:8000/console/published-projects/demowave/1.0.0/). The new section allows a citation to a paper to be added (but not edited for now).

Example of a project with an associated publication:

Screenshot 2023-07-14 at 5 32 06 PM

Example of a project without an associated publication:

Screenshot 2023-07-14 at 5 31 25 PM

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -708,7 +708,7 @@ class PublishedProjectContactForm(forms.ModelForm):
         return contact
 
 
-class PublishedProjectAddPublication(forms.ModelForm):
+class AddPublishedPublicationForm(forms.ModelForm):
     class Meta:
         model = PublishedPublication
         fields = ('citation', 'url')
@@ -717,10 +717,20 @@ class PublishedProjectAddPublication(forms.ModelForm):
         super().__init__(*args, **kwargs)
         self.project = project
 
-    def save(self):
+    def clean(self):
+        cleaned_data = super().clean()
+        existing_publication = PublishedPublication.objects.filter(project=self.project).first()
+
+        if existing_publication:
+            raise forms.ValidationError("A publication already exists for this project.")
+
+        return cleaned_data
+
+    def save(self, commit=True):
         publication = super().save(commit=False)
         publication.project = self.project
-        publication.save()
+        if commit:
+            publication.save()
         return publication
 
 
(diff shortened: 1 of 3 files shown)


bemoody:
Not your fault, but PublishedPublication is rather messy - we only allow people to set a single publication per project, but this isn't enforced by the model.

The Django-ish way to allow editing "zero or more" objects is to use a formset; you can see how PublicationFormSet works in `project.views.project_discovery` and `console.views.copyedit_submission`.  It's messy and hard to follow, but it might be possible to use more-or-less the same code for PublishedPublication.

Alternatively, you could add a `clean` method to prevent adding multiple publications.  TBH, though, I'd much prefer that if we add a form here, it should be easy to undo if you make a mistake.

```
class PublishedProjectAddPublication(forms.ModelForm):
```
If you create a form class, please include "Form" in the name, e.g. `AddPublishedPublicationForm`.

tompollard:
@bemoody I have updated the form name as suggested (now `AddPublishedPublicationForm`) and added a clean method to prevent multiple publications. Are you okay for this to be merged?

I haven't added the ability to edit publications mainly because it's extra work to implement this, but also partly because I'm not sure it is desirable to be able to edit the publications willy nilly.

Right now, I just want something that means we don't have to go into the shell so often to add publications. We should think more about what level of editing is actually desirable.

bemoody:
Not tested but looks fine to me.

tompollard:
> Not tested but looks fine to me.

I hereby promise that it works.

bemoody (approved):

PR #2046, opened 2023-07-25
Oauth Provider tests

### LINKED ISSUE #2040: Authorization: argument of type 'NoneType' is not iterable
Since the addition of https://github.com/MIT-LCP/physionet-build/pull/2003, we have been regularly seeing an `argument of type 'NoneType' is not iterable` error:

```
Internal Server Error: /content/autonomic-aging-cardiovascular/

TypeError at /content/autonomic-aging-cardiovascular/
argument of type 'NoneType' is not iterable

Traceback:

File "/physionet/python-env/physionet/lib/python3.9/site-packages/django/core/handlers/exception.py" in inner
 56.                 response = get_response(request)

File "/physionet/python-env/physionet/lib/python3.9/site-packages/oauth2_provider/middleware.py" in __call__
 32.                 user = authenticate(request=request)

File "/physionet/python-env/physionet/lib/python3.9/site-packages/django/views/decorators/debug.py" in sensitive_variables_wrapper
 42.             return func(*func_args, **func_kwargs)

File "/physionet/python-env/physionet/lib/python3.9/site-packages/django/contrib/auth/__init__.py" in authenticate
 77.             user = backend.authenticate(request, **credentials)

File "./user/backends.py" in authenticate
 15.         if '@' in username:

Exception Type: TypeError at /content/autonomic-aging-cardiovascular/
Exception Value: argument of type 'NoneType' is not iterable
Request information:
USER: AnonymousUser

GET: No GET data
POST: No POST data
FILES: No FILES data
COOKIES: No cookie data
```

The error can be reproduced with:

```
curl -H user-agent:'Microsoft Office Word 2014' -H accept-auth:'badger,Wlid1.1,Bearer,Basic,NTLM,Digest,Kerberos,Negotiate,Nego2' -H authorization:'Bearer' localhost:8000
```

See also: https://github.com/MIT-LCP/physionet-build/pull/2037.  For now a temporary fix is: https://github.com/MIT-LCP/physionet-build/pull/2039

@bemoody's suggestions/notes on this issue:

1. Our custom authorization (`DualAuthModelBackend`) does not seem to be behaving in the way that Django expects it to behave. We should look at the current authorization module and compare to our implementation.
2. It is possible that the fix in https://github.com/MIT-LCP/physionet-build/pull/2039 will break OAuth.
3. We need to better understand the OAuth module and work out what it is doing. We could consider temporarily removing the OAuth middleware in the meantime.

Rutvikrj26:
As Described by @bemoody in #2040, I've added the Test Case for OAuth. This test case allows access to a protected resource endpoint on the server. 

The Test first sets up an OAuth application and generates a token. This is the BaseTest, which is the parent and gets inherited for further tests. There are two testcases we want to set up - as mentioned in #2040. The first testcase checks the ability to access the protected resource once a user has a token. The second testcase focuses on authentication and achieving the token that can allow access. 

Testcase 1 : Once the person has a token, he can get authorized to access Hello - an endpoint emulating a protected resource. There are two tests added - One demonstrating a denial for unauthenticated access. The second one has an auth header with a bearer token and is allowed to access the protected resource. 

Testcase 2 : This test demonstrated the authorization code flow for two different kind of applications supported by OAuth Workflow. 

1. Confidential Application: OAuth2 client credentials is a grant flow that allows a web service or an app (confidential client) to use its own credentials, instead of impersonating a user, to authenticate when calling another web service or a web resource, such as a REST API. This is commonly used for server-to-server interactions that run in the background without immediate user interaction. The permissions are granted directly to the application itself by an administrator.

2. Public Applciation: A Public Application can not securely store client secrets, like a mobile application. These are untrusted and hence follow a [Proof Key for Code Exchange (PKCE) workflow](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow-with-proof-key-for-code-exchange-pkce).

There are two tests, each resembling one application type and the respective authentication flow to be used for implementing the authentication.

As can be seen, we can implement a direct channel between Physionet and Health Data Nexus, which is not exposed to the regular public. An authenticated user on Health Data Nexus can easily access a protected resource on Physionet(for Eg: Training), since the access will be granted to the HDN platform.

### DIFF
--- a/physionet-django/oauth/views.py
+++ b/physionet-django/oauth/views.py
@@ -2,6 +2,6 @@ from django.http import HttpResponse
 from oauth2_provider.views.generic import ProtectedResourceView
 
 
-class ApiEndpoint(ProtectedResourceView):
+class hello(ProtectedResourceView):
     def get(self, request, *args, **kwargs):
         return HttpResponse('Hello, OAuth2!')
--- a/physionet-django/oauth/tests.py
+++ b/physionet-django/oauth/tests.py
@@ -10,6 +10,7 @@ from oauth2_provider.models import get_access_token_model, get_application_model
 from django.urls import reverse
 from urllib.parse import parse_qs, urlparse
 from oauth2_provider.settings import oauth2_settings
+from django.utils.crypto import get_random_string
 
 Application = get_application_model()
 AccessToken = get_access_token_model()
@@ -67,18 +68,6 @@ class BaseTest(TestCase):
 
         return auth_headers
 
-    def get_random_string(
-        self,
-        length=12,
-        allowed_chars="abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789",
-    ):
-        """
-        Return a securely generated random string.
-        The default length of 12 with the a-z, A-Z, 0-9 character set returns
-        a 71-bit value. log_2((26+26+10)^12) =~ 71 bits
-        """
-        return "".join(random.choice(allowed_chars) for i in range(length))
-
 
 class TestOAuth2Authentication(BaseTest):
     def test_unauthenticated(self):
@@ -102,7 +91,7 @@ class BaseAuthorizationCodeTokenView(BaseTest):
         """
         Generate a code verifier and a code challenge according to the PKCE
         """
-        verifier = self.get_random_string(length=length)
+        verifier = get_random_string(length=length)
         if algorithm == "S256":
             challenge = (
                 base64.urlsafe_b64encode(hashlib.sha256(verifier.encode()).digest())
--- a/physionet-django/oauth/tests.py
+++ b/physionet-django/oauth/tests.py
@@ -1,56 +1,49 @@
-from django.test import TestCase
-from user.models import User
-from oauth2_provider.models import get_application_model, AccessToken
-from django.utils import timezone
 from datetime import timedelta
+from django.test import TestCase
+from django.utils import timezone
+from user.models import User
+from oauth2_provider.models import get_access_token_model, get_application_model
 
 Application = get_application_model()
+AccessToken = get_access_token_model()
 
-class OAuthTestCase(TestCase):
+
+class TestOAuth2Authentication(TestCase):
     def setUp(self):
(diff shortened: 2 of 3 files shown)


physionet-django/oauth/views.py
```diff
@@ -2,6 +2,6 @@
 from oauth2_provider.views.generic import ProtectedResourceView
 
 
-class ApiEndpoint(ProtectedResourceView):
+class hello(ProtectedResourceView):
```
tompollard, line 5: Did this happen intentionally?

Rutvikrj26: Yup - I changed it to Hello to keep the view name and the URL endpoint have the same name.

bemoody:
Sounds good - thanks!

It looks like your test cases aren't currently using the 'applications/register/' view, so is there a reason that you moved it?

(I don't know if we wanted to limit the use of OAuth, at least initially, by making those views only accessible in debug mode.  I'm not opposed in principle to opening the system up to anyone who wants to use it, but we'd need to understand the implications.)

Rutvikrj26:
Thanks for catching the Register Endpoint!
That one is supposed to be inside the debug and only be available via the Django Superuser Panel in production. Once the team agrees on the implementation, I can add these endpoints to the Admin panel, but until we start using Oauth, it is better not to have them in production.

bemoody:
This looks good, thanks!

One small nit - do you need to define `get_random_string()` here?  Couldn't you use `from django.utils.crypto import get_random_string` instead?

I think we also want to test the token returned in `test_basic_auth` and `test_secure_auth_pkce`, by doing something like:
```
        ...
        response = self.client.post(
            reverse("oauth2_provider:token"), data=token_request_data, **auth_headers
        )
        self.assertEqual(response.status_code, 200)

        self.client.logout()

        token = response.json()['access_token']
        auth = self._create_authorization_header(token)
        response = self.client.get("/oauth/hello", HTTP_AUTHORIZATION=auth)
        self.assertEqual(response.status_code, 200)
```
(or perhaps we don't need to call `client.login` to begin with, since we're using basic auth?)

bemoody:
Thanks!

Issue #2051, opened 2023-07-28
edit_ethics doesn't honor SYSTEM_MAINTENANCE_NO_UPLOAD


bemoody:
The `edit_ethics` view (project/views.py) appears to allow uploading and deleting files, but appears to ignore the SYSTEM_MAINTENANCE_NO_UPLOAD setting.

(Unless I've missed something.  I haven't followed the logic completely.)

If SYSTEM_MAINTENANCE_NO_UPLOAD is set, all site actions that write to the static or media directories must be disallowed.  The easiest way is to disable the entire view, as in `create_project`:

```
@login_required
def create_project(request):
    if settings.SYSTEM_MAINTENANCE_NO_UPLOAD:
        raise ServiceUnavailable()
    ...
```

but it'd be better to only reject POSTs.

PR #2062, opened 2023-08-18
Refactoring ProjectFiles creation and usage

bemoody:
Currently, project file storage is (mostly[*]) managed by the "ProjectFiles" classes: when you want to read or write or rename or delete a file, you call the appropriate ProjectFiles method rather than calling built-in `os` functions.  This works great!

However, in the long term we'd like to be able to support multiple project storage backends (which might mean distributing files across multiple local servers, and/or keeping some files entirely on remote servers.)  Details are very much TBD.

This means, however, we can't assume every project uses the same backend; although there will probably always be a single default backend used for new ActiveProjects, code will have to assume that each each existing project might use a different backend.

So, new code should not be calling the global `ProjectFiles` function, which will eventually go away; instead, assuming you have an existing project object, use `project.files` to access that project's files.  And don't assume that one project's files backend can be used to manipulate a different project.

[*] things I noticed:
- lightwave doesn't use ProjectFiles.
- PublishedProject.remove_zip is weird and should probably be rewritten.
- Some asymmetry with UploadedDocument.  I still don't understand the workings there, see issue #2051.

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -30,7 +30,6 @@ from project.models import (
     SubmissionStatus,
     exists_project_slug,
 )
-from project.projectfiles import ProjectFiles
 from project.validators import MAX_PROJECT_SLUG_LENGTH, validate_doi, validate_slug
 from user.models import CodeOfConduct, CredentialApplication, CredentialReview, User, TrainingQuestion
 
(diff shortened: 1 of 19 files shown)


tompollard:
looks great, thanks benjamin.

PR #2067, opened 2023-08-29
Remove version_order from UnpublishedProject

bemoody:
The ActiveProject and ArchivedProject classes have a `version_order` attribute, but this is *not* the same as the `version_order` of PublishedProject.

(For example, if you have already published version 1.0.0 (version_order=0) and 2.0.0 (version_order=1), and now you create a version 1.0.1, the new ActiveProject will have version_order=2.  Only when and if the new version is *published* will the correct version order be defined.)

(See `set_version_order` in publishedproject.py for details of how that works.)

Existing code like `if project.version_order > 0` is technically correct but looks wrong, because what we actually want to know is "are there any existing published versions?", not "is this the numerically-oldest version?"

And as it happens, UnpublishedProject already has a boolean field (is_new_version) to tell us exactly that.  Use it, and remove the confusing and redundant version_order field.


tompollard:
Thanks! I can't help thinking we ended up with the complication `version_order` logic for a reason, but I can't work out what that reason might have been. This new approach is simpler and worked well in testing, so I'll go ahead and merge.

PR #2069, opened 2023-08-31
Allow admin to be a project editor

bemoody:
For convenience in testing, the demo "admin" user should be
allowed both to assign editors to projects, and to serve as an
editor (i.e., they should be able to assign a project to
themselves.)


tompollard:
thanks!

PR #2070, opened 2023-09-01
Bump version `hdn-research-environment` to `2.3.6`

kshalot:
Zero impact on the PhysioNet deployment. We might have one or two similar PRs coming as well. It's not blocking us so it doesn't have to be merged ASAP.


tompollard:
The changes look straightforward, so I'm happy to merge. I haven't tested the updated package, so I'm assuming that it works for you!

Issue #1758, opened 2022-12-20
Access Policy cannot be changed from Credentialed / Restricted to Restricted / Open


briangow:
Once a project is submitted as Credentialed / Restricted it cannot be downgraded to Restricted / Open . Attempts to do so will result in an "Invalid submission. See errors below." message. No additional information is provided below. This will occur if a downgrade is attempted both for the editor in the copyedit stage or for the author in a revision stage. 

The error when trying to downgrade from Restricted to Open is:

```
self._errors
{'dua': ['This field is required.'], 'required_trainings': ['This field is required.']}
```

This should be fixed so that editors and authors can change the access policy as desired throughout the publication process.

superryeti:
Weirdly, when I tested locally, 
1. i was able to reproduce the error with `dua` only.( i tried to change accesspolicy from Credentialed --> Restricted, Credentialed --> open, Restricted --> Open)
2. I could reproduce this before project submission, during revision and during copyedit(so basically everywhere)
3. i could only reproduce this error message `Select a valid choice. That choice is not one of the available choices.`
I spent some time on trying to figure it out, and everything seems to be right

In every test, the input tag for `dua` is disabled by the `AccessMetadataForm` and the required  was also removed from the  `input` tag for `dua` by `AccessMetadataForm` was also removed and the `request.POST` doesn't have any data for `dua` and ` required_training` for open access policies.

the `dua` has a foreign key relation to `project.DUA` , so maybe  when the form is initialized on `project/views.py` and `console/views.py`  with ` access_form = forms.AccessMetadataForm(data=request.POST, instance=project, editable=editable)`

django might be doing some kind of check that is preventing us from adding None to the `dua` because of the existing foreign key relation.


The hacky way to fix this (that i found so far) is that we could just delete the fields in the form instead of disabling them
If we remove the input from the form entirely, Django removed the `dua` in the database as well.

```
  if self.access_policy not in {AccessPolicy.CREDENTIALED, AccessPolicy.CONTRIBUTOR_REVIEW}:
           # -self.fields['required_trainings'].disabled = True
           # - self.fields['required_trainings'].required = False
           # - self.fields['required_trainings'].widget = forms.HiddenInput()
           del self.fields['required_trainings']

        if self.access_policy == AccessPolicy.OPEN:
            # -self.fields['dua'].disabled = True
            # -self.fields['dua'].required = False
            # -self.fields['dua'].queryset = DUA.objects.none()
[...]

PR #2073, opened 2023-09-06
Allow a project's access policy to be changed from credentialed/restricted to open

briangow:
As discussed in #1758 attempting to switch a project's access policy from credentialed/restricted to open results in an "Invalid submission. See errors below." message. This underlying error is:
```
self._errors
{'dua': ['This field is required.'], 'required_trainings': ['This field is required.']}
```
These fields shouldn't be required for open projects though. Here I implement the change proposed by @amitupreti in #1758. This does a `del` on the fields which aren't required. 

I've also added a line to `check_integrity` that provides an error message when a credentialed or contributor review project doesn't have `self.required_trainings` set. It is necessary to have a training set for these project types.

### DIFF
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -869,11 +869,13 @@ class AccessMetadataForm(forms.ModelForm):
             self.fields['required_trainings'].disabled = True
             self.fields['required_trainings'].required = False
             self.fields['required_trainings'].widget = forms.HiddenInput()
+            self.initial['required_trainings'] = ''
 
         if self.access_policy == AccessPolicy.OPEN:
             self.fields['dua'].disabled = True
             self.fields['dua'].required = False
             self.fields['dua'].widget = forms.HiddenInput()
+            self.initial['dua'] = ''
 
         if not self.editable:
             for field in self.fields.values():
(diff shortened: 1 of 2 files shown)


bemoody:
1. If you set the access type to "Credentialed" and then change it to "Restricted", the project still has `required_trainings`.

2. If you set the access type to "Restricted" and then change it to "Open", the
 project still has a `dua`.

So unfortunately I don't think this is going to work.  We really need to ensure that dua is set to None and required_trainings is cleared if access type is Open.

bemoody:
try instead
```
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -869,11 +869,13 @@ def __init__(self, *args, **kwargs):
             self.fields['required_trainings'].disabled = True
             self.fields['required_trainings'].required = False
             self.fields['required_trainings'].widget = forms.HiddenInput()
+            self.initial['required_trainings'] = ''
 
         if self.access_policy == AccessPolicy.OPEN:
             self.fields['dua'].disabled = True
             self.fields['dua'].required = False
             self.fields['dua'].widget = forms.HiddenInput()
+            self.initial['dua'] = ''
 
         if not self.editable:
             for field in self.fields.values():
```

briangow:
Thanks @bemoody, I've implemented the change you suggested. It seems to be working in my testing.

tompollard:
Thanks @bemoody and @briangow, this looks good to me.

tompollard:
@briangow now on the production server!

Issue #2072, opened 2023-09-06
Can't change a project title for new versions


briangow:
While this may be by design, one cannot change a project's title. 

When attempting to do this while the project is in the active project phase, these lines of code prevent changing the title:

https://github.com/MIT-LCP/physionet-build/blob/5cef547f8a3e9b23be8ff89ae908a1d915f53a7d/physionet-django/project/modelcomponents/activeproject.py#L570-L581

`self.title` can be changed (in the database) leading up to this last line but `published_project.title` gets set to `title` before publication, which is the original / first title. 

When changing the title in the database after the project has been published, one is no longer able to download the .zip file since it is named based on the title at the time of publication.

tompollard:
I think it is useful for us to support changes in titles, because there are likely to be valid use cases (e.g. fixing a typo or tweaking the title in response to a change in data content between versions). I haven't checked, but I'm pretty sure there are already a few projects where the title differs between versions. 

If you would like to fix the title in a newer version of an active (i.e. pre-publication) project, one (hacky) way would be to:

1. temporarily update the title of `previous_published_projects.first().title` (side note, but it would make more sense for this to take the last project, not the first).
2. publish the new version of the project (which copies the title in 1).
3. switch `previous_published_projects.first().title` back to its original title.

I think this is a fairly safe approach, that should not break anything (other than temporarily changing the title of the original version of the project).

The main issue with title changes is that (as a result of a poor decision that we made in our early days!) the [title string is used for naming some objects](https://github.com/MIT-LCP/physionet-build/issues/1674), such as zipfiles and cloud buckets. As long as the title is changed before the project is published, this issue should not be a concern.

briangow:
Thanks @tompollard!

Here are some details on how to do what you proposed:

```
p = ActiveProject.objects.filter(slug=<slug>)
f = p.core_project.publishedprojects.first()
f.title = <new-title>
f.save()
```
After the project is published, reverse the change to `first().title`
```
p = PublishedProject.objects.filter(title__contains='<title-overlap>')[0]
f = p.core_project.publishedprojects.first()
f.title = <original-title>
f.save()
```

tompollard:
Thanks @briangow, we should also take the opportunity to switch `previous_published_projects.first().title` to something like `previous_published_projects.last().title`, otherwise I think we'll find that subsequent versions revert back to the old title.

bemoody:
[...]

PR #2074, opened 2023-09-06
Use title being shown in project preview for new publications

briangow:
As discussed in #2072 , the project title being shown in the project preview isn't used in the publication step for a new version. As @bemoody mentioned we should always use the version being displayed by the project preview (i.e. `self.title`). This updates the code to always use `self.title` for the publication of a new project version and removes the request to get the original title version since it is not needed in `publish`.


bemoody:
Thanks!  Can we get rid of `title` in line 540?  Looks like it is not used anywhere?

briangow:
Thanks @bemoody , I've removed the unused `title` parameter.

PR #2077, opened 2023-09-12
Mark move_files_as_readonly as an associated task

bemoody:
When a project is published, we create a SHA256SUMS.txt file which should contain hashes of all the project files (so people who download the files can check that they're all present and haven't been corrupted, etc.)

This can of course take a very long time, so we do it asynchronously.  Really it *ought* to be done long before publication (before or during the author approval stage) but it isn't.

While this task is still pending, admins should *not* be able to trigger other processes to run on the files.  In particular, admins should *not* be able to trigger zip file creation (via the "Make zip" button in the console) nor cloud synchronization (via the "Send files to GCP button") until the server has finished generating the SHA256SUMS.txt file.

For example, if you do the following:
- launch the demo server using `manage.py runserver` (but don't run `process_tasks`)
- log in to the demo server as "rgmark"
- submit "MIT-BIH Arrhythmia Database" for publication
- accept, approve, and publish the project
- go to the published project management page

Because the background-tasks daemon isn't running, you should see the message "Pending task: Read Only Files - MIT-BIH Arrhythmia Database v1.0.0", and the "Make checksum", "Make zip", and "Deprecate files" (and, if applicable, "Send files to GCP") buttons should all be grayed-out.

Then launch the background_tasks daemon (`manage.py process_tasks`), and the task should run (which takes less than a second).  Refresh the page in your browser and the buttons should no longer be grayed-out.

### DIFF
--- a/physionet-django/notification/utility.py
+++ b/physionet-django/notification/utility.py
@@ -12,8 +12,8 @@ from django.template import defaultfilters, loader
 from django.utils import timezone
 from django.urls import reverse
 
-from project.models import DataAccessRequest, PublishedProject, AccessPolicy
-from user.models import CredentialApplication
+import project.models
+import user.models
 
 RESPONSE_ACTIONS = {0:'rejected', 1:'accepted'}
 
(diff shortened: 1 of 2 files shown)


tompollard:
Thanks @bemoody, looks good to me. I'm going to come and ask you the reason for the switch in style on the imports!

Issue #2026, opened 2023-06-21
Archive old events in console once they have ended


mscanlan-git:
I think it would be a good idea if we archived old events in the console after they have reached their end date. This isn't a problem now, but eventually we will have many events that will clutter the 'events list' page. Having a separate page for 'archived events' may be a good idea.

tompollard:
I agree, good idea.

PR #2078, opened 2023-09-13
Added 'Archive Events' Page to filter events in the admin console

mscanlan-git:
This PR addresses #2026 by adding an "Archived Events" page to eliminate event clutter that will eventually occur over time as we add new events. 

EDIT: We will eventually need to change the normal events page to only include events that haven't been archived yet (active/future events). This can be done in a separate PR, though.

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -2952,6 +2952,20 @@ def event(request):
                    })
 
 
+@permission_required('user.view_archive_event', raise_exception=True)
+def archive_event(request):
+    """
+    List of archived events
+    """
+    archive_event = Event.objects.filter(all).order_by('archive_datetime')
+    archive_event = paginate(request, archive_event, 50)
+
+    return render(request, 'console/event_archive_event.html',
+                  {'event_archive_event': archive_event,
+                   'events_nav': True
+                   })
+
+
 @permission_required('user.view_all_events', raise_exception=True)
 def event_management(request, event_slug):
     """
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -2952,17 +2952,18 @@ def event(request):
                    })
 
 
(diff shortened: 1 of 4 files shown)


physionet-django/console/views.py
```diff
@@ -2952,6 +2952,21 @@ def event(request):
                    })
 
 
+@permission_required('user.view_all_events', raise_exception=True)
+def archive_event(request):
+    """
+    List of archived events
+    """
+    now = timezone.now()
```
tompollard, line 2960: Not sure that it is necessary to assign the `now` variable here. `Event.objects.filter(end_date__lte=timezone.now())` is just as readable.

tompollard (approved):
This is great @mscanlan-git, nice work! I have listed a few quick thoughts below. As this is a private part of the site I'd be happy to merge as it is, and perhaps these can be addressed in a couple of new pull requests?

- For the URL, instead of http://localhost:8000/console/archive_event/, we could consider switching to something like http://localhost:8000/console/event/archived/. This is consistent with the approach for the user pages in the console (e.g. http://localhost:8000/console/users/inactive/).
- For the menu, instead of "Event > Archived Events", we could do: "Event > Archived Events". We know that we are in the Events submenu, say no need to say this again. We might also want to change "Events > Event list" to something like "Events > Active" for consistency?

tompollard:
Now on the live site, thanks again! BTW, minor thing, but don't forget to clean up commit logs before opening a pull request (e.g in this case rolling back the `temp` commit with `git reset XXX` and then adding a fresh one).

mscanlan-git:
Thanks @tompollard , I will be looking to add those changes you mentioned in my next PR to address the active events page, to only allow currently actively events. 

Also, will keep note on cleaning commits, forgot about the temp commit from yesterday!

PR #2080, opened 2023-09-14
Added 'Active' Events to the Events submenu

mscanlan-git:
This PR piggybacks off of #2078 where we added an archive submenu for events in the admin console. This pull request continues that work by converting the 'events' tab to an 'active' tab which lists all events that are currently active or have future end dates beyond the current date.

Some other smaller additions:

- Removed now = timezone.now() and baked it into the .filter()
- Made function names more consistent across the board (event_active, event_archive)
- Adjusted URL titles to be event/archive or event/active
- Updated navbar name to be just 'active' or 'archived'


tompollard (approved):
you are good at this @mscanlan-git

tompollard:
Probably nothing to worry about, but I think that an event happening "now" will appear both in the active and archived pages.

mscanlan-git:
Thanks, Tom!



>  think that an event happening "now" will appear both in the active and archived pages.

This is something I considered as well. I suppose doing `event_archive = Event.objects.filter(end_date__lt=timezone.now())` for archived events would make more sense, this way any event older than the current date would be filtered out and events that occur on the current date or in the future can be left as event_active. Something to possibly consider in the future, but shouldn't be too problematic unless we had multiple events with the same end date.

PR #2083, opened 2023-09-14
Set new Google Cloud buckets as requester-pays

### LINKED ISSUE #2079: Setting requester-pays for google cloud buckets
When we create public Google Cloud buckets, we probably want them to be "requester-pays" unless we have a specific arrangement for somebody to cover download costs.

Is this currently being done by default (i.e. when the bucket is created through the PhysioNet admin console)?  Can we make this an explicit option in the console?

bemoody:
GCS buckets may be either "requester-pays" (the client accessing the data must specify a "project" that will be billed for egress costs), or "non-requester-pays" (any authorized client can access the data and the bucket's owner will be billed for egress costs.)

This option can be switched on or off at any time, and it would be nice to have a way to do so in the PhysioNet console.  For the time being, we want to set all newly created buckets as "requester-pays" by default.

This pull also cleans up the logic to create the bucket in a single API request rather than two.

**I don't anticipate any problems with this, but it hasn't been tested.  I'd suggest we push this to the live server and test with one or two small projects - check that it works and the resulting bucket settings are correct.  Please don't publish any big projects to GCP until we've tested this.**  We can wait to merge this if that's a problem.

Fixes #2079

### DIFF
--- a/physionet-django/console/utility.py
+++ b/physionet-django/console/utility.py
@@ -55,7 +55,8 @@ def create_bucket(project, version, title, protected=True):
     bucket_name, email = bucket_info(project, version)
     storage_client.create_bucket(bucket_name)
     bucket = storage_client.bucket(bucket_name)
-    bucket.iam_configuration.bucket_policy_only_enabled = True
+    # Only bucket-level permissions are enforced; there are no per-file ACLs.
+    bucket.iam_configuration.uniform_bucket_level_access_enabled = True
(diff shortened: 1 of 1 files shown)


tompollard:
Not tested, but looks good to me. Let's merge and see what happens!

tompollard:
@bemoody Now live, would you like to test on a couple of projects?

PR #2084, opened 2023-09-21
Quota manager for GCS

bemoody:
Today's yak-shaving has brought me to the fact that the GCS storage backend never properly implemented the `SubmissionInfo.quota_manager` method.

(Want to know what `quota_manager` is about?  See https://github.com/MIT-LCP/physionet-build/pull/1003#issuecomment-624179553 and issue #518)

This pull request rearranges logic a bit but shouldn't have any functional changes - storage usage is still measured by calling `GCSObject.size`.

This is necessary for exciting and wonderful things in the future.


bemoody:
Late last night I realized that despite how little functional code there is here, I still managed to make a small mistake.  That's why we need test cases ;)

tompollard:
Thanks Benjamin, looking forward to seeing where this is going!

PR #2086, opened 2023-09-22
PhysioNet-AWS Integration - Managing AWS S3 buckets and objects through PhysioNet

Chrystinne:
This PR introduces a set of functions for managing AWS S3 buckets and objects through PhysioNet. 

These functions include creating S3 client objects, determining bucket names, checking bucket existence, uploading files to buckets, configuring bucket policies based on the project's access policy, allowing the download of files using AWS command line tools, and more.

Each function is thoroughly documented with explanations of its purpose and any relevant notes. These additions enhance the project's ability to interact with AWS S3 for storage and access control.


tompollard:
Thanks @Chrystinne! Please hold off updating the requirements.txt for now because I'm about to bump a few packages, which will cause conflicts anyway.

Chrystinne:
@tompollard ok. Since you're updating it, could you also include 'boto3==1.28.33'?

tompollard:
@Chrystinne I added the boto dependency in: https://github.com/MIT-LCP/physionet-build/pull/2089

Please could you run the following steps to update this pull request to include the latest changes on dev?

1. Pull down the latest version of the dev branch (which now has boto in the requirements):
    `git switch dev && git pull`
3. Switch to your feature branch:
    `git switch physionet_aws_integration`
5. Remove your changes to requirements.txt by hard resetting to the penultimate commit: 
    `git reset 0ea7a1f3e4031766836ee0731ebab90c78792b2b --hard`
6. Rebase on the dev branch:
    `git rebase dev`
8. Push, overwriting the previous changes:
    `git push --force`

Once this pull request is rebased, I assume https://github.com/MIT-LCP/physionet-build/issues/2087 can be closed.

Chrystinne:
@tompollard Thank you, Tom! I will follow those steps!

tompollard:
Great, thanks @Chrystinne, I think this issue is resolved then. New error at https://github.com/MIT-LCP/physionet-build/pull/2086, but unrelated!

tompollard:
Whoops, sorry, ignore the message above. I'd thought I was posted at: https://github.com/MIT-LCP/physionet-build/issues/2087!

tompollard:
@Chrystinne Moving our Slack discussion to this pull request, so we can gather ideas from everyone:

> How do we manage secret keys on our servers? The tests on my PR are now failing because they can't find the AWS credentials. The AWS_PROFILE (that contains AWS credentials configured on my machine) is specified locally on my .env file. Do we need to add the files to the server? Or does our gitHub repo use some sort of program to receive secret keys/credentials?

The tests run independently of our servers, so adding configuration information to them won’t help. GitHub [does support secret variables](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions) (basically you add a variable for the key to the tests, and then you specify the private value of the variable in GitHub), but I don’t think this is a good solution because I’m pretty sure person/s among us (@bemoody 👀) won’t trust GitHub with private information!

I haven't looked over this pull request yet (or looked at the failing tests), but I think the solution is to find an approach that allows the code to run without the need to share keys. Some ideas to consider:

1) Only call your new functions if the AWS configuration is switched on (e.g. if `AWS_PROFILE` is `NULL`, then do not call the AWS functions). This is probably a requirement anyway because Health Data Nexus (and other platforms) may choose not to integrate with AWS. 
2) Add unit tests to the testing framework that check the behaviour of your new functions. You can use the `mocker` package (or whatever it is called) to simulate the API calls. We use this approach [elsewhere for testing](https://github.com/MIT-LCP/physionet-build/blob/c693f9bdee57914b4b10e53755f946858d6e5c6f/physionet-django/console/test_views.py#L416-L429), so you should be able to find examples.
3) Add an AWS configuration step [to the GitHub workflow](https://github.com/MIT-LCP/physionet-build/blob/dev/.github/workflows/physionet-build-test.yml) that sets up the AWS environment on the testing server, if possible. The nice thing about this would be that you could actually test the AWS integration, rather than just pretending to test it (like in 2).

1 will probably fix this pull request. 2 is important, so should be added shortly after we merge this pull request (or someone might argue before!). 3 would be great, but I'm unclear whether it is possible to do this without costing us money.

bemoody:
## Code style

Please don't add more functions to `console/utility.py`.

Instead, create a new file called (for example) `project/cloud/s3.py`, and put your new functions there.

## Storing bucket metadata

For GCP, we have the class `GCP` in `project/modelcomponents/storage.py`.  We'll need to define a similar class that will store similar information for S3 project mirrors.

(In particular, name of the bucket and the date/time it was created/updated.)

However, I think it would be better to define a single `CloudMirror` class that could be used for both GCP and S3, and eventually get rid of the `GCP` class.

bemoody:
For reference, how boto3 obtains credentials:
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html

bemoody:
https://pypi.org/project/moto/ sounds like it might be easy to use and sufficient for unit testing.

Otherwise, there appear to be quite a number of "S3-compatible" servers that could be used for testing, some of which are free software.

physionet-django/console/utility.py
```diff
@@ -13,8 +13,9 @@
 from django.contrib.sites.models import Site
 from django.conf import settings
 from django.urls import reverse
-
-from project.validators import validate_doi
```
tompollard, line 17: tests are failing because`from project.validators import validate_doi` was removed. it needs to be added back.

physionet-django/project/views.py
```diff
@@ -1909,6 +1918,8 @@ def sign_dua(request, project_slug, version):
     Page to sign the dua for a protected project.
     Both restricted and credentialed policies.
     """
+    from console import utility
```
tompollard: these imports belong at the top of the module.

physionet-django/console/views.py
```diff
@@ -818,17 +872,25 @@ def manage_published_project(request, project_slug, version):
     - Deprecate files
     - Create GCP bucket and send files
     """
+    from console.utility import get_bucket_name_and_prefix, check_s3_bucket_with_prefix_exists, has_aws_credentials, missing_S3_open_data_info
```
tompollard: imports should be at the top of the module.

physionet-django/console/views.py
```diff
@@ -763,6 +763,60 @@ def send_files_to_gcp(pid):
...
+        pid (int): The unique identifier (ID) of the project for which to update the bucket policy.
+
+    Returns:
+        bool: True if the bucket policy was updated successfully, False otherwise.
+
+    Note:
+    - Verify that AWS credentials and configurations are correctly set up for the S3 client.
+    - The 'updated_policy' variable indicates whether the policy was updated successfully.
+    """
+    from console import utility
```
tompollard: move import to the top of the module.

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+
+    project_name = project.slug + "-" + project.version 
+    aws_ids = get_aws_accounts_for_dataset(project_name)
+    if project.access_policy == AccessPolicy.OPEN:
+        s3 = create_s3_access_object()
+        if s3 is None:
+            return
+        put_public_access_block(s3, bucket_name, False)
+        bucket_policy = create_bucket_policy(bucket_name, aws_ids, True)
+    elif project.access_policy == AccessPolicy.CREDENTIALED or project.access_policy == AccessPolicy.RESTRICTED or project.access_policy == AccessPolicy.CONTRIBUTOR_REVIEW:
```
tompollard: you'll need to run style checks over the code. this line looks way to long for example (max width is 119 characters).

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+        bucket_name (str): The name of the AWS S3 bucket associated with the project.
+
+    Returns:
+        None
+
+    Note:
+    - Ensure that AWS credentials (Access Key and Secret Key) are properly configured for
+      the S3 client used in this function.
+    """
+    from project.models import AccessPolicy
```
tompollard: move import to top of module.

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+    client.put_public_access_block(
+        Bucket=bucket_name,
+        PublicAccessBlockConfiguration={
+            'BlockPublicAcls': configuration,
+            'IgnorePublicAcls': configuration,
+            'BlockPublicPolicy': configuration,
+            'RestrictPublicBuckets': configuration
+        }
+    )
+
```
tompollard: I might be wrong, but i'm pretty sure pep8 asks for two empty lines between functions: https://peps.python.org/pep-0008/#blank-lines

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+        dataset_name (str): The name of the dataset for which to retrieve AWS account IDs.
+
+    Returns:
+        list: A list of AWS account IDs associated with authorized users of the dataset.
+
+    Note:
+    - This function assumes that AWS account IDs are 12-digit numerical values.
+    - Users with the appropriate permissions and AWS account IDs are included in the result list.
+    """
+    from project.models import PublishedProject
```
tompollard: move imports to the top of the module.

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+
+    Returns:
+        bool: True if the S3 bucket (and optional prefix) exists and is accessible, False otherwise.
+
+    Note:
+    - Make sure that AWS credentials (Access Key and Secret Key) are properly configured
+      for this function to work.
+    - The S3 bucket should exist and be accessible based on the project's configuration.
+    """
+    from project.models import AccessPolicy
```
tompollard: move import to top of module.

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+    This function takes a list of common prefixes (often obtained from S3 bucket listings)
+    and extracts the 'Prefix' values from each element, returning them as a list.
+
+    Args:
+        common_prefixes (list): A list of common prefixes, typically from an S3 bucket listing.
+
+    Returns:
+        list: A list of extracted prefixes.
+    """
+    perfixes = [prefix.get('Prefix') for prefix in common_prefixes]
```
tompollard: `perfixes` should be `prefixes`

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+    - This function does not create or verify the existence of the actual S3 bucket; it only provides
+      the calculated bucket name based on project attributes.
+    """
+    from project.models import AccessPolicy
+
+    bucket_name = None
+
+    if project.access_policy == AccessPolicy.OPEN and has_S3_open_data_bucket_name():
+        bucket_name = settings.OPEN_ACCESS_DATA_BUCKET_NAME
+    elif project.access_policy == AccessPolicy.RESTRICTED or project.access_policy == AccessPolicy.CREDENTIALED or project.access_policy == AccessPolicy.CONTRIBUTOR_REVIEW:
```
tompollard: this line looks over 119 characters, the max width allowed according to our style checks

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+        project (project.models.Project): The project for which to determine the S3 bucket name.
+
+    Returns:
+        str: The S3 bucket name associated with the project.
+
+    Note:
+    - This function does not create or verify the existence of the actual S3 bucket; it only provides
+      the calculated bucket name based on project attributes.
+    """
+    from project.models import AccessPolicy
```
tompollard: move the import to the top of module.

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+    This function evaluates whether a project lacks the necessary S3 open data information,
+    based on its access policy and the availability of an S3 open data bucket name.
+
+    Args:
+        project (project.models.Project): The project to check for S3 open data information.
+
+    Returns:
+        bool: True if S3 open data information is missing, False otherwise.
+    """
+    from project.models import AccessPolicy
```
tompollard: move import to the top of the module.

physionet-django/console/utility.py
```diff
@@ -13,8 +13,9 @@
 from django.contrib.sites.models import Site
 from django.conf import settings
 from django.urls import reverse
-
-from project.validators import validate_doi
+import boto3
+import os
```
tompollard: Note that PEP8 suggests the following ordering for imports:

- Standard library imports (e.g. os, re, logging)
- Related third party imports. (e.g. boto3)
- Local application/library specific imports (e.g. from project.validators import validate_doi)

See: https://peps.python.org/pep-0008/#imports

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+
+    Returns:
+        str: A JSON-formatted string representing the initial bucket policy.
+
+    Note:
+    - This initial policy serves as a baseline and can be customized further to meet specific access
+      requirements for the bucket.
+    """
+    user = None
+    principal_value = '*' if public else {'AWS': [f'arn:aws:iam::{aws_id}:root' if user is None or user == '' else f'arn:aws:iam::{aws_id}:user/{user}' for aws_id in aws_ids]}
```
tompollard: the width of this line looks over the maximum of 119 characters.

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+        None
+
+    Note:
+    - Ensure that AWS credentials (Access Key and Secret Key) are properly configured for
+      the S3 client used in this function.
+    - The 'project_name' variable is created by concatenating the project's slug and version.
+    - The 's3_prefix' is set only for projects with an 'OPEN' access policy, providing
+      an optional prefix within the S3 bucket.
+    """
+    from project.models import AccessPolicy
```
tompollard: move import to top of module.

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+        None
+
+    Returns:
+        None
+
+    Note:
+    - Ensure that AWS credentials (Access Key and Secret Key) are properly configured for
+      the S3 client used in the 'upload_project_to_S3' function.
+    """
+    from project.models import PublishedProject
```
tompollard: move import to top of module.

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+
+    Returns:
+        None
+
+    Note:
+    - The function ensures that AWS S3 buckets remain intact but contain no files after execution.
+    - Ensure that AWS credentials (Access Key and Secret Key) with sufficient permissions
+      are properly configured for the S3 client used in this function.
+    """
+    from project.models import PublishedProject
```
tompollard: move import to top of module.

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+
+    Returns:
+        None
+
+    Note:
+    - The function leverages the 'delete_project_from_S3' method to perform project deletions.
+    - Ensure that AWS credentials (Access Key and Secret Key) with sufficient permissions
+      are properly configured for the S3 client used in this function.
+    """
+    from project.models import PublishedProject, AccessPolicy
```
tompollard: move imports to top of module.

physionet-django/console/utility.py
```diff
@@ -31,6 +32,828 @@ class DOIExistsError(Exception):
...
+    not_open_published_projects = PublishedProject.objects.filter(access_policy=AccessPolicy.RESTRICTED) | PublishedProject.objects.filter(access_policy=AccessPolicy.CREDENTIALED) | PublishedProject.objects.filter(access_policy=AccessPolicy.CONTRIBUTOR_REVIEW)
```
tompollard: this line is way over the maximum width of 119 characters. apply a formatter to reformat the code (be careful to apply the restyling to only your changes, not to existing code).

tompollard (commented):
@Chrystinne I took a quick look and added some suggestions. Mostly small things (e.g. imports go at the top!).

Chrystinne:
@bemoody I'm adding the functions to a new file (project/cloud/s3.py)

physionet-django/console/utility.py
Chrystinne: @tompollard I have added the import (from project.validators import validate_doi) and I'm still getting this error

physionet-django/console/views.py
Chrystinne: @tompollard I didn't add the import at the top of the module in some cases because I was encountering import errors like this:

ImportError: cannot import name 'get_file_forms' from partially initialized module 'project.views' (most likely due to a circular import)

Chrystinne: @tompollard Adding the AWS functions to a new file project/cloud/s3.py) solved most of the import errors, except for the one mentioned above.

bemoody:
In `physionet/settings/base.py`, we need to set and export AWS_SHARED_CREDENTIALS_FILE, like so:
```
if config('AWS_SHARED_CREDENTIALS_FILE', default=None):
    AWS_SHARED_CREDENTIALS_FILE = os.path.join(
        BASE_DIR,
        config('AWS_SHARED_CREDENTIALS_FILE'))
    os.environ['AWS_SHARED_CREDENTIALS_FILE'] = AWS_SHARED_CREDENTIALS_FILE
else:
    AWS_SHARED_CREDENTIALS_FILE = None
```

Add this to the `.env.example` accordingly:
```
# Path to the file containing credentials for AWS
# (https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#shared-credentials-file)
#AWS_SHARED_CREDENTIALS_FILE=aws_credentials.ini
```
(see issue #2095)

We still need to decide how buckets will be structured (see issues #2093, #2094).  But `OPEN_ACCESS_DATA_BUCKET_NAME` is not a good name for a global variable.  Please rename it to `S3_OPEN_ACCESS_BUCKET` or `AWS_S3_OPEN_ACCESS_BUCKET` or something like that.

Chrystinne:
@bemoody Thanks, Benjamin! I will see the issues.

bemoody:
We don't need both `has_aws_credentials()`/`has_s3_credentials` and `missing_s3_open_data_info()`/`missing_S3_open_data`.  Please just define one function that returns false if the required settings are missing.  I would write this:

```
#### s3.py

def has_s3_credentials():
    return all([
        settings.AWS_PROFILE,
        settings.S3_OPEN_ACCESS_BUCKET,
    ])

#### views.py

def manage_published_project(...):
    ...
    return render(
        ...,
        {
            ...,
            'has_s3_credentials': has_s3_credentials(),
            ...
        },
    )

#### manage_published_project.html

        {% if not has_s3_credentials %}
          <p>
            Support for Amazon S3 is not enabled.  (Check that
            <code>AWS_PROFILE</code> and <code>S3_OPEN_ACCESS_BUCKET</code>
            are set.)
          </p>
        {% elif not aws_bucket_exists %}
          ...
        {% endif %}
```

Chrystinne:
@bemoody The reason for analyzing both separately was that we could still upload projects with an access policy as credentialed, restricted, or contributor_review if the bucket name was not specified as S3_OPEN_ACCESS_BUCKET but the AWS credentials were provided.

I have made the changes you suggested though. Thanks!

physionet-django/console/views.py
```diff
@@ -985,7 +1074,7 @@ def gcp_bucket_management(request, project, user):
             LOGGER.info("The bucket {0} already exists, skipping bucket and \
                 group creation".format(bucket_name))
         else:
-            utility.create_bucket(project.slug, project.version, project.title, is_private)
+            create_s3_bucket(project.slug, project.version, project.title, is_private)
```
tompollard: this doesn't look right! an aws function is being called inside `gcp_bucket_management`

physionet-django/project/views.py
```diff
@@ -1909,6 +1918,8 @@ def sign_dua(request, project_slug, version):
     Page to sign the dua for a protected project.
     Both restricted and credentialed policies.
     """
+    from project.cloud.s3 import has_aws_credentials
```
tompollard: imports belong at the top of the module

physionet-django/project/views.py
```diff
@@ -1772,6 +1772,9 @@ def published_project(request, project_slug, version, subdir=''):
     """
     Displays a published project
     """
+    from project.cloud.s3 import get_bucket_name_and_prefix, check_s3_bucket_with_prefix_exists, has_aws_credentials
```
tompollard: imports don't belong here.

physionet-django/console/views.py
Chrystinne: @tompollard you are right! I mistakenly renamed 'utility.create_bucket' to 'create_s3_bucket'. Thanks for catching that!

physionet-django/project/views.py
Chrystinne: @tompollard I've moved import to the top of the module.

physionet-django/project/views.py
Chrystinne: @tompollard I've moved import to the top of the module.

bemoody:
In general, I think we want to reuse the same "session" and "client" object when we can.  For example, in `send_files_to_s3`, you are calling `create_s3_access_object` once for every file - we don't want that.

The name `create_s3_access_object` is a little unclear and I would rename it to `create_s3_client`.

I also think it would be better for this function to raise an exception, rather than returning None, if AWS_PROFILE is undefined.  (Note that `boto3.Session` will raise an exception if the specified profile isn't found in the credentials file.)

Chrystinne:
@bemoody I've made those changes related to the create_s3_client/send_files_to_s3 methods.

bemoody:
From what I understand of S3, I think that we want *all files to be stored in a prefix*, regardless of whether the bucket holds one project or many projects.

Doing this consistently should make the logic here much simpler (for us and for future clients that will access the data) and provide flexibility in the future.

I think it would be wise for the prefix to be `<slug>/<version>/`.  For example, an object path might be `mitdb/1.0.0/100.hea` or `mimiciv/2.2/hosp/admissions.csv.gz`.  From the S3 perspective there may be little difference between this and `<slug>-<version>/`, but it is useful to be compatible with existing PhysioNet URLs and may make other things simpler too (e.g. per-project storage metrics.)

bemoody:
(I realize you could not then use a single `s3.list_objects_v2` to list all projects and versions, but that's not something we want to be doing in most cases, and the code here already won't work for more than 1000 projects AFAIK.)

bemoody:
@Chrystinne, have you uploaded the demo projects to a public S3 bucket on the open data account?  If so can you share the location?

It's okay if the files are not in their final state, but it'd be nice to test how bucket access works from client's perspective.

bemoody:
What is `arn:aws:iam::724665945834`?

Is this the ID of the PhysioNet account?  (Is it *required* to include the bucket's owner in the bucket policy?)  Or is this another account ID that you are using for testing?

bemoody:
`create_s3_client`
`create_s3_resource`

* If the S3 credentials are undefined, these functions should raise an exception (do not return None.)

`get_bucket_name`

* Do not try to list all possible access policies.  Just have an "else" branch for all policies other than OPEN.

`get_bucket_name_and_prefix`

* If this function is not needed, please remove it.
* We *want* a function to compute the prefix for a project, but I don't think having a single function that returns "bucket name and prefix as a single string" is a great idea.

`get_all_prefixes`
`get_prefix_open_project`
`find_matching_prefix`
`get_list_prefixes`
`check_s3_bucket_with_prefix_exists`

* If these functions are not needed, please remove them.

`get_initial_bucket_policy`
`create_bucket_policy`

* `get_initial_bucket_policy` is just a more limited version of `create_bucket_policy`; no reason I can see to have two functions.
* `create_bucket_policy` should work if `aws_ids=[]` and `public=False` (should forbid access to everyone.)  If this doesn't work, please fix it.

`empty_project_from_S3`

* This function is dangerous (it takes a single project as argument but deletes everything from the entire bucket.)  Please remove it, or rewrite it to only operate on the specified project.

`empty_list_of_projects`

* This function is dangerous for the same reason as `empty_project_from_S3`.  Please remove it.

`delete_project_from_S3`
`delete_list_of_projects_from_S3`
`delete_all_projects_from_S3

* If these functions are not needed, please remove them.

Chrystinne:
@bemoody arn:aws:iam::724665945834 is the ID for the PhysioNet open data account. Also yes, it is required to include the bucket's owner in the bucket policy.

bemoody:
get_initial_bucket_policy says:

* Deny `s3:Get*` and `s3:List*` for everyone who is not `arn:aws:iam::724665945834:*`.

create_bucket_policy says:

* Allow `s3:Get*` and `s3:List*` for people in this list.
* Deny `s3:Delete*` for everyone who is not `arn:aws:iam::724665945834:*`.

get_initial_bucket_policy does not say anything about `s3:Delete*`, and neither of these policies grants *permission* for the 724665945834 account to do anything.  So I am very skeptical that these "deny" rules are doing anything.

The logic described here is very complicated:
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_evaluation-logic.html
but it seems to say that if there is no rule saying you *can* do something, then you can't.

bemoody:
I am rather concerned about what that `*` implies in `s3:Get*` and `s3:List*`.  A random authorized user should be allowed to retrieve objects and their metadata, and to list objects (beneath a certain prefix.)  But they should not be allowed to do *everything* starting with "Get" or "List".

Here, if I understand correctly, are the possible "Actions": https://docs.aws.amazon.com/AmazonS3/latest/userguide/list_amazons3.html

For example, random people should not be allowed to invoke `GetAccessPointPolicy`.

bemoody:
I was confused about how you were handling prefixes, but now I think I see why.

In `upload_project_to_S3`, you're doing the following:
```
    # upload files to bucket
    folder_path = project.project_file_root()

    # set the prefix only for the projects
    # in the open data bucket
    s3_prefix = ""
    if project.access_policy == AccessPolicy.OPEN:
        s3_prefix = f"{project.slug}/"
    send_files_to_s3(folder_path, s3_prefix, bucket_name)
```

The `project_file_root` method is poorly named - that method ought to be renamed or removed, and it shouldn't be used here.  What we want to do instead is:
```
    folder_path = project.file_root()
    s3_prefix = f"{project.slug}/{project.version}/"
    send_files_to_s3(folder_path, s3_prefix, bucket_name)
```

Each project version is managed independently; if we tell the system to upload version 2.0, we don't want to upload any other versions from the same project.

Chrystinne:
@bemoody The uploaded demo projects on the open data account are located at this location: s3://open-data-dev-server

bemoody:
In `sign_dua` (`project/views.py`), we do *not* want to invoke `update_aws_bucket_policy` if the project is not yet published to S3.

(We may still need to rework how we create and update bucket policies; there are a lot of tricky details, which I'm ignoring for now.)

The recommendation (which we unfortunately haven't followed in the past) is to define background tasks in a separate `tasks.py` file.  So `send_files_to_aws` and `update_aws_bucket_policy`) should ideally go in `project/tasks.py`, not in `console/views.py`.

Also, when you invoke `send_files_to_aws` or `update_aws_bucket_policy`, you should pass `verbose_name` and `creator` arguments (like how `gcp_bucket_management` invokes `send_files_to_gcp`).

Chrystinne:
@bemoody I've replaced the previous S3 actions with the following:

"Action": [
                "s3:GetObject",
                "s3:ListBucket"
            ]

bemoody:
In `upload_project_to_S3`, you always call `create_s3_bucket`.  There are at least three possibilities:

(a) The bucket doesn't exist, and we want to create it and then upload files.

(b) The bucket exists and it is owned by us, and we want to upload this project's files into it.

(c) The bucket exists and is **not** owned by us, and we want this function to crash and burn.

According to the boto3 documentation, (b) should result in an exception `s3.exceptions.BucketAlreadyOwnedByYou` and (c) should result in `s3.exceptions.BucketAlreadyExists`.

So I think we want this:
```
    try:
        create_s3_bucket(s3, bucket_name)
    except s3.exceptions.BucketAlreadyOwnedByYou:
        pass
```

bemoody:
Again in `upload_project_to_S3`, I think we should not call `update_bucket_policy` until *after* `send_files_to_s3` and `put_bucket_logging`.

Possibly we should call `put_bucket_logging` before `send_files_to_s3`.

bemoody:
We want to rename the zip file on upload, to use the slug instead of the original title (issue #2122).

So we should call `project.zip_name(full=True)` to get the local filename, but then the URL on S3 should be `f'{project.slug}/{project.slug}-{project.version}.zip`.

bemoody:
So here are the issues that I think still need fixing:

* Don't call update_aws_bucket_policy if there is no AWS bucket

* Handle uploading to a bucket that already exists (but be sure we're not uploading to a bucket owned by somebody else)

* Add `verbose_name` for background tasks

* Rename zip files to `{slug}-{version}.zip`

There are other minor improvements that would be nice to have, but these are the important ones.

Again, I'm setting aside issues around restricted access buckets, which I think still need work, but don't need to be addressed before uploading open access data.

Chrystinne:
@bemoody I worked on those issues in my last update:

- Don't call update_aws_bucket_policy if there is no AWS bucket;
- Handle uploading to a bucket that already exists (but be sure we're not uploading to a bucket owned by somebody else);
- Add verbose_name for background tasks;
- Rename zip files to {slug}-{version}.zip;

Now, if the bucket already exists and we are not the owner, we get the following error:

_"botocore.errorfactory.BucketAlreadyExists: An error occurred (BucketAlreadyExists) when calling the CreateBucket operation: The requested bucket name is not available. The bucket namespace is shared by all users of the system. Please select a different name and try again."_

.env.example
```diff
@@ -60,6 +60,17 @@ AWS_VALUE=secret
 AWS_VALUE2=secret
 AWS_CLOUD_FORMATION=url
 
+# AWS credentials (Access Key and Secret Key): Configure AWS credentials in the AWS CLI profile using the 'aws configure' command.
+AWS_PROFILE=
+BUCKET_OWNER_ID=
```
bemoody: Rename `BUCKET_OWNER_ID` to `S3_BUCKET_OWNER_ID` so it is clear we are talking about S3 and not some other kind of buckets.  Also please add a comment to say what should go here (the AWS account ID of the bucket owner.)

.env.example
```diff
@@ -60,6 +60,17 @@ AWS_VALUE=secret
...
+# Path to the file containing credentials for AWS
+# (https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#shared-credentials-file)
+AWS_SHARED_CREDENTIALS_FILE=
+# The default bucket name to store projects with an 'OPEN' access policy.
+S3_OPEN_ACCESS_BUCKET=
+# The default bucket name to store logs and metrics related to project usage.
+S3_SERVER_ACCESS_LOGS=
```
bemoody: Rename to `S3_SERVER_ACCESS_LOG_BUCKET`.

physionet-django/console/views.py
```diff
@@ -75,7 +76,15 @@
 from physionet.enums import LogCategory
 from console import forms, utility, services
 from console.forms import ProjectFilterForm, UserFilterForm
-
+from console import views
```
bemoody: This is not needed and doesn't belong here.

physionet-django/project/cloud/s3.py
```diff
@@ -0,0 +1,749 @@
...
+
+    Args:
+        client (boto3.client.S3): An initialized AWS S3
+        client object.
+        bucket_name (str): The name of the S3 bucket
+        to configure.
+        configuration (bool): The desired configuration for
+        public access. Set to 'True' to allow public access
+        or 'False' to restrict public access.
+
```
bemoody, line 556: I believe this is incorrect.  Please check.  Please rename the variable so that its meaning is clear:

If a `True` value means "forbid public access", then please rename `configuration` to `forbid_public_access`.

If a `True` value means "allow public access", then please rename `configuration` to `allow_public_access`.  (If that is the case, then the code in `update_bucket_policy` is wrong.)

physionet-django/console/views.py
```diff
@@ -764,6 +773,75 @@ def send_files_to_gcp(pid):
...
+    Returns:
+        None
+
+    Note:
+    - Verify that AWS credentials and configurations are correctly set
+    up for the S3 client.
+    """
+    project = PublishedProject.objects.get(id=pid)
+    # Create or get the associated AWS object
+    AWS.objects.get_or_create(project=project)
```
bemoody: The AWS object should either be created here, or in `aws_bucket_management`.  Not both.

physionet-django/console/views.py
```diff
@@ -1013,6 +1099,48 @@ def gcp_bucket_management(request, project, user):
...
+    """
+    is_private = True
+
+    if project.access_policy == AccessPolicy.OPEN:
+        is_private = False
+
+    bucket_name = get_bucket_name(project)
+
+    try:
+        AWS.objects.get(bucket_name=bucket_name)
```
bemoody: This is wrong.  Multiple projects will have the same bucket name.

physionet-django/console/views.py
```diff
@@ -1013,6 +1099,48 @@ def gcp_bucket_management(request, project, user):
...
+        messages.success(request, "The bucket already exists. Resending the \
+            files for the project {0}.".format(project))
+    except AWS.DoesNotExist:
+        if check_s3_bucket_exists(project):
+            LOGGER.info("The bucket {0} already exists, skipping bucket \
+                creation".format(bucket_name))
+        AWS.objects.create(project=project, bucket_name=bucket_name,
+                           is_private=is_private)
```
bemoody: You can remove the whole "try" block and just do:
```
if not AWS.objects.filter(project=project).exists():
    AWS.objects.create(project=project,
                       bucket_name=bucket_name,
                       is_private=is_private)
```
All of the logging and checking whether the bucket exists is not necessary here.

bemoody:
A couple of errors pop up when trying to test this using `moto`.

If the log bucket doesn't exist, then calling `upload_project_to_S3` gives:
```
botocore.exceptions.ClientError: An error occurred (InvalidTargetBucketForLogging) when calling the PutBucketLogging operation: The target bucket for logging does not exist.
```

If I create a brand-new bucket for logging, then `upload_project_to_S3` gives:
```
botocore.exceptions.ClientError: An error occurred (InvalidTargetBucketForLogging) when calling the PutBucketLogging operation: You must either provide the necessary permissions to the logging service using a bucket policy or give the log-delivery group WRITE and READ_ACP permissions to the target bucket
```

What should the bucket policy for the log bucket look like?  Or, what would you do to create a temporary log bucket for testing?

We might want to make S3_SERVER_ACCESS_LOG_BUCKET optional.

bemoody:
I don't think the `get_bucket_owner` works.  When I try it with an S3 bucket I created myself, it returns a 64-character hex string, not an account ID.  Have I missed something?

bemoody:
Also, when I said we want to be sure not to upload to a bucket owned by somebody else, my point was that *this should already be enforced, simply by attempting to create the bucket*.

From what I understand, if you call `s3.create_bucket` and the bucket already exists and is owned by somebody else, it'll raise `BucketAlreadyExists`.  That is what we want.  We want `upload_project_to_S3` to fail in that case.

If you call `s3.create_bucket` and the bucket already exists and is owned by you, it'll raise `BucketAlreadyOwnedByYou`.  We want to ignore that exception and proceed with uploading in that case.

So all I was trying to say was:
- Always call `s3.create_bucket` before uploading (don't skip that step just because you think the bucket already exists.)
- Catch the `BucketAlreadyOwnedByYou` exception and ignore it (but don't ignore any other types of exceptions.)
I don't think the logic needs to be any more complicated than that.

Chrystinne:
@bemoody The S3.Client.get_bucket_acl(**kwargs) method, which I invoke in the get_bucket_owner function, returns the 'Owner' object containing the 'DisplayName' and 'ID' (a 64-character hex string). 

In this context, the 'ID' refers to the Canonical User ID, a unique identifier associated with an AWS account. The Canonical User ID is a 64-character hexadecimal string linked to the AWS account's root user. It differs from the AWS account ID, a 12-digit number that uniquely identifies an AWS account. 

I opted for this approach as I couldn't find a method that directly returns the AWS account ID.

Chrystinne:
@bemoody The bucket policy for the log bucket should look like this (remember to change bucket_name in both 'Resource' and "XXXXXXXXXXXX" with the 12-digit account ID):

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "S3ServerAccessLogsPolicy",
            "Effect": "Allow",
            "Principal": {
                "Service": "logging.s3.amazonaws.com"
            },
            "Action": "s3:PutObject",
            "Resource": "arn:aws:s3:::bucket_name/*",
            "Condition": {
                "StringEquals": {
                    "aws:SourceAccount": "XXXXXXXXXXXX"
                }
            }
        },
        {
            "Sid": "S3PolicyStmt-DO-NOT-MODIFY-1697643261877",
            "Effect": "Allow",
            "Principal": {
                "Service": "logging.s3.amazonaws.com"
            },
            "Action": "s3:PutObject",
            "Resource": "arn:aws:s3:::bucket_name/*"
        }
    ]
}
```

So I would do something like this:

```
log_bucket_policy = create_log_bucket_policy(log_bucket_name)

if log_bucket_policy not in [None, ""]:
    set_bucket_policy(bucket_name, log_bucket_policy)
```

where **create_log_bucket_policy** would be implemented as follows:

```
def create_log_bucket_policy(bucket_name, account_id):
    log_bucket_policy = {
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "S3ServerAccessLogsPolicy",
            "Effect": "Allow",
            "Principal": {
                "Service": "logging.s3.amazonaws.com"
            },
            "Action": "s3:PutObject",
            "Resource": f"arn:aws:s3:::{bucket_name}/*",
            "Condition": {
                "StringEquals": {
                    "aws:SourceAccount": f"{account_id}"
                }
            }
        },
        {
            "Sid": "S3PolicyStmt-DO-NOT-MODIFY-1697643261877",
            "Effect": "Allow",
            "Principal": {
                "Service": "logging.s3.amazonaws.com"
            },
            "Action": "s3:PutObject",
            "Resource": f"arn:aws:s3:::{bucket_name}/*"
        }
    ]
    }

    # Convert the policy from JSON dict to string
    log_bucket_policy_str = json.dumps(log_bucket_policy)

    return log_bucket_policy_str
```

bemoody:
Thanks!  Looking at your `log_bucket_policy`, it looks like we only want the first statement ("S3ServerAccessLogsPolicy"), not the second ("S3PolicyStmt-DO-NOT-MODIFY-1697643261877").  We want to allow buckets owned by *our AWS account* to write logs into this bucket.

bemoody:
Following the most recent changes, the migration here will have to be updated.  You can rename the `0071_aws.py` to `0072_aws.py` and edit the dependency (`"0070_unpublishedproject_version_order_2"` to `"0071_alter_activeproject_options_and_more"`).  Or else rebase the branch and recreate the migration using `makemigrations`.

For the zip file, currently you're getting the file name/path as follows:
```
        zip_name = project.zip_name()
        zip_file_path = os.path.join(project.project_file_root(), zip_name)
```

But instead we want to rename the zip file on upload (see issue #2122):
```
        zip_name = project.zip_name(legacy=False)
        zip_file_path = project.zip_name(full=True)
```

PR #2090, opened 2023-09-23
Display DOI for current version and latest version.

tompollard:
When a new version of a project is published, it is assigned a version-specific DOI. We also assign a “canonical DOI”, which represents all versions of the project (and which always resolves to the latest version).

Currently, we only display the version-specific DOI on the project landing page. For example: https://physionet.org/content/icehg-ds/ only displays https://doi.org/10.13026/rfts-nv79.

The canonical DOI is sometimes what people want to know, so we should clearly display it too (for the project above, the canonical DOI is https://doi.org/10.13026/vrvd-mf28).

This pull request updates the Discovery box to display both DOIs, e.g.:

Screenshot 2023-09-23 at 12 20 25 AM

For the DataCite guidelines on versioning, see:
https://support.datacite.org/docs/versioning

Note: All projects should be assigned DOIs, so the if clauses in the template shouldn't be necessary. They are there for published projects because some legacy content may still be missing DOIs.

### DIFF
--- a/physionet-django/project/templates/project/project_preview.html
+++ b/physionet-django/project/templates/project/project_preview.html
@@ -84,7 +84,7 @@
     {% if project.version_clash %}
       <a style="color:red"><strong>{{ project.version}} * Clashing Version</strong></a>
     {% else %}
-      {{ project.version}}
+      {{ project.version }}
     {% endif %}
   {% else %}
     <a style="color:red"><strong>* Required field missing</strong></a>
@@ -177,7 +177,11 @@
       <div class="card my-4">
         <h5 class="card-header">Discovery</h5>
         <div class="card-body">
-          <p><strong>DOI:</strong><br>
+          <p><strong>DOI (version {{ project.version }}):</strong><br>
+            https://doi.org/10.13026/*****
+          </p>
+
+          <p><strong>DOI (latest version):</strong><br>
             https://doi.org/10.13026/*****
           </p>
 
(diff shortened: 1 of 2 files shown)


tompollard:
In cases where the current version \*is\* the latest version, I could see how the suggested wording ("this version", "latest version") might be a little confusing. Please suggest a better alternative if you have one! 

e.g.: 

- Instead of "DOI (this version)" and "DOI (latest version)", we could say "DOI (version 1.0)" and "DOI (latest version)", though I'm not sure this is really any better.
- Instead of "DOI (latest version)" we could say "DOI (all versions)". Not sure this is any better.

lilehman:
I vote for "DOI (version 1.0)" and "DOI (latest version)" --  I think it's always better to explicitly list the actual version (e.g. version 1.0) if possible.   The other option  "DOI (this version)" and "DOI (latest version)" would probably my second choice.

tompollard:
> I vote for "DOI (version 1.0)" and "DOI (latest version)" -- I think it's always better to explicitly list the actual version (e.g. version 1.0) if possible. 

I agree, on reflection I think listing the version is best. I have updated the pull request, example below:

Screenshot 2023-09-25 at 12 12 17 PM

tompollard:
I like the way Zenodo handles DOIs, with some clear explanatory text. Perhaps we can make improvements later (preferably after merging this pull request!).

Screenshot 2023-09-25 at 12 24 59 PM

alistairewj (approved):
LGTM

PR #2091, opened 2023-09-26
Fix notification for background tasks with no verbose_name

bemoody:
Background tasks can be created with a "verbose name" which provides a human-readable description.  This is what the existing background tasks (mostly implemented by Felipe) all do, and the verbose name is used when sending email notifications.

However, if a task is created without an explicit verbose_name (as is done in pull #2086), the verbose_name is set to None, which causes the email notification functions to crash.

It's best to use a verbose_name, but if we don't do so, we still want the notifications to work. :)


tompollard:
good catch!

Issue #2076, opened 2023-09-12
When changing a project's access policy to open, a file embargo can be applied without indication


briangow:
The file embargo option is not supposed to be applied to open access projects. There is code in place which prevents setting an embargo on an open access project. However, if a project was previously set with an access policy other than open and a new version of the project is created, the embargo is carried forward. This is the case even if the project is changed to open in a new version. In this case, the editor would assume that no embargo is being applied based on the message in the console "The files for an open access project cannot be placed under embargo."

We should fix this by doing one of these two things:

1. Prevent carrying the embargo forward when creating a new project version. This would require the editor to explicitly set the embargo each time a new version of a project is published. 
2. Update the code to clear `embargo_files_days` when the access policy is set to open.

tompollard:
> The file embargo option is not supposed to be applied to open access projects.

I might be misremembering, but I think the reason that we do not embargo open access projects is that it was more complicated to implement, because open access files live in a publicly-accessible folder? I think it is desirable to be able to embargo all projects, regardless of the license or usage terms.

bemoody:
> I think it is desirable to be able to embargo all projects, regardless of the license or usage terms.

Agreed that is something we should support and should open a separate issue about this :)

It also may be worth implementing a short term solution to avoid accidentally setting an embargo that won't be enforced.  This is more or less the same issue as #1759.

briangow:
Agreed, this is mostly covered in #1759 so I probably shouldn't have opened a separate issue for this. I will refer to this discussion there and close this issue.

Issue #1759, opened 2022-12-20
Handle attempts to change the access policy for a project under embargo to open


briangow:
Currently if a project is under embargo and an editor or author attempts to switch the access policy for the project to open during publication, we do not do any checks to prevent this or report an error.

In #1744 we indicate that there is a way to get access to files from open access projects without going through Django. Because of this, in #1740 we chose not to allow files in open access projects to be placed under embargo. The embargo would be meaningless if there was a way around it. 

In #1758 we indicate that it is not possible to change a project from credentialed / restricted to restricted / open. This is unrelated to whether a projects files are under embargo. 

Given this we need to do either a. or b.:

a. Implement a fix for #1758. When a project is under embargo and an attempt to switch the project's access policy to open is made, we need to either:
- prevent the user from switching by issuing a message such as "A project with files under embargo cannot be set as an open access project". 

or

- reset `embargo_files_days` to `None`, accept the request to switch the project to open, and display a message that indicates "The embargo on this project's files has been removed since it can not be set for an open access project"

b. Implement a fix for #1758. Implement a fix for #1744 which only allows users to go through Django to get to the files in open access projects. We can then remove the restriction that a project under embargo cannot be open access. In this case we do not need to add any new code related to access for projects under embargo. Regardless of whether a project is under embargo or not it will be possible to switch between open, restricted, and credentialed during the publication process.

briangow:
Please see #2076 for additional information around what will happen if a project is changed from restricted/credentialed with an embargo to open.

briangow:
Eventually we want to be able to embargo an open access project, as described in b. above. Until that is implemented, we want to prevent an embargo from being applied to an open access project. 

As discussed in #2076 , given the current state of the code we could prevent an embargo from being applied to a project that is switched to open access if we simply do not carry the embargo forward for new project versions. In a discussion with @tompollard , we agreed that this is the desirable behavior, regardless of this issue. Therefore, the editor would have to explicitly set the embargo for a new version of a project. If there is agreement on that, I will look into making this change. This will solve the short term problem by preventing an embargo from being accidentally applied to an open access project. 

We should leave this issue open (or create a new issue) until we come up with a solution for how to allow an embargo on an open access project.

PR #2092, opened 2023-09-27
Prevent a file embargo from being carried forward to new project versions

briangow:
In #1759 and #2076 we highlight what will happen if a restricted / credentialed project is published with an embargo and then subsequently a new version is created and the access policy is changed to open and published. When this is done the embargo will be carried forward from the previous version and the editor will not realize this (since we don't allow an embargo on an open project the console messaging indicates that no embargo will be put in place even though it will). Upon publication the projects files will be under embargo which shouldn't happen for an open access project as described in #1759.

This PR prevents carrying an embargo forward when a new project version is created. This will prevent a hidden embargo from being carried forward should a project with an embargo be switched to open in a new version. This is really the ideal behavior even for projects which don't change their access policy to open upon submitting a new version. We really want the editor to explicitly set an embargo for a given version instead of using the embargo setting for the previous version of the project.

### DIFF
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -429,7 +429,7 @@ class NewProjectVersionForm(forms.ModelForm):
 
         # Direct copy over fields
         for field in (field.name for field in Metadata._meta.fields):
-            if field not in ['slug', 'version', 'creation_datetime']:
+            if field not in ['slug', 'version', 'creation_datetime', 'embargo_files_days']:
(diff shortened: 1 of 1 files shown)


bemoody:
> We really want the editor to explicitly set an embargo for a given version instead of using the embargo setting for the previous version of the project.

I can see an argument either way.  But we don't have a lot of embargoed projects so whichever way we manage them seems fine.

bemoody (approved):

PR #2101, opened 2023-10-04
Validate AWS IDs when users set their AWS IDs in the Edit Cloud Details information

Chrystinne:
This pull request validates AWS IDs to ensure they match the 12-digit numerical values expected for AWS account IDs when users set their AWS IDs in the Edit Cloud Details information.


tompollard:
Looks good to me, thanks Chrystinne. A couple of quick thoughts:

- It would be nice if the error explained the pattern that we are expecting (12 digit number). 
- The validator belongs in a validators.py module (e.g. https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/user/validators.py or whichever app cloud code lives in).
- Following convention in validators.py, perhaps rename function to something like `validate_aws_id`?

tompollard:
> Following convention in validators.py, perhaps rename function to something like validate_aws_id?

Actually I see we have `is_institutional_email`, so perhaps something like `is_aws_id` fits after all.

bemoody:
This logic should go in the *form* class if not the *model* class.  It doesn't belong in views.py.

Probably the best thing would be to define it as a validator in the CloudInformation class.  Try that and see if it works.  If that doesn't work then define a `clean_aws_id` method in the CloudForm class.

physionet-django/user/models.py
```diff
@@ -1218,8 +1218,7 @@ class CloudInformation(models.Model):
                                 on_delete=models.CASCADE)
     gcp_email = models.OneToOneField('user.AssociatedEmail', related_name='gcp_email',
                                      on_delete=models.SET_NULL, null=True)
-    aws_id = models.CharField(max_length=60, null=True, blank=True, default=None)
-
+    aws_id = models.CharField(max_length=60, null=True, blank=True, default=None, validators=[validators.validate_aws_id])
```
tompollard: I think you've lost the blank line before `class Meta`? It needs to be added back.

Chrystinne: @tompollard I accidentally did. Thanks for catching that!

Chrystinne:
@bemoody Using a validator in the CloudInformation model class works. Thanks.

physionet-django/user/views.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import os
 import pdb
+import re
```
tompollard: this import is no longer needed.

tompollard:
Thanks @Chrystinne, looks good to me. My understanding is that validators won’t retroactively affect existing data, so there shouldn't be any problems with applying this on the live server. 

Please could you squash the commits before we merge? Perhaps one commit for the addition of the validator, and one for the migration?

tompollard:
thanks!

Issue #2056, opened 2023-08-08
Change Credential Time Length in Credential Email


mscanlan-git:
The email that users receive when submitting their credential application states "It may take several weeks to process your request".

![git-issue](https://github.com/MIT-LCP/physionet-build/assets/67070862/a1b36ce5-4c92-4926-a826-fd4a6f0230ea)

With some people needing access swiftly, this causes some users to send us emails asking for an expedited review. We can drastically cut down on these types of emails by stating "Please allow 3-5 days for your application to be reviewed and processed". Setting 3-5 days should suffice to add some buffer in case people need to wait due the weekends or holidays. 

I don't expect this change will eliminate all emails, but it should get rid of concerns people may have for a several week process.

tompollard:
Sounds good to @mscanlan-git. It is a bit more work, but we could add the time as a variable to our config file, so the template does something like:

```Please allow {{ estimated_time_for_credentialing }} for your application to be reviewed and processed.```

PR #2102, opened 2023-10-05
Update Credential and Training email time length 

mscanlan-git:
This PR addresses #2056 by adjusting the estimated wait users may expect to have their credential or training applications reviewed.


tompollard (commented):
Thanks again @mscanlan-git! Moving the estimated time to a variable is an improvement. Why not do the same for physionet-django/user/templates/user/edit_credentialing.html? 

Next step is to move the decision time into an environment variable, but this can be done in a later pull request.

tompollard:
Looks great, thanks @mscanlan-git!

PR #2111, opened 2023-10-16
files_panel dynamic navigation fixes

bemoody:
When you go to the "Files" section of a preview or published project, and click on a link to a subdirectory (with javascript enabled), then instead of reloading the entire page, a script will load just the files section for that directory.

I think this is generally a convenient thing, but there are a bunch of significant problems with the current implementation:

- After you navigate to a subdirectory in this way, the page URL is not modified (so you can't bookmark it.)

- After you navigate to a subdirectory, the back button doesn't take you back to where you were.

- After you navigate to a file within a subdirectory, the back button might or might not take you back to where you were.

- After you navigate to a subdirectory, other subdirectory links (if you right-click on one and say "copy link") are broken.

- The current implementation has theoretical XSS vulnerabilities.

- Inline scripts are a bad idea and should never be used.

The first three issues are variously fixed by calling `history.replaceState` or `history.pushState` before loading the new HTML.  The other issues are fixed by putting the logic in a separate, static, script file.

For example, if you go to `https://localhost:8000/content/demoecg/10.5.24/#files-panel`:

- If you click on "app", the page URL should change to `https://localhost:8000/content/demoecg/10.5.24/app/#files-panel`.

- If you then right-click on "Parent Directory" and click "Copy Link", the copied text should be `https://localhost:8000/content/demoecg/10.5.24/#files-panel` (not `https://localhost:8000/content/demoecg/#files-panel`.)

- If you then click on "12lead.pro", it should display the file; then, clicking the Back button should take you back to the "app" subdirectory (not to the toplevel directory of the project.)

- If you click the Back button again, it should return you to the toplevel directory (not back to some other page.)

It would be quite complicated to make this change in a truly backward-compatible way (i.e., returning the correct URLs and loading the new script when somebody invokes the *old* `navigateDir` function), and I thought this wasn't worth the effort.  So this is set up as a parallel template; old clients can still use the old one.  After a while the old template can be removed.

### DIFF
--- a/physionet-django/project/static/project/js/dynamic-files-panel.js
+++ b/physionet-django/project/static/project/js/dynamic-files-panel.js
@@ -1,0 +1,38 @@
+(function() {
+    'use strict';
+
+    var panel = $('#files-panel');
+    var cur_dir = panel.find('[data-dfp-cur-dir]').data('dfp-cur-dir');
+    var panel_url = panel.find('[data-dfp-panel-url]').data('dfp-panel-url');
+
+    function navigateDir(subdir, page_url, push_history) {
+        $.ajax({
+            type: 'GET',
+            url: panel_url,
+            data: {'subdir': subdir, 'v': '2'},
+            success: function(result) {
+                if (push_history)
+                    history.pushState(subdir, '', page_url);
+                else
+                    history.replaceState(subdir, '', page_url);
+                panel.html(result);
+                setClickHandlers();
+            },
+        });
+    }
+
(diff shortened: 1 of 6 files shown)


bemoody:
> The current implementation has theoretical XSS vulnerabilities

I should mention that `edit_files_panel` does too.  As I say, only a *theoretical* problem at present but should be fixed too.

tompollard:
looks great benjamin, thanks!

Issue #2105, opened 2023-10-10
List group membership for users in the admin console


tompollard:
We now handle permissions through "groups". e.g. an event host is a member of the "host" group. We should display group membership on the user pages in the admin console (https://physionet.org/console/user/manage/USERNAME/).

e.g. on the user page, we could show:

### Permission groups

The user is a member of the following groups:

- Host
- Web editor
- etc.

mscanlan-git:
I noticed in the code that we already display the user's group, however, I think putting this info in it's own subsection on the page would look better, especially if the user is apart of multiple groups:

<img width="881" alt="Screenshot 2023-10-11 at 2 42 57 PM" src="https://github.com/MIT-LCP/physionet-build/assets/67070862/1f49bd6c-3198-48cd-8f13-7390e9095dfc">

If this is what you had in mind, I can put a PR up for this.

PR #2113, opened 2023-10-17
Added permission groups to user management page in console

mscanlan-git:
This PR addresses #2105 by adding a permissions group section on the user's page in the console. 

Currently, the template is set up to list "no group" if the user does not have any group affiliations, however, I'm open to suggestions on maybe hiding the section entirely if the user has no group affiliation.

### DIFF
--- a/physionet-django/console/templates/console/user_management.html
+++ b/physionet-django/console/templates/console/user_management.html
@@ -111,21 +111,27 @@
     </div>
   {% endfor %}
 
-  {% if groups %}
-      <div class="row mb-1">
-        <div class="col-md-3">
-          User Groups:
-        </div>
-        <div class="col-md-9">
-          {% for group in groups %}
-              <a href="{% url 'user_group' group.name %}">{{ group.name }}</a>
-          {% empty %}
-            N/A
-          {% endfor %}
-        </div>
+  <br />
+  <h3>Permission groups</h3>
+  <hr>
+    <div class="row mb-1">
+      {% if groups %}
+      <div class="col-md-6">
+        The user is a member of the following groups:
       </div>
-
-  {% endif %}
+      <div class="col-md-9">
+        <ul>
+        {% for group in groups %}
+            <li><a href="{% url 'user_group' group.name %}">{{ group.name }}</a></li>
+        {% endfor %}
+        </ul>
+      </div>
+      {% else %}
+        <div class="col-md-6">
+          The user does not belong to a permission group.
+        </div>
+        {% endif %}
+    </div>
 
   <br />
   <h3>Projects</h3>


physionet-django/console/templates/console/user_management.html
```diff
@@ -111,21 +111,21 @@ <h3>Profile</h3>
...
-        </div>
-        <div class="col-md-9">
-          {% for group in groups %}
-              <a href="{% url 'user_group' group.name %}">{{ group.name }}</a>
-          {% empty %}
-            N/A
-          {% endfor %}
-        </div>
+  <br>
+  <h3>Permission Groups</h3>
```
tompollard: For consistency with other headers, we should probably use sentence case here: "Permission groups"

physionet-django/console/templates/console/user_management.html
```diff
@@ -111,21 +111,21 @@ <h3>Profile</h3>
...
-          User Groups:
-        </div>
-        <div class="col-md-9">
-          {% for group in groups %}
-              <a href="{% url 'user_group' group.name %}">{{ group.name }}</a>
-          {% empty %}
-            N/A
-          {% endfor %}
-        </div>
+  <br>
```
tompollard: Doesn't really matter, but nice to use `<br />` for consistency with other breaks in the document?

physionet-django/console/templates/console/user_management.html
```diff
@@ -111,21 +111,21 @@ <h3>Profile</h3>
...
+  <hr>
+    <div class="row mb-1">
+      <div class="col-md-6">
+        The user is a member of the following groups:
       </div>
-
-  {% endif %}
+      <div class="col-md-9">
+        {% for group in groups %}
+            <ul><li><a href="{% url 'user_group' group.name %}">{{ group.name }}</a></li></ul>
```
tompollard: `<ul>` should go above `{% for group in groups %}` and `</ul>` should go below `{% endfor %}`. Otherwise this creates a new list each loop (rather than a new item in the list).

physionet-django/console/templates/console/user_management.html
```diff
@@ -111,21 +111,21 @@ <h3>Profile</h3>
...
+        {% empty %}
+            <ul><li>No Groups</li></ul>
```
tompollard: It feels slightly unnecessary to have a "No groups" bullet here. Could we instead say `The user is a member of the following groups:...` if there is one or more groups and something like `The user does not belong to a permission group.` if not?

tompollard (commented):
Looks good, thanks Michael. I added a few very minor comments inline!

physionet-django/console/templates/console/user_management.html
tompollard: We now have:

```
      <div class="col-md-9">
        <ul>
        {% for group in groups %}
            <li><a href="{% url 'user_group' group.name %}">{{ group.name }}</a></li>
        {% empty %}
            The user does not belong to a permission group.
        {% endfor %}
        </ul>
      </div>
```

For people who aren't in a group, this step will output `The user does not belong to a permission group.` inside `<ul>` tags which I think isn't quite right. I haven't tested, but what about something like this:

```
  <br />
  <h3>Permission groups</h3>
  <hr>
    <div class="row mb-1">
      {% if groups %}
        <div class="col-md-6">
          The user is a member of the following groups:
        </div>
        <div class="col-md-9">
          <ul>
          {% for group in groups %}
              <li><a href="{% url 'user_group' group.name %}">{{ group.name }}</a></li>
          {% endfor %}
          </ul>
        </div>
      {% else %}
          <div class="col-md-6">
            The user does not belong to a permission group.
          </div>
      {% endif %}
    </div>
```

mscanlan-git: I switched over to the new format, it made the else statement make more sense by not including the `<ul>` tag

tompollard:
thanks, looks good to me!

PR #2115, opened 2023-10-18
Fix security bugs in edit_files_panel

bemoody:
As a followup to pull #2111, I noted that the `edit_files_panel` has the same security problem as the old `files_panel`.

(This affects the `project_files` view as well as `copyedit_submission`.)

For example, in `media/active-projects/SHuKI1APLrwWCqxSQnSk`, try renaming `subject-100` to `subject-100'+alert('Hello world')+'` .  Then go to the project files page (not the preview; that bug has now been fixed) and click the corresponding link.


tompollard:
Yoiks, good catch

Screenshot 2023-10-19 at 4 57 50 PM

Issue #2094, opened 2023-09-27
S3: restricted access bucket/policy structure


bemoody:
We want to provide mirrors of our published projects on Amazon S3 (similar to our existing Google Cloud mirrors) for purposes of providing faster and more convenient access.

For *restricted access projects*, we ideally want to allow each authorized individual to access the files via S3 API (so they can retrieve the data from Amazon directly, rather than going through our server to authorize each request.)

Amazon has a very low limit on the *size of the encoded policy document* (a JSON file) that can be attached to a single bucket.  Small enough that it is very likely there will be more MIMIC users than space to encode their Amazon account IDs.

https://stackoverflow.com/a/75936718

> After contacting AWS support on this it turned out that:
>
> 1. 20kB policy size limit is still valid and enforced.
>
> 2. AWS is doing some policy normalizations which reduce the total size of the policy, so the total size of the policy in characters may be bigger then what JSON.stringify(policy) may report.
>
> 3. To my understanding there are no means to calculate the size of the normalized policy besides contacting AWS S3 support. If JSON.stringify(policy).length < 20kB then you are safe, else consider a workaround soon.
>
> 4. Using multiple S3 Access Points is one of the ways how to bypass the policy size limits.

What are "S3 Access Points"?  How do we create them?  Is there a difference from a client's perspective between loading data from a shared bucket, versus loading data from a shared Access Point?

bemoody:
Also, I saw somewhere in the documentation that the policy supports something called "groups" but it sounded like a "group" is only a group of users *within* an AWS account, not a group of multiple AWS accounts.  Correct me if I'm wrong.

bemoody:
From what I understand, an access point is in effect an alias for a particular bucket, that has its own access policy.

So you can create a bucket `foo` and an access point `bar`, and the policy of `bar` specifies a list of account IDs and a list of prefixes that those accounts would be allowed to access.

(It sounds like an access point cannot point to a "subdirectory" of a bucket, or otherwise present a modified view of the keys/filenames in the bucket - but that's okay for our purposes.)

So it seems like a reasonable strategy would be to place *all* restricted project files in a single bucket, and create one or more access points for each one.


https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-points-restrictions-limitations.html

> You can create a maximum of 10,000 access points per AWS account per Region

which would be plenty if we wanted to assign one access point per 500 MIMIC users, though not enough to have one access point per user.


[...]

PR #2121, opened 2023-10-24
AWS identity verification

bemoody:
## Context

Pull #2086 added the ability to mirror project content on Amazon S3.  This is now working and we are in the process of uploading open-access projects from PhysioNet.

The changes here will be needed once we start uploading *restricted/credentialed* projects, so that we can securely grant access to authorized users.  (Identity verification aside, there are also some more significant changes that are needed for handling restricted/credentialed projects; see issue #2094.)

In brief: currently (in the old system Felipe set up), people are asked to *self-report* their AWS account number, and *any person or service within that account* would be allowed to access restricted data.

With these changes, in contrast, people will be asked to *verify* their personal AWS identity; subsequently, we'll be able to grant access *only to verified identities* (the latter part is yet to be implemented.)

## Why

DUAs for MIMIC and other databases require that data is only shared with authorized *individuals* (each person must register on PhysioNet and be credentialed.)  We want to enable cloud access for better performance, but complying with these DUAs requires knowing who is being granted permission to use these cloud services.

Moreover, although each user is ultimately responsible for data security, we want to encourage good practices.  People may be using AWS for all sorts of reasons unrelated to PhysioNet.  Giving themselves permission to access MIMIC through their personal account should not *also* grant permission to all of those unrelated and possibly-less-trusted services.

Some people may be using organizational AWS accounts rather than personal ones.  Maybe we want to discourage this, or maybe not, but we can't prevent it.  One member of an organization having access shouldn't grant access to everyone in the organization.

There is a lot about AWS authentication that is still a bit mysterious to me, but my gut feeling is that the "IAM user" level is the right level of authentication for PhysioNet and MIMIC.

It has been suggested that we could ask people to self-report their AWS username (or ARN?) in addition to their account number.  And yes, that would be an improvement; but it has the disadvantage that usernames are variable-length, and may not be long-term stable.  Better would be to ask people to self-report their *AWS userid*, but that's not easy for people to find and more likely to cause mistakes.

Finally, I can imagine that in the future there may be other reasons for wanting to associate a PhysioNet account with an AWS account, and having a strong verification process could enable more interesting forms of integration.

## How identity verification works

The concept is that we would have a special-purpose S3 bucket which allows access *only if the path matches the requester's AWS account and userid.*  To prove your identity, you generate a signature for a URL that can only be accessed by you, and paste that signed URL into a form on the site.

The process would be:

1. You go to your cloud settings page on PhysioNet.

2. We tell you to run the command `aws sts get-caller-identity`.

3. You copy the output into the form.

4. We then tell you to run a command like `aws s3 presign s3://asdfghjk/physionet.org-verification/email=root@example.org/userid=AIDAABCDEFGHIJKL/account=112233445566/username=barackobama/`.

5. You copy the output into the form.

6. We verify the format of the URL and submit it to AWS to verify the signature.

## Wait a minute, what's this "userid" thing you keep talking about?

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-unique-ids

## Setup and testing

Using this feature requires creating a special-purpose S3 bucket (a bucket which probably should not be used for anything else.)

*For the time being*, you can test this by setting `AWS_VERIFICATION_BUCKET_NAME` to `bm-uverify-test1`.  I will delete that bucket once we've set up a permanent replacement under the PhysioNet AWS account.

If you want to see exactly how the verification bucket is created, and test it yourself, see the instructions in `deploy/README.md`.

## Background

Although this implementation is guided by the needs of PhysioNet, my goal has been to design a general-purpose authentication protocol that could be used by any site that needs to verify cross-account AWS identities.

This is inspired in part by the technique used by Hashicorp Vault and discussed here:
* https://ahermosilla.com/cloud/2020/11/17/leveraging-aws-signed-requests.html
* https://www.hashicorp.com/resources/deep-dive-vault-aws-auth-backend

and similarly: https://stackoverflow.com/a/76099155

We could use the same method, but it would require the person to download and run a small program (and that program involves some pretty hairy digging into the AWS API.)

The method proposed here, in contrast, only requires the person to install the official AWS CLI and run a couple of commands.  I think that this is easier to understand and therefore paradoxically more secure (see if you can spot the security flaw in the StackOverflow answer.)

For information about *why* this works, see AWS documentation on policy variables:
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html

Also see the AWS CLI documentation:
https://docs.aws.amazon.com/cli/latest/reference/sts/get-caller-identity.html
https://docs.aws.amazon.com/cli/latest/reference/s3/presign.html


bemoody:
Just using the GetObject permission doesn't quite work - if the resource doesn't exist then we get a 403 even if we have permission to access it.

https://stackoverflow.com/a/19038017 explains why this is (which makes sense), though it's not 100% clear how the backend is interpreting the policy, and I'd like to find some official AWS docs clarifying the matter.

Adding ListBucket permission *seems* to make it work (and raises the question of whether the GetObject permission itself is needed.)

bemoody:
Unfortunately, patching in X-Amz-Expected-Bucket-Owner *doesn't* work.  Don't know if there's a way to include that header in the `aws s3 presign` command.  We can live without it, it just feels a little less safe.

bemoody:
Now with UI and should be working!

bemoody:
Nits:
- modal before deleting
- better error message for unsupported (non-AIDA) identity
- save original `Arn` so we can show it on the edit_cloud page?

bemoody:
Hmm.  Some sources say that Amazon presigned URLs have a maximum expiration time of one week.  Maybe that's true for the v4 signature?  It's definitely not true for the v2 signature, I just generated one that was valid for a year and it was accepted.

bemoody:
github-advanced-security, you're adorable.  It's a fair point that I should rework that regexp to reduce backtrackability (though I'm not sure it's possible to completely fix the issue; python `re` is annoying that way.)

As for the rest of it, get outta here. ;)

bemoody:
I believe this approach should work **if we use S3 Access Points** to grant access.  If we use **S3 Access Grants** then it is less clear - can we still grant access based on userid or do we have to use ARN?

So we may have to decide which approach to take before deciding how to validate identities. :(

bemoody:
> So we may have to decide which approach to take before deciding how to validate identities.

I'm still not clear about the access-grant thing and how it would work.

But just to note: this code could be changed to verify the requester's `aws:username` in addition to their `aws:userid`.  Arguably it makes sense for us to do that regardless.

So although I think it's questionable, security-wise, to rely on cross-account IAM ARNs, it's not impossible and also wouldn't preclude other approaches.

bemoody:
Note that using presigned URLs seems to be broken with bucket regions other than `us-east-1`: https://github.com/boto/boto3/issues/3015

This affects older versions of awscli.  Newer versions don't use boto3 and are possibly unaffected.

For the time being, PhysioNet should use us-east-1 for user verification, and other sites should do the same.

Unrelated to user verification, this might also cause problems if we want to store data in regions other than us-east-1.

bemoody:
I'm banning this bot.  I don't accept contributions from bots unless they are free and open source.

(Seriously.  I am not interested in providing free labor to Microsoft to help them develop better tools to lock people into GitHub.)

bemoody:
regarding boto3 being silly in test suite, see https://github.com/MIT-LCP/physionet-build/pull/2150#issuecomment-1843801417

PR #2126, opened 2023-11-06
Remove hard dependency on psycopg2

bemoody:
Importing the `django.contrib.postgres.search` module requires importing `psycopg2`.  Installing `psycopg2` can be a minor nuisance and is otherwise unnecessary if you're not using the postgres backend.


tompollard:
Feels odd to do this but...okay! Don't see any strong reason why not.

Issue #2128, opened 2023-11-08
Allow submitting authors to reassign a project to a different submitting author


tompollard:
Currently the submitting author of a project cannot be changed. We should consider updating the system to allow submitting author for a project to be switched (in a similar way to project reviewers). 

Options: 
1) Enable switching by project administrators, via the console.
2) Enable switching by submitting authors.

Perhaps begin with 1 and then introduce 2 later.

PR #2131, opened 2023-11-13
Allow submitting author status to be transferred to a co-author. Ref #2128

tompollard:
As discussed in #2128, currently the submitting author of a project cannot be changed. There are times when it would be helpful to allow submitting authors to transfer this status to co-authors. e.g. when:

- A co-author would like to take responsibility for editing content or uploading files.
- A project requires updating and the original submitting author is no longer available.

This change adds a 'transfer project' button to the author page of the project submission system.

### DIFF
--- a/physionet-django/notification/templates/notification/email/notify_submitting_author.html
+++ b/physionet-django/notification/templates/notification/email/notify_submitting_author.html
@@ -1,0 +1,11 @@
+{% load i18n %}{% autoescape off %}{% filter wordwrap:70 %}
+Dear {{ name }},
+
+You have been made submitting author of the project entitled "{{ project.title }}" on {{ SITE_NAME }}.
+
+You can view and edit the project on your project homepage: {{ url_prefix }}{% url "project_home" %}.
+
+{{ signature }}
+
+{{ footer }}
+{% endfilter %}{% endautoescape %}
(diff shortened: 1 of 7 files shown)


bemoody:
Not tested but sounds like a great idea.

If the project is published then only the original submitting author can create a new version.  (I have thought that it would be nice to allow any author to do that, but that can be another PR.)

However, we don't want to allow the original submitting author to create *two* new versions of the same project - does this still prevent that?

tompollard:
> we don't want to allow the original submitting author to create two new versions of the same project - does this still prevent that?

Good question! I tested this fairly thoroughly and, as far as I can see, it is not possible to create multiple new versions of a project. e.g. if the submitting author creates a new version and transfers the new version to a co-author, it is not possible to create a second new version.

I think the logic that prevents multiple new versions is defined here:

https://github.com/MIT-LCP/physionet-build/blob/cece7570ff004fb1542d3059bc1cf61610d17922/physionet-django/project/views.py#L317-L340

... and here:

https://github.com/MIT-LCP/physionet-build/blob/cece7570ff004fb1542d3059bc1cf61610d17922/physionet-django/project/modelcomponents/publishedproject.py#L254-L261

bemoody:
Sounds good, thanks for checking.

Could you add a test case in `project/test_views.py`?  It looks like the demo "MIMIC-III Clinical Database" (slug = T108xFtYkRAxiRiuOLEJ) has `rgmark` as submitting author and `aewj` as a coauthor, so we should be test reassigning it.

tompollard:
@bemoody Test case added!

bemoody:
```
+        self.assertTrue(self.submitting_author.user.email, self.AUTHOR_EMAIL)
+        self.assertTrue(self.coauthor.user.email, self.COAUTHOR_EMAIL)
```
You mean `assertEqual` here.

tompollard:
> You mean `assertEqual` here.

Good catch, now fixed.

PR #2133, opened 2023-11-16
Set default order of PublishedProject to `(title, version_order)` and ActiveProject to `(title, creation_datetime)`. Fixes #2132

### LINKED ISSUE #2132: List for selecting a parent project is not ordered well
When selecting a parent project for a publication, you can search by typing (which is great!), but you get all the projects in a really weird order (not so great):

Screenshot 2023-11-16 at 6 25 49 AM

The things that would make this easier are (in order of preference):

1. Versions for the same project are grouped together. So it shows MIMIC-IV-ED v1.0, MIMIC-IV-ED v2.0, MIMIC-IV-ED v2.1, MIMIC-IV-ED v2.2 sequentially, one after another.
2. Projects are ordered alphabetically, so MIMIC-IV v1.0, MIMIC-IV v2.0, ..., MIMIC-IV-ED v1.0, MIMIC-IV-ED v2.0, ...

tompollard:
As discussed in #2132, there is no default ordering for Published Projects. This means that tools that render lists of projects are messy. It is often difficult to find the project of interest because there is no clear order. One example is the "Select parent project" selection tool that appears during project submission.

This pull request sets the default ordering for Published Projects and Active Projects to `('title', 'version')`. I'd initially thought that we could set the ordering in the [MetaData](https://github.com/MIT-LCP/physionet-build/blob/80e589afda52d617c897e91528d0729d1996304e/physionet-django/project/modelcomponents/metadata.py#L24) class, but it turns out that the `Meta` element of Django models isn't inherited.


bemoody:
Don't sort by `version`, sort by `version_order`.  `2.0` comes before `10.0`.

That aside, it looks like this should address the unpredictability discussed in #2132.

However, this seemingly doesn't address *all* uses of unordered PublishedProject query sets (test_urls still generates one warning, I don't know from where.)

I also notice that there are two "project-autocomplete" views (one in `project`, one in `console`) that are exact duplicates of each other.  It looks like the one in `console` is not being used and should be removed.

tompollard:
> Don't sort by version, sort by version_order. 2.0 comes before 10.0.

I updated PublishedProject to use `version_order`. Looks like `version_order` doesn't exist for ActiveProject.

> I also notice that there are two "project-autocomplete" views (one in project, one in console) that are exact duplicates of each other. It looks like the one in console is not being used and should be removed.

I can fix this, but it seems only loosely related to this change. Shall I put in a separate PR?

bemoody:
> Looks like version_order doesn't exist for ActiveProject.

Hmm, yeah, for active projects it's a different situation.  I'm ambivalent about putting default orders on things, what makes sense in one case isn't necessarily desirable elsewhere.  If you want a default order on active projects, then shouldn't it be submission date or creation date or something?

> Shall I put in a separate PR?

Sure, please do.

tompollard:
> If you want a default order on active projects, then shouldn't it be submission date or creation date or something?

I think having some default sort order is helpful (e.g. I fairly regularly find myself searching for active projects in the Django console, usually on my dev server). I've switched to `creation_datetime` as suggested.

bemoody:
Thanks, looks good.

Issue #2122, opened 2023-10-25
S3: zip archives


bemoody:
We have to decide whether and how we are going to upload *zip archives* to S3.

Personally I would rather fix issue #1674 first, but that may take some time and it would probably be wiser to upload the zip files in their current imperfect state.

Currently, pull #2086 would upload the archive for `mitdb/1.0.0` as `s3://XXXXXX/mitdb/mit-bih-arrhythmia-database-1.0.0.zip` and the files in that archive would be named `mit-bih-arrhythmia-database-1.0.0/100.hea` etc.

Preferably the archive would be named `s3://XXXXXX/mitdb/mitdb-1.0.0.zip` and a file inside the archive would be named `mitdb-1.0.0/100.hea`.

Should we:

- keep the file as `s3://XXXXXX/mitdb/mit-bih-arrhythmia-database-1.0.0.zip`, plan to remove and replace it with `s3://XXXXXX/mitdb/mitdb-1.0.0.zip` eventually?

- upload the existing file as `s3://XXXXXX/mitdb/mitdb-1.0.0.zip`, plan to replace it eventually with a different version at the same URL?

- not upload zip files to S3 at all until issue #1674 is resolved?

@chrystinne @tompollard

tompollard:
> not upload zip files to S3 at all until issue https://github.com/MIT-LCP/physionet-build/issues/1674 is resolved?

This is tempting, but as I don't want to be responsible for holding @Chrystinne up, I think the best solution is:

> upload the existing file as s3://XXXXXX/mitdb/mitdb-1.0.0.zip, plan to replace it eventually with a different version at the same URL?

bemoody:
Addressed by pull #2086.

Issue #1674, opened 2022-09-23
Use project slug for generating name of zipfiles, instead of project title


tompollard:
We generate a zipfile containing all files within a project. At the moment, it looks like the filename of this zipfile is based on the project title. It would make much more sense for the filename to be based on the project slug, I think. 

The slug is unique, it doesn't have characters like spaces, apostrophes, etc, and it is short. Titles are much less suited to a filename. The function that needs a rethink is copied below:

https://github.com/MIT-LCP/physionet-build/blob/141d2a9262e6b9191e3bf81f3ccee9485c43e9cf/physionet-django/project/modelcomponents/publishedproject.py#L112-L126

superryeti:
Hi @tompollard , I had one concern about this.

Wouldn't changing the names of zipfile mess up the existing zips that were already created?

After checking the codebase, my understanding is that there are several parts of code that initiate zip of project files(which uses the `zip_name` function and create zip of the project files, and there are other part in code that call this `zip_name` function again when someone tries to retrieve the zip file for the project.

So that means when we update this function, the retrieval of zip will break for a lot of existing projects.

I assume to solve that, we need to retrigger the zip of project files, and that might take a lot of resources depending on the number of projects and the file size.. Plus i am not sure if we keep a entry in database, if the project files have been zipped

tompollard:
> Wouldn't changing the names of zipfile mess up the existing zips that were already created?

Yes, this is true. Part of the work would be fixing the current filenames (on our servers, and on the cloud servers). I don't think this would be a huge effort, and if we think it is necessary then it would be better to do sooner than later.

superryeti:
Good point. Do you think somekind of management command might be a good idea to do that? 
If feels weird to have a management command for a one off thing. (but i can't think of anything else right now)

any suggestions?

tompollard:
> Good point. Do you think somekind of management command might be a good idea to do that?
If feels weird to have a management command for a one off thing. (but i can't think of anything else right now). any suggestions?

@bemoody probably has thoughts on this, but I think a management command for the migration would be fine (and it could always be removed later).

bemoody:
Could be a data migration (RunPython) that creates a hard link if the
old zip file already exists.  Or we could just do it by hand.  Don't
want to delete the old files (at least not immediately) because
clients are in the process of downloading them.

bemoody:
Note this problem just came up again (vindr-cxr 1.0.0).

bemoody:
[...]

PR #2134, opened 2023-11-17
Add "legacy" parameter to PublishedProject.zip_name

bemoody:
We would like to avoid including the project title in zip file names as well as in the zip file contents, in favor of the project slug, which is more stable and machine-readable.  (See issue #1674 for more background.)

Eventually, therefore, the behavior of the zip_name function should be changed; as a first step in that direction, add an optional argument so callers can specify whether they want the old or new style.

Existing code continues to use the old style for now, but when we upload zip files to S3 I'd like to rename the zip files to the new style (see issue #2122).  Adding the code in PublishedProject should make that easier and cleaner, and allow the rest of the site to be migrated eventually.

### DIFF
--- a/physionet-django/project/modelcomponents/publishedproject.py
+++ b/physionet-django/project/modelcomponents/publishedproject.py
@@ -123,11 +123,21 @@ class PublishedProject(Metadata, SubmissionInfo):
         """
         return '-'.join((slugify(self.title), self.version.replace(' ', '-')))
 
-    def zip_name(self, full=False):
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks, looks good! As next steps, could we:

- Set the default argument to `legacy=False` (or remove it altogether?)
- Update all references to the zipfiles to a try/except (`try: new_zipfile; except legacy_zipfile`)?

We can discuss at: https://github.com/MIT-LCP/physionet-build/issues/1674

PR #2140, opened 2023-11-28
Fix permissions for the Pages section of the admin console. Closes #2060.

### LINKED ISSUE #2060: Permissions incorrectly set up for the Static Pages functionality in the console
There is an issue with the way permissions are set up for the Static Pages functionality in the console (https://physionet.org/console/static-pages/). It seems that we are unable to grant necessary permissions to navigate to the pages from the console menu.

https://github.com/MIT-LCP/physionet-build/blob/304e093dc550da8636552dc601d6545c07ffc771/physionet-django/console/templates/console/console_navbar.html#L316-L323

The menu requires users to have `perms.user.change_staticpage`, but this permission doesn't appear to exist (`perms.physionet.change_staticpage` apparently does however).

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/2060, there is a bug the "Pages" permissions. Currently, it is not possible to assign users the correct permissions to view the Pages tabs in the console.

Screenshot 2023-11-28 at 3 12 41 PM

There are two issues:

1. The navbar HTML template is looking for `user.change_staticpage` and `user.change_frontpagebutton` permissions, which don't exist. These permissions belong to the `physionet` app, not `users`.
2. The `view_redirects` view is restricting access to a `physionet.view_redirect` permission, which doesn't exist. The redirect permission belongs to Django's `redirects` app.

Side note, but  it isn't ideal that the navbar logic displays all three subitems {% if perms.physionet.change_staticpage or perms.physionet.change_frontpagebutton or perms.redirect.view_redirect %}, even if the user only has permission to view/edit one of the subitems. I think we can worry about this later.

It might be cleaner to display all tabs/pages admin users, even if the user doesn't have permission to edit/view the page. This way the user knows what functionality exists, and they can request access if needed.

### DIFF
--- a/physionet-django/console/templates/console/console_navbar.html
+++ b/physionet-django/console/templates/console/console_navbar.html
@@ -317,7 +317,7 @@
       {% endif %}
 
       <!-- Pages -->
-      {% if perms.user.change_staticpage or perms.user.change_frontpagebutton or perms.redirect.view_redirect %}
+      {% if perms.physionet.change_staticpage or perms.physionet.change_frontpagebutton or perms.redirect.view_redirect %}
       <li class="nav-item" data-toggle="tooltip" data-placement="right">
         {% if frontpage_buttons_nav or static_pages_nav %}
           <a id="nav_pages_dropdown" class="nav-link nav-link-collapse drop" data-toggle="collapse" href="#pagesComponent" data-parent="#sideAccordion" aria-expanded="true">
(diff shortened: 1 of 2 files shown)


bemoody:
> The view_redirects view is restricting access to a physionet.view_redirect permission, which doesn't exist. The redirect permission belongs to Django's redirects app.

It seems like our test suite should have caught this; probably it didn't because `admin` is a "superuser" and therefore is assumed to have all permissions, even those that don't exist. :/

PR #2146, opened 2023-11-29
AWS S3 sync command

Chrystinne:
This pull request adds a prefix to the generated AWS S3 sync command, including the project slug, and version.

### DIFF
--- a/physionet-django/project/modelcomponents/storage.py
+++ b/physionet-django/project/modelcomponents/storage.py
@@ -59,3 +59,9 @@ class AWS(models.Model):
 
     class Meta:
         default_permissions = ()
+
+    def s3_uri(self):
+        if self.is_private:
+            return f's3://{self.bucket_name}/{self.project.version}/'
+        else:
+            return f's3://{self.bucket_name}/{self.project.slug}/{self.project.version}/'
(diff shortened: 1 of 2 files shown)


bemoody:
We want to avoid putting logic in template files whenever we can.  Use Python functions and classes to define *what information should be shown*, and use templates to define *how it should be displayed as HTML*.

Also, a `<pre>` element should not contain unnecessary whitespace, and must not contain block elements such as `<p>`.

Instead of putting more logic into published_project.html, try defining a method in the `AWS` class that returns the S3 URI for the project.
```
class AWS(models.Model):
    ...
    def s3_uri(self):
        if self.is_private:
            return f's3://{self.bucket_name}/{self.project.version}/'
        else:
            return f's3://{self.bucket_name}/{self.project.slug}/{self.project.version}/'
```

Then in the template file we can simply write:
```
              <pre class="shell-command">aws s3 sync {{ project.aws.s3_uri }} DESTINATION</pre>
```

Chrystinne:
Thanks, @bemoody!

bemoody:
Thanks!  But please don't make changes in unrelated files (physionet-django/console/templates/console/manage_published_project.html).  Please make that change in a separate pull request.

PR #2150, opened 2023-12-01
Unit tests for S3

bemoody:
This adds a couple of test cases for the functions used to manage S3 buckets (added in pull #2086).

It uses moto as a mock server, and tests that we can "upload" a couple of example projects, and the buckets and objects are created as we expect.

A limitation is that this doesn't test permissions (i.e. to check that the policy lets people download and list files.)  moto has some ability to emulate AWS IAM policies, but I'm not sure if this extends to S3 bucket policies.

I think it should also be possible to adapt this to run tests against the *live* system.  I'm not going to try that at the moment.


tompollard:
@bemoody looks good to me. Please could you regenerate poetry.lock to deal with the conflict?

tompollard:
A bunch of the tests failed on what looked like a network issue (now passing). The following error looks legit:
https://github.com/MIT-LCP/physionet-build/actions/runs/7062428887/job/19226810113?pr=2150

I assume this has something to do with the config file not existing in the current production branch? Do we ignore the test or is there something that can be done to fix it?

bemoody:
The config file exists in the branch, for sure... is it possibly not loaded in the container?  Or is boto3 somehow not picking up the overridden environment variables?

The S3 code currently will not work at all in GCP deployments (send_files_to_s3 uses os.walk, for example) - and ideally that should be fixed!  But I don't think the "Docker" test *does* test GCP deployment, so I don't think that's the issue.

bemoody:
The error here is not when `test_create_log_bucket` is called.  The error is when the test_s3 module is *loaded*, and specifically I guess when the `mock_s3` decorator is evaluated.

Why is moto trying to create a boto3 session at the time `mock_s3` is called?

Why does this not happen (or why does it not cause an error) when running in a normal environment?

bemoody:
It seems that, if there are no credentials defined (`~/.aws/credentials` etc.), and:

- if `os.environ['AWS_PROFILE']` is set to an empty string, then `boto3.Session()` raises the exception `The config profile () could not be found`

- if `os.environ['AWS_PROFILE']` is unset, then `boto3.Session()` succeeds (andreturns a session object that can't be used.)


So there are three broken things happening here:

(a) boto3 being weird and inconsistent

(b) moto creating a session when the decorator is evaluated (relying on environment variables at the time the module is imported, not at the time the test methods are called)

(c) docker-compose.test.yml copying settings from `.env.example` into the environment (bad idea)


We could work around this by commenting-out AWS_PROFILE in `.env.example`.  But (b) makes me uneasy, I do not want the test code to be dependent on whatever might happen to be in ~/.aws or $AWS_PROFILE.

bemoody:
(and of course another option is to actually delete stuff from os.environ when test_s3.py is imported, but that also seems like a lousy idea)

bemoody:
So one alternative that I think should work is not to use mock_s3 as a decorator, but to invoke it in setUp and tearDown like http://docs.getmoto.org/en/latest/docs/getting_started.html#unittest-usage

*sure would be nice if I could test this without waiting 6 minutes every time...*

Issue #2147, opened 2023-11-30
Default reference emails to lowercase when comparing to user's email address in credential applications


mscanlan-git:
Currently, users are able to use the same email address for both themselves and their reference as the current check is not case sensitive.

Without publicly revealing the case I discovered this, here's an example:

Personal: test@mit.edu
Reference: Test@mit.edu

The code that checks the email addresses is as followed:

```
    def clean_reference_email(self):
        reference_email = self.cleaned_data.get('reference_email')
        if reference_email:
            if reference_email in self.user.get_emails():
                raise forms.ValidationError("""You can not put yourself
                    as a reference.""")
            else:
                validate_institutional_email(reference_email)
                return reference_email.strip()
```

Instead of `if reference_email in self.user.get_emails():` we should do a comparison with .lower() to default both email addresses to lowercase then compare. This would allow us to prevent this bug from occurring.

tompollard:
Good catch!

PR #2152, opened 2023-12-01
Default reference emails to lowercase when comparing to user's email addresses

mscanlan-git:
This PR fixes the bug that was found in #2147. Since `self.user.get_emails()` is a list, we go through the list to check that the reference's email does not match any of the potential emails that the user has.


tompollard:
thanks @mscanlan-git, looks good to me!

Issue #2117, opened 2023-10-18
Editor/Reviewer identity not properly anonymized


lilehman:
Please anonymize the identity of the editor assigned to review projects. Instead, please provide editor@physionet.org as a contact.

tompollard:
The name is displayed in the submission history template for active projects, e.g.:

<img width="800" alt="Screenshot 2023-12-02 at 2 54 21 PM" src="https://github.com/MIT-LCP/physionet-build/assets/822601/ad8ff04f-cb8b-4e82-b2f0-71cd5a8b64b7">

The code that needs editing is:

https://github.com/MIT-LCP/physionet-build/blob/54e57709490bb12a4b31e205a0f72cc03c469045/physionet-django/project/templates/project/active_submission_timeline.html#L190-L197

bemoody:
Hiding name of the editor makes sense.  But I think it would be better to have a *different contact address for each project*.

- Ideally I'd want discussions to be archived and attached to the project - ideally visible on the site even if people prefer email as a medium.

- We may want to have "guest editors" who wouldn't want (and shouldn't) receive communications about *all* submitted projects, and on the other hand it would be awkward for somebody to have to read `editor@` and manually redirect messages to the appropriate person.

One idea would be to say:

    To contact the editor, email <editor+SHuKI1APLrwWCqxSQnSk@physionet.org>.

At present, this would do exactly the same as writing to `editor@physionet.org`.  But in the future it would be possible for us to archive such messages and/or send them to the appropriate people.

PR #2156, opened 2023-12-02
Conceal the name of the editor during project submission. Ref #2117.

tompollard:
@li-lcp has asked to conceal the name of the editor for projects (see #2117). 

This pull request removes the name of the editor from the submission history page, and add general contact information for the platform.

In #2117, Li-wei asks for an "editor@physionet.org" email address to be listed. This address is not recorded in the config file, so for now I am using the general CONTACT_EMAIL.

Example:

Screenshot 2023-12-02 at 3 25 55 PM

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,6 +23,10 @@ CONTACT_EMAIL='PhysioNet Contact <contact@dev.physionet.org>'
 SERVER_EMAIL='PhysioNet System <root@dev.physionet.org>'
 ERROR_EMAIL='contact@physionet.org'
 
+# Contact address for project editors. This address may be viewable by authors.
+# Optionally, add "PROJECT-SLUG" to include the project slug.
+PROJECT_EDITOR_EMAIL='editor+PROJECT-SLUG@dev.physionet.org'
+
 # Admins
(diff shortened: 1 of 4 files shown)


bemoody:
I think we really don't want to use `contact@` and this behavior should be configurable.  Could be as simple as:
```
    if settings.PROJECT_EDITOR_EMAIL:
        contact_email = settings.PROJECT_EDITOR_EMAIL
    else:
        contact_email = project.editor.email
```

But we could also allow for a project-specific contact address (see my comment in #2117):
```
    if settings.PROJECT_EDITOR_EMAIL:
        contact_email = settings.PROJECT_EDITOR_EMAIL.replace('PROJECT-SLUG',
                                                              project.slug)
    else:
        contact_email = project.editor.email
```

bemoody:
Thanks!  The setting needs to be defined in physionet-django/physionet/settings/base.py (and it shouldn't default to physionet.org).  I don't think there is any need to define it separately in development/base.py, production.py, settings.py, and staging.py.

bemoody:
looks good!

Issue #2164, opened 2023-12-12
Renaming TrainingType to Course & Add Versioning to it.


Rutvikrj26:
What : 
This issue discusses the refactoring of TrainingType model to be renamed to course and add a version field to it.

Why :
The upcoming training PRs require versioning for TrainingType. The current setup implements a course module in the trainings app, which basically has a foreign key and a version. This leads to unnecessary complications and redundant tables. 
The proposed changes would eliminate an extra table and simplify the implementation logic. 

@kshalot @bemoody Please provide your views

PR #2165, opened 2023-12-13
Refactoring of TrainingType model to Course for better workflow in user Trainings

Rutvikrj26:
The PR is based on the issue #2164. 
This PR addresses the concerns over the trainings workflow in #1950 simplifying the implementation for training module. 

Changes : 
- Rename of the TrainingType model to Course in User Models
- Addition of two more fields to Course (required by the new trainings module)
- Respective refactoring of TrainingType at all the places it is referenced.

I have tested some simple trainings workflow, and they seem to work without breaking anything - however some tests are still breaking (Fixing in Progress). Would like for further help in testing the changes.

@tompollard @kshalot @bemoody Please review and provide your comments.


tompollard:
Thank Rutvik! Please could you:

- provide some justification for renaming `TrainingType`? what is the purpose of TrainingType, and why is `Course` a better name? 
- add a short docstring for the model, explaining its purpose?

Rutvikrj26:
Thanks @tompollard!

Currently the trainings are taken off-platform and only the completion document is added. Training PR #1950  tries to add a workflow where users can take courses on the platform. 

`TrainingType` model is a type of course/certification, which can be taken by users. For implementation in #1950 , there's essentially a shell table created named `course`, which is a foreign key to `TrainingType`, a version field, and is_active Flag. `TrainingType` itself can be updated to have the version and the is_active flag, which removes the requirement for an extra table. 

The word `TrainingType` is also semantically confusing, and hence, the requirement to change it to `Course`, which follows the regular conventions. 
A user takes a course and gets training. This course can have multiple versions, can be taken on a platform, or submitted via a PDF as required.

@tompollard Please add if I'm missing something from our discussion.

tompollard:
@Rutvikrj26 please could you rebase this on dev?

bemoody:
> For implementation in #1950, there's essentially a shell table created named course, which is a foreign key to TrainingType, a version field, and is_active Flag.

Okay, I have not been following #1950.  But my gut feeling is this is wrong.

What we currently call `TrainingType` is not at all the same thing as what 1950 calls `Course`.

For on-platform courses, we need a model that represents "a course that is available on the platform for people to complete" (with modules, and quizzes, and stuff.)

But a TrainingType is not that.  A TrainingType is a different beast - it's the representation of "a requirement that must be fulfilled before accessing a particular project" (and the object has associated fields regarding how we validate that the requirement was fulfilled.)

In fact, a TrainingType doesn't refer to only one particular thing: there are several different CITI courses which you might have taken, and the courses have been revised over the years, but PhysioNet generally accepts any of them as long as they were taken recently and include the required modules.

Likewise, when and if we start using on-platform courses, a single TrainingType will need to include *multiple* alternative Courses (for example, either "Human Subjects Training 1.0" or "Human Subjects Training 1.1" would satisfy the *same requirement* as far as accessing a particular project is concerned.)

I'm not seeing that it's necessary for a TrainingType to have a version, and it definitely isn't the same version as the Course version.

Nor does it make sense to link "whether or not a Course is active" (whether or not new users may enroll and complete the course) with "whether or not a TrainingType is active" (I'm not even sure what that would mean.)

Rutvikrj26:
Thanks @bemoody 

I see your point. We looked at `TraininingType` from an implementation perspective - thanks a lot for adding the usage perspective. It makes complete sense as to why we would want to avoid having versions in `TrainingType`, and have the course as a different table along with its own versioning that allows the addition, update, and expiry of different course versions. I'll be adding this information to the `TrainingType` docstrings to make the model's purpose and usage more clear.

@tompollard please let me know your thoughts.

tompollard:
@Rutvikrj26, Benjamin's explanation makes sense to me.

@bemoody so your suggestion is to keep TrainingType as it is, and close this PR?

tompollard:
@Rutvikrj26 should this be closed?

Rutvikrj26:
> should this be closed?

Yup, as per our discussion, this is no longer required and we can close this PR.

Issue #2162, opened 2023-12-08
Fix styling issues/improve event management page UI 


mscanlan-git:
Minor issue, nothing that really needs to be fixed immediately.

The description for an event shows the html syntax. I also think we could improve how the participants/application status is shown. Having the value -large gap- 'view' button is a bit jarring. 

![image](https://github.com/MIT-LCP/physionet-build/assets/67070862/c11950b7-77eb-4b5b-b849-bd3f8e51de7c)

PR #2167, opened 2023-12-15
Update 'manage' page in event management UI

mscanlan-git:
This PR updates the syntax errors that allows html to show in the description of an event on the event management page. 

Also mentioned in #2162, the way we present event participants is a bit off, I decided to merge the `{{ info.count }}` into the button itself, so it cleans up the page a bit. Open to other ways to present the information, but this looks a bit cleaner than before.

### DIFF
--- a/physionet-django/console/templates/console/event_management.html
+++ b/physionet-django/console/templates/console/event_management.html
@@ -56,7 +56,7 @@
 
       <div class="row mb-1">
         <div class="col-md-3">Description:</div>
-        <div class="col-md-9">{{ event.description|safe }}</div>
+        <div class="col-md-9">{{ event.description|escape }}</div>
       </div>
       
     </div>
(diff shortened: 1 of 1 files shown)


bemoody:
If you want to display the description as "plain text", and just ignore HTML formatting, then you could use `{{ event.description|striptags }}`.

If the description contains complicated markup (paragraphs, lists, subheadings, etc) then trying to remove the formatting will probably give ugly results no matter how we go about it.

Just some food for thought.

mscanlan-git:
Thanks @bemoody, I think `{{ event.description|striptags }}` might be the solution here. The issue seems to be difficult to replicate on my local server, but on the live server this is what is seen when clicking 'manage' on an event:

image

If striptags removes the html that's leaking on the page, that would be great. I did notice that this isn't an issue from the user's end, this is mainly for event hosts or admins with event access permissions that would see the html issue (no event participant would see this page).

PR #2169, opened 2023-12-21
Refactor console_navbar for maintainability

bemoody:
`console/templates/console/console_navbar.html` defines the structure of the navigation sidebar/popup for the Admin Console.

Currently this file contains a huge amount of boilerplate code, some of which is useless, some of which is buggy; it's full of logic that needs to be kept in sync with views.py, and that can't easily be tested by the test suite.  The UI design is good, but the complicated implementation makes it needlessly difficult to add new features to the console.

With a bit of magic, this pull request removes the logic from the template and defines the menu structure in Python instead - replacing 300-ish lines of messy template code with 70-ish lines of Python (see the bottom of `console/navbar.py`).

This fixes some small bugs (styling of the stats menu, noscript accessibility) but shouldn't significantly change the appearance or functionality.

A few pages are now longer treated as "children" of their "parent" page, because their URLs don't match.  I intend to fix this by changing the URLs, but I would like to do that in a separate pull request.

### DIFF
--- a/physionet-django/console/navbar.py
+++ b/physionet-django/console/navbar.py
@@ -1,0 +1,214 @@
+import functools
+
+from django.conf import settings
+from django.urls import get_resolver, reverse
+from django.utils.translation import gettext_lazy as _
+
+from physionet.settings.base import StorageTypes
+
+
+class NavLink:
+    """
+    A link to be displayed in the navigation menu.
+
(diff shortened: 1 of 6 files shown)


tompollard:
Full of surprises @bemoody. Big improvement!

Issue #2166, opened 2023-12-15
Migrate existing "Archived Project" to the ActiveProject model


tompollard:
In https://github.com/MIT-LCP/physionet-build/pull/2149, we essentially deprecated the `ArchivedProject` model. Before we can remove the ArchivedProject model, we will need to migrate existing Archived Projects to the ArchivedProject model.

I think this work probably requires:

- Migrating the database content in a database migration
- Moving files that currently exist in the archivedproject folder

Once the ArchivedProject model is removed, we can:

- Replace GenericForeignKeys on Reference, Author, etc with foreign keys
- Clean up awkward logic in our codebase that exists because of legacy GenericForeignKeys
- Look at whether removing GenericForeignKey helps us resolve current issues.

Other related issues (that I'll document here for now, but probably need their own issue):

- We need to do a better job at logging the status changes of projects
- We probably want to go through all submitted projects and set the status of many of them to "archived"
- We should implement automated archiving of inactive projects (with emails to authors explaining why the project was archived)
- Decide on a clear set of rules for when files for a project should be deleted.

PR #2170, opened 2023-12-22
Migrate ArchivedProject content to ActiveProject

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/2166, we would like to migrate content from the (now deprecated) `ArchivedProject` model to the `ActiveProject` model.

This pull request:

1) Adds an Author to the "Failed demo software for parsing clinical notes" (`slug="t2ASGLbIBoWaTJvPrM2A"`) fixture.
2) Creates a migration that migrates ArchivedProjects (and associated objects, like authors, references, logs, etc) to ActiveProjects with SubmissionStatus=ARCHIVED.

I have tested this fairly comprehensively and have not found any issues. 

Note that files are **not** migrated. Files for archived projects will remain in the `pn-media/archived-projects` subfolder. We should migrate them to `pn-media/active-projects` later. 

One way of testing is to:

1 . Reset to `c49b2a1c640a0a6c07ee135f3ff84d45e2baa4e3` (e.g. `with git reset c49b2a1c640a0a6c07ee135f3ff84d45e2baa4e3 --hard`) which is the pull request prior to the archived status being added to the ActiveProject object.
2. Create a bunch of interesting ArchivedProjects using the console. Add references, co-authors, copyediting history, etc.
3. Pull down the latest version of this branch.
4. Run the migrations to migrate ArchivedProjects to ActiveProjects.
5. Try viewing archived projects in the console, author project page, etc.
6. Change the submission status of the archived projects to unsubmitted (`p.submission_status = 0`) and save the change.
7. Check whether the now-active projects are correctly populated in the author submission system. You should find that all pages work correctly, except `http://localhost:8000/projects/SLUG/files/`, which will raise a `FileNotFoundError` as expected.

### DIFF
--- a/physionet-django/project/fixtures/demo-project.json
+++ b/physionet-django/project/fixtures/demo-project.json
@@ -340,6 +340,26 @@
   }
 },
 {
+  "model": "project.author",
+  "pk": 14,
+  "fields": {
+    "user": [
+      "admin"
+    ],
+    "display_order": 1,
+    "is_submitting": true,
+    "is_corresponding": true,
+    "approval_datetime": null,
+    "content_type": [
+      "project",
+      "activeproject"
+    ],
+    "object_id": 9,
+    "corresponding_email": 209,
+    "creation_date": "2020-04-10T16:08:49.337Z"
+  }
+},
+{
   "model": "project.publishedauthor",
   "pk": 1,
   "fields": {
--- a/physionet-django/project/migrations/0074_migrate_archived_to_active.py
+++ b/physionet-django/project/migrations/0074_migrate_archived_to_active.py
@@ -1,0 +1,69 @@
+from django.db import migrations
+from django.contrib.contenttypes.models import ContentType
+from project.models import SubmissionStatus
+
+
+def migrate_archived_to_active(apps, schema_editor):
+    """
+    Copy all ArchivedProject data to the ActiveProject model.
+    """
+    AnonymousAccess = apps.get_model("project", "AnonymousAccess")
+    ActiveProject = apps.get_model("project", "ActiveProject")
+    ArchivedProject = apps.get_model("project", "ArchivedProject")
+    Author = apps.get_model("project", "Author")
+    CopyeditLog = apps.get_model("project", "CopyeditLog")
+    EditLog = apps.get_model("project", "EditLog")
+    Log = apps.get_model("project", "Log")
+    Publication = apps.get_model("project", "Publication")
+    Reference = apps.get_model("project", "Reference")
+    Topic = apps.get_model("project", "Topic")
+    UploadedDocument = apps.get_model("project", "UploadedDocument")
+
+    # Get content types for both models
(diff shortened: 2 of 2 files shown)


bemoody:
Makes sense.  Thanks.

I think that we want to delete the ArchivedProject after migrating it.

tompollard:
> I think that we want to delete the ArchivedProject after migrating it.

Thanks Benjamin. Deleted!

bemoody:
Interesting error.

 call_command('purgeaccounts')
  -> user.delete()
  -> django.db.utils.OperationalError: no such table: project_archivedproject

Perhaps because of the `editor` field on ArchivedProject?

But I didn't mean deleting the whole table necessarily.

I meant, within your `migrate_archived_to_active`, after you've saved the new objects, then delete the old object.  So it all happens as a single transaction.

tompollard:
> I meant, within your migrate_archived_to_active, after you've saved the new objects, then delete the old object. So it all happens as a single transaction.

Sorry, misunderstood! Now done. I'll remove the table in a later pull request in this case.

bemoody:
I'm not sure if there is a reason for the particular set of fields being copied from ArchivedProject to ActiveProject (in line 28 onwards.)  As it is, this is losing some information from the old project (e.g. `installation`, `acknowledgements`, `release_notes`.)

It's instructive to look at the old archive() function which converted ActiveProjects into ArchivedProjects.  It did this:
```
        archived_project = ArchivedProject(archive_reason=archive_reason,
            slug=self.slug)

        modified_datetime = self.modified_datetime

        # Direct copy over fields
        for attr in [f.name for f in Metadata._meta.fields] + [f.name for f in SubmissionInfo._meta.fields]:
            setattr(archived_project, attr, getattr(self, attr))
```
In other words, it would copy all of the fields in Metadata or SubmissionInfo, plus `slug` and `modified_datetime`.

The only fields that weren't copied were `is_new_version`, `latest_reminder`, `doi`, and `submission_status`.  The first two don't seem important.  Missing `doi` seems like a bug but probably never came up.  `submission_status` doesn't exist for ArchivedProject.

Unfortunately we can't use the same approach in `migrate_archived_to_active` because the migration system doesn't know about abstract classes like Metadata and SubmissionInfo.  But I think the following would be good enough:
```
    for archived_project in ArchivedProject.objects.all():
        active_project = ActiveProject(
            submission_status=SubmissionStatus.ARCHIVED.value,
        )
        for attr in [f.name for f in ArchivedProject._meta.fields]:
            if attr != 'id':
                setattr(active_project, attr, getattr(archived_project, attr))
        active_project.save()
        ...
```

tompollard:
> I'm not sure if there is a reason for the particular set of fields being copied from ArchivedProject to ActiveProject (in line 28 onwards.)

I'd walked through the fields, picking ones that I thought were important to keep. I have now implemented your more-complete approach. Good suggestion, thanks.

bemoody:
thanks, looks great!

PR #2171, opened 2023-12-22
Remove the code introduced in PR #1987

Chrystinne:
This PR removes the code introduced in PR #1987, which provided a temporary solution for a Datathon in Spain—specifically, a JSON file of AWS user accounts available for download by administrators.


bemoody:
Looks good, thanks!

PR #2173, opened 2024-01-04
Updates to list of programming languages

bemoody:
We've had a request to add AWK as a programming language, and I also noticed that Fortran wasn't listed.

I've added these on PhysioNet, via the admin interface, but I think it makes sense to add them to the default site-data as well.


tompollard:
thanks!

PR #2174, opened 2024-01-08
Display credentialing status on event management page. Fixes #2163.

### LINKED ISSUE #2163: Credential column on event management page does not display a check mark for credentialed users
The `credential` column on the event management page does not properly display whether or not a user is credentialed. After discussing with an event host, we discovered that a few of the credentialed users who were event participants did not show up as credentialed. 

Here is a screenshot of the table in question (removed names & emails):
image

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/2163, the "credentialed" column on the event management page should display whether or not an event participant is credentialed.

As a result of a bug in the template, the credentialing status is not displayed. This pull request is a minor update to fix the bug. I also remove a couple of unused columns at the same time (e.g. "co-host").

### DIFF
--- a/physionet-django/events/templates/events/event_applications.html
+++ b/physionet-django/events/templates/events/event_applications.html
@@ -13,15 +13,15 @@
             </tr>
           </thead>
           <tbody>
-              {% for object in info.objects %}
+              {% for application in info.objects %}
                 <tr>
-                  <td>{{ object.user.username }}</td>
-                  <td>{{ object.user.get_full_name }}</td>
-                  <td>{{ object.user.email }}</td>
-                  <td>{{ object.comment_to_applicant }}</td>
-                  <td>{{ object.decision_datetime | date }}</td>
+                  <td>{{ application.user.username }}</td>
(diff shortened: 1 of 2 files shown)


tompollard:
@mscanlan-git please could you test this and merge if you are happy? To test:

1. Run `git pull` to pull down the branch
2. Check out the branch (`git switch tp/event_credentialed`)
3. Run the local server as usual. Check that the credentialing status is now displayed correctly (`rgmark` and `aewj` are credentialed users).

We need to think more about whether displaying the credentialing status is enough. For example, should we also display: 

1. training status?
2. details on which DUAs have been signed?

I think we should raise a new issue to discuss these points (and address them in a later pull request).

mscanlan-git:
Thanks @tompollard, this is working as intended. 

I do think we should have a column for training status and DUAs, as these are two additional requirements that users will need to complete. Having all 3 columns available to the event host will make it easier to pinpoint why a user is unable to access the datasets for that event. I will create an issue to note.

tompollard:
Thanks Michael! I have pushed the fix to the live server.

PR #2176, opened 2024-01-10
Cloud mirror overview page

bemoody:
This adds a page to the console which gives an overview of the active *cloud mirrors* (Google Cloud Storage and Amazon S3).

Currently this is just a table so that we can easily see which projects are uploaded and which aren't.  Since this stuff has to be managed by hand I'd like to have an easy way to see the status of all projects (and grouped by open vs restricted.)  Eventually it might be nice to add management functions to this page as well.

In order to view this page, you will need to have the `project.change_publishedproject` permission (the same permission as for manage_published_project.)  I think that there should be a separate permission for managing mirrors, but this will work for now.

The implementation is more fragile than I'd like, in a couple of ways.

- I want to retrieve the GCP/AWS information in *one SQL query*, and furthermore want to select only the relevant fields.  Using only() and select_related() lets us do that (and you can verify for yourself: this code is only making a single SQL query).  But this could easily be broken by code changes elsewhere, and suddenly this page would become much slower to load.

- I don't just want to display whether or not file access is allowed, I want to indicate the reason (deprecated, forbidden, embargoed).  There's no way to do this via `project.authorization.access` and I can't immediately think of a clean way to do so.

That said, I think this code works and should be useful.

### DIFF
--- a/physionet-django/console/navbar.py
+++ b/physionet-django/console/navbar.py
@@ -154,6 +154,10 @@ CONSOLE_NAV_MENU = NavMenu([
 
     NavLink(_('Storage'), 'storage_requests', 'cube'),
 
+    NavSubmenu(_('Cloud'), 'cloud', 'cloud', [
+        NavLink(_('Mirrors'), 'cloud_mirrors'),
+    ]),
+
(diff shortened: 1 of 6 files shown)


tompollard:
nice, thanks benjamin!

PR #2178, opened 2024-01-16
Handle missing contact address in project_submission

bemoody:
Pull #2156 introduced the PROJECT_EDITOR_EMAIL setting.

If this setting is not defined (or blank) then we want to fall back to the old behavior of showing the assigned editor's email address.

However, if the project has not yet been assigned to an editor, and PROJECT_EDITOR_EMAIL is blank, it would crash.

(For example, clear the PROJECT_EDITOR_EMAIL setting, then log in as rgmark and try to submit the "MIMIC-III Clinical Database" project.)


tompollard:
thanks for catching this

PR #2181, opened 2024-01-18
Update CITI Training instructions to include additional step regarding training report upload

mscanlan-git:
There is no issue associated with this PR, however, I decided to add an additional step to ask users to upload the training report rather than the certificate (hopefully to curb users from uploading the certificate). I included a sample report with sensitive info removed as reference.

I realized in the original instructions it wasn't clear which file the user should upload (it is stated on the settings/training page, but I figured it would cause no harm to reiterate it in the instructions).


bemoody:
This looks good.  It might be better if we could show a screenshot of the CITI web page that contains the links to download your report and certificate.

mscanlan-git:
Thanks for your feedback @bemoody, I've updated the screenshots to provide more context on how to obtain the reports and download them.

bemoody:
Looks great, thanks!

Issue #2179, opened 2024-01-17
Prevent users from uploading password-locked pdf files when submitting their CITI training report


mscanlan-git:
This hasn't really been a problem in the past, but lately I've noticed an uptick in the amount of training reports being submitted with this problem. If a user uploads a training report that is password locked, it looks like the page will throw a 500 error because it cannot load the pdf. At the moment, it is not possible to reject these applications as accessing the page isn't possible. I think either of these two solutions would be best:

1. Prevent users from uploading any pdf file that has a password requirement on it
2. Allow admins to view the training application, but have the pdf renderer hidden (this way, an admin can  reject & comment that the pdf file is password locked and should be re-uploaded without the password requirement).

Option 1 is likely the easiest option.

bemoody:
It seems very strange that somebody would do that intentionally.  I assume that this was not a genuine PDF file from CITI, but maybe somebody downloaded the CITI PDF and edited it before uploading to PhysioNet?

In the past I had proposed that we require people to upload only the genuine PDF file from CITI (which contains a verification code that we can parse), but I think some folks thought this would make things too complicated for applicants.  However, that was all prior to (a) the identity/training split, and (b) the CITI API support.

So one idea: have a single page that *either* lets people upload a PDF (and requires it to be a genuine CITI PDF), *or* lets people copy and paste the verification URL (so we can then get the completion report from CITI directly.)

Second idea is to add a form saying "if you have taken the CITI course as an MIT affiliate, select your (verified) institutional email here" (and then we can get the completion report through the CITI API.)

bemoody:
Okay, this looks like a possibly serious problem.

The three reports that triggered errors yesterday (that I noticed) all look like they belong to real people (two of them have already passed identity check, the third is pending.)  These three PDFs are indeed encrypted so I can't read them.

I just checked the 10 most recent submissions in /data/pn-media/training/.  All of them are using PDF "password protection" (i.e., not really encrypted but have the please-don't-edit-me flag), which also prevents our old method of extracting the verification code.  Some of these are completion reports while others are certificates, but all appear at first glance to be genuine.

bemoody:
Before I delve too deeply into this issue, how do people currently feel about the following options:

1. **Require every applicant to upload the original PDF file from CITI**, i.e. if the file has been modified in any way their submission is rejected.

[...]

PR #2182, opened 2024-01-19
Don't crash on encountering an encrypted PDF file

bemoody:
Some people have submitted PDF training reports to PhysioNet that are encrypted with a password.  I still don't know *why*, but at least three people did so.

We can't read them, so presumably these submissions ought to be rejected.  However, we currently aren't preventing these submissions, and trying to access the console page to view/reject the submission causes an error (PDFPasswordIncorrect, which is not a subclass of PDFSyntaxError).

This pull should treat encrypted PDF files the same as any other file that the server is unable to parse.

For more long-term solutions see issue #2179.

### DIFF
--- a/physionet-django/console/services.py
+++ b/physionet-django/console/services.py
@@ -4,7 +4,7 @@ import re
 from typing import Optional
 
 from pdfminer.high_level import extract_text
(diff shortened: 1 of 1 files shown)


tompollard:
(i have pushed to the live server)

PR #2186, opened 2024-01-31
Allow news items to be linked to all versions of a project

tompollard:
Currently, news items can only be linked to a single version of a published project, but in many or most cases we would actually like news items to link to all versions of a published project.

This pull request adds a flag to the news model that allows a news item to be linked to all versions of a project. Specifically, the changes are:

- Add a `link_all_versions` flag to the News model
- Allow the flag to be set in the form that is used to create news items
- Adds a `get_all_news()` method to the `PublishedProject` model that makes it straightforward to retrieve all news items for the project (both news items that are directly linked, and news items that are linked through the `link_all_versions` flag).
- Adds a migration that updates the News model. 

There may be better ways to implement this functionality, e.g. by:

- linking the News model to a CoreProject instead of PublishedProject
- setting a validator that requires linked CoreProjects to be associated with a PublishedProject (so that news items can only be added for published projects)
- adding `project_version` to the News item
- updating the form used to add news items so that the CoreProject and version (or "all versions" can be set).

I would prefer to stick with the implementation in this pull request for timeliness (so that @briangow can respond to a user request). We can review and improve the approach in future.


briangow (approved):

tompollard:
Thanks Brian

tompollard:
@briangow This is now live. I've set `link_all_versions` to `True` for the 2021 announcements.

```
p = PublishedProject.objects.get(...)
for n in p.news.all():
    n.link_all_versions = True
    n.save()
```

PR #2188, opened 2024-02-08
Relative image URLs

bemoody:
The link filter would transform:
```
<img src="https://physionet.org/static/images/background.jpg">
```
into:
```
<img src="/static/images/background.jpg">
```

But it would transform:
```
<img src="/static/images/background.jpg">
```
into:
```
<img>
```

which clearly is not what we want.

(~Amusingly~, it looks like ckeditor or browser (mis)feature would subsequently replace the sourceless image with `&nbsp;`.)

Y'all, I thought this feature was working years ago.  I'm sorry it wasn't.  People may have told me that it was broken and I didn't listen because it didn't come in the form of a bug report.

### DIFF
--- a/physionet-django/project/test_views.py
+++ b/physionet-django/project/test_views.py
@@ -520,7 +520,10 @@ class TestProjectEditing(TestCase):
         <mfrac><mn>22</mn><mn>7</mn></mfrac></math>,
         some ambiguous characters & < >,
         <form>invalid tags</form>,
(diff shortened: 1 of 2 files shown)


tompollard:
(pushed to the live server)

tompollard:
Side note, but if we are going to support images in descriptions then we should document the preferred approach for adding these images somewhere (either internally, or perhaps at: https://physionet.org/about/publish/). @li-lcp

PR #2192, opened 2024-02-12
Allow Hosts to Add Cohost (Re-Implemented)

Rutvikrj26:
This Pull Request implements the feature to add cohosts for an event. 

This PR is a continuation of the Cohost PR : #1925 

The feature has been implemented in a similar fashion to how the author Invitation is handled currently. 

Cohosts are supposed to be helper for hosts and can manage the participants(approving, rejecting participants)
To be cohost, one has to be first approved to join the event.

Worflow:

When a host opens the event page, the host will be able to see the option to invite a user as a cohost. 
event-cohost-form

The person invited gets an email for cohost Invitation:
cohost-invitation-email

The person sees the invitation on the specific event's page. 
cohost-request
cohost-request-accept

When the person accepts, the host gets an email that the person has accepted the cohost invitation.

The form runs the check if the user is a participant in the event and prohibits inviting a random user as the last one implemented.


Rutvikrj26:
Heyy @tompollard

This is the re-implemented version for inviting cohosts, following the implementation of author invitations in projects. 
I've made some design decisions as to where the invitation is sent from, where it shows up, and how the implementation allows only the host to invite, and a participant to accept. 
Please review and let me know your thoughts!

physionet-django/events/models.py
```diff
@@ -7,6 +7,7 @@
 from events.enums import EventCategory
 from events import validators
 from project.modelcomponents.fields import SafeHTMLField
+from project.modelcomponents.generic import BaseInvitation
```
tompollard: I don't think this import is needed (also, I'm not sure that having a dependency on the project app is a good idea...perhaps the reason that it was not eventually used?)

physionet-django/events/views.py
```diff
@@ -7,12 +7,17 @@
...
 from django.forms import modelformset_factory
 from django.urls import reverse
+from django.db import transaction
+from django.utils import timezone
 
 
 import notification.utility as notification
-from events.forms import AddEventForm, EventApplicationResponseForm
-from events.models import Event, EventApplication, EventParticipant
+from project import utility
```
tompollard: For consistency with the other imports (and for clarity with all of the `utility` modules), I think it would be better to do: `from project.utility get_form_errors`.

physionet-django/notification/templates/notification/email/invite_cohost.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n %}{% autoescape off %}{% filter wordwrap:70 %}
+{{ event_info }}
+
+Dear Colleague,
+
+You have been invited by {{ inviter_name }} ({{ inviter_email }}) to co-author an event entitled "{{ event.title }}" on {{ domain }}. Please accept or decline this request as appropriate.
```
tompollard: "co-author" an event isn't right. `co-host`?

physionet-django/notification/utility.py
```diff
@@ -77,6 +77,19 @@ def email_project_info(project):
 
     return header
 
+
+def email_event_info(event):
```
tompollard, line 81: It would be helpful to be a bit more specific here. "Header for the email" -> Which email?

tompollard (commented):
Thanks @Rutvikrj26 I prefer this to the previous implementation. I added a few comments inline. 

The "invite co-hosts" button and the "outstanding co-host invitations" content are hidden away in their current location.

Is there a reason why you didn't put them at http://localhost:8000/events/ (following a similar layout to http://localhost:8000/projects/)?

physionet-django/events/models.py
Rutvikrj26: Yup, Exactly!

physionet-django/notification/utility.py
Rutvikrj26: Added `emails sending out events related information`

Rutvikrj26:
> The "invite co-hosts" button and the "outstanding co-host invitations" content are hidden away in their current location.

That is a design decision I made to abstract out the related event invitations inside. 
I can have them on the general page as well. Since the events are a bit different from projects in nature, it made more sense that the person is getting any pending requests particular to that event inside the event-specific page. 

@tompollard let me know your thoughts. We can change it if you feel it is better to have them on the generic events page.

tompollard:
> The "invite co-hosts" button and the "outstanding co-host invitations" content are hidden away in their current location.

>> That is a design decision I made to abstract out the related event invitations inside.
I can have them on the general page as well. Since the events are a bit different from projects in nature, it made more sense that the person is getting any pending requests particular to that event inside the event-specific page.

Thanks for explaining. Even for projects, outstanding tasks displayed at /projects refer to a specific project, so I don't think what we are doing with events are much different? My preference is to display the requests on the /events page (in the same way as projects), making it clear which event the task relates to.

Rutvikrj26:
> My preference is to display the requests on the /events page

Hey @tompollard , thanks for adding your thoughts. I've switched the cohost invitation to show up on the event home page rather than the individual event page. 

image

bemoody:
Looks good, thanks!

Issue #2196, opened 2024-02-16
Upgrade to Django 5


tompollard:
Django 4 entered long term support in 2024:
https://www.djangoproject.com/download/

At some point, we should upgrade to Django 5. 

From the release details, it doesn't sound like upgrading should be too much effort:
https://docs.djangoproject.com/en/5.0/releases/5.0/

PR #2197, opened 2024-02-17
[WIP] Bump minimum version of Python to 3.10

tompollard:
This pull request bumps the minimum supported version of Python to 3.10, in preparation for upgrading to Django 5 (see: https://github.com/MIT-LCP/physionet-build/issues/2196). Alongside bumping the version here, we'll need to make sure that our servers are running Python 3.10 or greater.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9-bullseye
+FROM python:3.10-bullseye
 
 RUN apt-get update -y \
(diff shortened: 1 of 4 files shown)


bemoody:
When we upgrade to bookworm we will be using python 3.11.  So I do not forsee suporting python 3.10 and not supporting 3.9.

*After* we have upgraded to bookworm we can think about increasing the minimum suported version to 3.11.

PR #2199, opened 2024-02-26
[WIP] Allowing users to download project-related zip files from AWS S3 buckets

Chrystinne:
This PR generates presigned URLs for AWS S3 objects, allowing users to download project-related zip files from AWS S3 buckets.


bemoody:
To be honest, there are quite a number of problems with this code.

Please start by explaining the goal: what *new features* are you adding? How do they benefit users?

Then explain what changes are needed in order to achieve those goals.

PR #2204, opened 2024-03-13
Restrcting the AccessPolicy options allowed during the project setup

Rutvikrj26:
Why this is required:

- HDN does not support `Open` as one of the access policies, and hence ideally, it should not show up as one of the options allowed for the user to choose during the set-up of the project. 

What this PR does:

- Since the AccessPolicy names are strictly typed at a lot of places, and the access is based on the Integer comparisons, it is hard to remove the OPEN option as a part of the AccessPolicy class. 
- This PR modifies the AccessMetadataForm to restrict the choices available based on the environment variable `ALLOWED_ACCESS_POLICIES`, which has a default value set as the 4 available options, if the environment variable is not specified.


bemoody:
Use `config`, don't use `os.environ`.

This setting kind of goes together with `ENABLE_FILE_DOWNLOADS_OPTION`; I would list these two settings in the same place and perhaps name the new setting `ENABLED_ACCESS_POLICIES` (though "allowed" is also fine.)

I think it'd be better not to list the possible values in the physionet/settings/base.py.

Also the 'replace("_", " ").title()' (duplicating the logic of `AccessPolicy.choices`) doesn't really belong in the form class.

So it could be something more like:
```
    if settings.ALLOWED_ACCESS_POLICIES:
        self.fields['access_policy'].choices = [
            (value, label) for (value, label) in AccessPolicy.choices()
            if AccessPolicy(value).name in settings.ALLOWED_ACCESS_POLICIES
        ]
```

Or maybe it would be cleaner to define a method `AccessPolicy.allowed_choices` instead.  (I don't think we want to change the behavior of `AccessPolicy.choices` because those choices are baked into migrations.)

Rutvikrj26:
Heyy @bemoody 
Thanks for reviewing!

I've updated the code based on your suggestions except one pending change which I would like to understand the reason better before implementing.



> I think it'd be better not to list the possible values in the physionet/settings/base.py.

Wouldn't it be better to have that as a default value? The though was whichever deployment of Physionet wants to configure this setting, can just choose to add the environment variable, making it optional. 
Let me know your thoughts.

physionet-django/physionet/settings/base.py
```diff
@@ -644,3 +644,8 @@ class StorageTypes:
 
 # Emails
 PROJECT_EDITOR_EMAIL = config('PROJECT_EDITOR_EMAIL', default='')
+
+ALLOWED_ACCESS_POLICIES = config(
```
tompollard, line 648: We could consider:
- stripping white space from the list
- converting items to all caps

tompollard: We should also either:

tompollard:
We should also either: (1) Work out where we are going to document configurable settings or (2) Add the `ALLOWED_ACCESS_POLICIES` variable to the .env.example file.

tompollard:
Thanks @Rutvikrj26. I added a couple of suggestions above. Please address them in a new PR if necessary!

Issue #1801, opened 2023-01-11
Do we use django-cron at all?


alistairewj:
Just curious - I think we had planned to remove it, and Tom commented out its use in the settings:

https://github.com/MIT-LCP/physionet-build/blob/95aef7e30b503f5dad298ae037c9f0bad30b1855/physionet-django/physionet/settings/base.py#L35-L51

tompollard:
Good question, and I feel bad about the [PR where I introduced this change](https://github.com/MIT-LCP/physionet-build/pull/397/files). I have no memory of doing this. 

I don't see django-cron in requirements.txt, but I do see a `from django_cron` import in https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/cron.py.

Needs some investigation...sorry!

PR #2207, opened 2024-03-14
Add Django Q2 for task management and scheduling

tompollard:
## Why switch to Django Q2

We have a need to be able to:

1. Move tasks to the background, e.g. when uploading files to cloud and sending emails.
2. Run scheduled tasks, e.g. scrubbing spam accounts, logging metrics from services like CrossRef, DataCite, Google Scholar, etc.

Right now, we have:

- [Django Cron tasks](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/cron.py) that don't appear to run because [Django Cron is not installed](https://github.com/MIT-LCP/physionet-build/issues/1801).
- [Django Background Tasks](https://pypi.org/project/django-background-tasks/) which appears to have been [last updated 5 years ago](https://github.com/arteria/django-background-tasks/commits/master/)!

Django Q2 is [well documented](https://django-q2.readthedocs.io/en/master/index.html) and [under active development](https://github.com/django-q2/django-q2/commits/master/). It is compatible with Django 5 (which [I hope we can move to soon-ish](https://github.com/MIT-LCP/physionet-build/issues/2196)).

## What's in this pull request

This pull request:

1. Adds the Django Q2 dependency
2. Adds configuration to Django settings.
3. Sets the backend to the Django ORM (rather than external tools like Redis/Celery), for simplicity.
4. Provides instructions for running the Django Q2 server.

After the dependency is added, I would like to: 

- migrate all background-tasks and cron tasks to Django Q2.
- remove the dependency on the other django tasks management packages.


bemoody:
This all looks good and django-q2 sounds like a good way to go; thanks for find it.

> Django Cron tasks that don't appear to run because Django Cron is not installed.

Just to be clear: you make it sound like we're not running any periodic background tasks at the moment.  We *do* run periodic tasks using cron.  We *do not* use django-cron to do so.

I'm happy to merge this so you can experiment, but please don't change the existing cron jobs until I have a chance to look at them in detail.

tompollard:
Thanks!

> Django Cron tasks that don't appear to run because Django Cron is not installed.

> > Just to be clear: you make it sound like we're not running any periodic background tasks at the moment. We do run periodic tasks using cron. We do not use django-cron to do so.

Yep, I'm not saying that we don't run cron tasks. e.g. we're purging old accounts here:

https://github.com/MIT-LCP/physionet-build/blob/83c80c59ce618cbe9c9e2c239e143db7849d88fa/deploy/production/etc/cron.d/physionet#L11-L15

My point was just that we have django-cron tasks defined in https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/cron.py that (as far as I can see) are not being run:

https://github.com/MIT-LCP/physionet-build/blob/83c80c59ce618cbe9c9e2c239e143db7849d88fa/physionet-django/physionet/cron.py#L14-L73

PR #2213, opened 2024-04-02
Added the environment variable in the example file

Rutvikrj26:
This PR is a followup to #2204 where the access policies are getting restricted. This adds the configuration setting to the .env.example with usage instructions and it's effects on the platform.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -137,6 +137,10 @@ SSO_LOGIN_BUTTON_TEXT="Single Sign-On"
 ENABLE_FILE_DOWNLOADS_OPTION=true
 COPY_FILES_TO_NEW_VERSION=true
 
+# Allowed Access Policies [Optional] - This is a list of access policies that can be assigned to a newly submitted project on the 'Project Access' page.
+# The environment variable is a stringified list of access policies separated by commas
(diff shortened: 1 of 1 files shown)


bemoody:
1. This belongs together with the "ENABLE_FILE_DOWNLOADS_OPTION" setting.  Put it there, not at the end of the file.

2. Please try to explain what these settings are for, say something like "The following settings determine the options that can be selected for a newly-submitted project, on the 'Project Access' page.  These settings do not affect existing published projects."

3. The *default* (in .env.example) must be to enable all access policies, so that we can test them.

4. Always end text files with a line feed character.

Rutvikrj26:
Thankyou for your comments @bemoody!

I have implemented the changes suggested by you. 

> 1. This belongs together with the "ENABLE_FILE_DOWNLOADS_OPTION" setting.  Put it there, not at the end of the file.

Can you please expand on this a bit - why it belongs there. I could not directly see a pattern in the example env file that would suggest this. 

Thanks!

tompollard:
Thanks Rutvik, looks good to me.

> Can you please expand on this a bit - why it belongs there. I could not directly see a pattern in the example env file that would suggest this.

I think this is a fair question. If we have expectations for structure, we should outline them somewhere (e.g. perhaps in a comment at the top of the file).

PR #2219, opened 2024-04-17
Configurable production database name and credentials

bemoody:
Allow configuring the production database name (and username and hostname) by bringing these old settings modules in line with the new style config variables used by physionet.settings.settings.

(All these variables *should* already be set correctly on all our servers, but of course we should check before deploying.)


bemoody:
(And I think it should be "safe" to merge this because I think the git hooks should squawk if the variables are missing, but I'm testing that right now.)

tompollard:
thanks!

PR #2223, opened 2024-04-22
Updating the Certification Page for course based trainings

Rutvikrj26:
## What
This is a small PR making a tiny change in the queryset for the trainings. The logic is updated to also include custom logic for courses as their expiry is determined in slightly different way than the regular one. 

## Why
A TrainingType can have multiple courses attached to it, each being a unique version in itself. Each course version should have it's own valid duration - and hence it's own expiry date.

Whenever a user takes an on-platform course, it is the latest version of the course available at that time, but with time new versions can come in, and at some point, the version of the course that this training was attached to, will be deprecated. At that point all the trainings associated with this course version should be renewed. Thus, admin panel provides an option to set an expiry date to the associated trainings when deprecating a course version. That is how the expiry date for the trainings is decided for course version, which is different from the way it is determined for other types of trainings.

Thus, this PR adds onto the logic of querysets so that all the courses are accurately identified and put under active / expired versions. 

### This PR does not disturb the regular flow - it only add onto the logic to also fetch the course based trainings (A feature in the upcoming Trainings part-2 PR)


Rutvikrj26:
The model connection between Course and Trainings has now been updated to a one-to-many relationship. The initial plan of implementing the M2M was to avoid adding new fields to training. For example, the training_type is attached to course via a foreign key in course, thereby not changing the training_type model, but was not the best implementation and has hence been changed to the more logical implementation - one training can only have one course, but a course can have multiple trainings. 

That implementation requires a migration and a new table to be created, which is created in the #1951. 
**This PR should only be merged after merging #1951 else it will break the current Certifications Page**

Rutvikrj26:
@tompollard @bemoody 

When a user is taking a course, we only add a training in the system at the end of the course. The training directly gets to the Accepted stage. To show the courses in progress, I am thinking of making a training object at the start of the course and assign it the review status. When the user completes the course, that training object's status gets updated to Accepted. 

Furthermore, I'm also thinking of changing the `Under Review` to `In Progress` in the certifications page. Please let me know your thoughts on this. 

image

Rutvikrj26:
@tompollard @bemoody I've tried summarizing the larger goal & the smaller chunks the entire Course functionality was broken into #2299 

This is the third part of the functionality required to display the `in-progress`, `active` & `expired` courses.

physionet-django/training/views.py
```diff
@@ -41,7 +41,23 @@ def get_course_and_module_progress(user, course, module_order):
...
-    course_progress, _ = CourseProgress.objects.get_or_create(user=user, course=course)
+    # get the course progress of the user for the course
+    course_progress = CourseProgress.objects.filter(user=user, course=course).first()
+    if not course_progress:
+        course_progress = CourseProgress.objects.create(user=user, course=course)
+        # Initiate the training and set the status to review
+        slug = get_random_string(20)
+        while Training.objects.filter(slug=slug).exists():
+            slug = get_random_string(20)
+        training = Training()
```
tompollard: Is there a reason why the create method isn't used here? e.g.

```
training = Training.objects.create(
    slug=slug,
    training_type=course.training_type,
    user=user,
    course=course,
    process_datetime=timezone.now(),
    status=TrainingStatus.REVIEW,
)
```

physionet-django/user/managers.py
```diff
@@ -1,30 +1,57 @@
...
 from django.utils import timezone
 
-from user.enums import TrainingStatus
-
+from user.enums import TrainingStatus, RequiredField
+from training.models import Course
 
 class TrainingQuerySet(QuerySet):
     def get_review(self):
-        return self.filter(status=TrainingStatus.REVIEW)
```
tompollard, line 9: please could you add docstrings for these methods to explain how they should be used?

physionet-django/user/managers.py
```diff
@@ -1,30 +1,57 @@
...
+        return self.filter(
+            Q(status=TrainingStatus.REVIEW),
+            Q(training_type__required_field=RequiredField.DOCUMENT)
+            | Q(training_type__required_field=RequiredField.URL)
+        )
+
+    # adding a query to fetch the on-platform courses that are in the status REVIEW.
```
tompollard: this comment should be part of the docstring.

tompollard (commented):
I added a couple of suggestions, but overall this looks okay to me. The mixed tutorial types is fairly confusing, and I hope we can clear this up at some point.

physionet-django/training/views.py
Rutvikrj26: Nope, no reason in particular. The code evolved over time in a way that it ended up that way. Initially, the training was initialized, and then updated in the other parts of the code. I will re-factor it to use `create` method.

physionet-django/user/managers.py
Rutvikrj26: Sure, I've added the docstrings to the new one I added, and all the others - detailing what objects they will return.

physionet-django/user/managers.py
```diff
@@ -1,32 +1,71 @@
...
+            Q(status=TrainingStatus.REVIEW),
+            Q(training_type__required_field=RequiredField.DOCUMENT)
+            | Q(training_type__required_field=RequiredField.URL)
+        )
+
+    def in_progress(self):
+        """
+        Get the on-platform training objects for the user, that are in the status REVIEW.
+        """
+        return self.filter(
```
tompollard, line 23: it's odd that we are using the `TrainingStatus.REVIEW` here. Is there a reason why we don't introduce a new `TrainingStatus.IN_PROGRESS` status?

Rutvikrj26: Training has other mechanisms linked to it, and hence I did not want to introduce a new type of training status, which might introduce new bugs, given I might miss handling it somewhere. 

Furthermore, the REVIEW & the IN_PROGRESS are practically the same status, given one is for on-platform courses and other for document-submission based courses. Hence, to make sure we don't introduce unnecessary bugs accidentally, I went ahead with REVIEW.

Rutvikrj26: As discussed in the call with @tompollard & @bemoody, I've implemented a new `IN_PROGRESS` status, which is now used by on-platform trainings.

physionet-django/user/views.py
```diff
@@ -824,11 +824,14 @@ def edit_certification(request):
     )
     training_by_status = {
         "under review": training.get_review(),
+        "in progress": training.get_in_progress(),
         "active": training.get_valid(),
         "expired": training.get_expired(),
         "rejected": training.get_rejected(),
     }
 
+    print(training_by_status)
```
tompollard: Left over from debugging?

physionet-django/user/managers.py
```diff
@@ -1,32 +1,71 @@
...
+            Q(status=TrainingStatus.REVIEW),
+            Q(training_type__required_field=RequiredField.DOCUMENT)
+            | Q(training_type__required_field=RequiredField.URL)
+        )
+
+    def get_in_progress(self):
+        """
+        Get the on-platform training objects for the user, that are in the status REVIEW.
```
tompollard: Minor thing, but it would be good to update the docstring to match the new status.

Rutvikrj26:
@tompollard Cleaned up

tompollard:
thanks Rutvik!

PR #2224, opened 2024-04-23
Add a DB_PORT configuration variable

bemoody:
Similar to the DB_HOST and DB_NAME variables (see also pull #2219), add a configuration variable to control the port number for PostgreSQL.


bemoody:
@tompollard: could you please take a look at this?  (Since I already pushed it to the production server :P)

Note, we don't currently have CI checks for physionet.settings.staging and physionet.settings.production but we do for physionet.settings.settings.

tompollard:
Looks good, thanks!

PR #2234, opened 2024-05-13
Turn reference URLs into hyperlinks

bemoody:
References in the References section should include a link, if a URL is provided.  Fixes issue #658.

I changed my mind many times about how this should be implemented.  At first, I thought that we should separately provide "description" and "URL" fields for authors to edit.

But now I think, actually, the URL is part of the formal reference, and should be displayed as text for transparency's sake.  So I think it's better for the URL to be extracted automatically from the description.

On the other hand, I don't want to have ad-hoc parsing rules applied at runtime to published content.  (It's possible those rules might change in the future, which I think is okay for active but not for published projects.)  So I ended up with this approach:
- For active projects, the URL is extracted by a regular expression whenever it's needed.
- When a project is published, the automatically-extracted URL is saved as a model field.
- For published projects, it's possible to edit the `description` and `url` separately via Django admin, but it will only appear as a link if the `url` appears verbatim within the `description`.


tompollard:
looks great, thanks!

PR #2237, opened 2024-05-20
Fix resetdb after submitting UploadedDocuments

bemoody:
The original storage layout, designed by Chen, places all media files (and the dynamic static files) in subdirectories of subdirectories (`media/active-projects/SHuKI1APLrwWCqxSQnSk`, `media/credential-applications/YgqTQmcerj0gLx4MAaZV`, etc.)

Currently the UploadedDocument feature disregards this convention, and stores files in `media/ethics`.

Thus, it breaks the `resetdb` command, which assumes everything is stored in a subdirectory of a subdirectory.


tompollard:
looks good to me

Issue #1818, opened 2023-01-18
Multiple conflicting "view_event_menu" permissions


bemoody:
Currently, in the live and staging systems, there are two `django.contrib.auth.models.Permission` objects that both have `codename = "view_event_menu"`.  One that refers to the current `events.models.Event`, and one that refers to the now-deleted `user.models.Event`.

This results in a server error when trying to join an event, as @danamouk and @xborrat discovered.

Presumably we could go into the database and delete the obsolete permission object by hand, but this is weird; I don't understand why the object is there and wasn't deleted automatically by Django when applying the `0049_auto_20230112_1430` migration.

Running `resetdb` and `loaddemo` from scratch - either using sqlite or postgresql - seems to result in only one permission object.

bemoody:
A more robust way to write it would be something like...
```
Permission.objects.get(codename="view_event_menu",
                       content_type__app_label="events")
```

or else...
```
Permission.objects.get(codename="view_event_menu",
                       content_type=ContentType.objects.get_for_model(Event))
```

It seems that Django generally expects permission names to be unique within an app; most functions that handle permissions expect you to specify the name as a string `"{app_label}.{codename}"`.  Our `User.get_users_with_permission` should probably do the same thing.

At a guess, maybe the obsolete permission wasn't deleted because there were existing users with that permission?  Should test this.  It could be a problem if we think we've deleted a permission but it's still lingering in the production database.

bemoody:
> At a guess, maybe the obsolete permission wasn't deleted because there were existing users with that permission?

Just tried installing d6d195aaa27da54d14f31e7c59f3eef004537cd3 and pushing 3fe37a8e081100965a2286ebf616952d727c6abe.  I get:

```
physionet=> select * from auth_permission as p inner join django_content_type as c on p.content_type_id = c.id where p.codename = 'view_event_menu';
 id |               name                | content_type_id |    codename     | id | app_label | model 
----+-----------------------------------+-----------------+-----------------+----+-----------+-------
 42 | Can view event menu in the navbar |              25 | view_event_menu | 25 | user      | event
 67 | Can view event menu in the navbar |              66 | view_event_menu | 66 | events    | event
(2 rows)

physionet=> select * from auth_group_permissions as a inner join auth_group as g on a.group_id = g.id where permission_id in (42, 67);
 id | group_id | permission_id | id | name 
----+----------+---------------+----+------
 42 |        7 |            42 |  7 | Host
(1 row)

[...]

PR #2243, opened 2024-05-28
Adding cohost permissions to view/respond to participation requests

Rutvikrj26:
#2192 Implements the ability for an event host to add cohosts to the event. Currently cohosts don't have permissions to do anything on the platform. 

This PR implements the change in the events functionality for the cohost to be able to view the participant lists, edit the event, view/respond to participation requests. 

A scenario: Considering the event is held across the nation with multiple hubs, with each hub having an admin/lead. The admin for each hub is a cohost that can individually verify and approve the participant requests for the hub.


bemoody:
The key question here is what this means in terms of access control for event datasets.

I realize that for now, event datasets don't actually make a difference to PhysioNet, but I would like to support them eventually, and there are a few things about them that that still make me uneasy.

In issue #2244, I've tried to lay out what I think is needed to make this system reasonably secure and auditable - changes that I hope should be easy to implement and not cause problems for HDN either.  Could you please take a look?

Rutvikrj26:
Thanks for adding these @bemoody 
These are major security issues and need addressing before the next event happens. I'll be updating this PR to also solve the issues mentioned in #2244

bemoody:
* In console/views.py:

> @console_permission_required('user.add_event_dataset')

This should be `events`, not `user` - and the other lines referring to `user.view_all_events` should be `events.view_all_events` instead.  Oops!

`user.view_all_events` and `user.view_event_menu` are obsolete permissions that should have been removed (issue #1818).

* In events/templates/events/event_home.html:

> {% if user == event.host or user.id in event.get_cohost_ids %}

Couldn't this be `{% if user == event.host or user in event.get_cohosts %}` ?

It might be cleaner, instead, to define another function in events/templatetags/participation_status.py.

* In events/templates/events/event_notifications.html: the added text should be wrapped in `{% if participation_response_form.instance.event.datasets.exists %}`.

Rutvikrj26:
@bemoody  I've implemented these suggestion:

> This should be `events`, not `user` - and the other lines referring to `user.view_all_events` should be `events.view_all_events` instead. Oops!
> 
> `user.view_all_events` and `user.view_event_menu` are obsolete permissions that should have been removed (issue #1818).

Both these permissions were actually a part of Event model, I mistakenly put them to user. 



> * In events/templates/events/event_home.html:
> 
> > {% if user == event.host or user.id in event.get_cohost_ids %}
> 
> Couldn't this be `{% if user == event.host or user in event.get_cohosts %}` ?
> 
> It might be cleaner, instead, to define another function in events/templatetags/participation_status.py.

That is a much cleaner implementation - I implemented a filter in the templatetags that addresses this and removed the redundant method for cohost_ids.



> * In events/templates/events/event_notifications.html: the added text should be wrapped in `{% if participation_response_form.instance.event.datasets.exists %}`.

Wrapped the text in the condition to make it more secure.

physionet-django/project/managers/publishedproject.py
```diff
@@ -8,7 +8,7 @@
 
 
 class PublishedProjectManager(Manager):
-    def accessible_by(self, user):
+    def accessible_by(self, user, include_event_datatsets=True):
```
tompollard: typo `include_event_datatsets` should be `include_event_datasets`

physionet-django/console/templates/console/event_management_manage_dataset.html
```diff
@@ -4,6 +4,9 @@
     <ul class="list-group list-group-flush">
       <li class="list-group-item">
         <h5>Add a dataset to the event</h5>
+          <p>By adding a dataset, all participants (approved by {{ event.host.username }} or their cohosts) will be allowed to access this
```
tompollard, line 7: The title of the PR says that we are: "Adding cohost permissions to view/respond to participation requests". This change doesn't seem related to that task?

physionet-django/console/views.py
```diff
@@ -3134,7 +3134,8 @@ def event_archive(request):
                    })
 
 
-@console_permission_required('user.view_all_events')
+@console_permission_required('event.add_event_dataset')
```
tompollard, line 3137: `add_event_dataset` doesn't seem related to "Adding cohost permissions to view/respond to participation requests"

physionet-django/events/forms.py
```diff
@@ -72,7 +73,18 @@ class EventDatasetForm(forms.ModelForm):
     """
     A form for adding datasets to an event.
     """
-    dataset = forms.ModelChoiceField(queryset=PublishedProject.objects.all(),
+
```
tompollard, line 76: These changes do not (as far as I can see) relate to "Adding cohost permissions to view/respond to participation requests".

physionet-django/events/templates/events/event_notifications.html
```diff
@@ -53,6 +53,26 @@ <h5 class="modal-title">Respond to participation request</h5>
                                     <div class="modal-body author_participation-response_form">
                                         <p>{{ participation_response_form.instance.user.get_full_name }} is requesting
                                             to join Event {{ participation_response_form.instance.event.title }}.</p>
+
+                                        {% if participation_response_form.instance.event.datasets.exists %}
```
tompollard, line 57: These changes do not (as far as I can see) relate to "Adding cohost permissions to view/respond to participation requests".

physionet-django/events/views.py
```diff
@@ -22,9 +22,16 @@
 
 @login_required
 def update_event(request, event_slug, **kwargs):
+
     user = request.user
     can_change_event = user.has_perm('events.add_event')
     event = Event.objects.get(slug=event_slug)
+
+    # if the event has dataset added to it, it cannot be edited
```
tompollard, line 30: This is not intuitive behaviour, and doesn't (as far as I can see) relate to the original goal of the PR.

physionet-django/project/managers/publishedproject.py
```diff
@@ -52,14 +52,15 @@ def accessible_by(self, user):
...
-        # add projects that are accessible through events
-        events_all = Event.objects.filter(Q(host=user) | Q(participants__user=user))
-        active_events = set(events_all.filter(end_date__gte=datetime.now()))
-        accessible_datasets = EventDataset.objects.filter(event__in=active_events, is_active=True)
-        accessible_projects_ids = []
-        for event_dataset in accessible_datasets:
-            if has_access_to_event_dataset(user, event_dataset):
-                accessible_projects_ids.append(event_dataset.dataset.id)
-        query |= Q(id__in=accessible_projects_ids)
+        if include_event_datatsets:
```
tompollard: These changes do not (as far as I can see) relate to "Adding cohost permissions to view/respond to participation requests".

tompollard (commented):
@Rutvikrj26 I'm a little confused about this PR. The changes to the code don't seem to match the goal that you set out in the title and description ("Adding cohost permissions to view/respond to participation requests"). 

Side note but I don't remember why "invite co-host" form and notifications ended up being added to the "view event" page. We should definitely move them to http://localhost:8000/events/ (following a similar pattern to http://localhost:8000/projects/) at some point.

physionet-django/project/managers/publishedproject.py
Rutvikrj26: i've ran over it at least half a dozen times, but all three seem to be the exact same to me. Not sure what you're pointing to.

physionet-django/console/templates/console/event_management_manage_dataset.html
Rutvikrj26: This was a comment by Benjamin, requesting it to be made clear so that event host who adds the data acknowledges the fact that he is allowing the participants to access the data directly without any trainings for the duration of the event.

physionet-django/console/views.py
Rutvikrj26: The scope of the PR evolved after my discussion with Benjamin - where it not only included the permissions for the participants, but also manage the datasets that a host can access while making an event. 

Initially, the form was configured to show all the published projects on the platform. Later it was agreed upon that only the datasets that the person hosting the event has access to should be displayed in the form. (For Eg; I have access to only 3 out of 5 published projects and I plan to host an event - I should ideally not be allowed to select the other 2 to host an event)

physionet-django/events/forms.py
Rutvikrj26: The reason as mentioned above.

physionet-django/project/managers/publishedproject.py
Rutvikrj26: Above Reason

physionet-django/events/views.py
Rutvikrj26: This was a feature requested by Benjamin upon discussion.

physionet-django/events/templates/events/event_notifications.html
Rutvikrj26: This was added to make the cohosts approving the participant request. This was another point added by Benjamin during one of our discussions.

physionet-django/project/managers/publishedproject.py
tompollard: There is a spelling mistake, I think? Shouldn't "datatsets" be "datasets"?

Rutvikrj26: Oh, Thanks!!
I couldn't spot it😅

PR #2245, opened 2024-05-30
Move GCP functions from console.utility to project.cloud.gcp

bemoody:
Trying to organize things better, by putting the cloud mirror implementations together in the project.cloud package.

This doesn't change anything about the behavior of GCP mirrors - just moving code around.


tompollard:
good idea, thanks

tompollard:
Odd, why are these style checks failing after the PR was merged (but not before)?

Issue #2160, opened 2023-12-05
Updating DOI changes publication year


bemoody:
`console.utility.generate_doi_payload` generates the JSON metadata that is sent to the DataCite API, either to register, publish, or update a DOI.

This includes:
```
                "publicationYear": timezone.now().year,
```
which seems very wrong.

tompollard:
Good catch!

tompollard:
When creating the DOI for `core_project`s (which points to the latest version of the project), do you agree that we should set the `PublicationYear` to the latest version of the project?

bemoody:
Yeah, I think the core project DOI info should match the latest version, including publication year.

I'm not sure what happens right now if you go to a "non-latest-version" management page and click "Update metadata (core)".

PR #2246, opened 2024-06-07
Fix publicationYear when creating/updating DOIs

bemoody:
In the management page for a published project, there are buttons to create or update DOIs for the published project and its associated core project.  ("Update" if the project already has a DOI assigned, or "create" if it doesn't.)

The "publicationYear" in the DOI metadata should indicate the year when the resource was published, not the current year or the year the DOI was created.  This corresponds to the project's publish_datetime.

Previously, the page also gave you the option to create/update the core DOI even if you weren't looking at the latest project version.  This was not what you wanted - it would have erroneously applied the metadata of the old version.  This button is now disabled when inapplicable.

Fixes issue #2160.


tompollard:
Thanks Benjamin!

PR #2255, opened 2024-07-11
Allow editorial team to record internal notes for active projects

tompollard:
We are struggling to keep on top of reviewing and processing new project submissions to PhysioNet. One small feature that would be helpful is to allow people with editorial permissions to record short notes about a project.

e.g. "Contact the authors to arrange an embargo date"; "The references are incorrectly ordered. This should be fixed before publication"; "Takeshi, please could you review some of the images?"

This pull request adds functionality to create and delete notes about an active project.

Screenshot 2024-07-11 at 12 42 48 PM


bemoody:
In `console/views.py` line 280, notes should be ordered by `created_at`.

bemoody:
We might also want to say that you can't delete other people's comments.

tompollard:
> In console/views.py line 280, notes should be ordered by created_at.

Good catch. Comments are now sorted by `created_at` (newest at the top).

> We might also want to say that you can't delete other people's comments.

Comments can now only be deleted by the author of the comment.

bemoody:
`aria-labelledby="timeline-tab"` is wrong; you want "notes-tab".

`<li>` must be a child of a `<ul>` (or `<ol>`.)  Also, please indent the body of the loop.  So it should look like:
```
        <p>Notes posted here are visible only to the editorial team. They are not visible to authors.</p>
        <ul class="list-group">
          {% for note in notes %}
            <li class="list-group-item">
              <p>{{ note.content }}</p>
              ...
            </li>
          {% endfor %}
        </ul>
```

tompollard:
> `aria-labelledby="timeline-tab"` is wrong; you want "notes-tab".

Fixed (along with a couple of unrelated mislabeled aria tags).

> `<li>` must be a child of a `<ul>` (or `<ol>`.)  Also, please indent the body of the loop. 

Fixed, thanks for picking this up.

bemoody:
Why list-inline?

tompollard:
> Why list-inline?

By default the list is indented, which wastes valuable horizontal space (and also doesn't look very nice). I considered adding a no-indent style, but this would have just been duplicating `list-inline`. I'm happy to change to another non-indented style/layout if you have suggestions.

bemoody:
Ah, yeah.  list-group is what you want AFAICT.

bemoody:
Another small thing, I'd use `{{ note.content|linebreaks }}`.

tompollard:
> Ah, yeah. list-group is what you want AFAICT.

Switched to `list-group`, thanks (for future, how do I know this is what I want? because preference for native bootstrap styles?)

> Another small thing, I'd use `{{ note.content|linebreaks }}`.

Done!

bemoody:
Searching for list-group-item, I found this: https://getbootstrap.com/docs/4.0/components/list-group/ - that shows at least how the classes are "intended" to be used.

"list-inline" is... not documented?  But we can guess that "list-inline" is supposed to be used with "list-inline-item".

Anyway, this looks fine to me.

tompollard:
> Searching for list-group-item, I found this: https://getbootstrap.com/docs/4.0/components/list-group/ - that shows at least how the classes are "intended" to be used.

> "list-inline" is... not documented? But we can guess that "list-inline" is supposed to be used with "list-inline-item".

Makes sense, thanks for the explanation! CSS is still a bit of a mystery for me.

Issue #2258, opened 2024-07-17
Replicate "Internal notes" functionality for managing project reviews


tompollard:
We could consider replicating the "internal notes" functionality that was introduced in https://github.com/MIT-LCP/physionet-build/pull/2255 as a simple first step towards implementing peer review functionality.

i.e. We would add a "Reviews" tab to the editorial tools page for Active Projects, like this:

![Screenshot 2024-07-17 at 7 26 20 PM](https://github.com/user-attachments/assets/af87dc58-8f34-4aa3-8edb-8c0600d2a372)

Initially project editors would need to request reviews via email and then post the reviews on behalf of reviewers. We could then work towards more complete peer review functionality by:

1. Allowing PhysioNet users to post reviews to a project upon invitation.
2. Allowing PhysioNet editors to send email invitations to reviewers
3. Introducing an "Awaiting reviews" step to the editorial workflow (between "Awaiting Editor Assignment" and "Awaiting Decision") that would encourage editors to seek external reviewers on all submissions.

@bemoody @lilehman Please let me know your thoughts. 

We can implement Step 1 very quickly if you agree this is a good direction.

PR #2261, opened 2024-07-19
WIP: Add functionality for managing project reviews 

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/2258, it would be helpful to be able to record reviews of datasets in the editorial system.

This pull request takes a simple approach to addressing this issue by replicating the "Internal notes" functionality that was added in  https://github.com/MIT-LCP/physionet-build/pull/2255 to support posting of reviews.

image

Initially project editors would need to request reviews via email and then post the reviews on behalf of reviewers. We could then work towards more complete peer review functionality by:

- Allowing PhysioNet users to post reviews to a project upon invitation.
- Allowing PhysioNet editors to send email invitations to reviewers
- Introducing an "Awaiting reviews" step to the editorial workflow (between "Awaiting Editor Assignment" and "Awaiting Decision") that would encourage editors to seek external reviewers on all submissions.


lilehman:
This is useful, especially if we can capture external reviewers comments. Thanks!

tompollard:
Assuming we would like to go ahead with this change, the pull request will need to be updated to follow the approach used in https://github.com/MIT-LCP/physionet-build/pull/2264

Issue #2262, opened 2024-07-23
Internal notes under `/info` not shown under `/edit`


briangow:
In https://github.com/MIT-LCP/physionet-build/pull/2255 a tab was added to allow adding notes to a submitted project which are only visible internally. This tab shows under the `/info` and `/edit` pages for the project. However, when I add notes under `/info` they are not shown under `/edit`.

bemoody:
Should be fixed by pull #2264

PR #2264, opened 2024-07-24
Ensure internal notes are displayed to editors.

tompollard:
In https://github.com/MIT-LCP/physionet-build/pull/2255 we added a feature to store "internal notes" for projects. There is a bug in the current implementation that means notes are not displayed to people "editing" a project (see: https://github.com/MIT-LCP/physionet-build/issues/2262). This fixes that bug.

To reproduce the bug:

- Login as `tompollard`
- Go to http://localhost:8000/console/submitted-projects/p7TCIMkltNswuOB9FZH1/info/ and add an internal note
- Go to http://localhost:8000/console/submitted-projects/p7TCIMkltNswuOB9FZH1/edit/. No note!

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -271,6 +271,43 @@ def submission_info_redirect(request, project_slug):
     return redirect('submission_info', project_slug=project_slug)
 
 
+def submission_info_card_params(project,
+                                authors,
+                                author_emails,
+                                storage_info,
+                                edit_logs,
+                                latest_version,
+                                url_prefix,
+                                bulk_url_prefix,
+                                reassign_editor_form,
+                                embargo_form,
+                                notes,
+                                internal_note_form):
+    """
+    Parameters used across submission_info_card.html pages, including:
+
+    awaiting_authors.html
+    copyedit_submission.html
+    edit_submission.html
+    publish_submission.html
+    submission_info.html
+    """
+    return {
+        'project': project,
+        'authors': authors,
+        'author_emails': author_emails,
+        'storage_info': storage_info,
+        'edit_logs': edit_logs,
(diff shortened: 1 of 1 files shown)


tompollard:
Side note, the implementation of the /info/ and /edit/ pages is very muddled. I added a comment about this in the docstrings of the relevant views. We should refactor at some point.

bemoody:
That's just one of many.  awaiting_authors.html, copyedit_submission.html, edit_submission.html, publish_submission.html, submission_info.html all include submission_info_card.html.

Notice all the common stuff (ReassignEditorForm, EmbargoFilesDaysForm, ...) that is getting passed to render() in every one of those view functions.

We should at the very least factor out that stuff into one function, e.g.:
```
def submission_info_card_params(request, project):
    return {
        # template parameters used by submission_info_card.html
        'project': project,
        'authors': authors,
        'author_emails': author_emails,
        'storage_info': storage_info,
        ...
        'reassign_editor_form': reassign_editor_form,
        'embargo_form': embargo_form,
        'notes': notes,
        'internal_note_form': internal_note_form,
    }

def edit_submission(request):
    ...
    return render(request, {
        **submission_info_card_params(project),
        # any additional template parameters used by edit_submission.html
        ...
    }
```

(It looks like that was the original intent of `project.info_card`, but that function wasn't designed in an extensible way.)

tompollard:
Thanks for your suggestion Benjamin. I updated the templates that I missed, and added a `submission_info_card_params` function in an attempt to start consolidating some of the common function calls. More refactoring/cleaning needed, but if possible I'd like to save this for a new PR.

bemoody:
Looks good, thanks!

Issue #2263, opened 2024-07-23
List of training applications in admin console not in chronological order


cpitsch:
The training applications at https://physionet.org/console/training/review/ seem to be ordered by "Identity Verified" before the submission date. This makes it tedious to handle the applications by submission date.

For me, it seems more intuitive to first order by submission date and then by their verification status.

tompollard:
@cpitsch let me know when you have spare 15 min and we can walk through the fix together.

PR #2265, opened 2024-07-24
Change the sorting of training applications in the admin console

cpitsch:
As outlined in #2263, the list of training applications is currently sorted first by credentialing status and then by the timestamp of the application submission. This makes it difficult to handle the applications in the order they came in.

I propose to order the training applications by the keys (and order):

1. Date
2. Credentialed Status
3. Time

Like this, the applications can be handled by the day they came in, while still handling applications of credentialed users first (as these are the ones where an accepted training is immediately useful).  Within these groups, the applications are ordered by their exact timestamp.

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -1789,11 +1789,11 @@ def training_list(request, status):
     """
     List all training applications.
     """
-    trainings = Training.objects.select_related(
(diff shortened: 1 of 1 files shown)


tompollard:
Looks good to me, thanks Cameron!

PR #2270, opened 2024-08-07
Add "downloads" page to console with link to CSV files

tompollard:
This pull request adds a new page at "/console/downloads/" that allows administrators (`@console_permission_required('project.can_view_stats')`) to download the following CSV files:

- "Users": Contains information on all active users.  Example: [users_example.csv](https://github.com/user-attachments/files/16530731/users_example.csv)
- "Projects": Contains information on all published projects. Example: [projects_example.csv](https://github.com/user-attachments/files/16530736/projects_example.csv)

Some additional notes:

- I sneaked in a change to the the menu item from "Usage stats" to "Metrics".
- We may want to introduce a new permission for these tools later (preferably in a new pull request).
- We will want to introduce new variables and datasets later.
- This pull request is needed to generate metrics for our U24 proposal, so a quick review/merge would be appreciated!

### DIFF
--- a/physionet-django/console/navbar.py
+++ b/physionet-django/console/navbar.py
@@ -204,7 +204,7 @@ CONSOLE_NAV_MENU = NavMenu([
         NavLink(_('Project review'), 'guidelines_review'),
     ]),
 
-    NavSubmenu(_('Usage Stats'), 'stats', 'chart-area', [
+    NavSubmenu(_('Metrics'), 'stats', 'chart-area', [
         NavLink(_('Editorial'), 'editorial_stats'),
         NavLink(_('Credentialing'), 'credentialing_stats'),
         NavLink(_('Submissions'), 'submission_stats'),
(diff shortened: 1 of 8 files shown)


tompollard:
@bemoody sorry for the churn. Please could you take a look?

The user data was generating a timeout on the staging server, so I ended up switching to an approach using `StreamingHttpResponse`. 

I'm sure we will need to make some additional tweaks once this is running on the live server, but I'm hoping this is acceptable as a starting point?

bemoody:
> `@console_permission_required('project.can_view_stats')`

Viewing credentialing info should require credentialing permissions, I would use user.change_credentialapplication for now.  Yes, we can refine this later.

tompollard:
> Viewing credentialing info should require credentialing permissions, I would use user.change_credentialapplication for now. Yes, we can refine this later.

Switched, as suggested.

bemoody:
Nice.  Thanks for figuring out how to make the streaming response work, we may want to adopt that for other API-type views.

Issue #2137, opened 2023-11-20
Reference order field is not unique per project


bemoody:
Originally, the `Reference` and `PublishedReference` objects had no explicit order, which was a mess since it led to references not being displayed in the intended order.

There was, however, a database-level constraint that no project could have two references with the same *text content* ("description".)

All of this was changed through a series of pull requests (#1602, #1817, #1819) until arriving at the current behavior:

```
class Reference(models.Model):
    class Meta:
        unique_together = (('description', 'object_id', 'order'),)

class PublishedReference(models.Model):
    class Meta:
        unique_together = (('description', 'project', 'order'))
```

Let's understand what this is saying.  For `PublishedReference`, the above code is saying that there can be no two references with the same description *and* the same project *and* the same order.

There can, however, be:
- Two references with the same description and the same order (and different projects.)
- Two references with the same description and the same project (and different orders.)
- Two references with the same project and the same order (and different descriptions.)

For `Reference`, it's saying that there can be no two references with the same description, same object_id, and the same order.

There can, however, be:
- Two references with the same description and the same order (and different projects, if they have different object_ids.)
- Two references with the same description and the same project (and different orders.)
- Two references with the same project and the same order (and different descriptions.)

And there *cannot* be two references with the same description, same order, and different projects that happen to have the same object_id (e.g., ActiveProject number 1 and ArchivedProject number 1.)

This is of course nonsense and not what we want.  What we want is:
```
class Reference(models.Model):
    class Meta:
        unique_together = (
            ('content_type', 'object_id', 'description'),
            ('content_type', 'object_id', 'order'),
        )

class PublishedReference(models.Model):
    class Meta:
        unique_together = (
            ('project', 'description'),
            ('project', 'order'),
        )
```
or else:
```
class Reference(models.Model):
    class Meta:
        unique_together = (
            ('content_type', 'object_id', 'order'),
        )

class PublishedReference(models.Model):
    class Meta:
        unique_together = (
            ('project', 'order'),
        )
```
because I'm not sure it was ever all that helpful to prevent duplicate references at the database level.

We also, I think, want to make the `order` field non-nullable.

Note that currently on PhysioNet, there is one published project that contains duplicate `order`s (`physiotag-1.0.0` has two references with `order=4`) as well as one active project (`RPv08rcu97FdkJhIft3f` has two references with `order=9`), and a few other projects with unusual order sequences.

```
for p in project.models.PublishedProject.objects.all():
[...]

PR #2271, opened 2024-08-08
Don't scramble references when creating a new project version

bemoody:
When a new project version is created, references are copied from the latest published version.

Copy these references in the order they are displayed on the published project page.

- Reference "id"s must be in sequence (this is what the `order_by('order')` does); this determines the order the references will appear in the Project Content page.

- Reference "order"s must be in sequence (this is what the `order=p_reference.order` does); this determines the order the references will appear in the Project Preview page and ultimately in the published project.

**This pull request does not fix any existing projects.**  This pull request *only* affects what will happen when you click the "new version" button.  Many existing active and published projects have scrambled lists of references.  This pull request does not and cannot fix them.

This pull request also does not fix the fact that for some active projects, the order shown in the Content page doesn't match the Preview page.  If the two pages don't match, that's actually a good sign because in that case we *know* something is wrong.

This pull request also does not fix the lack of database constraints (issue #2137), which we can't fix until we have a way to migrate the broken projects.  Doing nothing is better than doing the wrong thing.

### DIFF
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -493,10 +493,12 @@ class NewProjectVersionForm(forms.ModelForm):
                     author=author)
 
         # Other related objects
-        for p_reference in self.latest_project.references.all():
+        for p_reference in self.latest_project.references.order_by('order'):
             reference = Reference.objects.create(
                 description=p_reference.description,
-                project=project)
+                order=p_reference.order,
+                project=project,
+            )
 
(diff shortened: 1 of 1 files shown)


bemoody:
Oh yeah, and this also won't fix cases where the published references have no order (presumably because they were copied from a previous version and never edited, and meaning that "order they are displayed on the published project page" is not fixed.)

tompollard:
Thanks Benjamin!

PR #2273, opened 2024-08-14
Don't allow "completing copyedit" with outstanding integrity errors

### LINKED ISSUE #2272: Copyedit page allows "completing copyedit" while project contains errors
There are various requirements that must be satisfied before publishing a project.  For example, certain fields cannot be blank.

These automatic checks are performed by `ActiveProject.check_integrity`, and they are used in three primary places:

- In `project_preview`, any errors are shown at the top of the page.

- In `project_submission`, the author is not allowed to submit the project if there are any errors.

- In `publish_submission`, the editor is not allowed to publish the project if there are any errors.

However, these checks are not performed by `copyedit_submission`.  They should be.  For example, if a required field is blank (because the editor deleted it) and the editor clicks "complete copyedit", they should receive an error.

bemoody:
There are various requirements that must be satisfied before publishing a project. For example, certain fields cannot be blank.

If a project contains integrity errors, the author is not allowed to submit it.  However, it's possible that integrity errors can be introduced *after* the project is submitted:

- the editor accidentally broke something during the copyedit process
- things that were not treated as errors *at the time the project was submitted* are treated as errors *now*, and should be addressed before publication

If the project is still in its "awaiting copyedit" stage, then the best thing is to require the editor to correct those errors before they can "complete copyedit" and advance the project to "awaiting author approval".

Fixes #2272

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -281,6 +281,9 @@ class CopyeditForm(forms.ModelForm):
     def clean(self):
         if self.errors:
             return
+        project = self.instance.project
+        if not project.check_integrity():
+            raise forms.ValidationError(project.integrity_errors)
         if self.cleaned_data['made_changes'] and not self.cleaned_data['changelog_summary']:
             raise forms.ValidationError('Describe the changes you made.')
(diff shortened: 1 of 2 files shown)


tompollard:
Thanks Benjamin

PR #2276, opened 2024-08-20
Add login using ORCID oauth

matkaczmarek:
Change introduces three new endpoints:

- `authorcid_login` -> which exchanges token with ORCID using Oauth and logs in the user if ORCID is already connected to the existing user or redirects to `orcid_register` for new users
- `orcid_register` -> present a form to register new users and links ORCID to it once registered
- `orcid_init_login` -> builds URL and redirects to ORCID for authorization

This PR also introduces changes in the login view and SSO login view:
image


physionet-django/sso/templates/sso/login.html
```diff
@@ -17,11 +17,20 @@ <h6 class="card-subtitle mb-2 text-muted">Login through an external institute</h
         <a
           type="button"
           class="btn btn-primary center p-2 px-3"
-          href="{% url 'sso_login' %}"
+          href="{% url 'login' %}"
```
rafalkosla101: Out of curiosity - why this link has changed?

matkaczmarek: it shouldn't, debug leftover, good catch!

bemoody:
There are a few issues I see here - some small, some larger - but first we need to understand the high level functionality.

We currently have the option to add an ORCID ID to a user's profile.  Is it your intention that *someone who has already done that* should now be able to log in via ORCID?

What does "log in via ORCID" mean, exactly?  What is the process that the user goes through (explain the steps for somebody who doesn't understand OAuth) and how does this process prove that *the person sitting at the keyboard* is the person with that ORCID ID?

You also need to explain how "registration" works: describe at a high level what steps are taken, from when a new user first visits the site, until they are fully registered and logged-in.

matkaczmarek:
> We currently have the option to add an ORCID ID to a user's profile. Is it your intention that _someone who has already done that_ should now be able to log in via ORCID?
> 

Yes, it was my intention that someone who has already linked an ORCID ID profile should now be able to log in using ORCID. In my opinion, those shouldn't be separate, user still have the option to log in using a username and password. Moreover, linking ORCID ID uses the same Oauth mechanism to authorize a user. It would be a bad user experience that once accounts are linked, users cannot use all linked account benefits and have to link ORCID one more time just to log in.


> What does "log in via ORCID" mean, exactly? What is the process that the user goes through (explain the steps for somebody who doesn't understand OAuth) and how does this process prove that the person sitting at the keyboard is the person with that ORCID ID?
>

From a user's perspective, once the user clicks `Log in via ORCID` (maybe we should rephrase it to `Log in using ORCID`), they are redirected to the ORCID authorization page where the user logs in to ORCID platform using ORCID credentials (here they are proving that `the person sitting at the keyboard is the person with that ORCID ID`). Once it's done we are redirected to our Physionet page and logged in if authorization was correct.

From a technical point of view, when the user is redirected to Physionet page after authorization, Physionet and Orcid exchange first the code, then the token, which consists of encrypted Orcid user data, such as ORCID ID. This token could also be used to access more ORCID user data, but we are not using this functionality. On our site, we are validating the token against malicious behavior. If everything is correct and this ORCID ID is linked to some user in Physionet then this user is logged in.


> You also need to explain how "registration" works: describe at a high level what steps are taken, from when a new user first visits the site, until they are fully registered and logged-in.
>

If, after the authorization process, we see that this ORCID ID is not linked to any account, then we redirect users to a standard registration form, where they have to create a Physionet account providing a username, email, and password. Orcid profile is automatically linked to the new account using data from the Orcid authorization token. If an account is not created then nothing happens.

bemoody:
> On our site, we are validating the token against malicious behavior.

What is validated?  What kind of malicious behavior?  Please be specific.

This article seems like a good place to start: https://oauth.net/articles/authentication/

matkaczmarek:
> What is validated? What kind of malicious behavior? Please be specific.
> 
> This article seems like a good place to start: https://oauth.net/articles/authentication/

Thanks for a great article. Referring to it, when I was writing ` On our site, we are validating the token against malicious behavior,` I had in mind `Injection of access tokens`  which was resolved by retrieving the access token directly from ORCID endpoint not from request parameters: https://github.com/MIT-LCP/physionet-build/pull/2276/files#diff-2f424d21ad5aad51520be5d93166608c26e0df7a678f42851d72ad8b62fd9f15R536-R547

Also when implementing ORCID authentication, I followed their documentation (`Implement OAuth, 3 legged OAuth`): https://info.orcid.org/documentation/api-tutorials/api-tutorial-get-and-authenticated-orcid-id/ which results in authenticated ORCID user.

matkaczmarek:
Hi @bemoody, after our last call I two changes:
- I added `id_token` verification by checking the signature of the ID token using ORCID JSON Web Key Set (ORCID public keys used to verify JWT)
- Log in using ORCID is now only presented to the user if ORCID_SCOPE contains an `openid` scope.

bemoody:
If somebody's ORCID account was compromised on Monday, and they reset their ORCID password on Tuesday, and they try to log in to PhysioNet using ORCID on Wednesday, how do we know whether it's the real person or the attacker?

If they logged in to PhysioNet using ORCID on Monday, and their ORCID account was compromised on Tuesday, and they reset their ORCID password on Wednesday, and they visit a PhysioNet page on Thursday using the session cookie they received on Monday, how do we know whether it's the real person or the attacker?

matkaczmarek:
> If somebody's ORCID account was compromised on Monday, and they reset their ORCID password on Tuesday, and they try to log in to PhysioNet using ORCID on Wednesday, how do we know whether it's the real person or the attacker?
> 
If the password is changed on Tuesday then all the active ORCID sessions should be invalidated including this Monday's one, therefor if the attacker tries to log in to Physionet on Thursday they will have to go through the OAuth authentication flow again which requires knowledge of the new ORCID password

> If they logged in to PhysioNet using ORCID on Monday, and their ORCID account was compromised on Tuesday, and they reset their ORCID password on Wednesday, and they visit a PhysioNet page on Thursday using the session cookie they received on Monday, how do we know whether it's the real person or the attacker?

It depends on how long we keep the session on our side. When the ORCID account is compromised, Physionet itself cannot automatically know this. Unfortunately, ORCID (like most of the OAuth providers) does not notify relying parties about password resets or compromises. It is on our side to invalidate session cookies as often as possible (24 hours seems reasonable). If the attacker gets the session cookie to Physionet, he/she will be able to access the attacked account until the cookie expires. Also, if someone's ORCID account is compromised, good practice is to change the passwords to all websites this ORCID account was linked.

Those are very extreme scenarios. As website creators, we often cannot determine whether the account was compromised. Currently, in our flow, if an attacker gets access to a user's email, then Physionet cannot do anything, as the attacker can change the password (using reset password form) and log in to Physionet using a new password. In both scenarios, we trust the users that they can keep their email/ORCID accounts safe.

bemoody:
> If the password is changed on Tuesday then all the active ORCID sessions should be invalidated including this Monday's one, therefor if the attacker tries to log in to Physionet on Thursday they will have to go through the OAuth authentication flow again which requires knowledge of the new ORCID password

Right.  So how do you *know* that the person went through the ORCID login - how do you know that the "code" they gave you was generated recently?

> It is on our side to invalidate session cookies as often as possible (24 hours seems reasonable). If the attacker gets the session cookie to Physionet, he/she will be able to access the attacked account until the cookie expires

PhysioNet session IDs are valid for 14 days.  This is not a security problem, because when you change your password, all other sessions are invalidated.

Having sessions that expire without warning or recourse *is* a significant usability problem (which is why I've insisted that we don't change from the Django defaults.)  For example, I don't think it's okay if, after somebody spends 8 hours uploading a file, the upload is rejected because they had logged in 31 hours earlier.  In fact, we could and should do better than we're doing now.

bemoody:
I've now done a bunch of reading on OIDC, looked at some existing libraries and was unimpressed.  ;)  But it seems like there are a couple of possibilities:

- OIDC providers *may* allow the RP to refresh the id_token in the same way as refreshing the access_token.  This seems ideal if it works.

- Embed in the page a hidden iframe that, every N minutes, redirects itself to the ORCID authentication page (using the query parameter `prompt=none`) - thereby ensuring that as long as you keep the browser open, the tokens are regularly refreshed.

matkaczmarek:
> Right. So how do you _know_ that the person went through the ORCID login - how do you know that the "code" they gave you was generated recently?
> 

It doesn't say in the documentation but I checked it experimentally and code is valid for no more than 10 minutes. After this time I got `The authorization code is invalid or has expired` error and it prevented me from exchanging it for access token and as a result I was not able to log in to Physionet


> PhysioNet session IDs are valid for 14 days. This is not a security problem, because when you change your password, all other sessions are invalidated.
> 
> Having sessions that expire without warning or recourse _is_ a significant usability problem (which is why I've insisted that we don't change from the Django defaults.) For example, I don't think it's okay if, after somebody spends 8 hours uploading a file, the upload is rejected because they had logged in 31 hours earlier. In fact, we could and should do better than we're doing now.

That's a valid case, but as said on a meeting the attackers will have access to the Physionet as long as they will have valid session. If we want to keep session alive as long as the ORCID session they it will last couple of hours (no more than one day).

bemoody:
To elaborate on my previous comment:

Naturally, we only know what happens on our server.  Verifying the `id_token` lets us know that the client was authenticated to ORCID servers at a particular time, but it doesn't tell us whether the client is still authenticated.

We want to allow *authenticated* clients to remain logged-in to PhysioNet for a reasonably long time (say, at least 24 hours), and we also want *unauthenticated* clients to be rejected within a reasonably short time period after they cease to be authenticated (say, at most one hour.)

That means that we need to regularly receive updated information from the ORCID service to tell us whether that client is still authenticated from their point of view.

There is no clever way to make this work without polling the ORCID server in some way.

## making the client do the polling

One way to poll the ORCID server is to tell the client to go and get a new authorization code.  It's possible that this could be done by embedding an iframe (e.g., in navbar.html, with appropriate CSS to make it invisible), where the content of the iframe contains:

`<meta http-equiv="refresh" content="600; url=https://orcid.org/oauth/authorize?response_type=code&scope=openid&prompt=none&client_id=...&redirect_uri=...&state=...">`

Then, after 600 seconds, the client will request a new auth code from orcid.org; orcid.org will give either a new auth code or an error; and we tell the client to do the same thing 600 seconds later.

Some logic is needed to ensure that if the user is authenticated via ORCID, and we have *not* received a positive response from the ORCID server recently, then the user is treated as anonymous.  It might be possible to do this by expiring the Django session, but it might be better to use a separate middleware for this purpose.

Of course, we can also add logic to the *user login* view to say: if the client was previously authenticated via ORCID and has now expired, then redirect directly to ORCID without showing a PhysioNet login form.

## polling on the server side

The other way to poll the ORCID server is to use the refresh_token, as described here: https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokens

This might or might not work, depending on the OIDC provider.  The advantage is that it's seamless from the client's POV - it works in places where client-side redirection is impossible.

This would have to be done by middleware, I would think.  The Django auth backend API doesn't seem to provide a way to do it.

matkaczmarek:
Thanks @bemoody for explaining what you want to achieve. It makes sense but let me verify what options we have and whether they are programmaticaly doable. Thanks also for the tip with iframe and refresh_token. I will come back to you with potential solutions.

rafalkosla101:
Since i have partly taken over this one:
As far as I have tested the implementation orcid is preventing his pages from being loaded in any kind of iframes to prevent security issues such as clickjacking attacks. We could be probably instead of iframe using a popup window that is somewhere hidden and checking what happens inside of that window but it is definitely not a elegant way to handle it and it would still be limiting the time of usage to a couple of hours


> Of course, we can also add logic to the _user login_ view to say: if the client was previously authenticated via ORCID and has now expired, then redirect directly to ORCID without showing a PhysioNet login form.

That in a way is quite limiting since a user would be trapped into "orcid login" way of accessing our platform whenever he/she connects orcid to the physionet itself, we could be adding some sort of disconnect feature but that kind of defeats a purpose. We could be also adding a button to `end all active sessions` which would to some extent minimalize the risk of attacker having 14 days of free physionet access



> ## polling on the server side
> The other way to poll the ORCID server is to use the refresh_token, as described here: https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokens

Refresh token is not deactivated when user changes password and the token by itself has quite long expiry date which means that the attacker would be able to access the site anyway.

physionet-django/sso/templates/sso/login.html
```diff
@@ -21,8 +21,19 @@ <h6 class="card-subtitle mb-2 text-muted">Login through an external institute</h
           aria-disabled="true"
         >
           <i class="fa fa-university fa-lg mr-3"></i>
-          <span class="h6">{{ sso_login_button_text }}</span>
+          <span class="h6">login using you institution</span>
```
rafalkosla101: Maybe we could use there a capital letter?

rafalkosla101 (approved):

physionet-django/sso/templates/sso/login.html
matkaczmarek: I think there is also a small typo I will refactor it together with changes after Tom's or Benjamin's review

physionet-django/physionet/settings/base.py
```diff
@@ -281,11 +281,14 @@
 # Tags for the ORCID API
 ORCID_DOMAIN = config('ORCID_DOMAIN', default='https://sandbox.orcid.org')
 ORCID_REDIRECT_URI = config('ORCID_REDIRECT_URI', default='http://127.0.0.1:8000/authorcid')
+ORCID_LOGIN_REDIRECT_URI = config('ORCID_LOGIN_REDIRECT_URI', default='http://127.0.0.1:8000/authorcid_login')
 ORCID_AUTH_URL = config('ORCID_AUTH_URL', default='https://sandbox.orcid.org/oauth/authorize')
 ORCID_TOKEN_URL = config('ORCID_TOKEN_URL', default='https://sandbox.orcid.org/oauth/token')
 ORCID_CLIENT_ID = config('ORCID_CLIENT_ID', default=False)
 ORCID_CLIENT_SECRET = config('ORCID_CLIENT_SECRET', default=False)
 ORCID_SCOPE = config('ORCID_SCOPE', default=False)
+ORCID_LOGIN_ENABLED = config('ORCID_LOGIN_ENABLED', default=("openid" in ORCID_SCOPE))
```
tompollard: Shouldn't `ORCID_LOGIN_ENABLED` just default to False?

physionet-django/physionet/settings/base.py
```diff
@@ -150,7 +150,7 @@
     },
 ]
 
-AUTHENTICATION_BACKENDS = ['user.backends.DualAuthModelBackend']
+AUTHENTICATION_BACKENDS = ['user.backends.DualAuthModelBackend', 'user.backends.OrcidAuthBackend']
```
tompollard: No need to add the backend if it isn't used, so perhaps this instead?

```
if ORCID_LOGIN_ENABLED:
    AUTHENTICATION_BACKENDS.append('user.backends.OrcidAuthBackend')
```

.env.example
```diff
@@ -103,6 +103,7 @@ ORCID_REDIRECT_URI=http://localhost:8000/authorcid
 ORCID_CLIENT_ID=SECRET
 ORCID_CLIENT_SECRET=SECRET
 ORCID_SCOPE='/read-limited,/activities/update'
```
tompollard, line 105: Should `ORCID_OPEN_ID_JWKS_URL` also be added to the `.env.example` file? Might want to include a brief explanation for this variable, because the name isn't very friendly.

physionet-django/physionet/settings/base.py
```diff
@@ -281,11 +281,14 @@
 # Tags for the ORCID API
 ORCID_DOMAIN = config('ORCID_DOMAIN', default='https://sandbox.orcid.org')
 ORCID_REDIRECT_URI = config('ORCID_REDIRECT_URI', default='http://127.0.0.1:8000/authorcid')
+ORCID_LOGIN_REDIRECT_URI = config('ORCID_LOGIN_REDIRECT_URI', default='http://127.0.0.1:8000/authorcid_login')
```
tompollard: Check these variables are included in .env.example.

.env.example
```diff
@@ -103,6 +103,7 @@ ORCID_REDIRECT_URI=http://localhost:8000/authorcid
...
+ORCID_LOGIN_ENABLED=True
```
tompollard: `ORCID_LOGIN_REDIRECT_URI` also seems to be missing

physionet-django/sso/templates/sso/login.html
tompollard: It isn't clear why this has changed at all.

physionet-django/sso/templates/sso/login.html
```diff
@@ -21,8 +21,19 @@ <h6 class="card-subtitle mb-2 text-muted">Login through an external institute</h
...
         </a>
+        <br>
+        {% if enable_orcid_login %}
+          <h6 class="card-subtitle mb-2 mt-3 text-muted">or using ORCID iD</h6>
+          <a id="orcid_login"
+             type="button"
+             class="btn btn-secondary center p-2 px-3"
+             href="{% url 'orcid_init_login' %}">
+            <img src="https://orcid.org/sites/default/files/images/orcid_24x24.png" />
```
tompollard: Feels like it would be safer to host the image locally.

physionet-django/user/forms.py
```diff
@@ -930,3 +931,29 @@ def save(self):
...
+
+class OrcidRegistrationForm(RegistrationForm):
+    """
+    Form to register new user after signing in with ORCID.
+    This saves user as the same way RegistrationForm but also stores
+    orcid_token and
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.orcid_token = kwargs.pop('orcid_token', None)
```
tompollard, line 944: Does the token get validated before saving?

physionet-django/user/forms.py
```diff
@@ -930,3 +931,29 @@ def save(self):
         TrainingQuestion.objects.bulk_create(training_questions)
 
         return training
+
+
+class OrcidRegistrationForm(RegistrationForm):
+    """
+    Form to register new user after signing in with ORCID.
+    This saves user as the same way RegistrationForm but also stores
+    orcid_token and
```
tompollard: "but also stores orcid_token and..."?

physionet-django/user/templates/user/login.html
```diff
@@ -27,6 +27,20 @@ <h2 class="form-signin-heading">Account Login</h2>
...
+  {% if enable_orcid_login %}
+    <div class="separator">
+      <span>or</span>
+    </div>
+    <div class="form-signin">
+      <a id="orcid_login"
+         type="button"
+         class="btn  btn-lg btn-secondary btn-block"
+         href="{% url 'orcid_init_login' %}">
+        <img src="https://orcid.org/sites/default/files/images/orcid_24x24.png" />
```
tompollard: should we use local image?

physionet-django/user/urls.py
```diff
@@ -18,6 +18,9 @@
     path("settings/cloud/aws/", views.edit_cloud_aws, name="edit_cloud_aws"),
     path("settings/orcid/", views.edit_orcid, name="edit_orcid"),
     path("authorcid/", views.auth_orcid, name="auth_orcid"),
+    path("authorcid_login/", views.auth_orcid_login, name="auth_orcid_login"),
```
tompollard: Should these paths only be available if ORCID is enabled, e.g.

```
if settings.ORCID_LOGIN_ENABLED:
    urlpatterns += [
        ...
    ]
```

physionet-django/user/views.py
```diff
@@ -459,30 +469,26 @@ def auth_orcid(request):
...
-        messages.error(request, 'Invalid Grant Error: authorization code may be expired or invalid.')
-        token_valid = False
+    token_valid, token = _fetch_and_validate_token(request, code, oauth)
 
     if token_valid:
-        orcid_profile, _ = Orcid.objects.get_or_create(user=request.user)
-        orcid_profile.orcid_id = token.get('orcid')
+        orcid_id = token.get('orcid')
+        orcid_profile = Orcid.objects.filter(orcid_id=orcid_id).first()
+        if orcid_profile and orcid_profile.user != request.user:
```
tompollard, line 486: My understanding here is that we:

1. Get the ORCID
2. Check if this ORCID is associated with a user
3. If the ORCID is not associated with a user, create a new user account.

This is quite problematic. Most (or at least many) users will have an ORCID but the ORCID will not be linked to their PhysioNet account. In these cases, the user will end up with duplicate accounts. We do not have a method for merging/linking PhysioNet accounts right now.

tompollard (commented):
Aside from minor comments, this all looks good to me, except for the following issue which I think is quite problematic:

- Most (or at least many) users will have an ORCID but the ORCID will not be linked to their PhysioNet account. In these cases, the user will end up with duplicate accounts.

Do you have any thoughts on how we can manage this issue?

physionet-django/user/views.py
tompollard: One solution might be to implement a method for joining accounts, perhaps in a later PR.

matkaczmarek:
> Aside from minor comments, this all looks good to me, except for the following issue which I think is quite problematic:
> 
> * Most (or at least many) users will have an ORCID but the ORCID will not be linked to their PhysioNet account. In these cases, the user will end up with duplicate accounts.
> 
> Do you have any thoughts on how we can manage this issue?

I tried retrieving email from ORCID and using it to check whether we already have a user in our database with the same email. Unfortunately, by default email in ORCID is set as a private attribute by default which means you cannot read it using API and access token. Users can set it to public in ORCID and only then we can see this email. Moreover, you need to have read access set in `ORCID_SCOPE` and then make additional API calls to retrieve this email. In my opinion, it is too much overhead to do when it is almost guaranteed to fail.

The only thing the user can do is to link the ORCID account to the existing Physionet account using Settings > Orcid view:
image
Moreover, you can always unlink the ORCID account from your Physionet account and link it with another one. So if users by mistake duplicate accounts they can always unlink and link them to the correct one.

physionet-django/user/forms.py
matkaczmarek: Token is validated in this [line](https://github.com/MIT-LCP/physionet-build/pull/2276/files#diff-2f424d21ad5aad51520be5d93166608c26e0df7a678f42851d72ad8b62fd9f15R525) before redirecting to `OrcidRegistrationForm`

physionet-django/sso/templates/sso/login.html
matkaczmarek: It was a debug left over, changed it back.

physionet-django/physionet/settings/base.py
matkaczmarek: That's true, I set it to False.

.env.example
matkaczmarek: Added it to `.env.example` with short explanation

physionet-django/sso/templates/sso/login.html
matkaczmarek: Added local image and used it here and in other login view.

physionet-django/user/forms.py
matkaczmarek: I think I got distracted and forgot to end this sentence. Fixed!

physionet-django/user/urls.py
matkaczmarek: Good idea! Changed

physionet-django/user/views.py
matkaczmarek: You can see my answer [below](https://github.com/MIT-LCP/physionet-build/pull/2276#issuecomment-2548216926)

tompollard (approved):

physionet-django/user/templates/user/login.html
tompollard: looks like images are now hosted locally, thanks.

matkaczmarek:
Hi @tompollard, I don't have write access to merge this PR. As I said in yesterday's meeting, I checked it one more time on HDN staging, and everything seems to work as it should. Could you merge it for me or give me and @rafalkosla101 write access so that we can merge PRs in the future?

Issue #2277, opened 2024-08-20
Add variables to the "Download data" section of the console 


tompollard:
We provide a tool for downloading data at console/downloads/. 

This issue is a placeholder for variables that we should add to the download (and related changes).

## Projects
- [ ] Rename to "Published Projects"
- [ ] Add approximate storage size, if possible (`storage_volume`) 
- [ ] Add number of signed DUAs (`signed_dua_count`)

## Users

- [ ] Add the date when the user was (last?) credentialed (`credentialed_date`)
- [ ] Add project type (`project_type`)

## Other
- [ ] Add a "All active projects" sheet that exports all active projects.
- [x] Add an "PublishedAuthors" sheet that exports a list of all published authors
- [x] Rename "Download data" to "Data export"

tompollard:
- Rename "all_emails" to "other_emails" in the user table (or make sure "all_emails" really does contain all emails).

PR #2279, opened 2024-08-21
Add option to download a published authors CSV from the console

briangow:
This adds another download option on top of what was introduced in #2270 . It gives the option to download all of the published author objects and related data. 

I've included all the base `PublishedAuthor` fields and have linked to a couple of other relevant fields (`user_id`, `project_id`, etc.). I'm happy to add / remove fields as desired.

### DIFF
--- a/physionet-django/console/templates/console/downloads.html
+++ b/physionet-django/console/templates/console/downloads.html
@@ -29,6 +29,14 @@
         </a>
       </div>
 
+      <div class="mb-4">
+        <h6>Published authors</h6>
+        <p>Download a complete list of published authors, including their names, email addresses, and affiliations.</p>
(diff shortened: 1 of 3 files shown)


tompollard:
Thanks! Looks good to me, once the style issue is resolved.

Not sure we need `author.pk` (it might even be the same as `author.id`?).

tompollard:
Oh sorry, one small thing:

`', '.join([a.name for a in author.affiliations.all()]),` might need a bit more thought because affiliations include commas. It will be difficult to work out where one affiliation ends and the other starts.

briangow:
Thanks @tompollard , I've removed the primary key from the CSV output and now delimit multiple affiliations by using a semi-colon.

tompollard:
Addresses one of the points in: https://github.com/MIT-LCP/physionet-build/issues/2277

PR #2280, opened 2024-08-23
Add additional fields to published projects CSV download

briangow:
This PR adds fields to indicate the resource/project type to the published projects CSV download available in the console. It also addresses these suggested updates for projects from https://github.com/MIT-LCP/physionet-build/issues/2277 :
- Rename to "Published Projects" (renamed the button, everything else seemed to have both words already)
- Add approximate storage size, if possible (`storage_volume`) (added `main_storage_size` which is what the entire project takes up across versions. Could add `incremental_storage_volume` also if desired).
- Add number of signed DUAs (`signed_dua_count`) Add this by counting the `DUASignature` objects for a given `project_id`.


tompollard:
Looks good, thanks. Please could you convert bytes to something easier to work with (perhaps mb or gb?). 

I'm not quite sure what the size is telling us right now, so we need to look into this (part of reason I hadn't implemented). 

Is it the storage size allocated to the project? If so, I'm not sure whether or not this is an accurate reflection of storage?

bemoody:
main_storage_size = total size of this version's file content in bytes.  The number of bytes that you will download - not counting directories or metadata - if you download the project using wget.

This is calculated after the project is published.  It's not equal to physical disk usage, nor is the sum of main_storage_size for all versions equal to the total quota usage.  And of course, the value might be wrong if the project was published in a non-standard way.  But this is the most relevant number *for users*.  If the number is wrong, it should be fixed.

briangow:
@tompollard , I'm converting the storage size to MB now. It sounds like we want to continue reporting `main_storage_size`.

tompollard:
> main_storage_size = total size of this version's file content in bytes. The number of bytes that you will download - not counting directories or metadata - if you download the project using wget.

Nice, thanks Benjamin. 

> And of course, the value might be wrong if the project was published in a non-standard way. But this is the most relevant number for users. If the number is wrong, it should be fixed.

Does "non-standard way" refer just to projects that are symlinked to a non-standard location? I think this is one or two MIMIC projects (ECG, CXR?).

tompollard:
@briangow sorry, this is just a personal thing really, but please could you change “storage_size(MB)” to “storage_size_mb” for consistency with other headers?

tompollard:
@briangow sorry, also, probably want to round to the nearest whole number?

briangow:
@tompollard , we don't anticipate needing storage size information on projects < 0.5 MB for any reason? If we might then perhaps I should include a couple of decimal places.

tompollard:
> @tompollard , we don't anticipate needing storage size information on projects < 0.5 MB for any reason? If we might then perhaps I should include a couple of decimal places.

Perhaps let's go for one decimal place? I can't see us ever wanting more fine-grained than that.

briangow:
@tompollard , I've made these updates.

tompollard:
Thanks Brian, looks good to me!

Issue #2103, opened 2023-10-10
S3: scrounging bandwidth for uploads


bemoody:
In pull #2086, one thing that worries me a little is that it will initially be very slow to upload projects.  At present, the upload process would have to compete for bandwidth with all of the clients currently downloading data.

I can think of a couple of workarounds:

1. We could do the uploads from the backup (physionet-production) server.  One advantage is that it's in a completely different physical location.  However, it would be a messy manual process and we would probably need to manually update the database on physionet-live.

2. We could configure the S3 client to use an HTTP proxy (via a separate network link, albeit from the same building.)  In fact we could set a *single* proxy server for everything (GCP, DataCite, ORCID, as well as AWS) but I think it might be preferable to configure S3 separately.

One thing I don't want to do is to prioritize uploads over client requests.

bemoody:
It'd also be super nifty if we had a way to track the progress of background tasks in the admin console.

tompollard:
> It'd also be super nifty if we had a way to track the progress of background tasks in the admin console.

Good idea, this would be helpful!

bemoody:
Any thoughts on how we should manage the network traffic?

Uploading hundreds of projects will require some automation in any event.  But I'd prefer to do so using Chrystinne's code rather than trying to script it in some other way.

tompollard:
Sorry, not my area of expertise! Personally I think I'd just take a short term hit on our network, perhaps alongside a news item explaining why downloads are slow.

tompollard:
Thinking about this a little more, my preference would be:

> We could configure the S3 client to use an HTTP proxy (via a separate network link, albeit from the same building.) In fact we could set a single proxy server for everything (GCP, DataCite, ORCID, as well as AWS) but I think it might be preferable to configure S3 separately.

This seems like an approach that may be useful in the longer term (rather than a one-off, just for the initial batch of uploads to AWS).

bemoody:
> Personally I think I'd just take a short term hit on our network, perhaps alongside a news item explaining why downloads are slow.

I don't think that's practical, though.  When I say "competing for bandwidth", I mean that uploading to Amazon would be limited to the same speed as everyone else; uploading 30 TB would take months.

It's true that *in theory* we could monkey with the traffic control settings to prioritize certain connections over others, but that's difficult and finicky and I don't want to try to deal with it.

bemoody:
to set a custom proxy, something like this should work
```
[...]

Issue #1903, opened 2023-02-28
"Create bucket and send files" should sync files intelligently


bemoody:
The "Create bucket and send files" button in the console is somewhat inefficient, and also not as robust as I would like.

We want the published project bucket to be an exact mirror of the published project files.  If they don't match, then we want to be able to click this button and have all files synchronized.

In particular:
- If a file exists in the bucket and not in the published project, it should be deleted.
- If a file exists in the project and not in the bucket, it should be uploaded.
- If a file *differs* between the project and the bucket, it should be re-uploaded.
- If a file already exists in both places and is identical (i.e., has the same MD5 hash), we should not waste time/bandwidth uploading it again.

(Possibly, there should be an option not to delete files nor to replace "newer" files with "older" ones, but to log a warning instead.)

Currently, however, the function merely *uploads every existing file*, regardless of whether it's been uploaded already, and ignores files that have been deleted.

This is implemented by `console.utility.upload_files` (which should be better named).  Currently, this uses `google.cloud.storage` and doesn't use `physionet.gcs`, but it looks like we could probably port it to use `physionet.gcs` without much difficulty.

This functionality might partially overlap with the need to copy files *between two buckets* (as when publishing or creating a new version), which is currently done by `physionet.gcs.GCSObject.cp_dir_content`.  So it might be worth considering having one function to handle both cases.

tompollard:
Thanks for raising this Benjamin. Tagging @Chrystinne because this might be a good starting point for some of the cloud storage refactoring.

bemoody:
Note that S3 doesn't store a conventional hash of the file (unlike GCS, AFAIK.)  Currently, the ETag in S3 is calculated by something approximately like this:
```
def aws_hash(fileobj, chunksize=8*1024*1024):
    meta_hash = hashlib.md5()
    n = 0
    while True:
        part = fileobj.read(chunksize)
        part_hash = hashlib.md5(part)
        if n > 0 and not part:
            break
        meta_hash.update(part_hash.digest())
        n += 1
    if n == 1:
        return part_hash.hexdigest()
    else:
        return '{}-{}'.format(meta_hash.hexdigest(), n)
```

PR #2289, opened 2024-09-12
Optimize S3 re-upload

bemoody:
The "Create AWS bucket and send files" button currently causes *every* file in the project to be uploaded.  Instead, it's better to check whether the destination files already exist, and only upload the files that are necessary.

- If the background task gets interrupted and restarted (e.g., because we deployed a new version to the server) then it can avoid repeating a lot of the work that was done previously.

- With care, we should be able to upload a large database to S3 from a secondary server (faster than uploading from the live server); after that's finished, click the "send files" button and it will simply check that everything was transferred correctly.

There are a bunch of caveats here; this only partially addresses the issues I raised in #1903 and #2103, but this is a modest change that should significantly speed up the ongoing process of releasing data on S3.

Limitations to note:

- We still do not ever *delete* files from S3.  If a file is removed from the published project, "send files to AWS" **will not** remove that file from the S3 bucket.  If the deletion was intentional, it'll have to be removed from S3 (and GCS) manually.  This is a long-standing issue with both GCS and S3.

- Listing files per subdirectory (using `Delimiter`) isn't as efficient as listing all the files at once.  This is a compromise to avoid downloading a potentially huge amount of data from S3 all at once and keeping it in memory.  However, this could be improved by using `sorted_tree_files` instead of `os.walk`.

- And of course this only addresses S3, not GCS; it would be better to do this in a way that's portable across cloud providers.

### DIFF
--- a/physionet-django/project/cloud/s3.py
+++ b/physionet-django/project/cloud/s3.py
@@ -586,7 +586,10 @@ def send_files_to_s3(folder_path, s3_prefix, bucket_name, project):
         raise ValueError("AWS_PROFILE is undefined. Please set it in your settings.")
 
     s3 = create_s3_client()
-    for root, _, files in os.walk(folder_path):
+    for root, subdirs, files in os.walk(folder_path):
+        subdirs.sort()
+        files.sort()
+
         if root == folder_path:
             dir_prefix = s3_prefix
         else:
--- a/physionet-django/project/cloud/s3.py
+++ b/physionet-django/project/cloud/s3.py
@@ -1,18 +1,20 @@
+import json
+from math import ceil
+import os
+import re
+
 import boto3
 import botocore
-import re
-import os
-import json
+from botocore.exceptions import ClientError
 from django.conf import settings
+from django.db.models import Q
+
+from project.authorization.access import can_view_project_files
 from project.models import PublishedProject, AWS, AccessPolicy, AWSAccessPoint, AWSAccessPointUser
 from user.models import (
     User,
     CloudInformation
 )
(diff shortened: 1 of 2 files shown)


bemoody:
Draft because I want to wait for the current batch of S3 uploads (~120 GB) to finish, so I can test with a small project.

bemoody:
@Chrystinne can you please take a look at this?

A good way to test it is:
- create a demo AWS bucket
- upload a project such as `demoecg` to the bucket
- manually add another file to the `static/demoecg/10.5.24` directory
- upload the project to AWS a second time

You should see that the new file is added to the bucket, but the *modification times* of the existing files don't change.

Chrystinne:
@bemoody I'm running the tests by following these steps:

1. I created a new credentialed-access project named creden-20.
2. Published and submitted it to AWS.
However, when managing the new project in the admin console, I see the following:

AWS
The files have been sent to AWS. The bucket name is: creden-20.

However, this is what we should see instead:

AWS
The files have been sent to AWS. The bucket name is: controlled-data-dev-server.

Also, the bucket named creden-20 is not being created on AWS S3, either inside controlled-data-dev-server as it should be or elsewhere.

Chrystinne (approved):
@bemoody I've tested the re-upload and it looks great!

We still need to address the issue below, though:

When re-uploading files to AWS S3 through PhysioNet, the deleted files on the server aren't deleted from S3 buckets with the current re-upload functionality.

I've opened an issue for this: https://github.com/MIT-LCP/physionet-build/issues/2482

We can merge your PR and address that in a new PR.

PR #2290, opened 2024-09-13
Disallow assigning a project author as the editor

bemoody:
We want to avoid assigning a project author as the editor, to avoid the appearance of a conflict of interest, and to ensure that there's always another set of eyes on the project before it's published.

I think these changes should cover all the possibilities, but it's been a couple of months since I wrote this, so take a careful look. :)

- When project is "awaiting editor assignment", trying to assign it to one of the authors should give an error.

- When project is "awaiting decision", trying to reassign it to one of the authors should give an error.

- When project is "awaiting author revisions", if the submitting author sends an author invitation to the editor, and they try to accept that invitation, that should give an error.  (They should be able to reassign to another editor and then accept the invitation.)

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -105,6 +105,20 @@ class AssignEditorForm(forms.Form):
             raise forms.ValidationError("Incorrect project selected.")
         return pid
 
+    def clean(self):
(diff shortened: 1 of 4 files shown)


tompollard:
Thanks!

tompollard (approved):

PR #2297, opened 2024-09-19
Fix loading/clearing of media files

### LINKED ISSUE #2247: loaddemo installing weird "ethics" file
The file `demo-files/media/ethics/Ethics_Approval_567b029d-9ea6-41b8-b738-bf45675b24ce.txt` was added in pull #2241.

For some reason, `loaddemo` is copying this file, not to `media/ethics/Ethics_Approval_567b029d-9ea6-41b8-b738-bf45675b24ce.txt`, but to `media/ethics`.

This clearly doesn't work, and also breaks `resetdb`.

In contrast, `TestMixin` apparently installs the file in the correct place in the *testing* media directory - which is why test_urls works.

`TestMixin` should be doing the same thing as `loaddemo` when setting-up demo files, and it should be doing the same thing as `resetdb` when tearing-down.  (That said, it'd be good to avoid doing this setup/teardown for each individual test function.)

bemoody:
Currently, `manage.py loaddemo` incorrectly creates a file (not a directory) called `media/ethics`.

Even worse, `manage.py resetdb` currently crashes if that file exists.

Fixes #2247

### DIFF
--- a/physionet-django/user/management/commands/loaddemo.py
+++ b/physionet-django/user/management/commands/loaddemo.py
@@ -99,13 +99,15 @@ def copy_demo_media():
     for subdir in os.listdir(demo_media_root):
         demo_subdir = os.path.join(demo_media_root, subdir)
         target_subdir = os.path.join(settings.MEDIA_ROOT, subdir)
+        os.makedirs(target_subdir, exist_ok=True)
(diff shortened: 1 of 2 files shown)


tompollard:
looks good to me, thanks

PR #2298, opened 2024-09-19
Rename demo credential-applications directory to training

bemoody:
The directory "media/credential-applications/" was used for training reports long ago.  Nowadays, training reports are uploaded to "media/training/" instead.  Make the demo files reflect that (and update the fixtures accordingly.)


tompollard:
thanks!

PR #2300, opened 2024-09-20
Require valid reference order for submission/publication

bemoody:
Many existing projects have an invalid order of entries in their References section, due to issue #2137.  "Invalid" means:

(a) One or more references have `order=None`, typically because it was copied from a past published version.

(b) Two references have the same `order` value, typically because a reference was deleted from the middle of the list and then a new one was added.

(c) `order` is distinct but doesn't match `id` order, typically because *two* references were deleted from the middle of the list and then a new one was added.

Whatever the reason for having an invalid order of references, it is *not feasible to fix the order automatically*, so we want to require the author/editor to review and fix it by hand before the project is published.

The demo project "MIMIC-III Clinical Database" currently has `order=None` for its three references.  With these changes, you should see:

- If you open the preview page, there is an error message "Order of references may be incorrect".

- If you open the Project Content page, there is a warning at the bottom of the page.

- If you add or edit references, and click "Save Description" (without checking the checkbox), the references should be saved but the warning should still be present.

- If you check the checkbox and then click "Save Description", the warning should go away.

- Alternatively, if you *delete* all three of the original references and then refresh the page, the warning should go away.

### DIFF
--- a/physionet-django/console/templates/console/copyedit_submission.html
+++ b/physionet-django/console/templates/console/copyedit_submission.html
@@ -54,6 +54,7 @@
           {% include "project/content_inline_form_snippet.html" with form=description_form %}
           {% include "project/content_inline_form_snippet.html" with form=ethics_form %}
           {% include 'project/item_list.html' with item="reference" item_label=reference_formset.item_label formset=reference_formset form_name=reference_formset.form_name add_item_url=add_item_url remove_item_url=remove_item_url %}
+          {% include "project/confirm_reference_order_form.html" %}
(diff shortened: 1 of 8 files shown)


tompollard:
@bemoody sorry, conflict with https://github.com/MIT-LCP/physionet-build/pull/2301. Please could you rebase on dev (or just fix in a new commit if you prefer)?

bemoody:
oops - should be fixed!

tompollard:
Nice, thanks!

PR #2303, opened 2024-09-25
Remove the ArchivedProject model.

tompollard:
This pull request removes the unused `ArchivedProject` model.

We now track the "Archived" status of projects on the ActiveProject model (see: https://github.com/MIT-LCP/physionet-build/pull/2149).

In https://github.com/MIT-LCP/physionet-build/pull/2170 we migrated all ArchivedProject content to ActiveProjects.


bemoody:
Thanks.  This just needs MIGRATE_AFTER_INSTALL=True.

Unfortunately after I tried to fix loaddemo in #2297, it's now broken in a different weird way. :/

tompollard:
> This just needs MIGRATE_AFTER_INSTALL=True.

Thanks, now added.

> Unfortunately after I tried to fix loaddemo in https://github.com/MIT-LCP/physionet-build/pull/2297, it's now broken in a different weird way. :/

That explains the error I was getting yesterday after trying to run `loaddemo`. Before running this command, I needed to clear the `media/active-projects/` directory.

PR #2305, opened 2024-09-27
More detailed instructions for configuring AWS CLI

bemoody:
Add more detailed instructions to the /settings/cloud/ page.

(Remember that the AWS stuff won't appear on that page unless you set `AWS_VERIFICATION_BUCKET_NAME` in `.env`.)

It's important to tell people that they need to enable the `AmazonS3FullAccess` policy.  Perhaps that part is obvious to people who use AWS regularly, but to me it is extremely strange and unintuitive that if you *create a user with the default settings*:
- That user is allowed to access S3 buckets in its own account, if the bucket's policy says it's okay.
- That user is *not* allowed to access S3 buckets in other AWS accounts, regardless of the bucket's policy.

(Pull #2292 addresses another facet of the same issue.)

### DIFF
--- a/physionet-django/user/templates/user/edit_cloud.html
+++ b/physionet-django/user/templates/user/edit_cloud.html
@@ -27,6 +27,27 @@
   <form action="" method="post">
     {% csrf_token %}
     {% if user.cloud_information.aws_verification_datetime %}
+      <hr>
+      <p>
+        To access {{ SITE_NAME }} resources via Amazon Web Services:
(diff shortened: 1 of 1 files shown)


bemoody:
The lack of documentation is really the only blocking issue for enabling AWS verification on PhysioNet, so I'd like to enable it once we make this change.

tompollard:
This looks great, thanks Benjamin.

PR #2318, opened 2024-10-23
Fixed the kwargs passing issue for adding event to the dataset.

Rutvikrj26:
This is a fix for a small bug - introduced while implementing the feature updating the allowed datasets while creating an event. 

It's a small change adding a `user` kwargs to the `eventDatasetForm`


matkaczmarek (approved):
Make sense, should fix problem with `KeyError: 'user'` in [this line](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/events/forms.py#L78)

PR #2321, opened 2024-12-02
[T-CAIREM 1136] Use affiliation attribute to automatically credential faculty users.

matkaczmarek:
[sc-1136](https://app.shortcut.com/tcairem/story/1136/add-automatic-credentialing-of-users-sourced-from-trusted-edugain-idps)

**What:**
Automatically apply additional permissions to the account based on information retrieved from identity providers. The more experienced the researcher is, the more permissions we should give them. We have to discuss this more with January once we know what information we can retrieve from edugain IDPs.

**Why:**
Enhance user experience.

**How:**
Add automatic credentialing for faculty members (training won't be applied automatically)
Reference: https://docs.feide.no/reference/schema/attributes/edupersonaffiliation.html#saml-attribute-edupersonaffiliation

### DIFF
--- a/physionet-django/sso/views.py
+++ b/physionet-django/sso/views.py
@@ -6,7 +6,7 @@ from django.contrib import messages
 from django.contrib.auth import authenticate
 from django.contrib.auth import login as auth_login
 from django.contrib.auth.tokens import default_token_generator
-from django.db import transaction
+from django.db import transaction, DatabaseError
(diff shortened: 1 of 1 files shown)


tompollard:
We are working on improving the credentialing workflow, and this includes better use of information available from integrated accounts. I'm closing for now, but we can reopen later if needed.

PR #2323, opened 2024-12-13
Fix install-pn-test-server for recent Git security issues

bemoody:
A recent-ish security update (git 1:2.39.5-0+deb12u1) broke the ability to clone a local directory that is owned by another user.  This broke `ssh-install-pn-test-server`:

```
Cloning into bare repository 'physionet-build.git'...
fatal: detected dubious ownership in repository at '/tmp/physionet-tmp.git'
To add an exception for this directory, call:

	git config --global --add safe.directory /tmp/physionet-tmp.git
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

### DIFF
--- a/deploy/test-server/install-pn-test-server
+++ b/deploy/test-server/install-pn-test-server
@@ -61,6 +61,9 @@ apt-get install -y \
 ################################################################
 
 if [ -n "$WFDB_REPOSITORY" ]; then
+    if [ -d "$WFDB_REPOSITORY" ]; then
+        git config --global --add safe.directory "$WFDB_REPOSITORY"
+    fi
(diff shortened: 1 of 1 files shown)


bemoody:
(I don't fully understand the original issues and I'm fuzzy about whether this is a Git problem or a Debian problem.  Regardless, though, I'm pretty annoyed about what seems to be a major regression in a stable package.  Obviously, the ability to clone/fetch repositories that you don't trust is fundamental to what Git is meant to do, and to the best of my knowledge Git *can* securely fetch untrusted repositories via (dumb)HTTP, yet somehow it's impossible to do the same thing via the local filesystem?)

tompollard (approved):

PR #2325, opened 2025-01-14
[T-CAIREM 1243] Hotfix issue with boolean conversion and use string comparison instead

matkaczmarek:
Unfortunately `enable_orcid_login` variable is passed as string and therefore it is evaluated to true showing orcid log in all the time.


bemoody (changes requested):
No, this is not right.  The templates were correct, it is `base.py` that is wrong.  The variable ORCID_LOGIN_ENABLED must be defined as a boolean value, not as a string.

matkaczmarek:
> No, this is not right. The templates were correct, it is `base.py` that is wrong. The variable ORCID_LOGIN_ENABLED must be defined as a boolean value, not as a string.

Good catch, I casted env variable to bool and it works this time. Thanks for this tip.

bemoody (approved):

Issue #710, opened 2019-08-27
Should merging to dev trigger Shippable to re-run tests on pull request branches?


tompollard:
Suggested by @lbulgarelli. Should merging a branch to dev trigger Shippable to re-run tests for all pull request branches? 

If the tests are not triggered upon merges to dev, then a pull request may appear to be passing tests, but may actually fail when merged (e.g. because migrations are out of sync).

tompollard:
It looks like we can configure hooks using the instructions at: http://blog.shippable.com/configure-web-hooks-to-trigger-continuous-integration

Oddly though, I don't see the "subscriptions" item in the menu. It definitely used to be there, but isn't now. Looks like other people have had this issue too: https://github.com/Shippable/support/issues/3850

I've checked the third party integration status for the github organization, which says "No restrictions" so this isn't the problem. Also tried resyncing as suggested, but I get a time out. Will try again later...

PR #2326, opened 2025-01-14
Configure workflows to allow use of merge queue

bemoody:
GitHub's "merge queue" provides a way to run tests at the time a
branch is merged (i.e. tests run on the exact tree that is pushed to
dev; if the tests fail it doesn't get pushed.)

I don't know how well this will work, but it seems like it's worth
trying.

In order to do this, both the workflows and the repository settings
need updating.  Documentation is here:
- https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue
- https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#merge_group

### DIFF
--- a/.github/workflows/docker-build-test.yml
+++ b/.github/workflows/docker-build-test.yml
@@ -17,6 +17,9 @@ on:
   pull_request:
     branches:
       - dev
+  merge_group:
+    types:
+      - checks_requested
   workflow_dispatch:
     inputs:
       publish_tag:
--- a/.github/workflows/physionet-build-test.yml
+++ b/.github/workflows/physionet-build-test.yml
@@ -9,6 +9,9 @@ on:
   pull_request:
     branches:
       - dev
+  merge_group:
+    types:
+      - checks_requested
 
 jobs:
   test:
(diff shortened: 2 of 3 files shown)


bemoody:
@tompollard, I'd appreciate your thoughts.  Hard to know if this will cause problems until we try it, but this addresses an old issue (#710)

tompollard:
> @tompollard, I'd appreciate your thoughts. Hard to know if this will cause problems until we try it, but this addresses an old issue (https://github.com/MIT-LCP/physionet-build/issues/710)

I like the idea of having a system that helps to make sure that tests are run on the latest version of the dev branch.

I took a quick read of https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue and I don't fully understand how the queues are managed, but I'd be up for giving it a try.

I'd also be okay with the following "Require branches" option as an alternative:

> The merge queue provides the same benefits as the Require branches to be up to date before merging branch protection, but does not require a pull request author to update their pull request branch and wait for status checks to finish before trying to merge.

tompollard (approved):

PR #2327, opened 2025-01-14
Adding fixtures for setting up load tests on HDN

Rutvikrj26:
T-CAIREM is deploying an end-to-end testing pipeline for load-testing the platform. We are adding 100 load test users to simultaneously access and modify resources on the platform. 

The PR makes the following changes:
- Adds a `demo-load-test-users.json` containing fixtures for user, profile, training-type, credential application, etc required for each of the 100 users
- Adds a command in the sequence of loaddemo to load all the data from the above fixture.


bemoody (changes requested):
You can't load a fixture containing "environment.cloudidentity" objects if the "environment" app isn't enabled.  We need to ensure that the loaddemo command works regardless of the ENABLE_CLOUD_RESEARCH_ENVIRONMENTS setting.

Probably the best thing would be:
- Move the objects for the *user* app (`"user.user"`, `"user.associatedemail"`, etc.) from `load-test-users.json` into `physionet-django/user/fixtures/demo-user.json`
- Move the `"environment.cloudidentity"` entries into a separate file called `physionet-django/environment/fixtures/demo-environment.json`

Alternatively you could load the entire `load-test-users.json` conditionally, only if `ENABLE_CLOUD_RESEARCH_ENVIRONMENTS` is true, but that will make it more complicated to manage the fixtures in the future.

Rutvikrj26:
Thanks! I completely missed that this was failing. 
I'll move the user related fixtures in the `demo-users.json` and the environment related ones into to a new folder as you suggested.

Rutvikrj26:
I moved the fixtures related to environment into the actual hdn-research-environment package, and kept the `loadtest-user` & `demo-user` separate. The intent was to keep the demo data clean and not bloat the actual demo data with 200 load test users, and hence I've kept the entire block the loads the loadtest data optional.

bemoody:
Looking at the data, it looks like it is programmatically generated and it would be easy to write a function in loaddemo.py to create the objects programmatically.  For maintenance purposes, that would much better than dealing with a bunch of JSON files.

(The demo files are a mess and the way of handling them badly needs improvement.  Adding additional JSON files - that are liable to conflict with the demo files - would make it harder, rather than easier, to make those improvements.)

bemoody:
Also, do not merge dev into a feature branch.  Rebase instead.

Rutvikrj26:
I get your point about the conflicts. 
To generate the users for load tests in a way they don't conflict with the default demo data is to have primary keys that come after all the user data. Since the current user data is in json format, the values have to be manually put into the script. For eg. 209 is where the users start from. So, we either reform the entire thing, or it's kinda dangerous to add a bunch of magic numbers into that generation code. This would make it more complicated to maintain (partly generated and partly json files).

Reframing the entire way in which the demo data is handled is beyond the scope of this pr, and we need to get onto the load testing, for the february datathon. So, I'm suggesting we merge this pr, and then open an issue to rework how the loaddemo command.

bemoody:
That's just the point - by generating the users programmatically you *avoid* hardcoding those magic numbers.  For example:

```
    for i in range(1, 201):
        user = User.objects.create(
            password="pbkdf2_sha256$600000$mazq87qhgvhG0GZ27nGLb6$Yj2uxK4FkIhU5iiKqa913dc84XLec1IRTmleRHfalW4=",
            is_superuser=False,
            email=f"load_test_user_{i}@example.com",
            username=f"loadtestuser{i}",
            sso_id=None,
            join_date="2025-01-09",
            last_login="2025-01-09T17:50:00Z",
            registration_ip=f"127.0.0.{i}",
            is_active=True,
            is_admin=False,
            is_credentialed=True,
            credential_datetime="2025-01-09T18:14:00Z",
        )
        profile = Profile.objects.create(
            user=user,
            first_names=f"Load{i}",
            last_name="Tester",
            affiliation="",
            location="",
            website="",
            photo="",
        )
        # [You should be able to follow the same pattern to create
        # CredentialApplication, Training, etc.  You can skip
        # AssociatedEmail because that's created automatically.]
```

That way, we have perhaps 150 lines of code to maintain, rather than 22,000 lines of code.

bemoody:
But even if it were *really necessary* to generate a 22,000-line JSON file, I would much prefer that you commit a 300-line script instead.  Even if that script is fragile, kludgy, and poorly documented.

Rutvikrj26:
Well, two points:

1. The idea is to not mix & match the way in which these are handled. Handling half of data by generating it and the other half using json files is a very bad idea (atleast I think so) since that makes it confusing and error prone. For Eg. consider the scenario, a user went to the fixtures, and just added a new user with a new pk and other models, without knowing that there's generation happening in the loaddemo command file - it can be a conflict. I want to avoid this kind of mix and match. Hence, my suggestion to open a new issue that shifts everything to generation based code rather than json, keeping things uniform.

2. (debatable, but) I don't think the pk values in the "json" file are "magic numbers" & "json" does not have lines of "code"

bemoody:
> For Eg. consider the scenario, a user went to the fixtures, and just added a new user with a new pk and other models, without knowing that there's generation happening in the loaddemo command file - it can be a conflict.

Sure, let's consider that scenario.  Suppose developer Charlie wants to add a new demo user, as part of a new test case.  Based on looking at `demo-user.json`, the obvious thing to do would be to add a new `"user.user"` entry to that file, setting its `"pk": 209`.

If there is a `loadtest-users.json` file and a `loadtest-identities.json` file, as you propose, what conflicts would arise?  What would Charlie need to do to resolve those conflicts?

If there is no `loadtest-users.json` or `loadtest-identities.json` file, but rather a Python function as I propose, what conflicts would arise?  What would Charlie need to do to resolve those conflicts?

Let me know if I have misunderstood what you mean by conflicts.

Rutvikrj26:
If the loadtest-users live in the same directory as the `demo-users.json`, then it becomes apparent that there are users apart from the ones listed in the `demo-users.json`, whereas if they are in the script, it is not that apparent. Charlie, with some basic understanding of the file structure and a sense of the project knows that he needs to keep in mind the users in the `demo-users.json` as well as the ones in the `loadtest-users.json`. 

However, if the users generated from the script, they are hidden away in the loaddemo command - it is not that apparent for a new developer who is not aware about how the custom loaddemo command functions. Charlie will not expect that are hidden "generated" users when the rest of the demo data is all loaded up through json fixture files. What I meant by conflict was the way the demo data is made available to the system (either in json files or generated at run time).

Rutvikrj26:
Summarizing the discussion today morning with tom, 
The PR needs to be broken down into two PRs - one for bumping the hdn-research-enironment version, and another handling the addition of the load-test user fixtures into the loaddemo command. 

Taking on the suggestion to use a different variable specific for loadtests, I am planning to add a new env variable `ADD_LOADTEST_FIXTURES` which defaults to false, and can be set to enable to add the `loadtest-users.json` fixtures. Since this requires change from the very early commits - it did not make much sense to rebase, and hence I'm closing this PR, and will be opening up two new PSs.

PR #2333, opened 2025-01-28
Feature: Add server side cursors option

rafalkosla101:
This option is responsible for a specific django behaviour about querying the database, if enabled the database calls will be sectioned and fetched in a few smaller calls, to which we need constant database connection. We would like to change our database management tool(pgbouncer) to transaction mode and to do that we would need to disable that default django behaviour. The implementation only adds an option to change it which defaults to the same behaviour as it is now. We will change that on our side by env variables

### DIFF
--- a/physionet-django/physionet/settings/production.py
+++ b/physionet-django/physionet/settings/production.py
@@ -18,7 +18,7 @@ DATABASES = {
         'PASSWORD': config('DB_PASSWORD'),
         'HOST': config('DB_HOST'),
         'PORT': config('DB_PORT', default=''),
-        'DISABLE_SERVER_SIDE_CURSORS': config('DISABLE_SERVER_SIDE_CURSORS', default=False),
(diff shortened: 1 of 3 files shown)


physionet-django/physionet/settings/production.py
```diff
@@ -18,6 +18,7 @@
         'PASSWORD': config('DB_PASSWORD'),
         'HOST': config('DB_HOST'),
         'PORT': config('DB_PORT', default=''),
+        'DISABLE_SERVER_SIDE_CURSORS': config('DISABLE_SERVER_SIDE_CURSORS', default=False),
```
matkaczmarek: I would add `cast=bool` as for example in here: https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/settings/base.py#L47

matkaczmarek (approved):

tompollard (approved):

PR #2335, opened 2025-01-28
Added the loadtest fixtures files & Updated the loaddemo Command

Rutvikrj26:
This PR introduces two changes:

- Adds a new environment variable that is used for adding loadtest data when we want to perform load tests.
- Updates the loaddemo command to introduce a conditional block at the end of the handle - loading the loadtest fixture data if the `ADD_LOADTEST_FIXTURES` is set in environment variables.


bemoody:
My understanding from the previous meeting is that this is not needed.

PR #2340, opened 2025-02-03
Do not promote commercial social media

bemoody:
It's not necessary or productive for PhysioNet to advertise, endorse, or encourage the use of commercial social media.  Furthermore, these buttons add clutter and provide negligible value.  Let's not do that.


tompollard:
I'm not against removing these, but I do have a slightly different perspective to you. I think it is our role as a publisher to help distribute knowledge with the community (including making it simple for people to share through social media). It is standard practice for most publishing platforms. Let's discuss tomorrow!

PR #2345, opened 2025-02-06
Fix CSS kludge that breaks JavaScript modal boxes

bemoody:
An old kludge by Lucas Bulgarelli (to make the splash-screen fill the browser window) currently interferes with CKEditor modal boxes and will interfere with TinyMCE modal boxes.

For example, open one of your "Project Content" pages, scroll to the bottom and in the CKEditor toolbar, click the "equation" button.  The equation editor modal box pops up, as expected - but in the background, "behind" the modal box, the page has scrolled back to the top.

The same problem and worse problems occur with TinyMCE.

I don't know why this is, and don't really care to debug it further.  It's likely to be browser-dependent.

But I think the best thing is not to mess around with the dimensions and overflow behavior of the html and body elements.  On current browsers, we can use `100vh` to get Lucas's desired effect, and it doesn't really hurt anything if the effect doesn't work on outdated browsers.

### DIFF
--- a/physionet-django/static/custom/css/physionet.css
+++ b/physionet-django/static/custom/css/physionet.css
@@ -1,6 +1,3 @@
-html,body{height:100%; width: 100%;}
-html{ overflow: overlay; }
-
 /* Open in new window icon */
 a[target='_blank']::after {
   content: '\29C9';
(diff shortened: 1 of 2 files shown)


bemoody:
See also pull #2324.

If you're trying to test this, remember that you need to run `manage.py compilestatic`.

tompollard (approved):
Looks good to me!

PR #2349, opened 2025-02-19
Add small UX adjustment and redirect to `login` page instead of `sso_login`

matkaczmarek:
When the ORCID user clicks the activate link it is activated and redirected to ORCID login. When the user clicks it again it is already activated and redirected to SSO login (EduGain) which is a bad UX as its account is linked with ORCID, not with EduGain.

This PR changes this behaviour and redirects to the login page.


rafalkosla101 (approved):

tompollard (approved):

tompollard:
Looks good, feel free to merge!

PR #2351, opened 2025-03-10
Implemented a model-level unique enforcement between Event & Users.

Rutvikrj26:
The Event Application model currently does not enforce one application per event per user. This PR fixes the problem by including a unique_together constraint between event & user on the event application model - allowing one user to apply to an event at most once.


bemoody:
Makes sense to me.

Is there a sequence of steps you can perform on the website that would result in two applications for same person?  (For example, loading the same page in two tabs and submitting the same form twice?)

Rutvikrj26:
Further adding the sequence to handle pre-existing event-user duplicate event application pairs in the database. 
Added a function in the migration to be run before adding the constraint, which fetches the duplicate pairs, and keeps only one using the following logic:
- If there was a decision made on one of the applications - only the one with decision is kept (most recent one if multiple with decision)
- if none of the applications had a decision made - the most recent one is kept.

I've tested the event application deletion functionality by adding dummy duplicate events via fixtures and then introducing  
 the migration - which successfully executed the above logic, removed redundant duplicates, and applied the unique constraint.

Rutvikrj26:
> Is there a sequence of steps you can perform on the website that would result in two applications for same person? (For example, loading the same page in two tabs and submitting the same form twice?)

This issue was surfaced during the datathon, when the server was not scaled and the system was running at full-capacity. Due to the lag in between the user submitting applications on the webpage & the database processing it, a user was able to submit multiple requests to join the event, and since there was no constraint on the database level restricting uniqueness between user & event in an event application, we ended up with duplicate applications. 


Theoretically, the method you suggested might work out. Sending two requests simultaneously might also trigger this.

bemoody:
No, now that I think about it, this does not sound right at all.

One person can only be a *participant* once - EventParticipant is already unique per event and user.

But I don't see a reason one person can't have multiple *applications*, and possibly good reasons for remembering if the same person applied to the same event in the past.

bemoody:
Please work out, in detail, exactly what the bug is that you're trying to fix - preferably with a precise sequence of steps to reproduce it.  Open an issue so that we can discuss how best to fix it.

If the ultimate fix for the bug involves enforcing that each person only ever has one event application, then it might additionally make sense to add a database constraint.  But a database constraint *by itself* is not enough to fix the bug, whatever it is.

PR #2355, opened 2025-03-13
Modify order of references in the "When using this resource" box

tompollard:
This is a minor change to keep the seas calm by modifying the order of references in the "When using this resource, please cite:" box at the top of the landing pages for the annual CinC challenges.


Chrystinne (approved):
Looks good to me!

PR #2356, opened 2025-03-13
Data Access Request form changes for Bridge2AI

Chrystinne:
This PR modifies the Data Access Request (DAR) form for the Bridge2AI project. The changes focus on adding new information collected during the data access request process.

Changes:

-   Renamed "Research Project Details" field to "Scientific Abstract"
-   Added a new optional "Lay Summary" field to collect non-technical project descriptions
-   Added database migration file for the model change


physionet-django/project/templates/project/data_access_request_view.html
```diff
@@ -87,6 +87,9 @@ <h3>Project Description</h3>
   <div class="alert alert-secondary">
     <h4>{{ access_request.data_use_title }}</h4>
     {{ access_request.data_use_purpose|safe }}
+    {% if access_request.lay_summary %}
+    {{ access_request.lay_summary|safe }}
```
tompollard: maybe indent for readability.

physionet-django/project/templates/project/data_access_request_table_snippet.html
```diff
@@ -21,6 +21,7 @@ <h2>Previous Requests</h2>
                     <td>{{ req.decision_datetime|date }}</td>
                     <td>{{ req.data_use_title }}</td>
                     <td>{{ req.data_use_purpose | safe }}</td>
+                    <td>{% if req.lay_summary %}{{ req.lay_summary | safe }}{% endif %}</td>
```
tompollard: I think in general, even if data is not provided I think it is better to display the field to the user (so that they know it was not provided). Otherwise they have a form that changes based on the application.

physionet-django/project/templates/project/data_access_request_status_detail.html
```diff
@@ -8,5 +8,8 @@ <h1>Data Access Request - {{ project }}</h1>
         <h2>{{ access_request.data_use_title }}</h2>
         <small class="text-muted">submitted on {{ access_request.request_datetime }}{% if access_request.is_accepted %}, valid {% if access_request.valid_until %}until {{ access_request.valid_until }}{% else %}forever {% endif %}{% endif %}{% include 'project/data_access_request_badge.html' with data_access_request=access_request %}</small>
         <p>{{ access_request.data_use_purpose|safe }}</p>
+        {% if access_request.lay_summary %}
```
tompollard: even if data is not provided I think it is better to display the field to the use

physionet-django/project/modelcomponents/access.py
```diff
@@ -75,6 +75,7 @@ class DataAccessRequest(models.Model):
 
     data_use_title = models.CharField(max_length=200, default='')
     data_use_purpose = SafeHTMLField(blank=False, max_length=10000)
+    lay_summary = SafeHTMLField(blank=True, null=True, max_length=5000)
```
tompollard, line 78: Can you double check that we want blank = True and null = True here? My understanding is that this is sometimes discouraged because it can lead to confusing query results (e.g. where you query for empty strings and the query doesn't return results where lay_summary is null).

physionet-django/project/forms.py
```diff
@@ -1123,6 +1125,9 @@ def __init__(self, project, requester, template, *args, **kwargs):
         self.project = project
         self.requester = requester
 
+        # Make lay_summary optional in the form
+        self.fields['lay_summary'].required = False
```
tompollard: I think we should make this a required field because it is probably a good thing for everyone.

tompollard (commented):
thanks! added a few minor comments.

physionet-django/project/modelcomponents/access.py
Chrystinne: Thanks for catching this, @tompollard. Here’s what I found:

In Django models, **blank=True** allows the field to be optional in forms and **null=True** allows the database to store NULL as a value.

Having both **blank=True** and **null=True** means the field can either be an empty string ("") or NULL in the database, which can lead to inconsistent query results. For example:

   `MyModel.objects.filter(lay_summary="")`
This will not return objects where lay_summary is NULL, only those with an empty string.

   `MyModel.objects.filter(lay_summary__isnull=True)`
This will only return records where lay_summary is NULL, not those with an empty string.

Since this creates two different ways to represent "no data," Django best practices recommend:

For CharField or TextField, use **blank=True**, **null=False**, treating empty strings as missing values.
For ForeignKey or DateTimeField, use null=True, since empty values cannot be represented as an empty string.

**Suggested fix for our case:**
Since lay_summary is a string-based field (SafeHTMLField), it’s best to use:

   `lay_summary = SafeHTMLField(blank=True, null=False, max_length=5000)`

This ensures missing values are consistently stored as an empty string ("") and prevents query inconsistencies.

Chrystinne: However, since lay_summary will now be a required field, we will define it this way for consistency, as Django assumes null=False by default:

`lay_summary = SafeHTMLField(blank=False, max_length=5000)`

tompollard:
The error is: `sqlite3.IntegrityError: NOT NULL constraint failed: project_dataaccessrequest.lay_summary` (I assume because we are now requiring data in a column that wasn't populated in the past). We could add a default value (e.g. "Summary not provided" or "N/A") or require the field in the form but not enforce this requirement on the model. 

At the moment it looks like `blank=False` on the model, but the default value in the migration is `""`. I think it would be fine just to set `blank=True`.

bemoody:
**Do not merge until issue #2357 is resolved.**

Chrystinne:
We also need to resolve issue #2359

tompollard (approved):

Issue #2362, opened 2025-03-19
Misreporting filesize for ALOTT dataset


tompollard:
The filesize of the ALOTT dataset is misreported as 0B:
https://physionet.org/content/alott/1.0.0/telemetry-data/#files-panel

![Image](https://github.com/user-attachments/assets/234a1680-fd3c-4a0e-a7b5-94ed1f6e3dcc)

tompollard:
I have fixed the value in the database. I assume we bypassed the calculation step in our workflow when we manually transferred the files.

bemoody:
The actual problem here is:

1. The total file size is not calculated until after the project is published.  (This is issue #519.)

2. The published project page displays the size as zero bytes if it hasn't yet been calculated.

> I have fixed the value in the database.

Please don't do that.  Now I don't know whether the value currently stored in the database is the value that you wrote or the value that the server calculated.

PR #2363, opened 2025-03-20
Hide "total uncompressed size" if unknown

bemoody:
If a published project's download size is not yet known, the site displays "Total uncompressed size: 0 B."

This is obviously wrong and makes it look like the server is broken.  But in fact, it is *expected* that the size is unknown when the project is first published.

This fixes issue #2362.


tompollard:
Nice, thanks!

tompollard (approved):

Issue #2359, opened 2025-03-18
Error when 'Research Project Details' is NULL in Data Access Request Form (Contributor Review Datasets)


Chrystinne:
I encountered an error when submitting a data access request for "HiRID, a high time-resolution ICU dataset" (https://physionet.org/content/hirid/1.1.1/)  while leaving the "Research Project Details" field empty. 

Steps to Reproduce:

- Go to the data access request form for HiRID v1.1.1 (https://physionet.org/request-access/hirid/1.1.1/);
- Fill in the required fields except for "Research Project Details" (leave it NULL);
- Submit the form.

tompollard:
```
Traceback:

File "/physionet/python-env/physionet/lib/python3.11/site-packages/django/core/handlers/exception.py" in inner
 55.                 response = get_response(request)

File "/physionet/python-env/physionet/lib/python3.11/site-packages/django/core/handlers/base.py" in _get_response
 197.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/physionet/python-env/physionet/lib/python3.11/site-packages/django/contrib/auth/decorators.py" in _wrapper_view
 23.                 return view_func(request, *args, **kwargs)

File "/home/physionet/physionet-build/physionet-django/project/views.py" in request_data_access
 2093.                   {'project_request_form': project_request_form,

Exception Type: UnboundLocalError at /request-access/hirid/1.1.1/
Exception Value: cannot access local variable 'project_request_form' where it is not associated with a value
Request information:
```

PR #2364, opened 2025-03-20
Fix UnboundLocalError in DataAccessRequest form submission

Chrystinne:
This PR fixes issue #2359 where users encounter an UnboundLocalError when submitting a data access request form for contributor review access type projects while leaving the "Research Project Details" field empty.

Fixed the request_data_access function to ensure project_request_form is always defined before being referenced.


bemoody (approved):

tompollard (approved):

tompollard:
(Looks like we need to fix https://github.com/MIT-LCP/physionet-build/issues/2366 before merging)

PR #2371, opened 2025-03-27
Increase maximum length of news item titles from 100 to 150 characters

tompollard:
We are required by the NIH to post a news item with a title that is ~110 characters long. This pull request increases the maximum length of news item banners to 150 characters.


Chrystinne (approved):
Looks good to me.

PR #2375, opened 2025-03-31
Added the functionality to record a GCP Billing Id attached to an event in the event model

Rutvikrj26:
This feature allows addition of the gcp billing id to events for a more streamlined workflow of attaching shared billing ids to the cloud identities of the users participating in the events. This addresses the t-cairem [issue #18](https://github.com/T-CAIREM/physionet-build/issues/18)

The following key changes have been implemented:
- Added a `gcp_billing_id` field in the events model & the migration for it. 
- Added a conditional rendering of Billing Accounts to choose from if the user has a cloud identity (only HDN users have cloud identities currently)
- Implemented signals in the hdn-research-environment package to handle the changes in the EventApplication & CloudIdentity.

Another part to be implemented alongside for the feature to work is : Bumping the hdn-research-environment-package(#2377).


bemoody (dismissed):
You need to explain the problem in a lot more detail.  Do not explain what you're proposing to do, explain **why**.  Open an issue first so that we can discuss possible solutions and implementation strategies.  Open a pull request only after there is a consensus about what the strategy should be.

It is unlikely that I will approve of changes whose functionality is implemented by an undocumented API that you are developing in private.  New features should be developed in the open (meaning, there is a public record of the discussions, including the motivation for the feature and the reasoning behind the implementation.)

Furthermore:
- Do not make unrelated changes, such as fixing typos.  Review your own diffs before committing and before pushing.
- Do not merge dev into a feature branch.
- There is never a good reason to write `except Exception: pass`.

Rutvikrj26:
> You need to explain the problem in a lot more detail. Do not explain what you're proposing to do, explain **why**. Open an issue first so that we can discuss possible solutions and implementation strategies. Open a pull request only after there is a consensus about what the strategy should be.

We discussed during one of the tuesday meetings and is documented in a [issue](https://github.com/T-CAIREM/physionet-build/issues/18).

> * Do not make unrelated changes, such as fixing typos.  Review your own diffs before committing and before pushing.

I fixed typo in a comment as I felt it did not deserve an issue-pr of it's own. However, I can revert it back if you'd like that. 

> * There is never a good reason to write `except Exception: pass`.

Good point - I've added logging statement to be able to debug in case there is an unknown issue in the `research-environment-api`

tompollard (approved):
Thanks @Rutvikrj26. I'm okay merging this, but as discussed I think we need to improve the way we handle custom functionality (i.e. in this case, functionality that requires the HDN package). 

Towards this goal, I think it is a good idea to make some progress on moving the repository codebase into a separate entity (so that both HDN and PhysioNet are equal, downstream projects).

Issue #2242, opened 2024-05-22
Too many open files


bemoody:
The project file upload handler will fail (EMFILE: Too many open files) if there are too many files included in the upload.

"DATA_UPLOAD_MAX_NUMBER_FILES" is set to 1000, but the process's FD limit is presumably set to the default of 1024.  If we want to allow 1000 files at a time, we should raise the rlimit.

Also note that the inline script in edit_files_panel.html doesn't check if the number of files will be permitted.

bemoody:
Having 18 to 20 FDs open when idle seems typical:
```
$ sudo lsof -a -c uwsgi -d 0-999999999 | awk '{print $2}' | uniq -c
      1 PID
      6 485021
     38 485022
     18 815631
     20 815844
     19 815877
     20 816536
     18 818355
     19 818500
     20 819132
     20 819671
     20 820690
     20 825215
```
The one process with 38 is presumably the "emperor" that is managing the others, and isn't handling requests itself.

bemoody:
See also pull #2036 which added the option.

We should have a custom upload handler, and not create all these temporary files.

bemoody:
keywords: ulimit, setrlimit, getrlimit, RLIMIT_NOFILE

tompollard:
@bemoody the 1000 files is arbitrary, so are you okay with me reducing DATA_UPLOAD_MAX_NUMBER_FILES as a temporary fix?

tompollard:
Just looking, and I see that this has been done already.

bemoody:
I changed it but I see that this isn't much of an improvement (it still gives a 500 error.)

Currently, it appears to be raising an exception inside the csrf middleware handler (because it must parse the post body to extract the token.)

We could possibly work around this by
- marking the view as "csrf_exempt"
- manually invoking "csrf_protect"
- catching the TooManyFilesSent exception
(similar to this: https://docs.djangoproject.com/en/5.1/howto/csrf/#protecting-a-view-for-only-one-path)

Another idea would be to use a secret URL to avoid the need for a CSRF token; I'm not completely sure if this can be done securely.

PR #2036, opened 2023-07-03
Allow DATA_UPLOAD_MAX_NUMBER_FILES and DATA_UPLOAD_MAX_MEMORY_SIZE to be configured

tompollard:
This pull request addresses https://github.com/MIT-LCP/physionet-build/issues/2035.  When users attempt to upload more than 100 files (which is fairly common when data contributors are submitting new projects) the following error is raised:

```
The number of files exceeded settings.DATA_UPLOAD_MAX_NUMBER_FILES.

TooManyFilesSent at /projects/SLUG/files/
The number of files exceeded settings.DATA_UPLOAD_MAX_NUMBER_FILES.
``` 

The "100" file limit is Django's default (see: https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-DATA_UPLOAD_MAX_NUMBER_FILES). 

This pull request:

1) Allows the `DATA_UPLOAD_MAX_NUMBER_FILES` to be configured in the .env file.
2) Also allows `DATA_UPLOAD_MAX_MEMORY_SIZE` to be configured in the .env file.

PR #2376, opened 2025-04-04
Show error message if too many files are selected for upload

bemoody:
Django's (default) file upload mechanism is limited in the number of files it can handle (independent of limitations on file size.)

Consequently, Django enforces a hard limit on the number of files and raises an exception if this is exceeded.  (Pull #2036 made this limit configurable.)

Note that HTML doesn't allow specifying any limit for multiple-file upload fields.

Try to prevent people hitting the limit by accident, by disabling the submit button through JavaScript if too many files are selected.

For thoughts about better long-term solutions, see issue #2242.


tompollard (approved):
Thanks, looks good to me

PR #2378, opened 2025-04-07
Update project serializer to include publish_date, doi, version

tompollard:
We have an API at `http://localhost:8000/api/v1/project/published/search` that allows users to export metadata about published projects. 

The API is potentially useful for people who are developing tools that interface with PhysioNet content (e.g. the MEDS project: https://github.com/Medical-Event-Data-Standard.

The metadata returned from queries is defined in `PublishedProjectSerializer`. This pull request updates this serializer to include the following fields:

- `version` (important, because otherwise a search returns what look like duplicate results, one result for each version).
- `core_doi`
- `version_doi` -> (from `doi`)
- `is_latest_version`
- `short_description`
- `publish_date`

Also adds a citation string to the `ProjectVersionsSerializer` method.

### DIFF
--- a/physionet-django/export/serializers.py
+++ b/physionet-django/export/serializers.py
@@ -23,11 +23,20 @@ class DUASerializer(serializers.ModelSerializer):
 class PublishedProjectSerializer(serializers.ModelSerializer):
     license = LicenseSerializer()
     dua = DUASerializer()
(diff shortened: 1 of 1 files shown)


Chrystinne (approved):
Looks good!

PR #2379, opened 2025-04-09
Make training an optional field

Chrystinne:
This PR allows the waiving of training submissions for credentialed access type. This modification has been implemented to support the release of the Bridge2AI project.

The PR modifies the handling of required trainings for credentialed access policy in the `AccessMetadataForm`.

The empty_field option doesn't work for the current AccessMetadataForm as it triggers the error "" is not a valid value (see image below):

Issue_PR

To fix that issue, we needed to add a custom `CustomModelMultipleChoiceField` to allow 'No training required' as a valid option to be selected by the authors when no training is required.  

All the following tests passed:

1. Creation and publication of a new project P1 with credentialed access type with no training required, and testing the access with different types of users: 
   - User 1 with **credentialing = False** and **training = False**
   - User 2 with **credentialing = True** and **training = True**
   - User 3 with **credentialing = True** and **training = False**
   - User 4 with **credentialing = False** and **training = True**

2. Creation and publication of a new project P2 with credentialed access type with CITI training required, and testing the access with different types of users: 
   - User 1 with **credentialing = False** and **training = False**
   - User 2 with **credentialing = True** and **training = True**
   - User 3 with **credentialing = True** and **training = False**
   - User 4 with **credentialing = False** and **training = True**

3. Creation and publication of a new version of project P1 keeping the same information as the previous version, and testing the access with different types of users: 
   - User 2 with **credentialing = True** and **training = True**
   - User 3 with **credentialing = False** and **training = False**

4. Creation and publication of a new version of project P2 keeping the same information as the previous version, and testing the access with different types of users:  
   - User 2 with **credentialing = True** and **training = True**
   - User 3 with **credentialing = False** and **training = False**

5. Creation and publication of a new version of project P1 changing the information of the previous version (i.e., required training information), and testing the access with different types of users: 
   - User 2 with **credentialing = True** and **training = True**
   - User 3 with **credentialing = False** and **training = False**

6. Creation and publication of a new version of project P2 changing the information of the previous version (i.e., required training information), and testing the access with different types of users:  
   - User 2 with **credentialing = True** and **training = True**
   - User 3 with **credentialing = False** and **training = False**

### DIFF
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -42,6 +42,7 @@ from project.models import (
 )
 from user.models import User, TrainingType
 from user.validators import validate_affiliation
+from django.forms import ModelMultipleChoiceField
 
 INVITATION_CHOICES = (
     (1, 'Accept'),
@@ -924,11 +925,24 @@ class AccessMetadataForm(forms.ModelForm):
             access_policy=self.access_policy
         )
 
-        if self.access_policy not in {AccessPolicy.CREDENTIALED, AccessPolicy.CONTRIBUTOR_REVIEW}:
+        if self.access_policy in {AccessPolicy.OPEN, AccessPolicy.RESTRICTED}:
             self.fields['required_trainings'].disabled = True
             self.fields['required_trainings'].required = False
             self.fields['required_trainings'].widget = forms.HiddenInput()
             self.initial['required_trainings'] = ''
+
+        if self.access_policy == AccessPolicy.CREDENTIALED:
+            self.fields['required_trainings'].disabled = False
+            self.fields['required_trainings'].required = True
+
+            # Replace the original field with a custom version
+            original_field = self.fields['required_trainings']
+            custom_field = CustomModelMultipleChoiceField(
+                queryset=original_field.queryset.order_by('name'),
+                required=True,
+                widget=original_field.widget
+            )
+            self.fields['required_trainings'] = custom_field
 
         if self.access_policy == AccessPolicy.OPEN:
             self.fields['dua'].disabled = True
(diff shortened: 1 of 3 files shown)


physionet-django/project/forms.py
```diff
@@ -924,12 +925,25 @@ def __init__(self, *args, **kwargs):
             access_policy=self.access_policy
         )
 
-        if self.access_policy not in {AccessPolicy.CREDENTIALED, AccessPolicy.CONTRIBUTOR_REVIEW}:
+        if self.access_policy in {AccessPolicy.OPEN, AccessPolicy.RESTRICTED}:
```
tompollard, line 928: The logic here could be clearer. It's unclear how the current set of statements are organised. My preference would be a set of if else statements, working through each access policy, e.g.

if open: do x
elif restricted: do y
etc

Chrystinne:
@tompollard, I've modified the code to preselect the "No training required" option when the field is empty or was previously set to no training. 

Now it appears like this when editing a project that does not require training:

Screen Shot 2025-04-09 at 3 46 42 PM

tompollard:
Looks good to me @Chrystinne. One minor thing is that the published project includes an empty "Training required:" section.

Screenshot 2025-04-11 at 4 46 17 PM

This section should either:

1. Not be displayed if there are no training requirements, or
2. It should say something like "None required" or "No training required".

Chrystinne:
@tompollard Thanks, Tom. I've updated this section to display "No training required" if there are no training requirements:

Screen Shot 2025-04-14 at 4 26 02 PM

tompollard (approved):
Thanks Chrystinne, looks good to me!

Issue #2380, opened 2025-04-10
Allow users to generate tokens for managing API access


tompollard:
I would like to add support for tokens to enable secure, programmatic use of protected API endpoints (e.g. downloading credentialed datasets or querying metadata), without prompting for login credentials.

Once support for tokens has been developed, we can then add features to our currently-unused [Python package](https://github.com/MIT-LCP/physionet) that make use of them for tasks like:

- Searching for PhysioNet projects
- Loading metadata and datasets

There have been requests for this functionality from the [MEDS developers](https://github.com/Medical-Event-Data-Standard). Use of tokens is fairly common across similar data sharing platforms (e.g. HF, GitHub, etc). 

### Proposed steps for implementation

My plan would be to implement this in the following steps:

1. Add interface for users to create tokens (but hide this interface from users). Each token should:
    - Be tied to a user
    - Expire automatically after a fixed duration (e.g. 60 days)
    - Be revocable via the UI
2. Add token-based authentication to the backend. 
    - Accept `Authorization: Token <token>` headers
    - Check that the token is valid and not expired
    - Permit access to protected endpoints if the user is authorized
 3. Add features to the [PhysioNet Python package](https://github.com/MIT-LCP/physionet) that make use of the tokens, e.g.
     - Download a list of files and checksums (requested by MEDS).

PR #2381, opened 2025-04-10
WIP: Add interface for generating access tokens (ref #2380)

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/2380, I would like to add support for tokens to enable secure, programmatic use of protected API endpoints (e.g. downloading credentialed datasets or querying metadata), without prompting for login credentials.

This pull request is part 1 of the implementation plan. It adds an interface for users to create tokens. Each token: is tied to a user; expires after 60 days; and is revocable via the UI. For now, the interface is hidden from users.


tompollard:
@bemoody I'm looking into whether we can use OAuth tokens, instead of generating our own. I'll make this a WIP for now.

tompollard:
Superseded by https://github.com/MIT-LCP/physionet-build/pull/2389, which is similar functionality but using OAuth tokens.

Issue #2383, opened 2025-04-10
Updating the User information returned in OAuth Response


Rutvikrj26:
When calling /oauth/token with the access token, it currently returns a response body like the following (tokens anonymized for clarity):

{
  "access_token": "NSTRI_ACCESS_TOKEN",
  "expires_in": 36000,
  "token_type": "Bearer",
  "scope": "read write",
  "refresh_token": "NSTRI_REFRESH_TOKEN"
}

We need a stable identifier that can be used across physionet (provider) & NSTRI (client). Currently there are three options based on the user model - `pk`,`email`, `username`.

Since email is something that can change & is not preferred - `username` might be a good idea. Username is both required & unique. Furthermore, I don't think there is any workflow that allows modification of username (@tompollard can comment) 

Also, Scope needs to be updated & views need to be implemented that provide the requested data post proper authentication & authorization checks. 

@westernsalt can add more information on scope part & the specific fields they would like to request.

tompollard:
> Since email is something that can change & is not preferred - username might be a good idea. Username is both required & unique. Furthermore, I don't think there is any workflow that allows modification of username (@tompollard can comment)

Username *can* currently be modified. Not sure if this is desirable or not. I'm not particularly keen on using the primary key publicly, so my preference would be to create a new, unique, public "PhysioNet User ID" (without "PhysioNet" in the name, to allow for general use).

tompollard:
@bemoody we need a way of being able to uniquely identify a user on PhysioNet. Okay with the idea of creating a new ID for this purpose?

tompollard:
@Rutvikrj26 I just had a quick chat with Benjamin. He said he was happy with creating something like a unique user ID (e.g. `public_uuid`) for this purpose.

westernsalt:
@Rutvikrj26

Thanks - I read through all the updates, and I’m really glad to hear that a `public_uuid` will be added!

We’d love to receive user info in the `/oauth/token` response based on the following scopes.  
Here’s what we’re hoping for:

**Requested scopes and corresponding data fields:**

- `profile` → `username`, `full_name`
- `email` → `email`
- `institution` → `institution`
- `credentialing` → `training_status`, `credentialing_status`
- `orcid` → `orcid`
- `public_id` → `public_uuid`

Would it be possible to support this structure?

tompollard:
@westernsalt I think these are all reasonable fields to share

tompollard:
[...]

PR #2385, opened 2025-04-11
Add `public_user_uuid` field to User model

tompollard:
It would be helpful to have a unique field to identify user accounts. One use case for this ID is for our project to allow PhysioNet users to login to the NSTRI platform (see: https://github.com/MIT-LCP/physionet-build/issues/2383).

The unique ID would allow the NSTRI platform to maintain a persistent link between logins and PhysioNet users (see: https://github.com/MIT-LCP/physionet-build/issues/2383).

PhysioNet users can update their email, username, etc, so these fields do not meet this need. It seems weird to be sharing the primary key in our database, so instead we are creating a new `public_user_uuid` field.

This pull request:

- Adds a `public_user_uuid` to the User model
- Adds a 3-step migration:
    1. create the field, first nullable;
    2. backfill existing user data
    3. add the non-null constraint.

### DIFF
--- a/physionet-django/user/migrations/0064_user_public_user_uuid.py
+++ b/physionet-django/user/migrations/0064_user_public_user_uuid.py
@@ -1,0 +1,21 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("user", "0063_alter_training_status"),
+    ]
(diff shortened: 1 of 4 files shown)


Chrystinne (approved):
@tompollard The code looks good to me!

PR #2387, opened 2025-04-16
Adds an `/oauth/userinfo` endpoint with updated OAuth scopes

tompollard:
This pull request introduces the following updates to support OAuth2 access to data (ref https://github.com/MIT-LCP/physionet-build/issues/2383):

1. Adds the following read only OAuth2 scopes:

  - profile:read – username and full name
  - email:read – email address
  - institution:read – institutional affiliation
  - credentialing:read – credentialing
  - orcid:read – ORCID iD
  - public_id:read – persistent public user UUID

2. Adds an `/oauth/userinfo` endpoint that returns selected user profile information depending on the scopes granted to the access token.

3. Adds/updates tests for the new scopes and endpoint.

### DIFF
--- a/physionet-django/oauth/tests.py
+++ b/physionet-django/oauth/tests.py
@@ -46,7 +46,7 @@ class BaseTest(TestCase):
 
         self.access_token = AccessToken.objects.create(
             user=self.test_user,
-            scope="read write",
+            scope="profile:read email:read public_id:read",
             expires=timezone.now() + timedelta(seconds=300),
             token="secret-access-token-key",
             application=self.application,
@@ -113,7 +113,7 @@ class BaseAuthorizationCodeTokenView(BaseTest):
         authcode_data = {
             "client_id": self.application.client_id,
             "state": "random_state_string",
-            "scope": "read write",
+            "scope": "profile:read email:read public_id:read",
             "redirect_uri": "http://example.org",
             "response_type": "code",
             "allow": True,
@@ -132,7 +132,7 @@ class BaseAuthorizationCodeTokenView(BaseTest):
         authcode_data = {
             "client_id": self.application.client_id,
             "state": "random_state_string",
-            "scope": "read write",
+            "scope": "profile:read email:read public_id:read",
             "redirect_uri": "http://example.org",
             "response_type": "code",
             "allow": True,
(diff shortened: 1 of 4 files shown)


Chrystinne:
@tompollard The tests on my local branch pass when using the following scope:
`profile:read email:read public_id:read`

However, it didn't work when I used the full scope as follows:
`profile:read email:read institution:read credentialing:read orcid:read public_id:read`

Am I possibly doing something wrong when testing?

Chrystinne:
Maybe you can double check the attributes 'institution', 'training_status', 'credentialing_status', and 'orcid' for the user?

Chrystinne:
I'm not sure this is exactly what you want, but this code works for me:

```
if "institution:read" in scopes:
    data["institution"] = user.profile.affiliation

if "credentialing:read" in scopes:
    data["training_status"] = user.get_first_training_status()
    data["credentialing_status"] = user.get_credentialing_status()

if "orcid:read" in scopes:
    data["orcid"] = user.get_orcid_id()
```
    
I needed to add the following method to the User model:

```
def get_first_training_status(self):
    """Returns the status of the user's first training, or None if no trainings exist."""
    training = self.trainings.first()
    return training.status if training else None
```

Rutvikrj26:
A couple of suggestions:

- Currently, the view verifies the request but does not validate the specific required scopes upfront or does not have any default scopes. A cleaner approach would be to atleast a couple of default scopes for the processing of the request to be meaningful.

- Secondly, hardcoding of the scopes and loosely checking if the string exists in the scope is not ideal. One potential option could be to have a dictionary mapping of the scope to fields. That will make it more maintainable, and set a pattern for other resource views to be developed similarly. Something like the following maybe:
```
SCOPES_MAPPING = {
    "profile:read": ["username", "full_name"],
    "email:read": ["email"],
    "institution:read": ["institution"],
    "credentialing:read": ["training_status", "credentialing_status"],
    "orcid:read": ["orcid"],
    "public_id:read": ["public_user_uuid"],
}

class UserInfoView(ScopedProtectedResourceView):
    def get(self, request, *args, **kwargs):
        valid, r = get_oauthlib_core().verify_request(request, scopes=[])
        if not valid:
            return JsonResponse({"error": "Invalid or missing token"}, status=403)

        token = r.access_token
        scopes = token.scope.split()
        user = token.user

        # Dynamically build the response data
        data = {}
        for scope, fields in SCOPES_MAPPING.items():
            if scope in scopes:
                for field in fields:
                    data[field] = getattr(user, field, None)

        return JsonResponse(data)
```
Further tightening of the system can be done by validating that the fields added in the scopes mapping are actually a part of the user model, in the tests(assuming that the pattern is userInfoView gives data from the user table, ProjectInfoView might do for the project, etc).

Rutvikrj26 (commented):
A couple of suggestions:

- Currently, the view verifies the request but does not validate the specific required scopes upfront or does not have any default scopes. A cleaner approach would be to atleast a couple of default scopes for the processing of the request to be meaningful.

- Secondly, hardcoding of the scopes and loosely checking if the string exists in the scope is not ideal. One potential option could be to have a dictionary mapping of the scope to fields. That will make it more maintainable, and set a pattern for other resource views to be developed similarly. Something like the following maybe:
```
SCOPES_MAPPING = {
    "profile:read": ["username", "full_name"],
    "email:read": ["email"],
    "institution:read": ["institution"],
    "credentialing:read": ["training_status", "credentialing_status"],
    "orcid:read": ["orcid"],
    "public_id:read": ["public_user_uuid"],
}

class UserInfoView(ScopedProtectedResourceView):
    def get(self, request, *args, **kwargs):
        valid, r = get_oauthlib_core().verify_request(request, scopes=[])
        if not valid:
            return JsonResponse({"error": "Invalid or missing token"}, status=403)

        token = r.access_token
        scopes = token.scope.split()
        user = token.user

        # Dynamically build the response data
        data = {}
        for scope, fields in SCOPES_MAPPING.items():
            if scope in scopes:
                for field in fields:
                    data[field] = getattr(user, field, None)

        return JsonResponse(data)
```
Further tightening of the system can be done by validating that the fields added in the scopes mapping are actually a part of the user model, in the tests(assuming that the pattern is userInfoView gives data from the user table, ProjectInfoView might do for the project, etc).

tompollard:
@Chrystinne good catch, thanks for the comments. I've fixed the data fields (and removed training status for now, because it needs a little more thought).

tompollard:
Thanks @Rutvikrj26, I addressed your suggestions by (1) moving the scopes into a mapping file (2) adding `profile:read` as a required scope.

Chrystinne (approved):
Looks good, @tompollard! And the tests are now passing for the full scope:
profile:read email:read institution:read credentialing:read orcid:read public_id:read

Chrystinne:
@tompollard Maybe you can remove "training status" from the PR description for the credentialing
(credentialing:read – credentialing/training status).

PR #2388, opened 2025-04-17
Fix section order in /projects/X/content/ for model projects

bemoody:
Different resource types have different sets of optional and required sections - some sections are only required for certain resource types, and the section titles don't match the database field names.

For Model projects (resource type 3), the "Model Description" section (`project.content_description`) appears third, before "Technical Implementation" (`project.methods`).  For example, see http://localhost:8000/projects/AGlSZa5VTv5Dg09CSpoh/preview/.

However, in the Project Content page (http://localhost:8000/projects/AGlSZa5VTv5Dg09CSpoh/content/), "Technical Implementation" is currently shown before "Model Description".  This inconsistency is confusing.

Also, there should be a red star next to "Technical Implementation", to indicate that this section is required for Model projects.

### DIFF
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -564,7 +564,7 @@ class ContentForm(forms.ModelForm):
          'conflicts_of_interest',
          ),
         # 3: Model
-        ('title', 'abstract', 'background', 'methods', 'content_description',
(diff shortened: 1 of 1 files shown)


tompollard (approved):
thanks, looks good to me!

PR #2389, opened 2025-04-23
Allow users to generate personal OAuth access tokens

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/2380, we would like to add support for tokens to enable secure, programmatic use of protected API endpoints (e.g. downloading credentialed datasets or querying metadata), without prompting for login credentials.

This pull request is part 1 of the implementation plan. It adds an interface for users to create up to 3 personal (OAuth) tokens. Each token: is tied to a user; expires after 60 days; and is revocable via the UI. For now, the interface is hidden from users.

Important notes:

- The tokens require a local OAuth Application to be created. This can be done using the management command `python manage.py create_default_oauth_app  --username <username>`
- The tokens are currently given a single `data:download` scope. This scope has no function right now, but later it could allow users to request files.
- The tokens do not support a "name". This can be later, if needed.
- For now, the settings page is hidden from users.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -230,3 +230,14 @@ CITI_SOAP_URL="https://webservices.citiprogram.org/SOAP/CITISOAPService.asmx"
 # See https://docs.djangoproject.com/en/4.2/ref/settings/
 DATA_UPLOAD_MAX_NUMBER_FILES=1000
 DATA_UPLOAD_MAX_MEMORY_SIZE=2621440
+
+# OAUTH_CLIENT_APP_NAME is the name of the default OAuth application used
+# when programmatically generating access tokens (e.g., via the /settings/tokens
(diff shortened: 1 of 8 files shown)


Chrystinne:
@tompollard 

Just a small comment: It might be useful to update the PR's description to add the --username parameter to the management command? (i.e., `python manage.py create_default_oauth_app --username <username>`)

I got the error "unrecognized arguments:" when using the management command without it.

tompollard:
> Just a small comment: It might be useful to update the PR's description to add the --username parameter to the management command? (i.e., python manage.py create_default_oauth_app --username <username>)

Done!

Chrystinne (approved):
Looks good, @tompollard!

PR #2390, opened 2025-04-24
Generic handling of project content sections

bemoody:
Projects have a lot of HTML content sections (abstract, background, methods, and so on.)  The *list* of sections - which sections are permitted, which are required, what are their titles, and what order do they appear in - is complicated, and different for each project type.

Nonetheless, this information has been duplicated in many places:
- X_content.html
- X_content_preview.html
- published_project.html
- ContentForm
- ActiveProject

This duplication creates inconsistencies and makes maintenance difficult, and at the same time, none of the above provide a generic interface that new features can make use of.

So this pull request defines such a generic interface, and replaces all of the above with calls to that interface.

This pull request shouldn't make any visible changes to the appearance or behavior of the site.  One exception is that the error messages from check_integrity are slightly different: some fields are listed in a different order, and "Conflicts Of Interest" is now "Conflicts of Interest".


tompollard:
Thanks @bemoody, this is great. It looks like the form submission is failing in `TestProjectEditing.test_content` because the project is missing a title:

```
(Pdb) print(response.context['form'].errors)
<ul class="errorlist"><li>title<ul class="errorlist"><li>This field is required.</li></ul></li></ul>`
```

bemoody:
Thanks for debugging!  That `response.context['form'].errors` is a useful thing to know about.

tompollard (approved):
Big improvement, thanks!

PR #2398, opened 2025-05-07
Remove direct link to static zip file

bemoody:
If a zip archive is available for a published project we display a link in the Files section.

Currently, if the project is *open*, we display a direct link like `https://physionet.org/static/published-projects/mitdb/mit-bih-arrhythmia-database-1.0.0.zip`.

Instead, we can use the "get-zip" link like `https://physionet.org/content/mitdb/get-zip/1.0.0/`, which works for all projects regardless of access policy.

### DIFF
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -305,17 +305,12 @@
         {% endif %}
 
         <ul>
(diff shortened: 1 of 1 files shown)


tompollard (approved):
thanks, looks good to me.

PR #2400, opened 2025-05-09
Improved sandboxing for LightWAVE published projects

bemoody:
Change the behavior for the LightWAVE server so that the server only has access to the specific project requested.  Furthermore, call `can_view_project_files` to check that we're authorized.

Up to now, the LightWAVE server for published projects runs with access to all open-access projects at once - e.g. if you query a record from `mimic3wdb-matched`, the LightWAVE server is allowed to retrieve data from `mimic3wdb`.  (This was intentional, since these projects were built using tons of symlinks on the original PhysioNet system.)

This is not optimal because:

- We want to move all projects (both open and controlled) into `pn-media` storage, for better control over authorization, which means there won't be a single chroot directory that contains only open-access projects.

- The old system never worked for controlled-access projects and we'd eventually like to support that.

- Allowing symlinks itself brings security risks, and we'd eventually like to disable them.

On the PhysioNet server, I've manually replaced all the symlinks inside `pn-static/published-projects` with hard links.  So hopefully this should work now.

Things not done here:

- Supporting controlled-access projects would not be too difficult but isn't done here.

- I'd like to do away with the symbolic link to `/usr/local/database/wfdbcal`, and all the kludgery around it, because this keeps tripping me up in testing.

### DIFF
--- a/physionet-django/lightwave/tests.py
+++ b/physionet-django/lightwave/tests.py
@@ -33,12 +33,18 @@ class TestPublished(TestMixin):
         for q in test_queries:
             qstr = q.format(db='demobsn/1.0', record='231')
 
-            response = self.client.get(server + '?' + qstr)
+            # Each request (both 'dblist' and non-'dblist' requests)
+            # should make exactly one database query; see
+            # views.lightwave_server.
+
+            with self.assertNumQueries(1):
+                response = self.client.get(server + '?' + qstr)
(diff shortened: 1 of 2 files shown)


tompollard (approved):
looks good to me, thanks benjamin (though i know you are still working on it!)

tompollard (approved):

PR #2416, opened 2025-06-05
Add SHA256SUMS.txt download endpoint

tompollard:
This pull request adds a new API endpoint that allows authorized users to download the SHA256SUMS.txt file for a specific project:

`/v1/projects/published/<project_slug>/<version>/sha256sums/`

The file contains SHA256 checksums for all files in the dataset, which is useful for verifying data integrity. If the file doesn't exist, a not found error is returned.

On the dev server, the following endpoint should return a file if logged in: 
http://localhost:8000/api/v1/projects/published/demopsn/1.0/sha256sums/

The following endpoint returns a not-found error:
http://localhost:8000/api/v1/projects/published/demowave/1.0.0/sha256sums/

### DIFF
--- a/physionet-django/export/tests.py
+++ b/physionet-django/export/tests.py
@@ -1,3 +1,0 @@
-from django.test import TestCase
-
-# Create your tests here.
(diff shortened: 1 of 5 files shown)


Chrystinne (approved):
@tompollard Looks good to me, Tom!

PR #2417, opened 2025-06-05
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.


physionet-django/user/models.py
```diff
@@ -279,7 +279,7 @@
     ("UZ", _("Uzbekistan")),
     ("VU", _("Vanuatu")),
     ("VE", _("Venezuela (Bolivarian Republic of)")),
-    ("VN", _("Viet Nam")),
+    ("VN", _("Viet Name")),
```
tompollard: Should be Vietnam

physionet-django/project/migrations/0051_auto_20211129_0427.py
```diff
@@ -20,13 +20,13 @@ def migrate_forward(apps, schema_editor):
...
...HI-report@physionet.org, citing the\r\n   location of the specific information in question.  \r\n6. The LICENSEE will use the data for the sole purpose of lawful use in\r\n   scientific research and no other.  \r\n7. The LICENSEE will be responsible for ensuring that he or she\r\n   maintains up to date certification in human research subject\r\n   protection and HIPAA regulations.\r\n8. The LICENSEE agrees to contribute code associated with publications\r\n   arising from this data to a repository that is open to the research community.\r\n9. This agreement may be terminated by either party at any time, but the\r\n   LICENSEE's obligations with respect to PhysioNet data shall continue\r\n   after termination.  \r\n\r\nTHE DATA ARE PROVIDED \"AS IS\", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR\r\nIMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,\r\nFITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL\r\nTHE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR\r\nOTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,\r\nARISING FROM, OUT OF OR IN CONNECTION WITH THE DATA OR THE USE OR OTHER\r\nDEALINGS IN THE DATA.",
```
tompollard: We should leave the migration - might cause issues to modify.

tompollard (changes requested):

physionet-django/user/models.py
yarikoptic: I completely skipped "Viet Nam" for now since implicated in migrations and might need one if there is decision to go for ISO name "Vietnam"

physionet-django/project/migrations/0051_auto_20211129_0427.py
yarikoptic: yeap, I ignored `requestor` and `requestors` you would stay consistent to those choices , and thus migration was avoided.

tompollard (approved):
Looks good, thanks!

tompollard:
@yarikoptic there is a trailing space somewhere. please could you take a look?

tompollard (approved):

PR #2420, opened 2025-06-07
[WIP] Add token-based authentication for file downloads

tompollard:
This PR adds token-based authentication for downloading files from published projects. Users can now download files using an OAuth access token with the 'data:download' scope.

Key changes:

- Add new file download endpoint at /api/v1/projects/published/{project_slug}/{version}/files/{filepath}/
- Require OAuth token with 'data:download' scope for access
- Support streaming downloads for large files

Testing:

- Added unit tests covering unauthorized access, missing files, successful downloads
- Verified token scope requirements and DUA signing requirements

### DIFF
--- a/physionet-django/export/tests/test_files.py
+++ b/physionet-django/export/tests/test_files.py
@@ -4,8 +4,13 @@ from django.test import TestCase
 from django.urls import reverse
 from rest_framework.test import APIClient
 from rest_framework import status
-from project.models import PublishedProject, DUASignature, AccessPolicy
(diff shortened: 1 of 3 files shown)


bemoody:
This breaks existing access policy, doesn't use appropriate storage
APIs, dumps massive I/O burden onto uwsgi, and also fails to handle
numerous HTTP features.

PhysioNet has a perfectly good API for downloading files.  We don't
need a second one and we definitely do not need a second one that is
badly implemented.

tompollard:
Let's chat about this. It is marked as a work in progress, and we do need a token based method for supporting downloads, I think.

Issue #2397, opened 2025-05-06
Modify handling of data for open access projects


tompollard:
Open access projects are currently made available to users in a slightly different way to controlled access projects, and they are stored in a different location. 

To help us to manage access to all datasets - including open access datasets - through Django, we would like to make some modifications to this approach.

tompollard:
@bemoody can this be closed?

tompollard:
Side note, all georestricted open datasets in the AWS Open Bucket were moved to a private bucket. This is only a short term solution.

PR #2424, opened 2025-06-11
Migrate static/published-projects to media/published-projects

bemoody:
Historically, public data files (i.e. published projects with an "open" access policy) were moved into Django's "static" storage area, to avoid querying Django every time a file was accessed.  (Non-public projects, on the other hand, are kept in Django's "media" storage.)

This creates a lot of needless complexity *and* limits our flexibility for access control.  It's also confusing to developers because this isn't the way Django's static storage is intended to be used.

This pull request implements a data migration to move published project directories and zip archives from the static to the media area.  (There shouldn't be any conflicts, because currently there's only one copy of each project version.)  The static area will still exist, with appropriate symbolic links pointing into the media area.

This is a large piece of issue #2397.

Later (hopefully after this is deployed and tested) we'll add rules in nginx to redirect clients from /static/published-projects/ to /files/, and finally the symlinks in /static/published-projects/ will be removed.


bemoody:
I believe this'll work but I haven't tested it extensively in a "real" server.  Let's try to plan a time to do this deployment when I'll be online.

tompollard:
thanks @bemoody. any objections to me running this on the staging server (assuming local testing all seems fine)?

bemoody:
Yeah, go ahead but I don't remember if the staging server *has* any project files.

tompollard:
> Yeah, go ahead but I don't remember if the staging server *has* any project files.

Thanks. Just took a look - yes, there are files. e.g. `mitdb/1.0.0/` contains a bunch of records in pn-static.

bemoody:
This doesn't really work because the git hook scripts run as `pn`, not as `www-data`.  That should be fixed.

But a quick workaround would be
- change ownership of `pn-static/published-projects/*` and `pn-static/published-projects/*/*` from `www-data:www-data` to `pn:pn`
- deploy the migration
- change ownership of `pn-media/published-projects/*` and `pn-media/published-projects/*/*` from `pn:pn` to `www-data:www-data`

tompollard (approved):

tompollard:
@bemoody this has been deployed to the production server. Please check!

`os.symlink(new_path, new_path)` raised an error when the target directory already existed in `pn-media` (i.e. where the project had both open and restricted versions), so I fixed these projects manually (examples include `kinecal`, `pmd`, and `tremordb`). Aside from this, everything appeared to run as expected.

rgmark:
Thanks!

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

bemoody:
From now on, please do not push anything to the server without discussing with me first.

bemoody:
> Please check!
Yes, there are now file permissions that are now incorrect, and there are files that are in the wrong places, and I don't have time to clean this up.

tompollard:
Projects that existed in both pn-static and pn-media:

```
ct-ich
kinecal
ltrsvp
lyapunov
meditation
mhd-effect-ecg-mri
mimic2-iaccd
mimic-iv-ecg
multi-gait-posture
multimodal-nback-music
multimodal-satellite-data
perg-ioba-dataset
pmd
propofol-anesthesia-dynamics
rcvsim
sleep-dep-hemo-cog
tremordb
wearable-exam-stress
widsdatathon2020
```

PR #2428, opened 2025-06-17
Bump test server to Debian 12

tompollard:
This pull request updates the scripts for configuring a demo PhysioNet server in a virtual machine to use Debian 12, rather than Debian 11 (for consistency with the change introduced in https://github.com/MIT-LCP/physionet-build/pull/2421)


tompollard:
Also updates the Dockerfile to use Debian 12 as its base image (bullseye -> bookworm).

briangow (approved):

PR #2430, opened 2025-06-18
Add Homepage redesign

fpdrozd:
This PR implements the homepage redesign using bootstrap and custom styles. All of the new styles are wrapped in a `modern-ui` class to isolate them from the rest of the site. The redesign includes:
- a whole homepage rewrite
- implementing new navigation and footer on subpages
- layout changes on subpages
- mobile support

[Figma designs](https://www.figma.com/design/t2CVcw2XvgAMGqqUrmKiDr/PhysioNet---Platform-Designs?node-id=230-4266&t=2xfdEukK3ZJtkPrb-0)


physionet-django/old/templates/footer.html
```diff
@@ -0,0 +1,18 @@
+<!-- This footer template is from the bootstrap 'blog' example -->
```
tompollard: @fpdrozd What is the purpose of the "old" folder in these paths (e.g. `physionet-django/old/templates/footer.html`)? If this content is no longer needed, can't it just be deleted?

physionet-django/templates/navbar.html
```diff
@@ -1,22 +1,195 @@
...
-		<form class="form-inline" action="{% url 'content_index' %}">
-	    	<input name="topic" class="search-input" type="text" placeholder="Search">
-		    <span class="input-group-btn">
-		      <button id="search-button" type="submit" class="btn-search my-2 my-sm-0" title="Search {{ SITE_NAME }}"><span class="fa fa-search"><span class="visually-hidden">Search {{ SITE_NAME }}</span></span></button>
-		    </span>
-	 	</form>
-	</div>
-</nav>
+<!-- move this to the right place -->
+<script>
```
tompollard, line 170: I'm not sure we want this embedded here? Same for other scripts.

physionet-django/templates/navbar.html
```diff
@@ -1,22 +1,195 @@
...
+});
+dom.searchCloseButton.addEventListener('click', () => {
+  dom.nav.classList.remove('nav--search');
+});
+
+dom.hamburger.addEventListener('click', (ev) => {
+  ev.stopPropagation();
+  dom.navWrapper.classList.toggle('nav-wrapper--mobile-open');
+});
+</script>
```
tompollard: Missing newline at the end of the file.

physionet-django/templates/home.html
```diff
@@ -15,71 +15,128 @@
...
-    {% for news in front_page_banner %}
-    <li>
-      <a href="{% url 'news_by_slug' news.slug %}">{{ news.title }}</a>
-    </li>
-    {% endfor %}
-    </ul>
+  <div class="banner relative" style="height: 650px; background-image: linear-gradient(to right, white 40%, rgba(255,255,255,0) 80%), url({% static 'images/banner.png' %});">
+    <div class="container absolute my-0">
+      <div class="banner__column d-flex flex-column justify-content-center h-100">
+        <h1 style="font-weight: 600;">Advancing Medical Research<br>Through Open Data</h1>
```
tompollard: The banner text and slogan should be managed as environment variables. None of this should be hardcoded in the template.

physionet-django/templates/footer.html
```diff
@@ -1,18 +1,60 @@
...
-          <p>For more accessibility options, see {{ FOOTER_ACCESSIBILITY_PAGE|safe }}.</p>
-        {% endif %}
-        <p><a href="#">Back to top</a></p>
+<div class="modern-ui">
+  <footer class="footer w-100 secondary-bg py-5">
+    <div class="container my-0 py-0 py-md-5">
+      <img src="{% static 'images/logo-white.png' %}" width="180px" alt="PhysioNet Logo" class="pb-2" />
+      <div class="footer__columns light-text">
+        <div>
+          <p class="dark-light-text mb-2">PhysioNet is a repository of freely-available medical research data, managed by the MIT Laboratory for Computational Physiology.</p>
```
tompollard: The codebase is used by other repositories (not just Physionet) so content like "PhysioNet is a repository of freely-available medical research data, managed by the MIT Laboratory for Computational Physiology" should not be hardcoded. I would suggest moving all of this text to environment variables (and actually I think some of the variables already exist, e.g.:

- `SITE_NAME`
- `STRAPLINE`
- `FOOTER_SUPPORTED_BY`
- `FOOTER_MANAGED_BY`
- `FOOTER_ACCESSIBILITY_PAGE`

physionet-django/templates/footer.html
```diff
@@ -1,18 +1,60 @@
...
+          <p class="dark-light-text mb-2">Supported by the National Institute of Biomedical Imaging and Bioengineering (NIBIB) under NIH grant number R01EB030362.</p>
+          <p class="dark-light-text mb-2">For more accessibility options, see the <a class="dark-primary-text" href="https://accessibility.mit.edu/">MIT Accessibility Page.</a></p>
+        </div>
+        <div class="d-flex flex-column pl-0 pl-md-5 pt-4 pt-md-0">
+          <h2 class="text-white font-weight-bold">Navigation</h2>
+          <a href="/content" class="pb-3 dark-light-text">Discover Data</a>
```
tompollard, line 16: These menu items ("about", "share data" etc) should be pulled in from the static pages model, not hardcoded. One the static page management page (e.g. https://physionet.org/console/static-page/2/edit/) there is a item called "nav bar" that determines whether the page appears in the nav bar.

physionet-django/templates/navbar.html
```diff
@@ -1,22 +1,195 @@
...
+  <nav class="nav">
+    <div class="container my-0 h-100 d-flex justify-content-between align-items-center">
+      <button type="button" class="nav__hamburger circle-button mr-2">
+        <i class="nav__hamburger__list bi bi-list"></i>
+        <i class="nav__hamburger__x bi bi-x"></i>
+      </button>
+      <a href="/"><img src="{% static 'images/logo.png' %}" width="200px" alt="PhysioNet Logo" /></a>
+      <div class="nav__links align-items-center">
+        <a href="/content"><button type="button" class="nav-link">Discover Data</button></a>
+        <a href="/about/publish"><button type="button" class="nav-link">Share Data</button></a>
```
tompollard: These menu items ("about", "share data" etc) should be pulled in from the static pages model, not hardcoded. One the static page management page (e.g. https://physionet.org/console/static-page/2/edit/) there is a item called "nav bar" that determines whether the page appears in the nav bar.

tompollard (changes requested):
Thanks @fpdrozd, the page looks great. Some work is needed to avoid hardcoding content in the templates. I've made some notes on this in the code.

physionet-django/physionet/context_processors.py
```diff
@@ -32,3 +33,15 @@ def cloud_research_environments_config(request):
     return {
         "ENABLE_CLOUD_RESEARCH_ENVIRONMENTS": settings.ENABLE_CLOUD_RESEARCH_ENVIRONMENTS,
     }
+
```
tompollard, line 37: Needs extra newline before `homepage_config` to pass style checks.

tompollard:
I noticed a couple of other things while testing:

1. It looks like the console pages have not been updated (e.g. try logging in as admin and going to http://localhost:8000/console/).

Screenshot 2025-06-19 at 12 36 32 AM


2. The "Publish your data" button on the front page is always highlighted. Is this intentional?

 
Screenshot 2025-06-19 at 12 35 38 AM

physionet-django/templates/home.html
tompollard: Currently we have `STRAPLINE` which could hold the "Advancing Medical Research Through Open Data" bit. Add new variable for the "A comprehensive platform for discovering, sharing, and analyzing complex physiologic signals and healthcare datasets." (e.g. `SITE_DESCRIPTION`)

physionet-django/templates/navbar.html
fpdrozd: I left a comment here in case there is a dedicated place for this kind of scripts. In my opinion it makes more sense to keep it in one file and be able to use the navbar as a self-contained module.

physionet-django/templates/home.html
fpdrozd: Fixed!

physionet-django/templates/footer.html
fpdrozd: Fixed!

physionet-django/templates/footer.html
fpdrozd: Fixed!

fpdrozd:
@tompollard I started working on the nav but still need to fix up the sidebar before pushing it.

I couldn't reproduce the text highlight bug but I added styles that completely disable text selection on buttons. Hope that helps!

tompollard:
I had a play around with this locally. Minor things that we need to address, ideally before making live:

1. The `Publish your data` buttons link to a list of published projects. Instead, they should link to `/about/publish/`
2. The descriptions in the menu don't really make sense ("Upcoming workshop and conferences in medical informatics"; "Innovative projects using medical datasets to advance healthcare" etc).
3. The "Publish your data" button on the front page is always highlighted (Chrome Version 138.0.7204.93), as mentioned above.

rafalkosla101:
@tompollard 
1. Fixed but i will wait to push with the rest of the items
2. Do you have any example texts how would you like it to be? should i only remove the medical part or restructure the whole thing ?
3. Do you mean that the button is always pink? if so i think that was the initial design but we can change it to the same behaviour as the button next to it

tompollard:
> Fixed but i will wait to push with the rest of the items

Thanks, please could you push?

> Do you have any example texts how would you like it to be? should i only remove the medical part or restructure the whole thing ?

Please leave this for now. I will add the text in a separate PR.

> Do you mean that the button is always pink? if so i think that was the initial design but we can change it to the same behaviour as the button next to it

Ah, okay, thanks for explaining. I hadn't realised that this was a design decision. Let's leave it as it is for now then.

bemoody:
**All fonts and images must be hosted locally** (and must be under a free/open-source license.)

**All scripts and stylesheets must be hosted locally**, must be free/open-source, and (if compiled/minified) the corresponding source code must also be included so that we can debug it.

bemoody:
The logo images should be configurable.  There should be appropriate variables defined in the settings ("SITE_HEADER_LOGO" and "SITE_FOOTER_LOGO"?)  If the variable is undefined, then the SITE_NAME should just be displayed as text.  So instead of this:
```
<a href="/"><img src="{% static 'images/logo.png' %}" width="200px" alt="PhysioNet Logo" /></a>
```
do this:
```
<a href="/">
  {% if SITE_HEADER_LOGO %}
    <img src="{{ SITE_HEADER_LOGO }}" width="200px" alt="{{ SITE_NAME }}">
  {% else %}
    {{ SITE_NAME }}
  {% endif %}
</a>
```
It's okay to include the PhysioNet logos in this repository, but they shouldn't be displayed by default.  Please rename `logo.png` to `physionet-logo.png` or something similar.

bemoody:
There shouldn't be `<button>` nested inside an `<a>`.  Having two focusable items requires pressing tab twice to move between them, and the focus indicators look weird and inconsistent.  Instead of this:
```
<a href="{{ page.url }}"><button type="button" class="nav-link">{{ page.title }}</button></a>
```
Just do this:
```
<a href="{{ page.url }}" class="nav-link">{{ page.title }}</a>
```
(there are many similar instances in navbar.html and in home.html.)

bemoody:
On the front page, the links to news items ("Preparing New Challenges", etc.) and to projects ("Self Managed Access Database Demo", etc.) do not have any color or decoration to indicate that they are links.  Perhaps they should be the same pink color as the buttons.

Also, when one of these links is focused, the focus indicator appears as a thin line at the right-hand side, which is hard to see.

rafalkosla101:
@bemoody do you by any chance have any video/screenshot for the last comment you left? Im not 100% sure what are you referring to

bemoody:
That problem is due to the `overflow: hidden` here:
```
.modern-ui .card-title {
  font-size: 1.25rem;
  margin-top: 0.5rem;
  margin-bottom: 0.5rem;
  display: -webkit-box;
  -webkit-line-clamp: 3;
  -webkit-box-orient: vertical;
  overflow: hidden;
  cursor: pointer;
}
```
A better way to accomplish what you're trying to do would be:
```
.modern-ui .card-title {
  font-size: 1.25rem;
  margin-top: 0.5rem;
  margin-bottom: 0.5rem;
}
.modern-ui .card-title > a {
  display: -webkit-box;
  -webkit-line-clamp: 3;
  -webkit-box-orient: vertical;
  overflow: hidden;
}
```
(`cursor: pointer` doesn't belong here at all.)

tompollard:
Continued in: https://github.com/MIT-LCP/physionet-build/pull/2444

PR #2431, opened 2025-06-18
Handle errors in pipelines in git hooks

bemoody:
If something goes wrong in the installation process (for example, one of the "manage.py" commands fails), the git hooks should immediately stop what they're doing and not blindly keep going.

In particular, if there is an early migration that fails for any reason, then the old (working) server code should be left running, rather than installing the new (presumably dependent on the migration, and thus broken) server code.  That's certainly how it was intended to work, but it didn't work because the "migratetargets" command had its output redirected to a pipe.

Use bash's "pipefail" mode to ensure that we catch errors from all non-final commands in a pipeline.  Currently the "migrate" and "migratetargets" commands are the only ones where this should make a difference.

### DIFF
--- a/deploy/post-receive
+++ b/deploy/post-receive
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -e
+set -o pipefail
 
 install_dir=/physionet/physionet-build
 working_dir=/physionet/physionet-build.new
--- a/deploy/update
+++ b/deploy/update
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -e
+set -o pipefail
 
 install_dir=/physionet/physionet-build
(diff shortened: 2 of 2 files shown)


bemoody:
For the avoidance of doubt: the scripts in the `deploy` directory are convenience copies of the git hook scripts that we install on all of our servers.  These scripts don't automatically install/update themselves.

I've now deployed this change; this pull request is to bring the repository in sync with what's on the servers.

bemoody:
And yes, it would be wise to be more cautious about checking migrations: we probably could and probably should apply early migrations during the `update` script, and we also probably could and probably should apply any *residual unapplied late migrations* or throw an error.

tompollard (approved):

PR #2434, opened 2025-07-01
Add support for georestriction to `can_access_project`

tompollard:
As part of work to address incoming policy changes on data sharing, we would like to have the ability to restrict access to data based on the location of a request.

This pull request updates the `can_access_project()` function to handle `request` as an additional argument. If the request comes from a restricted country and the project is labelled as `georestricted` then `can_access_project` returns False.

This is a work in progress (in particular, the GeoIP2 package requires a file that is not currently included in this pull request).


bemoody:
Assume that I'm clueless, explain what files I would need to download and what settings I would need to configure.

tompollard:
> Assume that I'm clueless, explain what files I would need to download and what settings I would need to configure.

@bemoody Are you referring to the geolocation database file? Yes, I'll add some detail on this once tests are passing.

tompollard:
@bemoody This is ready for review. Please could you take a look? 

I have tested using the `GeoLite2-Country.mmdb` file downloaded from https://dev.maxmind.com/geoip/geolite2-free-geolocation-data/ and the blocking works as expected.

To test using the IP file:
- Download `GeoLite2-Country.mmdb` from https://dev.maxmind.com/geoip/geolite2-free-geolocation-data/
- Set `BLOCKED_REGIONS` in `.env` file with the 2-digit ISO 3166-1 country code/s
- Set `GEOIP_PATH=`  to the path of the folder containing the `GeoLite2-Country.mmdb` file
- Start the server and set `georestricted=True` on a `PublishedProject` you'd like to test.
- Use VPN to connect from the relevant country (I use [ngrok](https://ngrok.com/) to temporarily make my local server available on a public IP).

bemoody:
Documentation:
```
+The path to the folder containing GeoLite2-Country.mmdb.gz file should be assigned to `GEOIP_PATH`:
```
Should say `GeoLite2-Country.mmdb`, not `GeoLite2-Country.mmdb.gz`.

bemoody:
physionet-django/physionet/utility.py:
```
+geoip = None
```
Use all-uppercase for globals (like `LOGGER`).

```
+    try:
+        geoip = GeoIP2(path=settings.GEOIP_PATH)
+    except Exception as e:
+        raise e
```
The argument `path=settings.GEOIP_PATH` is unnecessary, that appears to be Django's default.  Also, remove the useless try/except block.

```
+def get_client_ip(request):
+    """
+    Get the client's IP address from the request.
+    Handles cases where the request goes through a proxy or load balancer.
+    """
+    x_forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR')
+    if x_forwarded_for:
+        ip = x_forwarded_for.split(',')[0]
+    else:
+        ip = request.META.get('REMOTE_ADDR')
+    return ip
```
We must not accept X-F-F in production mode.  It might be useful to support X-F-F in *debug mode only*:
```
    if settings.DEBUG:
        x_forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR')
        if x_forwarded_for:
            return x_forwarded_for.split(',')[0]
    return request.META.get('REMOTE_ADDR')
```

bemoody:
```
+    try:
+        return geoip.country(ip)['country_code']
+    except (GeoIP2Exception, KeyError, TypeError):
+        return None
```
These exceptions don't work, AFAIK.
- TypeError or KeyError means there is a logic error, and shouldn't be caught.
- GeoIP2Exception is a Django exception (not a geoip2 exception) and means the database wasn't installed correctly.

If `ip` is a *valid IP address* that the database doesn't know about, then we get `geoip2.errors.AddressNotFoundError`.  We probably should only be catching `geoip2.errors.AddressNotFoundError` or possibly `geoip2.errors.GeoIP2Error`.

If there is a reason that you need to catch multiple exception types here, please add a comment to explain the reasons.

bemoody:
We need some test cases to ensure that the geoip logic is sound.  It might make sense to add a set of unit tests for `physionet.utility.get_country_code` alongside the test cases you've already added.

A very small test database can be found here: https://github.com/maxmind/MaxMind-DB/raw/main/test-data/GeoLite2-Country-Test.mmdb

I would add this (as `GeoLite2-Country.mmdb`) into `demo-files/geoip` and use that for testing.

1. Test using addresses that are in the demo database (e.g., 2.125.160.216 = GB, 50.114.0.0 = US, 89.160.20.112 = SE).

2. Test using addresses that are NOT in the demo database (e.g., 192.168.1.1, 10.1.2.3, 18.18.42.54).  In particular, RFC1918 addresses should be treated as "country=None" and should be allowed.

tompollard:
Thanks Benjamin - I have addressed all of the points below:

> Should say GeoLite2-Country.mmdb, not GeoLite2-Country.mmdb.gz.

Fixed, good catch!

> Use all-uppercase for globals (like LOGGER).

Switched to uppercase. 

> The argument path=settings.GEOIP_PATH is unnecessary, that appears to be Django's default. Also, remove the useless try/except block.

The path is referencing the GEOIP_PATH in the env file. Isn't it clearer to keep it?  

> Also, remove the useless try/except block.

Removed.

> We must not accept X-F-F in production mode. It might be useful to support X-F-F in debug mode only

Switched to your suggested approach.

> TypeError or KeyError means there is a logic error, and shouldn't be caught.

Removed all error catching for now. Will look at adding something to handle IP not found later.

bemoody:
Thanks.  It is failing because GEOIP isn't refreshed when settings are changed by the test case.  We could work around it like this:
```
diff --git a/physionet-django/physionet/test_utility.py b/physionet-django/physionet/test_utility.py
index 8041261f4..ea4cf2cb0 100644
--- a/physionet-django/physionet/test_utility.py
+++ b/physionet-django/physionet/test_utility.py
@@ -91,6 +91,7 @@ def setUp(self):
         test_db_path = os.path.join(os.path.dirname(__file__), '..', '..',
                                    'demo-files', 'geoip', 'GeoLite2-Country.mmdb')
         print(f"Test GeoIP database path: {test_db_path}")
+        utility._reload_geoip_data()
 
     def test_localhost_ips(self):
         """
diff --git a/physionet-django/physionet/utility.py b/physionet-django/physionet/utility.py
index 8cb2cc737..dbcfd5cbb 100644
--- a/physionet-django/physionet/utility.py
+++ b/physionet-django/physionet/utility.py
@@ -28,14 +28,23 @@ def is_blocking_countries():
     return any(region != 'localhost' for region in getattr(settings, 'BLOCKED_REGIONS', []))
 
 
-if not settings.GEOIP_PATH:
-    if is_blocking_countries():
-        raise RuntimeError(
-            "BLOCKED_REGIONS is set to block real countries, but GEOIP_PATH is not configured. "
-            "Please set GEOIP_PATH to the directory containing your GeoIP2 database files."
-        )
-else:
-    GEOIP = GeoIP2(path=settings.GEOIP_PATH)
+def _reload_geoip_data():
+    """
+    Reload GeoIP data after settings were modified.
+    This function is used for testing.
+    """
+    global GEOIP
+    if not settings.GEOIP_PATH:
+        if is_blocking_countries():
+            raise RuntimeError(
+                "BLOCKED_REGIONS is set to block real countries, but GEOIP_PATH is not configured. "
+                "Please set GEOIP_PATH to the directory containing your GeoIP2 database files."
+            )
+        GEOIP = None
+    else:
+        GEOIP = GeoIP2(path=settings.GEOIP_PATH)
+
+_reload_geoip_data()
 
 
 CONTENT_TYPE = {
```
(of course if you can think of a cleaner solution that would be welcome.)

tompollard:
Thanks @bemoody. Instead of reloading the GEOIP data, I moved initialization to a `_get_geoip()` function that skips initialization if the object already exists. I'm happy to switch to the _reload_geoip_data` approach if you prefer!

bemoody (approved):

PR #2435, opened 2025-07-07
Redirect /static/published-projects/ to new URLs

bemoody:
Add rules to redirect old /static/published-projects/ URLs to the current, supported URLs.  This includes published project zip files as well as published project files.

(Once this is done, the actual files under /pn-static/published-projects/ should no longer be accessible, and we should be able to remove that directory.)


tompollard (approved):

tompollard:
Looks good, thanks!

Issue #2436, opened 2025-07-07
Check for broken geoip configuration


bemoody:
From the discussion of pull #2434

> Instead of reloading the GEOIP data, I moved initialization to a _get_geoip() function that skips initialization if the object already exists.

This is a bit of a problem because now the server **will not fail immediately** if it's misconfigured (e.g., BLOCKED_REGIONS is set and GEOIP_PATH is not).

Better, I would think, to move the check-for-broken-config into settings/base.py.

PR #2437, opened 2025-07-08
Move check for GEOIP installation to settings.py. Fixes 2436

tompollard:
This pull request addresses https://github.com/MIT-LCP/physionet-build/issues/2436

> From the discussion of pull https://github.com/MIT-LCP/physionet-build/pull/2434

>> Instead of reloading the GEOIP data, I moved initialization to a _get_geoip() function that skips initialization if the object already exists.

> This is a bit of a problem because now the server will not fail immediately if it's misconfigured (e.g., BLOCKED_REGIONS is set and GEOIP_PATH is not).

> Better, I would think, to move the check-for-broken-config into settings/base.py.

## Changes:

- Updates `settings/base.py` to check that `GEOIP_PATH` is assigned if there are `BLOCKED_REGIONS`.
- Updates `_get_geoip()` to handle initialization without configuration checks
- Removed the now unused `is_blocking_countries()`

### DIFF
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -790,3 +790,11 @@ OAUTH2_PROVIDER = {
 
 # Path to GeoIP2 database directory
 GEOIP_PATH = config('GEOIP_PATH', default=None)
+
+# Validate GeoIP configuration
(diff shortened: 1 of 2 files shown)


bemoody (approved):
Thanks.  Nice to simplify things a little.

tompollard:
> Thanks. Nice to simplify things a little.

I agree, thanks for picking this up!

PR #2439, opened 2025-07-08
Rewrite /physiobank/database/ to new URLs

bemoody:
Update the old WFDB API to go through Django rather than accessing files in pn-static/published-projects (which is going away.)

Simplifying the nginx logic is a good thing in any case.


tompollard (approved):
thanks, looks good to me!

Issue #2438, opened 2025-07-08
Update submission/editorial process so that the georestricted flag is set appropriately


tompollard:
We need to update the submission and/or editorial process so that the `georestricted` flag is set appropriately.

PR #2440, opened 2025-07-08
Allow georestricted flag to be set by editor prior to publication. Ref #2438

tompollard:
Currently there is no way of setting the `georestricted` flag for projects to True via the platform. 

This pull request responds to https://github.com/MIT-LCP/physionet-build/issues/2438 by adding a form item in the "Publish Submission" page of the editorial system that allows the editor to set the flag to True or False.

Screenshot 2025-07-08 at 3 58 21 PM

This is not a long-term solution, but it does address the need for now. It is a flexible approach that is easily reversible.

### DIFF
--- a/physionet-django/console/forms.py
+++ b/physionet-django/console/forms.py
@@ -328,6 +328,11 @@ class PublishForm(forms.Form):
     """
     Form for publishing a project
     """
+    georestricted = forms.ChoiceField(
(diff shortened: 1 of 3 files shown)


tompollard:
(@bemoody sorry for the churn. now ready for review)

bemoody (approved):

PR #2441, opened 2025-07-10
Add configuration link when user lacks verified AWS credentials

Chrystinne:
We are currently managing AWS CLI commands for downloading datasets from S3 buckets based on access policies:

* **Open-access datasets**: AWS CLI download commands are always displayed, and users don't need to configure credentials
* **Controlled-access datasets**: Users must configure their AWS credentials in the platform before they can download datasets

This PR informs users when AWS credential configuration is required to download a given dataset from S3, and provides a direct link to the configuration page (https://physionet.org/settings/cloud/).

### DIFF
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -337,10 +337,10 @@
               Download the files using AWS command line tools:
               <pre class="shell-command">aws s3 sync {{ s3_uri }} DESTINATION</pre>
             </li>
-          {% elif project.access_policy != AccessPolicy.OPEN and not user.cloud_information.aws_verification_datetime %}
+          {% elif show_aws_configuration_link %}
             <li>
               To download the files using AWS command line tools, first
-              <a href="https://physionet.org/settings/cloud/">configure your AWS credentials</a>.
+              <a href="{% url 'edit_cloud' %}">configure your AWS credentials</a>.
             </li>
           {% endif %}
           {% endif %}
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -330,11 +330,19 @@
               <pre class="shell-command">wget -r -N -c -np{% if project.access_policy %} --user {{ user }} --ask-password{% endif %} {{ bulk_url_prefix }}{% url 'serve_published_project_file' project.slug project.version '' %}</pre>
             </li>
           {% endif %}
-          {% if has_s3_credentials and project.aws.sent_files and s3_uri %}
+
+          {% if has_s3_credentials and project.aws.sent_files %}
+          {% if s3_uri %}
             <li>
               Download the files using AWS command line tools:
               <pre class="shell-command">aws s3 sync {{ s3_uri }} DESTINATION</pre>
             </li>
+          {% elif project.access_policy != AccessPolicy.OPEN and not user.cloud_information.aws_verification_datetime %}
+            <li>
+              To download the files using AWS command line tools, first
+              <a href="https://physionet.org/settings/cloud/">configure your AWS credentials</a>.
+            </li>
+          {% endif %}
           {% endif %}
 
         </ul>
--- a/physionet-django/project/views.py
+++ b/physionet-django/project/views.py
@@ -63,6 +63,7 @@ from project.cloud.s3 import (
 )
 from django.db.models import F, DateTimeField, ExpressionWrapper
 from physionet.utility import get_client_ip, get_country_code
+from user.awsverification import aws_verification_available
 
 LOGGER = logging.getLogger(__name__)
 
@@ -1964,6 +1965,21 @@ def published_project(request, project_slug, version, subdir=''):
     except AWS.DoesNotExist:
         s3_uri = None
 
+    # Determine when to show AWS configuration link
+    try:
+        show_aws_configuration_link = (
+            not s3_uri
+            and project.aws.sent_files
+            and project.aws.is_private
+            and aws_verification_available()
+            and (
+                not hasattr(user, 'cloud_information')
+                or not user.cloud_information.aws_verification_datetime
+            )
+        )
+    except AWS.DoesNotExist:
+        show_aws_configuration_link = False
+
     context = {
         'project': project,
         'authors': authors,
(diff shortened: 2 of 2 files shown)


physionet-django/project/templates/project/published_project.html
```diff
@@ -331,10 +331,17 @@ <h5>Access the files</h5>
             </li>
           {% endif %}
           {% if has_s3_credentials and project.aws.sent_files and s3_uri %}
+          {% if project.access_policy == AccessPolicy.OPEN or user.cloud_information and user.cloud_information.aws_verification_datetime %}
```
tompollard: @Chrystinne please could you give a little more detail on the changes here? e.g. 

- why drop `has_s3_credentials` and `s3_uri`? Are these incorrect or not needed?
- what is `user.cloud_information` checking? does this return true if GCP details are provided but not AWS?
- if `project.access_policy == AccessPolicy.OPEN` is true, it will always display the AWS command won't it? What happens if the data isn't on AWS?

tompollard (changes requested):

tompollard (commented):
Currently, my understanding is that (one of?) the bugs is that there is no way of knowing about the AWS integration unless you have already added your AWS credentials. Is this pull request attempting to address that issue or a different issue?

Chrystinne:
> Currently, my understanding is that (one of?) the bugs is that there is no way of knowing about the AWS integration unless you have already added your AWS credentials. Is this pull request attempting to address that issue or a different issue?

Thanks for checking, @tompollard! I believe this change improves the user experience. The PR was created to inform PhysioNet users that they’ll be able to download datasets from S3 if their AWS credentials are configured on the platform. I've updated the PR description to make this clearer.

tompollard (commented):
> > Thanks for checking, @tompollard! I believe this change improves the user experience. The PR was created to inform PhysioNet users that they’ll be able to download datasets from S3 if their AWS credentials are configured on the platform. I've updated the PR description to make this clearer.

Thanks @Chrystinne, in this case I don't think this logic is correct. As far as I can see, users still require AWS credentials in order to see the download link. Everything sits inside the following if clause:

```
      {% if has_s3_credentials and project.aws.sent_files %}
            ....decide what content to display
      {% endif %}
```

Can you review the logic. It doesn't seem right to me. Also, `project.access_policy != AccessPolicy.OPEN` is in an odd place.

Chrystinne:
> > > Thanks for checking, @tompollard! I believe this change improves the user experience. The PR was created to inform PhysioNet users that they’ll be able to download datasets from S3 if their AWS credentials are configured on the platform. I've updated the PR description to make this clearer.
> 
> Thanks @Chrystinne, in this case I don't think this logic is correct. As far as I can see, users still require AWS credentials in order to see the download link. Everything sits inside the following if clause:
> 
> ```
>       {% if has_s3_credentials and project.aws.sent_files %}
>             ....decide what content to display
>       {% endif %}
> ```
> 
> Can you review the logic. It doesn't seem right to me. Also, `project.access_policy != AccessPolicy.OPEN` is in an odd place.

Actually, the has_s3_credentials method you're referring to checks PhysioNet’s account AWS credentials (the ones configured on the server side) used to interact with S3 via the Boto3 library.

tompollard (approved):
Thanks for checking, @tompollard! I believe this change improves the user experience. The PR was created to inform PhysioNet users that they’ll be able to download datasets from S3 if their AWS credentials are configured on the platform. I've updated the PR description to make this clearer.

Thanks @Chrystinne, in this case I don't think this logic is correct. As far as I can see, users still require AWS credentials in order to see the download link. Everything sits inside the following if clause:

      {% if has_s3_credentials and project.aws.sent_files %}
            ....decide what content to display
      {% endif %}
Can you review the logic. It doesn't seem right to me. Also, project.access_policy != AccessPolicy.OPEN is in an odd place.

> Actually, the has_s3_credentials method you're referring to checks PhysioNet’s account AWS credentials (the ones configured on the server side) used to interact with S3 via the Boto3 library.

Got it, thanks for explaining. Makes sense then!

bemoody:
Hmm.  There's a deeper problem here: if somebody has already signed DUA, then adding their AWS credentials won't help, AFAIK.

Before we start telling people they need to add their AWS credentials, we should ensure they will *be able to gain access to the project* after doing so.

bemoody:
```
{% elif project.access_policy != AccessPolicy.OPEN and not user.cloud_information.aws_verification_datetime %}
```
Checking `project.access_policy` here is not right.  What we want is to check `project.aws.is_private`.  We *also* probably want to check `user.awsverification.aws_verification_available()` before displaying this message.

The logic here is *complicated*, and therefore it would be better to define the logic in Python, not in the template file.

bemoody:
```
To download the files using AWS command line tools, first
<a href="https://physionet.org/settings/cloud/">configure your AWS credentials</a>.
```
Should be `href="{% url 'edit_cloud' %}"`

tompollard:
> Hmm. There's a deeper problem here: if somebody has already signed DUA, then adding their AWS credentials won't help, AFAIK.

This sounds independent of whether or not the link is displayed? Do we have that issued tracked?

bemoody:
See issue #2442.

Chrystinne:
> Before we start telling people they need to add their AWS credentials, we should ensure they will _be able to gain access to the project_ after doing so.
@bemoody Thanks for your comments! I don't think we will have this issue, as the AWS CLI command won't be displayed if the user isn't included in the Access Point policy — the s3_uri check will return None, and the command won't be shown.

Chrystinne:
> ```
> {% elif project.access_policy != AccessPolicy.OPEN and not user.cloud_information.aws_verification_datetime %}
> ```
> 
> Checking `project.access_policy` here is not right. What we want is to check `project.aws.is_private`. We _also_ probably want to check `user.awsverification.aws_verification_available()` before displaying this message.
> 
> The logic here is _complicated_, and therefore it would be better to define the logic in Python, not in the template file.

Yes, I agree with these comments, and I’ll check the user.awsverification.aws_verification_available() method. Thanks!

Chrystinne:
> > ```
> > {% elif project.access_policy != AccessPolicy.OPEN and not user.cloud_information.aws_verification_datetime %}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Checking `project.access_policy` here is not right. What we want is to check `project.aws.is_private`. We _also_ probably want to check `user.awsverification.aws_verification_available()` before displaying this message.
> > The logic here is _complicated_, and therefore it would be better to define the logic in Python, not in the template file.
> 
> Yes, I agree with these comments, and I’ll check the user.awsverification.aws_verification_available() method. Thanks, @bemoody!

Chrystinne:
> See issue #2442.

I've mande some comments there.

Chrystinne:
@bemoody could you please have another look?

Chrystinne:
@bemoody Could you please have another look?

bemoody (approved):

PR #2446, opened 2025-07-15
Enable users to be added to access point policies after they have signed DUA

Chrystinne:
PhysioNet users are not able to access projects from S3 buckets if they sign the DUA before configuring their AWS credentials on PhysioNet (by verifying their identity).

This PR adds the button "Enable AWS access" that users will click to access the files using AWS command line tools. In that case, they will be added to the access point policy linked to the project.

### Changes Made:

Added method to AWS model to check if a user is already included in any access point policy for a project;

Added "Enable AWS access" link in project templates that appears when users can't see the S3 URI for private projects but are eligible for AWS access.

### Prerequisites:

- User must have signed the DUA for the project;
- User must have verified AWS credentials in their cloud settings;
- Project must be private and have files sent to S3.

### DIFF
--- a/physionet-django/project/modelcomponents/storage.py
+++ b/physionet-django/project/modelcomponents/storage.py
@@ -68,6 +68,21 @@ class AWS(models.Model):
         """
         return f's3://{self.bucket_name}/{self.project.slug}/{self.project.version}/'
 
+    def user_included_in_ap_policy(self, user):
+        """
+        Check if a user has access to this AWS project through any access point.
+
+        Args:
+            user (User): The user to check
+
+        Returns:
+            bool: True if user has access, False otherwise
+        """
+        return AWSAccessPointUser.objects.filter(
+            user=user,
+            aws=self
+        ).exists()
+
     def __str__(self):
         return f"AWS instance for project: {self.project.slug}"
 
--- a/physionet-django/project/modelcomponents/storage.py
+++ b/physionet-django/project/modelcomponents/storage.py
@@ -68,7 +68,7 @@ class AWS(models.Model):
         """
         return f's3://{self.bucket_name}/{self.project.slug}/{self.project.version}/'
 
-    def user_included_in_ap_policy(self, user):
+    def user_in_access_point_policy(self, user):
         """
         Check if a user has access to this AWS project through any access point.
 
--- a/physionet-django/project/views.py
+++ b/physionet-django/project/views.py
@@ -2272,6 +2272,11 @@ def enable_aws_access(request, project_slug, version):
         messages.error(request, 'Project not found.')
         return redirect('project_home')
 
+    # Verify if the user has access to the project
+    if not can_access_project(project, user, request):
+        messages.error(request, 'You do not have permission to access this project.')
+        return redirect('published_project', project_slug=project_slug, version=version)
+
     try:
         result = add_user_to_access_point_policy(project, user)
         if result:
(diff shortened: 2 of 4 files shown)


physionet-django/project/modelcomponents/storage.py
```diff
@@ -68,6 +68,21 @@ def public_s3_uri(self):
         """
         return f's3://{self.bucket_name}/{self.project.slug}/{self.project.version}/'
 
+    def user_included_in_ap_policy(self, user):
```
tompollard: i think it is clearer to include "access_point" here. e.g. `user_in_access_point_policy`

uWSGI-2.0.19.1-cp39-cp39-macosx_12_0_arm64.whl
tompollard, line 1: should not be here?

uWSGI-2.0.22-cp39-cp39-macosx_12_0_arm64.whl
tompollard, line 1: should not be here?

your_database.db
tompollard, line 1: should not be here?

physionet-django/project/views.py
```diff
@@ -2252,6 +2260,39 @@ def data_access_requests_overview(request, project_slug, version):
                    'is_additional_reviewer': proj.is_data_access_reviewer(reviewer)})
 
 
+@login_required
+def enable_aws_access(request, project_slug, version):
```
tompollard, line 2264: where are you checking whether the user should have access to the dataset?

tompollard (changes requested):
added a few quick notes

uWSGI-2.0.22-cp39-cp39-macosx_12_0_arm64.whl
Chrystinne: Thanks for spotting that, Tom!

physionet-django/project/views.py
Chrystinne: > where are you checking whether the user should have access to the dataset?

I haven't included this verification before since the modification is in the "Access the Files" section, which users only see after signing the DUA. However, I added the verification for extra security. Thanks for your comments, @tompollard!

physionet-django/search/urls.py
```diff
@@ -14,6 +14,11 @@
     path('challenge/', views.challenge_index, name='challenge_index'),
     path('model/', views.model_index, name='model_index'),
     path('content/', views.content_index, name='content_index'),
+    path(
+        'content/<project_slug>/<version>/enable-aws-access/',
+        project_views.enable_aws_access,
+        name='enable_aws_access'
+    ),
```
bemoody, line 22: For consistency with `sign-dua`, `request-access`, etc., let's use `'enable-aws-access/<project_slug>/<version>/'` instead of `'content/<project_slug>/<version>/enable-aws-access/'`.

Chrystinne: @bemoody I've made that change. I've also tested locally the inclusion in APs and it's working.

bemoody (approved):
All looks good, thanks!

Chrystinne:
> All looks good, thanks!

Thanks for your review!

Chrystinne (commented):
@tompollard I've addressed your comments. Thanks for your review, Tom!

tompollard (approved):

bemoody:
Unfortunately, this code is badly broken, so I have rolled back the server.

Issue #2317, opened 2024-10-23
the `--diff` option for flake8 was removed in version 6.0.0


tompollard:
Currently we use `flake8 --diff` in our style checks:

https://github.com/MIT-LCP/physionet-build/blob/5fdcd7729f9f7f709482ab291082da20e28fdf4b/.github/workflows/physionet-build-test.yml#L118-L137

These tests are likely to start failing sometime soon, because the `--diff` option was removed from flake8 in version 6.0.0: https://flake8.pycqa.org/en/latest/release-notes/6.0.0.html

bemoody:
flake8 is a combination of two mostly-unrelated things: pyflakes (check for faulty/suspicious logic) and pycodestyle (check for incorrect formatting.)

The diff option was removed from flake8 because it doesn't work reliably for pyflakes errors.  (Of course, unreliable doesn't mean it's useless, but we probably shouldn't be relying on it.)

pycodestyle still supports differential checking (and of course this is not perfect either, but nothing about code style checking is perfect.)

I think the best thing would be:
- fix all of the existing pyflakes errors (currently 259 of them)
- require new pull requests to have zero errors from pyflakes
- continue to use differential checks for pycodestyle

In the short term of course we can continue using flake8 v5.

PR #2447, opened 2025-07-16
Cleanup and simplification of GitHub workflows

bemoody:
Assorted minor improvements to GitHub workflows:

- Now that checks are always peformed *before* pushing (using `merge_group`), I don't see a point in running exactly the same checks on the same commits, *after* pushing (using `push`).  It just wastes time and energy.

- Separate the code style (flake8) checks into their own workflow.  They only need to run once (not once for every matrix configuration), and it's nice for developers not to have style errors conflated with test failures.

- Simplify the logic of the flake8 diff check and remove the duplicated code.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -3,10 +3,12 @@
 name: Codespell
 
 on:
-  push:
-    branches: [dev]
   pull_request:
-    branches: [dev]
+    branches:
+      - dev
+  merge_group:
+    types:
+      - checks_requested
 
 permissions:
   contents: read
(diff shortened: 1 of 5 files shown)


bemoody:
I considered also changing style.yml to check for style issues in the *pre-merged* rather than *post-merged* code (diff the pull-request head against the merge base, rather than the merge against the target branch.)

- If "your code" is acceptable to flake8, but `git merge` introduces style errors, then I tend to think that's not your problem and it should just be allowed.

- It would also be easier to fix style errors if they were displayed in the pre-merged context (e.g., using pre-merged line numbers.)

- If we don't care about post-merge style errors, it'd be one fewer thing that needs to run in `merge_group`.

But I've left it as-is for now.

bemoody:
> flake8: error: unrecognized arguments: --diff

> github doesn't use -o pipefail

:|

bemoody:
see also issue #2317

tompollard (approved):
Nice, thanks!

PR #2449, opened 2025-07-17
fix: add missing variable initialization

Chrystinne:
This PR adds missing variable initialization (user_in_access_point_policy).

### DIFF
--- a/physionet-django/project/test_views.py
+++ b/physionet-django/project/test_views.py
@@ -24,7 +24,8 @@ from project.models import (
     PublishedAuthor,
     PublishedProject,
     StorageRequest,
-    SubmissionStatus
+    SubmissionStatus,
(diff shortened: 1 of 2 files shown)


bemoody (changes requested):
Typo in the test case: `AWS.object.create` should be `AWS.objects.create`.  Otherwise this looks good, the new test case catches the error in the previous code.

bemoody (approved):

Chrystinne:
> Typo in the test case: `AWS.object.create` should be `AWS.objects.create`. Otherwise this looks good, the new test case catches the error in the previous code.

Great! Thanks, Benjamin!

bemoody (approved):

bemoody (approved):

PR #2450, opened 2025-07-17
Add authentication check before access point policy verification

Chrystinne:
This PR checks if user is authenticated before checking if they are in access point policy in the user_in_access_point_policy method.


bemoody (approved):

bemoody:
This is needed because otherwise the published_project view would crash if the project is *controlled* and *has an AWS mirror* and *the user is not logged in.*

Issue #2330, opened 2025-01-17
Automated handling of expired AWS IDs


Chrystinne:
Currently, our system requires an automated solution for managing expired AWS IDs in access point policies. This will improve reliability and reduce maintenance overhead by:

- Automatically detecting expired AWS IDs at regular intervals;

- Removing expired AWS IDs from policies without disrupting users with valid AWS IDs;

- Synchronizing access point policies with user authorization status through periodic updates by adding/removing/updating users whose authorization status has changed or whose AWS ID has changed.

When users remove their AWS account from PhysioNet, they must be removed from the access point policies they were previously added to. Similarly, when users add a new AWS account, they must be re-added to the same access points they were previously using.

bemoody:
A couple questions that we should try to answer:

1. What happens when an invalid userid is specified in `PutAccessPointPolicy`?  Does it behave the same way as `PutBucketPolicy` (i.e., fail with an error message that explicitly states which ID was invalid)?

2. If an invalid userid is present in an existing access point policy, what does `GetAccessPointPolicy` return?  Does it behave the same way as `GetBucketPolicy` (i.e., valid userids are converted to ARNs while invalid userids are left as `AIDA` strings)?

I know how bucket policies work in the case of userids that have been deleted from their AWS account.  I haven't checked whether access-point policies work the same way.

I also don't know what happens if an *entire AWS account is deleted*, particularly if we were to grant access to `root` ARNs.

I know that there are (or were, at one time) some people on PhysioNet whose "aws_id" was a 12-digit number and was not a valid AWS account number (i.e., PutBucketPolicy would reject it.)  But since those IDs were unverified, I don't know whether they were *formerly valid* or *never valid*.

bemoody:
Answer to question 1:

> 1. What happens when an invalid userid is specified in `PutAccessPointPolicy`?  Does it behave the same way as `PutBucketPolicy` (i.e., fail with an error message that explicitly states which ID was invalid)?

Yes.  In this situation, `s3_control.put_access_point_policy` fails with an exception like this:
```
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/physionet/physionet-build/physionet-django/project/cloud/s3.py", line 1046, in initialize_access_points
    insert_access_point_policy(access_point, data_access_point_name, project, subset_aws_ids)
  File "/home/physionet/physionet-build/physionet-django/project/cloud/s3.py", line 1012, in insert_access_point_policy
[...]

PR #2451, opened 2025-07-18
Implement retry mechanism for malformed S3 access point policies

Chrystinne:
This PR aims to solve the issue reported in #2330 

We implement a retry mechanism to insert S3 access point policies to prevent errors when updating them after users delete their AWS credentials on PhysioNet after having been added to policies.

Our strategy is to parse out the invalid users, mark them as unverified, and retry inserting the access point policy.

### Changes Made:
- Detect and remove invalid AWS user IDs from policies until all of them are processed;
- Preserve new users being added during retry attempts;
- Add safety limit to prevent infinite loops.

### DIFF
--- a/physionet-django/project/cloud/s3.py
+++ b/physionet-django/project/cloud/s3.py
@@ -478,7 +478,7 @@ def get_aws_accounts_for_access_point(access_point_name):
             if user.cloud_information.aws_userid
             else None
         )
-        if re.search(aws_id_pattern, aws_id):
+        if aws_id and re.search(aws_id_pattern, aws_id):
             aws_accounts.append({'aws_id': aws_id, 'aws_userid': aws_userid})
 
     return aws_accounts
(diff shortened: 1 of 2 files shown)


bemoody:
Thanks.  This probably *works* but the logic is hard to follow.

As things stand, `insert_access_point_policy` receives a list of users (`subset_aws_ids`) who should be granted access.  The function that calls `insert_access_point_policy` (`initialize_access_points` or `add_user_to_access_point_policy`) is responsible for deciding which users should be included in that list.

I think that if an error occurs, we should be able to:
- Catch the exception in `insert_access_point_policy`
- Call `mark_user_unverified`
- Remove the invalid user from `subset_aws_ids`
- Try to call `create_data_access_point_policy` and `set_data_access_point_policy` again

That should be much simpler.  There shouldn't be a need to call `get_aws_accounts_for_access_point` a second time.

The idea is to keep logic localized.  Errors should be handled in the function that "knows how" to handle the error.  `insert_access_point_policy` has the necessary information (the list of `subset_aws_ids`.)  `set_data_access_point_policy` doesn't (all it has is a JSON policy.)

Chrystinne:
Thank you for your comments, @bemoody! It makes total sense to handle the logic for dealing with malformed policy errors in `insert_access_point_policy` instead of `set_data_access_point_policy`. Could you have another look, please?

Chrystinne:
@bemoody I have also included in the tests users who have deleted their credentials after signing the DUA, and then re-entered them. They are being added again to the access point policy associated with the same access point they were associated with before deleting their credentials, which is what we want.

bemoody (approved):
Looks great!

Chrystinne:
> Looks great!

Great! Thank you for your contributions!

Issue #2454, opened 2025-07-22
Fix rendering issues on new front page design


tompollard:
The new front page design doesn't look nice on @bemoody's machine (Firefox, no JS). A couple of the major issues are:

1. No truncation of news item content.
2. The multicolumn cards for news appear as a single, large card.

Fix this.

PR #2444, opened 2025-07-15
Upsidelab homepage redesign

tompollard:
This pull request is a continuation of https://github.com/MIT-LCP/physionet-build/pull/2430.

The pull request implements a homepage redesign using bootstrap and custom styles. All of the new styles are wrapped in a modern-ui class to isolate them from the rest of the site. The redesign includes:

- a whole homepage rewrite
- implementing new navigation and footer on subpages
- layout changes on subpages
- mobile support

[Figma designs](https://www.figma.com/design/t2CVcw2XvgAMGqqUrmKiDr/PhysioNet---Platform-Designs?node-id=230-4266&t=2xfdEukK3ZJtkPrb-0)

## New changes in response to feedback from @bemoody 

> The logo images should be configurable. 

These are now configurable in the .env file with `SITE_HEADER_LOGO` and `SITE_FOOTER_LOGO`

> On the front page, the links to news items ("Preparing New Challenges", etc.) and to projects ("Self Managed Access Database Demo", etc.) do not have any color or decoration to indicate that they are links.

Changing the text to pink looked pretty ugly, so instead I added a pink underline on the hyperlinks.

> There shouldn't be `<button>` nested inside an `<a>`

Fixed.

> All fonts and images must be hosted locally (and must be under a free/open-source license.). All scripts and stylesheets must be hosted locally.

I have add the fonts (Inter), stylesheets (Swiper), and images (Bootstrap Icons) to the static folder.

PR #2455, opened 2025-07-22
Add `news_preview` template tag to truncate front page news items.

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/2454, the new design added in https://github.com/MIT-LCP/physionet-build/pull/2444 does not truncate news items. As a result, the front page doesn't look great.

This pull request:

1. adds a new `news_preview` template tag that processes news content for before displaying on the front page (removes images, strips HTML, truncates content).
2. Adds a new templatetags test module and adds tests for the new template tag.

These changes partially address https://github.com/MIT-LCP/physionet-build/issues/2454.


bemoody:
The fundamental concept (mandatory mechanical truncation) is not a particularly great idea.  And then, the implementation is essentially reinventing Django functionality.

(A similar server-side effect, but less bad, would be to use `truncatewords_html`.)

Visually clipping the content, without modifying it, should be doable in CSS in the same way as titles are clipped (-webkit-line-clamp stuff.)

*Modifying* the content and formatting would be better done under the editor's control, e.g. by adding a "short_description" field, or splitting "content" into "brief_content" and "extended_content".

bemoody:
> (A similar server-side effect, but less bad, would be to use truncatewords_html.)

Ahh, I see this is "truncating words measured by characters" which is a little different and better than "truncating to N words" or "truncating to N characters".

Even so: client side is simpler and better; editor control is superior to both.

tompollard:
Thanks for looking.

> Modifying the content and formatting would be better done under the editor's control, e.g. by adding a "short_description" field, or splitting "content" into "brief_content" and "extended_content".

I agree this would be nice, but it's a new feature we would need to add. 

> (A similar server-side effect, but less bad, would be to use truncatewords_html.)

Truncate Words looks like it is doing something fairly different? https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#truncatewords-html. 

I guess there may be a way of combining multiple Django tags, but it seems best to strip images, HTML tags etc. Is it that bad to use our own function?

> Visually clipping the content, without modifying it, should be doable in CSS in the same way as titles are clipped (-webkit-line-clamp stuff.)

Can you explain why sending all the data and then clipping with CSS is better? Happy to do that, just want to understand the reasoning before I do.

bemoody:
> it seems best to strip images, HTML tags etc.

That's a very different question than whether the content should be clipped/truncated (either mechanically or by hand).

I'd say no, HTML shouldn't be forbidden: if a news item talks about `CO<sub>2</sub>` then forcing that to be displayed as `CO2` (prominently on the front page!) looks ugly.  Judicious use of paragraph breaks, italics, links, lists can all be good for readability, even in a short blurb.

> Can you explain why sending all the data and then clipping with CSS is better?

In comparison to truncatewords_html or something similar - CSS would clip at the end of a line rather than clipping somewhere more-or-less random, and it'd clip based on the visual size of the content so it'd appear consistent across multiple news items.

tompollard:
Thanks Benjamin, I can make these changes but quite honestly it feels like we're creating extra work with limited reward. If we're adding a short description later, all this will get replaced anyway.

tompollard:
@bemoody Just added `truncatewords_html` for now. Does what we need and can be improved later.

bemoody:
This is pretty funny, actually.  style.css is *already doing what it should*, except that they failed to consider the fact that `<p>` don't nest.

```
.modern-ui .card--news p {
  font-size: 0.875rem;
  margin: 0;
  display: -webkit-box;
  -webkit-line-clamp: 7;
  -webkit-box-orient: vertical;
  overflow: hidden;
  color: var(--gray-color);
}
```

PR #2456, opened 2025-07-22
Improve rendering of news with no JS.

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/2454, when Javascript is disabled the new front-page news items are displayed as a single full-width block instead of individual cards. This pull request falls back to a CSS grid for displaying the news items if Javascript is disabled.


bemoody (approved):
Looks nice!

PR #2458, opened 2025-07-24
Display front_page_banner items at top of page. Fixes #2457.

### LINKED ISSUE #2457: Front page news duplicated
With the new home page template (pull #2444), all of the news items are displayed twice.  Once above the featured resources, once below the featured resources.

I'm guessing that the intention here was that the "front page banner" items would be displayed at the top of the page, and "regular" news items would be displayed at the bottom.  But that's not what the code is doing currently.

(Note that there aren't any front page banner items in demo-notification.json.)

tompollard:
The new front page displays two news carousels. As discussed in https://github.com/MIT-LCP/physionet-build/issues/2457, both carousels currently display the same news items.

This pull request fixes this issue:

- The top carousel now only displays news items marked as "front page banner".
- The lower carousel displays all news items (we could exclude front page banner items, but I prefer to display them in both places).

### DIFF
--- a/physionet-django/physionet/views.py
+++ b/physionet-django/physionet/views.py
@@ -24,7 +24,7 @@ def home(request):
     """
     featured = PublishedProject.objects.filter(featured__isnull=False).order_by('featured')[:6]
     latest = PublishedProject.objects.filter(is_latest_version=True).order_by('-publish_datetime')[:6]
-    news_pieces = News.objects.all().order_by('-publish_datetime')[:5]
+    news_pieces = News.objects.filter(front_page_banner=False).order_by('-publish_datetime')[:5]
     front_page_buttons = FrontPageButton.objects.all()
     front_page_banner = News.objects.filter(front_page_banner=True)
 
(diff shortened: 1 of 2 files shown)


bemoody:
I think that we also want to *exclude* front page banner items from news_pieces.  If an item appears at the top, it shouldn't appear again at the bottom.

tompollard:
As mentioned, I prefer the banner news items to appear in both places, but okay!

tompollard:
> I think that we also want to exclude front page banner items from news_pieces. If an item appears at the top, it shouldn't appear again at the bottom.

Banner items are now excluded from the lower carousel.

bemoody:
Sorry for posting without reading.  I think I understand the desire to keep the chronological list chronological.  I would still argue that nobody wants to read the same thing twice, and it gives a good impression if every element on the page contributes something of value. 

I would think of it like any blog or forum software that lets you mark items as "sticky".  Of course, people can still click "More news" if they want to see a full chronological list.

bemoody (approved):

tompollard:
> I would still argue that nobody wants to read the same thing twice, and it gives a good impression if every element on the page contributes something of value.

No problem, I think you're right on reflection - not duplicating the item probably gives a better experience for the user.

PR #2465, opened 2025-07-30
Display more information about cloud mirrors

bemoody:
In the Manage Project page (/console/published-projects/X/Y/) for a published project, there are buttons to create cloud mirrors (Google Cloud Storage and Amazon S3).

Currently, there's very little information shown here (or anywhere else on the site) about the status of the cloud mirror.  Add some basic information about the mirror if it exists:

**Location of the data:** the URI for the data in the cloud system

**Whether project files have been uploaded,** which is also the flag that controls whether the cloud mirror is advertised to users

**Whether the project zip archive has been uploaded**

**Creation time:** the time that somebody first clicked the "send files" button

**Last upload completion time:** the time that the most recent successful upload process finished

**Access policy type** (public or private)

**Associated access groups:** identifiers for user groups (GCP) or access points (AWS) that can be used to access the mirror

### DIFF
--- a/physionet-django/console/static/console/css/cloud-mirrors.css
+++ b/physionet-django/console/static/console/css/cloud-mirrors.css
@@ -42,14 +42,14 @@
     content: "\f375";           /* aws */
 }
 
-.project-site-status-open,
-.project-site-status-restricted,
-.project-site-status-embargo,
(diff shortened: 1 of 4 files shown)


tompollard (approved):
Looks good, thanks!

Issue #1076, opened 2020-06-18
Explore how we can add flexibility to the storage location of project files.


tompollard:
There is an LCP project where we would like to make data available only on BigQuery (not on PhysioNet servers or cloud buckets).

Modifying PhysioNet so that users have more flexibility in where files are hosted could be helpful more generally, e.g. 

- data is too large to host on our local servers.
- specific requirements around where data can be physically hosted (e.g. only on New Zealand territory).
- benefits in terms of controlling and auditing access.

briangow:
@tompollard , @bemoody , we should revive the discussion around this. We have identified a number of projects where we would like to have the project tables on BigQuery, but don't see the need to have the project files in a GCP bucket. 

Since we have agreed that the same permission group should be used for buckets and BigQuery, we should use the same code for generating the access group. Currently to create the `access_group` you need to create or reference a bucket. Developing independent code for creating the `access_group` could be beneficial even if we continue to add files to the BigQuery dataset manually. We would probably also want to standardize how the BigQuery dataset gets named. 

A separate discussion addresses the larger picture around integrating the platform with BigQuery, including the sending of files to BigQuery in an automated fashion: https://github.com/MIT-LCP/physionet-build/issues/2058 . 

Also, this discussion should be considered, with respect to refactoring this code: https://github.com/MIT-LCP/physionet-build/issues/1927.

tompollard:
(It's not at all clear from my initial post, but this issue also related to the requirement for local storage. I wanted us to be able to host data on the cloud, without a public local copy).

PR #2466, opened 2025-07-31
Add management command for linking projects to already existing buckets and groups

briangow:
Historically, projects and groups have been created manually (in addition to being created via the PN platform). This makes maintenance of these resources via the PN platform challenging since they don't have entries in the `GCP` and `DataAccess` objects. 

The management command being introduced here can be used to link a project with an existing bucket and group. 

I've tested this from the local demo database:
```
python3 manage.py link_existing_gcp_bucket \
  --project-slug demopsn \
  --project-version 1.0 \
  --group-email  sccmdatathon2023@physionet.org \
  --bucket-name reflacx-xray-localization-1.0.0.physionet.org \
  --user-email admin@mit.edu
```

checking the GCP and DataAccess objects shows:
```
print(gcp_obj.__dict__)
{'_state': <django.db.models.base.ModelState object at 0x107c32060>, 'id': 1, 'project_id': 5, 'bucket_name': 'reflacx-xray-localization-1.0.0.physionet.org', 'access_group': 'sccmdatathon2023@physionet.org', 'is_private': True, 'sent_zip': False, 'sent_files': True, 'managed_by_id': 1, 'creation_datetime': datetime.datetime(2025, 7, 31, 12, 1, 20, 499251, tzinfo=datetime.timezone.utc), 'finished_datetime': None}
```

```
data_access=DataAccess.objects.get(project=project)
for field, value in data_access.__dict__.items():
	if not field.startswith('_'):  
		print(f"{field}: {value}")
id: 1
project_id: 5
platform: 3
location: sccmdatathon2023@physionet.org
```
Please note that to test this from the local demo database, I had to allow passing of `user_project` in `project.gcp.add_email_bucket_access`. This should not be necessary from the production server since it knows what account to bill. 

For consistency with how the buckets and groups are currently created via the PN platform, I elected to:
- create a DataAccess object and set the location when doing this linkage. This is done as part of "Create bucket and send files". 
- set `sent_files` to `True` in the GCP object. Since the bucket was already created, we assume it was populated with files. If this isn't done, the admin console will show the message below. When this is done, the admin console will show "Resend files", as desired. 
  - The files are being sent to GCP. The bucket name is: reflacx-xray-localization-1.0.0.physionet.org. If this message is here for a long time, check the Django "process_tasks"

### DIFF
--- a/physionet-django/project/management/commands/link_existing_gcp_bucket.py
+++ b/physionet-django/project/management/commands/link_existing_gcp_bucket.py
@@ -1,120 +1,86 @@
+
 import google.auth.transport.requests
 from django.core.management.base import BaseCommand, CommandError
 from project.cloud.gcp import ROLES, add_email_bucket_access
-from project.models import PublishedProject, DataAccess
 from project.modelcomponents.storage import GCP
+from project.models import PublishedProject, DataAccess
 from user.models import User
 
 
 class Command(BaseCommand):
-    help = "Link an existing GCP bucket and group to a project."
+    help = 'Link an existing GCP bucket and group to a project.'
 
     def add_arguments(self, parser):
-        parser.add_argument("--project-slug", required=True)
-        parser.add_argument("--project-version", required=True)
-        parser.add_argument("--group-email", required=True)
-        parser.add_argument("--bucket-name", required=True)
-        parser.add_argument(
-            "--user-email", required=True, help="Email of user to set as manager"
-        )
-        parser.add_argument(
-            "--dry-run",
-            action="store_true",
-            help="Show what would be changed, but do not modify anything",
-        )
+        parser.add_argument('--project-slug', required=True)
+        parser.add_argument('--project-version', required=True)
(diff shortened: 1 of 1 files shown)


physionet-django/project/management/commands/link_existing_gcp_bucket.py
```diff
@@ -0,0 +1,86 @@
...
+import google.auth.transport.requests
+from django.core.management.base import BaseCommand, CommandError
+from project.cloud.gcp import ROLES, add_email_bucket_access
+from project.modelcomponents.storage import GCP
+from project.models import PublishedProject, DataAccess
+from user.models import User
+
+
+class Command(BaseCommand):
+    help = 'Link an existing GCP bucket and group to a project.'
```
tompollard, line 11: Could you add a bit more detail here about what the command is doing?

briangow:
Closing this. Follow #1076 for related information.

PR #2469, opened 2025-08-07
Prevent cloud uploads when an upload task is already pending

bemoody:
In the published project management page there are buttons to upload a project to Google Cloud and/or AWS.  If the project has already been uploaded then it can be re-uploaded.

We don't want to upload the same project twice accidentally; that would be pointless and wasteful.

Previously, therefore, the button would be hidden if the `sent_files` flag was false (on the grounds that, if `sent_files` was false, that presumably meant that somebody had clicked the button but the task had not yet completed.)  Even then, the button being hidden wouldn't prevent you from submitting the same form twice.

`sent_files` can be false even though there is no pending upload task - the upload task may have failed, or the sent_files flag may have been manually cleared to indicate the bucket was removed.  (In the future, we should add console functions to remove cloud buckets.)

And there can be a pending upload task even though `sent_files` is true - the project is currently being re-uploaded.

So instead of using `sent_files`, use `get_associated_tasks` to check whether there is actually a task pending.  If there is a task pending, then don't create a new one if the form is submitted again; and also, if there is a task pending, then make the HTML button disabled (but still visible.)

### DIFF
--- a/physionet-django/console/tasks.py
+++ b/physionet-django/console/tasks.py
@@ -100,7 +100,7 @@ def associated_task(model, param, *, read_only=False, field='pk'):
     return decorate
 
 
-def get_associated_tasks(instance, *, read_only=None):
+def get_associated_tasks(instance, *, read_only=None, name=None):
     """
     Find pending tasks associated with a model instance.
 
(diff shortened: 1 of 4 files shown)


tompollard (approved):
looks good to me

PR #2470, opened 2025-08-11
Replace banner.png with WebP version

bemoody:
The "banner.png" image shown on the front page is quite large.  Replace it with a compressed version.


tompollard (approved):
Thanks!

PR #2473, opened 2025-08-20
Provide access to users for open-geo-restricted datasets on new bucket using Access Points

Chrystinne:
We need to provide access to users for open-geo-restricted datasets on AWS. This PR implements the new functionality.

Since open-geo-restricted datasets don't have a DUA associated with them, users can request access on AWS by clicking the "Enable AWS access" button on the project's page. 

The "Enable AWS access" button first checks whether the user can access files. If the user has access, we add them to the access point.

**Updates needed:**

**- [Done] AWS console:**
- Create the new bucket physionet-geo-restricted;
- Set up policy on the new bucket to delegate access management to Access Points.

**- [Done] Code:**
For providing access to users for open-geo-restricted projects via CLI:
- Removing the has_signed_dua verification from the S3 URI creation should be sufficient;
- Set is_private = True for open-geo-restricted datasets in the AWS object.

To upload open-geo-restricted projects to S3:
- Set the bucket policy if the bucket physionet-geo-restricted was newly created [this step won't be necessary as it was done via AWS console, but it's important to check whether the code is able to do that];
- Avoid updating open bucket policy if projects are geo-restricted.

**- [TO-DO] Database:**

is_private = True for all open and geo-restricted datasets;
bucket_name = 'physionet-geo-restricted' for all open and geo-restricted datasets.

**- [TO-DO] Server:**

Add configuration for the new bucket: S3_GEO_RESTRICTED_ACCESS_BUCKET=physionet-geo-restricted;

### DIFF
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -291,7 +291,7 @@ S3_SERVER_ACCESS_LOG_BUCKET = config('S3_SERVER_ACCESS_LOG_BUCKET', default=None
 # Bucket name for the S3 bucket containing the controlled access data
 S3_CONTROLLED_ACCESS_BUCKET = config('S3_CONTROLLED_ACCESS_BUCKET', default=None)
 
-# Bucket name for the S3 bucket containing the geo-restricted access data
+# Bucket name for the S3 bucket containing geo-restricted projects with an 'OPEN' access policy.
 S3_GEO_RESTRICTED_ACCESS_BUCKET = config('S3_GEO_RESTRICTED_ACCESS_BUCKET', default=None)
 
 # Header tags for the AWS lambda function that grants access to S3 storage
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -292,7 +292,7 @@ S3_SERVER_ACCESS_LOG_BUCKET = config('S3_SERVER_ACCESS_LOG_BUCKET', default=None
 S3_CONTROLLED_ACCESS_BUCKET = config('S3_CONTROLLED_ACCESS_BUCKET', default=None)
 
 # Bucket name for the S3 bucket containing geo-restricted projects with an 'OPEN' access policy.
-S3_GEO_RESTRICTED_ACCESS_BUCKET = config('S3_GEO_RESTRICTED_ACCESS_BUCKET', default=None)
+S3_OPEN_ACCESS_BUCKET_WITH_LOGIN = config('S3_OPEN_ACCESS_BUCKET_WITH_LOGIN', default=None)
 
 # Header tags for the AWS lambda function that grants access to S3 storage
 AWS_HEADER_KEY = config('AWS_KEY', default=False)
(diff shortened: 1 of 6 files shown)


physionet-django/physionet/settings/base.py
```diff
@@ -290,6 +290,9 @@
 # Bucket name for the S3 bucket containing the controlled access data
 S3_CONTROLLED_ACCESS_BUCKET = config('S3_CONTROLLED_ACCESS_BUCKET', default=None)
 
+# Bucket name for the S3 bucket containing the geo-restricted access data
+S3_GEO_RESTRICTED_ACCESS_BUCKET = config('S3_GEO_RESTRICTED_ACCESS_BUCKET', default=None)
```
tompollard: The name "S3_GEO_RESTRICTED_ACCESS_BUCKET" doesn't feel right. I don't think there is anything special about the bucket that makes it "georestricted"?

Georestriction is something implemented on the PhysioNet side (and as a result, all buckets could be said to be "georestricted").

bemoody: There isn't anything special about the bucket.  It is "the bucket into which the PhysioNet server will deposit projects that the PhysioNet server has designated as georestricted".

If you're wondering why we would want to have different buckets for different access policies, the answer is to provide future flexibility.  There are a lot of things that S3 lets us control at the bucket level that cannot be controlled at access-point level.

tompollard: Thanks @bemoody. So all georestricted data will live in this bucket?

Chrystinne:
@bemoody I believe the PR covers all the necessary changes we discussed yesterday. Please let me know if I miss something or if you find anything that needs to be added.

physionet-django/physionet/settings/base.py
bemoody: This bucket would be for projects that are georestricted and have an "open" access policy (as the comment in .env.example says).  The comment in base.py is a little ambiguous.

bemoody:
We also need to set `is_private` in the AWS object.  This is set in `aws_bucket_management` (console/views.py).  `is_private` determines whether or not users see instructions for using an access point.

physionet-django/physionet/settings/base.py
tompollard: I would prefer something like: `S3_OPEN_ACCESS_BUCKET_WITH_LOGIN` for the variable name.

tompollard (approved):
Looks good to me!

physionet-django/project/views.py
```diff
@@ -1956,7 +1956,7 @@ def published_project(request, project_slug, version, subdir=''):
     s3_uri = None
     try:
         if project.aws.is_private:
-            if has_signed_dua and request.user.is_authenticated:
```
tompollard, line 1959: Doesn't this `has_signed_dua` need to be there for restricted and controlled projects?

Chrystinne: Not anymore. We're now checking project.aws.is_private to create the appropriate S3 URI.

tompollard: Looks like you added `has_signed_dua` back in a force push? Bit confused about what happened here!

Chrystinne: I'm solving conflicts with the dev branch before removing has_signed_dua.

Chrystinne:
@bemoody To solve the merge conflicts, I've done the rebase with dev and reapplied the last commits. Could you please have another look?

bemoody (approved):

physionet-django/project/views.py
bemoody: I believe the has_signed_dua there was always redundant.  If you don't have permission to view the project files then the S3 instructions are always hidden in the template.

PR #2476, opened 2025-08-28
Fix the navigation bar

bemoody:
This is an assortment of fixes and improvements to the new navigation bar, which currently has some serious problems.

- Always display the user menu, which currently is inaccessible on < 1200px windows.

- Use popover dialogs (which require a fairly new browser); these don't require JavaScript and work better than the current script does.  Existing behavior is preserved for old browsers.

- Remove the nested menu inside the hamburger menu.

- Remove fixed positioning.

- Add accessible text in various places.


physionet-django/templates/navbar.html
```diff
@@ -120,86 +144,48 @@
...
+        {% endif %}
+      {% endfor %}
+      <hr>
+      <div class="nav-sublinks">
         {% for button in FRONT_PAGE_BUTTONS %}
           <a class="nav-sublink" href="{{ button.url }}">
             <div class="nav-sublink__icon">
               {% if button.associated_image_path %}
-                <img src="{% static button.associated_image_path %}" alt="{{ button.label }}" height="32px">
+                <img src="{% static button.associated_image_path %}" alt="" height="32px">
```
tompollard, line 179: Is button.label not generating something useful here?

physionet-django/templates/navbar.html
```diff
@@ -24,50 +27,71 @@
...
+            </span>
+          </button>
+          <dialog class="nav-link__popup" id="nav__user_popup" popover
+                  aria-label="Account: {{ user }}">
+            <div class="nav-sublinks">
               {% if user.is_authenticated and perms.events.view_event_menu %}
                 <a class="nav-sublink" href="{% url 'event_home' %}">
                   <div class="nav-sublink__icon">
-                    <img src="{% static 'images/icons/events.png' %}" height="32px" />
+                    <img src="{% static 'images/icons/events.png' %}" alt="" height="32px" />
```
tompollard, line 94: If we are adding alts, shouldn't they be populated?

tompollard (approved):
Thanks Benjamin, looks fine to me. 

> Remove the nested menu inside the hamburger menu.

What is the motivation for removing the nest menu? Works nicely for me, but I guess you've found it difficult to navigate?

bemoody:
> What is the motivation for removing the nest menu?

I didn't like the visual inconsistency and I didn't like having to click twice.  When there are a grand total of six items, this feels silly.

It could be kept.  Maybe a better visual design would be if the `.nav-link`s were arranged horizontally.

physionet-django/templates/navbar.html
bemoody: button.label is the text that is displayed immediately afterwards.  If the image were omitted, the link should read "Data", not "Data Data".

physionet-django/templates/navbar.html
bemoody: The image is not standing in for text.  It does not provide any additional information.  It is purely decorative.

MDN gives an excellent explanation of what `alt` means: https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/alt

tompollard:
> I didn't like the visual inconsistency and I didn't like having to click twice. When there are a grand total of six items, this feels silly.

Fair enough, feel free to merge!

PR #2477, opened 2025-08-28
Initial GitHub workflow testing for Debian 13

bemoody:
Test compatibility with Debian 13, by running the `physionet-upgrade-test` workflow on a `debian:13` container.

This is a preliminary test, and I'm not proposing to change *all* of the scripts and GitHub workflows yet, because it'll probably be another couple of months before PhysioNet gets around to upgrading.  No point in running an excessive number of workflows in the meantime.


tompollard:
Looks like Pillow 10.3.0 doesn’t support Python 3.13.

tompollard:
@bemoody please could you rebase this?

bemoody:
This is still not working because the new glibc broke sandboxed-lightwave.

Issue #1324, opened 2021-04-20
Adding Upload Agreement 


li-lcp:
Please add the following upload agreement as part of the data submission process. The contributors should go through the agreement prior to uploading actual data.

Upload Agreement

By clicking "I Agree" and uploading files to PhysioNet, you agree that:

 1. You are the original creator, or copyright holder, or you have
    permission from the creators and copyright holders to share these
    files.

 2. These files do not contain information that could identify individual human subjects.

Please check all that apply:

 [ ] This project does not contain any data derived from human subjects.

 [ ] This project contains data derived from other de-identified
     datasets published on PhysioNet or elsewhere.

     You will need to cite these datasets in your project description,
     and explain how you created the derived data.

     Even if you are using data previously published elsewhere, we
     expect you to take all reasonable steps to ensure *the files you
     are uploading* are free of personally identifiable information.

 [ ] This project contains data obtained from human subjects, and all
     personally identifiable information has been removed.

     Your project description will need to explain how you collected
     the data, and the processes that you followed to remove any
     personally identifiable information before uploading.

     You should also indicate the IRB (Institutional Review Board) that
     approved your research, or document the reasons that your
     research is exempt from IRB approval.

tompollard:
See #1698

tompollard:
@Chrystinne I'm taking a look at this now.

PR #2479, opened 2025-08-29
Add upload agreement to project submission workflow. Ref #1324

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/1324, we would like to add an upload agreement as part of the project submission process.

This pull request adds a new "Upload Agreement" step to the project submission workflow. The submitting author must agree to our terms and complete one or more checkboxes. It is not possible to upload files until this step is complete.

The screenshot below shows this new step in the submission process:
 
Screenshot 2025-08-28 at 11 22 23 PM


physionet-django/project/views.py
```diff
@@ -1150,6 +1208,10 @@ def project_files(request, project_slug, subdir='', **kwargs):
      move_items_form, delete_items_form) = get_file_forms(
          project=project, subdir=subdir, display_dirs=display_dirs)
 
+    # Check if upload agreement has been accepted
+    from project.models import UploadAgreement
```
tompollard: didn't mean to leave this here. tom, look at moving.

tompollard:
@bemoody please could you take a look at this if you have the opportunity?

bemoody:
Haven't tested this yet.

I think that if a project is *reassigned* to a different submitting author, the new author should have to fill out the form again.  Could be done by deleting the agreement upon reassignment or by recording the user who signed it.

I think maybe we also want to require the form to be completed before *submitting*, as well as before uploading files.

bemoody:
Instead of 390d15bbf7628ac1cba90d5fe8cee41ecfad96cd, please just add an UploadAgreement to demo-project.json.

tompollard:
> Instead of https://github.com/MIT-LCP/physionet-build/commit/390d15bbf7628ac1cba90d5fe8cee41ecfad96cd, please just add an UploadAgreement to demo-project.json.

Done!

physionet-django/project/forms.py
```diff
@@ -1196,6 +1200,85 @@ def __init__(self, responder, *args, **kwargs):
...
+        # Check if we already have an instance (for updates)
+        if self.instance and self.instance.pk:
+            # Update existing agreement
+            agreement = self.instance
+            agreement.no_human_subjects = self.cleaned_data['no_human_subjects']
+            agreement.derived_data = self.cleaned_data['derived_data']
+            agreement.human_subjects_deidentified = self.cleaned_data['human_subjects_deidentified']
+            agreement.accepted = True
+            if self.user:
+                agreement.accepted_by = self.user
```
bemoody: I don't understand what this is for.  How is this different than just calling `super().save(commit=False)` ?

physionet-django/project/views.py
```diff
@@ -890,6 +891,61 @@ def project_discovery(request, project_slug, **kwargs):
...
+    if is_submitting and project.author_editable():
+        editable = True
+    else:
+        editable = False
+
+    existing_agreement = UploadAgreement.get_active_agreement(project)
+
+    if request.method == 'POST':
+        # Pass the existing instance if it exists
+        if existing_agreement:
```
bemoody: This "if" statement is useless.  The two branches are the same.

physionet-django/project/views.py
```diff
@@ -890,6 +891,61 @@ def project_discovery(request, project_slug, **kwargs):
...
+
+        if upload_agreement_form.is_valid():
+            upload_agreement_form.save()
+            messages.success(request, 'Upload agreement has been accepted.')
+            return redirect('project_files', project_slug=project.slug)
+        else:
+            messages.error(request, 'Invalid submission. See errors below.')
+    else:
+        # Get existing agreement or create new form
+        if existing_agreement:
```
bemoody: This "if" statement is useless.  The two branches are the same.

bemoody:
Your help string says "Each project can have one active upload agreement."  But that doesn't appear to be the case.

If there is one UploadAgreement per project, there should be a OneToOneField (or a ForeignKey having unique=True.)

If there is one UploadAgreement per project and accepted_by, there should be a unique_together.

bemoody (dismissed):
We do not want to permit multiple UploadAgreements for the same author and the same project.  A good way to do that would be to remove `project` and `accepted_by`, and replace them with a single foreign key linking to `project.Author` and having `unique=True`.

`get_active_agreement` is full of debugging code that should be unnecessary.  I would get rid of this method entirely.  Possibly add a method to Author that returns the author's UploadAgreement.  You do not need to do `project.authors.get(is_submitting=True)` to obtain the submitting author, just call `project.submitting_author()`.

Currently it looks like this will forbid editors from copyediting a project if the submitting author has not accepted the agreement.  We don't want that.

Non-submitting authors should not see the message "You must accept the upload agreement" on the files page.

physionet-django/project/forms.py
tompollard: you're right, updated to use your suggested syntax.

physionet-django/project/migrations/0100_uploadagreement_and_more.py
```diff
@@ -0,0 +1,84 @@
...
+                    to='project.author',
+                )),
+            ],
+            options={
+                'verbose_name': 'Upload Agreement',
+                'verbose_name_plural': 'Upload Agreements',
+                'db_table': 'project_uploadagreement',
+            },
+        ),
+        migrations.AddConstraint(
```
lukepayyapilli, line 72: This creates the table and the constraint, but there's no data migration for projects that already exist. On deploy, every current `ActiveProject` has a submitting author with no agreement row, so:

- `check_integrity()` adds "You must accept the upload agreement before submitting" for every project currently in an author-editable state. That's the behaviour @bemoody asked for, but worth warning authors before it lands.
- `process_files_post` refuses all uploads for those projects, including from editors (see the comment on `views.py`).

**The wrinkle:** the `CheckConstraint` means a backfill can't stay silent. Every row must assert one of the three data-type claims, so backfilling forces you to fabricate a statement about data nobody reviewed. Setting `no_human_subjects=True` across the board would be untrue for a good number of projects, and this is meant to be a record of what an author asserted.

Two ways out, depending on what you want the record to mean.

**Option A, grandfather without asserting anything.** Add a fourth boolean (`legacy`, default `False`, not shown in the form) and include it in the constraint, so backfilled rows are honestly marked as predating the agreement:

```python
def backfill_upload_agreements(apps, schema_editor):
    ContentType = apps.get_model('contenttypes', 'ContentType')
    Author = apps.get_model('project', 'Author')
    UploadAgreement = apps.get_model('project', 'UploadAgreement')

    content_type = ContentType.objects.get(app_label='project', model='activeproject')

    UploadAgreement.objects.bulk_create([
        UploadAgreement(author=author, accepted=True, legacy=True)
        for author in Author.objects.filter(content_type=content_type, is_submitting=True)
    ])
```

(`Author` is a generic FK onto the project, hence the `content_type` filter, and `is_submitting` narrows it to one row per project, which keeps the `unique=True` relationship intact.) `accepted_datetime` stays null, which reads correctly: accepted by policy, not by a person clicking a button.

**Option B, don't backfill.** Treat projects created before the deploy as exempt, so the rule only applies going forward:

```python
def upload_agreement_required(self):
    return self.creation_datetime >= settings.UPLOAD_AGREEMENT_START_DATE
```

Cheaper, and avoids writing rows nobody asserted, at the cost of a date in settings.

Either way, existing projects in the copyedit stage need to keep working for editors.

physionet-django/project/views.py
```diff
@@ -1102,6 +1155,12 @@ def process_files_post(request, project):
         raise ServiceUnavailable()
 
     if 'upload_files' in request.POST:
+        # Check if upload agreement has been accepted
+        agreement = getattr(project.submitting_author(), 'upload_agreement', None)
+        if not agreement or not agreement.accepted:
+            messages.error(request, 'You must accept the upload agreement before uploading files.')
```
lukepayyapilli, line 1148: This check has no editor exemption, but the two places that decide whether to *show* the upload control do:

```python
has_accepted_agreement = (
    is_editor
    or (agreement is not None and agreement.accepted)
)
```

This looks like the remaining half of @bemoody's September point ("this will forbid editors from copyediting a project if the submitting author has not accepted the agreement. We don't want that"). `f9c0467` added the editor branch to `files_editable`, so the button renders enabled for an editor in the copyedit stage, but this check still refuses the upload. With no backfill migration, that's the state for every pre-existing project.

The rule is currently written out in four places (`check_integrity`, here, `project_files`, `project_files_panel`) and two of them already disagree. Suggest putting it on the model once:

```python
# ActiveProject
def upload_agreement_accepted(self):
    agreement = getattr(self.submitting_author(), 'upload_agreement', None)
    return agreement is not None and agreement.accepted

def can_upload_files(self, user):
    if user == self.editor and self.copyeditable():
        return True
    return self.author_editable() and self.upload_agreement_accepted()
```

Then this block becomes:

```python
if 'upload_files' in request.POST:
    if not project.can_upload_files(request.user):
        messages.error(request, 'You must accept the upload agreement before uploading files.')
        return ''
```

and both views pass `'can_upload_files': project.can_upload_files(request.user)` into the template context, so the button and the server agree by construction. `check_integrity` calls `upload_agreement_accepted()` and drops its `try/except`.

physionet-django/project/modelcomponents/uploadagreement.py
```diff
@@ -0,0 +1,76 @@
...
+    # Metadata
+    created_datetime = models.DateTimeField(auto_now_add=True)
+    updated_datetime = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        db_table = 'project_uploadagreement'
+        verbose_name = 'Upload Agreement'
+        verbose_name_plural = 'Upload Agreements'
+        constraints = [
+            models.CheckConstraint(
```
lukepayyapilli, line 69: Good instinct enforcing this at the DB level, but it's also implemented in `UploadAgreementForm.clean()`, and in Django 4.2 the two stack rather than one taking over.

`BaseModelForm._post_clean` calls `instance.full_clean(exclude, validate_unique=False)`, and `validate_constraints` defaults to `True`, so `CheckConstraint.validate()` runs on every model form save. It runs *after* `_clean_form()`, and both errors accumulate. Submit the page with nothing checked and `{{ upload_agreement_form.non_field_errors }}` renders two messages:

> Please select at least one option that applies to your project.
> Constraint "at_least_one_data_type_selected" is violated.

(I'd expected an `IntegrityError` here before checking the 4.2 source, so this is milder than it first looks, but the second line is still user-visible and not something we want to show.)

Cleanest fix is to let the constraint own both the rule and its wording, which Django 4.1+ supports directly:

```python
models.CheckConstraint(
    check=(
        models.Q(no_human_subjects=True)
        | models.Q(derived_data=True)
        | models.Q(human_subjects_deidentified=True)
    ),
    name='at_least_one_data_type_selected',
    violation_error_message='Please select at least one option that applies to your project.',
)
```

Then delete `UploadAgreementForm.clean()` entirely. One rule, one message, and `UploadAgreementAdmin` picks up the same friendly error for free instead of the raw constraint name (it currently has all three booleans editable with no validation of its own). Needs a matching `AlterConstraint` in the migration.

physionet-django/project/forms.py
```diff
@@ -1357,3 +1365,66 @@ def __init__(self, *args, **kwargs):
...
+                'Please select at least one option that applies to your project.'
+            )
+
+        return cleaned_data
+
+    def save(self):
+        agreement = super().save(commit=False)
+        agreement.author = self.author
+        agreement.accepted = True
+        # Reset the timestamp so it reflects the latest acceptance
```
lukepayyapilli: This line exists to defeat the `save()` override on the model:

```python
# UploadAgreement.save()
if self.accepted and not self.accepted_datetime:
    self.accepted_datetime = timezone.now()
```

So the timestamp is maintained cooperatively by a form that nulls it and a model that repopulates it, and the coupling isn't visible from either file alone. Suggest setting it here and deleting the `save()` override:

```python
def save(self):
    agreement = super().save(commit=False)
    agreement.author = self.author
    agreement.accepted = True
    agreement.accepted_datetime = timezone.now()
    agreement.save()
    return agreement
```

Separately, `accepted` and `accepted_datetime` encode the same fact twice and can disagree. If the timestamp is only ever set on acceptance, the boolean can go:

```python
@property
def accepted(self):
    return self.accepted_datetime is not None
```

That does mean the console template's `upload_agreement.accepted` keeps working unchanged, and the field drops out of the admin `list_filter`. Worth it only if you don't need to query on `accepted` in SQL, so take or leave.

physionet-django/project/modelcomponents/uploadagreement.py
```diff
@@ -0,0 +1,76 @@
...
+            )
+        ]
+
+    def __str__(self):
+        return (
+            f"Upload Agreement for {self.author} - "
+            f"{'Accepted' if self.accepted else 'Pending'}"
+        )
+
+    def save(self, *args, **kwargs):
```
lukepayyapilli: Two things this `save()` override gives up, both fixable while keeping the one-agreement-per-author shape @bemoody asked for.

**The original acceptance time is discarded on update.** `UploadAgreementForm.save()` sets `accepted_datetime = None` specifically so this override restamps it, so re-accepting overwrites when the author first agreed. If the button is going to say "Update Agreement", it's worth keeping both:

```python
first_accepted_datetime = models.DateTimeField(null=True, blank=True)
accepted_datetime = models.DateTimeField(null=True, blank=True)
```

setting `first_accepted_datetime` only when it's null and `accepted_datetime` on every acceptance. The console card can then show both, which is more useful to an editor than the latest timestamp alone.

**There's no record of which terms were agreed to.** The binding wording lives in `project_upload_agreement.html`, so editing that template retroactively changes what everyone who already accepted appears to have agreed to. A `terms_version = models.CharField(max_length=16)` set from a constant at save time fixes that without adding rows, and gives you a way to deliberately re-prompt authors when the wording changes (treat an agreement matching an older version as not accepted).

Minor, but `on_delete=CASCADE` on the author also means the record disappears with the author row. `PROTECT` or a soft delete may be more appropriate for something that reads as a legal agreement, depending on how author removal is handled elsewhere.

physionet-django/project/views.py
```diff
@@ -1043,6 +1088,13 @@ def project_files_panel(request, project_slug, **kwargs):
      move_items_form, delete_items_form) = get_file_forms(
          project=project, subdir=subdir, display_dirs=display_dirs)
 
+    submitting_author = project.submitting_author()
+    agreement = getattr(submitting_author, 'upload_agreement', None)
+    has_accepted_agreement = (
```
lukepayyapilli: This three-line block is duplicated verbatim in `project_files` further down (around line 1249). Both also call `project.submitting_author()`, which is `self.authors.get(is_submitting=True)`, so each render costs two extra queries.

The `can_upload_files(user)` method suggested on `process_files_post` replaces both copies with one call, and gives a single place to attach `select_related('upload_agreement')` if query count becomes a concern. `console.views.submission_info_card_params` does the same traversal and could use `upload_agreement_accepted()` too.

lukepayyapilli (commented):
_Updated after reading the earlier thread with @bemoody properly. I'd raised two things that were already settled there, so I've withdrawn them and sharpened the one that turns out to still be open._

Read through the full diff plus the surrounding code on `main` (`project_auth`, `submitting_author()`, `TransferAuthorForm`, the context processors). The feature makes sense and the structure follows existing conventions well: splitting into `modelcomponents/uploadagreement.py` matches how the rest of the app is organised, enforcing "at least one option" at the DB level is the right instinct, surfacing acceptance status on the editorial submission card is a genuinely useful addition, and `preparation_checklist.html` got updated (easy to forget).

Five things worth a look, left inline with suggested fixes.

**1. The editor exemption is only half applied.** @bemoody asked for this in the September review ("this will forbid editors from copyediting a project if the submitting author has not accepted the agreement. We don't want that"). `f9c0467` added the editor branch to `files_editable` in `project_files`, and `project_files_panel` computes `has_accepted_agreement = is_editor or ...`, but `process_files_post` still checks only the submitting author's agreement with no editor exemption. So an editor copyediting a project whose submitting author hasn't accepted gets an enabled Upload Files button and an error toast on use. I think that requirement is still half open.

**2. No backfill migration.** Migration `0100` creates the table but doesn't populate it, so every `ActiveProject` that exists at deploy time has a submitting author with no agreement row. Combined with #1 that's the default state for the entire existing estate. There's a wrinkle in the fix worth deciding on, noted inline.

The rest are about stating the rule once, and about what the agreement record is meant to preserve.

Smaller items I didn't leave inline, with what I'd suggest:

- **No test covers the new rule.** Both test files gained `UploadAgreement.objects.create(...)` so existing tests keep passing, but nothing asserts that upload is blocked without one. Worth adding the inverse: post to `project_files` with `upload_files` for a project whose submitting author has no agreement, and assert the file wasn't created. Same shape for `clean()` rejecting zero checkboxes, for a non-submitting author being refused, and for the editor case in #1.
- **Three idioms for one lookup.** `try/except UploadAgreement.DoesNotExist` in `activeproject.py`, `getattr(..., 'upload_agreement', None)` in the views, `hasattr(...)` in `forms.py`. Once the model method suggested inline exists, all three call sites collapse into `project.upload_agreement_accepted()` and the question goes away.
- **`has_accepted_agreement` is `True` for editors who haven't accepted anything.** Renaming it `can_upload_files` in `project_files`, `project_files_panel`, `project_files.html` and `project_files_form.html` makes the templates read correctly.
- **The verbose boolean @bemoody flagged twice in September is back** in `project_upload_agreement`: `if is_submitting and project.author_editable(): editable = True else: editable = False` is just `editable = is_submitting and project.author_editable()`.
- **Step numbers are now in three places.** `project_files.html` hardcodes `7.`, `project_proofread.html` hardcodes `8.`, and `project.html` has the `{% with %}` list plus a special case for the `upload_agreement` title. Replacing the space-separated string with a list of tuples in the view context, for example `[('authors', 'Authors'), ..., ('upload_agreement', 'Upload Agreement'), ...]`, lets `forloop.counter` own the numbering, drops the `{% if %}`, and lets the two `<h2>` headings use the same source instead of literals.
- **The agreement wording exists in four places:** the template's `alert-info` block, `Meta.labels`, model `help_text`, and again frozen in migration `0100`. They have already diverged slightly, and the template's blanket statement about identifiable information reads oddly next to the "no human subjects" option. Defining the three strings once as module-level constants in `uploadagreement.py` and importing them into the form would leave only the migration copy, which is inert.

Happy to talk any of these through, and glad to be wrong on the backfill if there's a data task planned elsewhere.

physionet-django/project/views.py
tompollard: Added a `upload_agreement_accepted()` method to `ActiveProject` (2f5aa286). This replaces the duplicated blocks in `project_files`, `project_files_panel`, and `process_files_post`. `console.views.submission_info_card_params` still does its own lookup since it needs the full agreement object for the template, not just a boolean.

physionet-django/project/modelcomponents/uploadagreement.py
tompollard: > The original acceptance time is discarded on update. UploadAgreementForm.save() sets accepted_datetime = None specifically so this override restamps it, so re-accepting overwrites when the author first agreed.

I removed the `save()` override. `accepted_datetime is now set directly in the form (0eea3210).  I kept just a single accepted_datetime for now, reflecting the latest acceptance. I think this is sufficient for our purposes. 

> There's no record of which terms were agreed to. The binding wording lives in project_upload_agreement.html, so editing that template retroactively changes what everyone who already accepted appears to have agreed to. 

Good point. I thought about different approaches for tracking terms_version (e.g. hashing), but I prefer to keep it as is for now. I'll raise an issue to suggest idea of tracking specific text in future.

> Minor, but on_delete=CASCADE on the author also means the record disappears with the author row. PROTECT or a soft delete may be more appropriate for something that reads as a legal agreement, depending on how author removal is handled elsewhere.

Published author objects cannot be deleted, so the agreement record is preserved for published projects. For unpublished projects, deleting the author (and its agreement) when removed from a project is reasonable.

physionet-django/project/forms.py
tompollard: Addressed in in 0eea3210. `accepted_datetime` is now set directly in the form and the model save() override is removed.

physionet-django/project/modelcomponents/uploadagreement.py
tompollard: Addressed in 1a21c5fd — added `violation_error_message` to the CheckConstraint and removed `UploadAgreementForm.cleanz`

physionet-django/project/views.py
tompollard: I added an editor exemption in aaf14a72. Duplicated logic extracted into `upload_agreement_accepted()` method on ActiveProject in 2f5aa286.

physionet-django/project/migrations/0100_uploadagreement_and_more.py
tompollard: I went with option B in 074e853f — added UPLOAD_AGREEMENT_START_DATE setting. Projects created before this date are exempt. I clarify in the docstring of the `upload_agreement_accepted()` that it will return true for pre-implementation uploads.

tompollard:
> 1. The editor exemption is only half applied.

Editor exemption completed in aaf14a72. `process_files_post` now skips the agreement check for editors. 

> 2. No backfill migration.

Existing projects handled via `UPLOAD_AGREEMENT_START_DATE` setting (074e853f). Projects created before the cutoff are exempt, no backfill needed.

> No test covers the new rule.

Added `TestUploadAgreement` with tests covering upload blocked/allowed, non-submitting author warning, and empty form rejection (aa479af9).

> Three idioms for one lookup.

Collapsed into `upload_agreement_accepted()` on ActiveProject (2f5aa286).

> The verbose boolean

Simplified to `editable = is_submitting and project.author_editable()` (2c368631).

> Step numbers are now in three places.

Defined once via project_settings_steps template tag, used in both the sidebar and page headings (11c7176b).

> The agreement wording exists in four places

Checkbox labels defined as constants in `uploadagreement.py` and imported in the form (8759743d).

lukepayyapilli:
@tompollard: re-reviewed the latest state, this time running it locally rather than reading. Summary up front: the constraint/message rework and the `upload_agreement_accepted()` extraction both check out, but the editor exemption has landed in the wrong places and I think #1 below is a shipping regression.

Re-reviewed at `25f7aa7c`. This time I ran it: built the image from `Dockerfile`, brought up `db`/`gcs`/`test` from `docker-compose.yml`, `resetdb` + `loaddemo`, and drove each point through the real views with the demo fixtures. Output below is copied from those runs.

Good news first. `makemigrations --dry-run --check` reports **No changes detected**, the PR's own `TestUploadAgreement` is **4/4 green**, and the `violation_error_message` rework does exactly what it was meant to. Posting the form with nothing ticked gives:

```
[5] friendly message shown: True
[5] raw constraint name leaked: False
[5] row created: False
```

`check_integrity` is wired up correctly too (`['You must accept the upload agreement before submitting.']`, `is_submittable: False`). The step-number deduplication is a real improvement.

The editor exemption is still the problem, and it's worse than half-applied: it was added to the three places an editor can't reach and left out of the one place they can.

## Blocking

### 1. Editors cannot upload files during copyediting

`project_files_form.html` now gates the button on `has_accepted_agreement`. That template is reached from `console/copyedit_submission.html:48` → `edit_files_panel.html:18` → `project_files_form.html`, and `console.views.copyedit_submission` never puts the variable in its context. Undefined → `''` → falsy.

Driving a project to `NEEDS_COPYEDIT` and loading the console page as the handling editor, on a project whose agreement **is** accepted:

```
[1] has_accepted_agreement in console context: False
[1] files_editable in console context: True
[1] console button: <button id="upload-files-button" type="button" class="btn btn-success"
              disabled title="You must accept the upload agreement first">
[1] console button disabled: True
```

The same project's ajax panel, which goes through `project_files_panel` and *does* set the variable:

```
[1] panel button: <button id="upload-files-button" type="button" class="btn btn-success"
              data-toggle="modal" data-target="#upload-files-modal" onclick="prepareUpload()">
[1] panel button disabled: False
```

So the button is dead on page load and comes back to life after clicking into a folder. Note it also loses `data-toggle`/`data-target`/`onclick` in the disabled branch, so the modal can't be opened at all. This isn't a cosmetic `disabled` attribute the user can work around.

The server side is fine. With the agreement deleted outright, the editor's upload still succeeds:

```
[1b] editor upload POST status: 200
[1b] editor upload landed on disk: True
[1b] button after POST disabled: True
```

`process_files_post`'s exemption works; the only thing standing between an editor and copyediting files is a button this PR disabled. Nothing in the suite covers it, which is why it's green.

### 2. The editor branch added to `project_files` is unreachable

```python
is_editor = request.user == project.editor
if is_submitting and project.author_editable():
    files_editable = True
elif is_editor and project.copyeditable():
    files_editable = True
```

`project_files` is `@project_auth(auth_mode=0, post_auth_mode=2)`. `auth_mode=0` is `allow = is_author`, and `post_auth_mode=2` is `allow = is_submitting and project.author_editable()`, followed by `if not is_submitting: raise Http404()` in the view body.

A normal editor can't even open the page:

```
[2] editor (non-author) GET project_files: 403
```

Forcing the only case where the branch can fire (an editor who is also an author of the project, in copyedit):

```
[2b] GET project_files as editor+author: 200
[2b] files_editable: True
[2b] upload button rendered enabled: True
[2b] POST project_files as editor+author: 403
```

A full enabled toolbar (upload, create folder, rename, move, delete), every button of which 403s.

That said, I had to construct that case by hand, and it's worth being straight about how reachable it is: `console.forms.ReassignEditorForm` excludes author user ids from its queryset and `console/forms.py:114` rejects an editor who is an author outright, so an editor being an author of their own project is structurally prevented. The practical harm is therefore close to nil. The honest statement is that this branch is unreachable dead code that reads as though it implements the editor exemption and cannot. That matters mostly because it's presumably why #1 was believed to be fixed. I'd revert it and drop `is_editor` from `project_files`; editors copyedit through the console, which is where #1 actually needs fixing.

### 3. `UPLOAD_AGREEMENT_START_DATE` is a config landmine

```python
_upload_agreement_start = config('UPLOAD_AGREEMENT_START_DATE', default=None)
UPLOAD_AGREEMENT_START_DATE = (
    datetime.fromisoformat(_upload_agreement_start)
    if _upload_agreement_start else None
)
```

**It crashes on the obvious input.** `USE_TZ = True`, so `creation_datetime` is aware. The docstring's `'2026-10-01T00:00:00Z'` parses to an aware datetime on 3.11+, but the form anyone will actually type does not:

```
[3] fromisoformat("2026-10-01") -> datetime.datetime(2026, 10, 1, 0, 0)
[3] tzinfo: None
[3] upload_agreement_accepted() raised TypeError: can't compare offset-naive and offset-aware datetimes
[3] GET project_files raised TypeError: can't compare offset-naive and offset-aware datetimes
```

`upload_agreement_accepted()` is called from `check_integrity`, `project_files`, `project_files_panel` and `process_files_post`, so a plausible env-file typo takes out the files page and submission for every project, with nothing failing at startup. Worth being explicit:

```python
_start = config('UPLOAD_AGREEMENT_START_DATE', default=None)
if _start:
    UPLOAD_AGREEMENT_START_DATE = parse_datetime(_start)
    if UPLOAD_AGREEMENT_START_DATE is None:
        raise ImproperlyConfigured('UPLOAD_AGREEMENT_START_DATE must be an ISO 8601 datetime')
    if timezone.is_naive(UPLOAD_AGREEMENT_START_DATE):
        raise ImproperlyConfigured('UPLOAD_AGREEMENT_START_DATE must include a timezone offset')
else:
    UPLOAD_AGREEMENT_START_DATE = None
```

`ImproperlyConfigured` is already imported in `base.py`.

**The default is the unsafe one.** `[3b] settings.UPLOAD_AGREEMENT_START_DATE default: None`, which means "require the agreement of every project ever created". Simulating a project that was already in flight when the rule lands (submitted, in copyedit, no agreement row):

```
[dead-end] status: 40 author_editable: None
[dead-end] files page status: 200
[dead-end] red banner shown: True
[dead-end] agreement page status: 200
[dead-end] "cannot be edited" shown: True
[dead-end] POST agreement status: 403
```

The submitting author gets a red *Upload Agreement Required!* banner, follows its button to a page that tells them the project cannot be edited, and any attempt to accept is refused with a 403. A dead end on a project that was fine yesterday, for the entire existing estate. Going forward the state is unreachable, since `check_integrity` blocks submission without an agreement and `submit()` raises, so this is purely a deploy-day problem, which is exactly what the cutoff is for. Replacing the backfill with a cutoff is a reasonable call; the safe default is the other way round.

**It isn't in `.env.example`.** Given the failure mode above, it should be there with the intended value and a comment.

## Should fix before merge

### 4. The `transfer()` hook flips the wrong author's record

```python
old_author = self.project.submitting_author()
...
if hasattr(old_author, 'upload_agreement'):
    old_author.upload_agreement.accepted = False
    old_author.upload_agreement.save()
```

Transferring MIMIC-III from `rgmark` to `aewj`:

```
[4] old author agreement.accepted: False
[4] old author accepted_datetime: 2018-09-04 17:26:24.820000+00:00
[4] new submitting author: aewj
[4] new author has agreement: False
[4] upload_agreement_accepted() after transfer: False
[4] with old flag restored to True, upload_agreement_accepted(): False
```

Two things there. The last line is the point: `upload_agreement_accepted()` reads the *new* submitting author, who has no row, so the answer is `False` whether or not the hook ran. For the case @bemoody asked about (reassign to someone new, they should fill the form in again), the `OneToOneField` on `Author` already gives you that, and this block changes nothing.

What it does do is rewrite a record of something a person asserted. `accepted: False` sitting next to `accepted_datetime: 2018-09-04 17:26:24` is the redundancy I flagged last time diverging in practice, one commit after we agreed to keep both fields. I'd delete the block (and `old_author`) and let the OneToOne do the work. The only case where the flip has any effect is transfer-away-and-back, where "they already agreed to these terms" seems like the right answer anyway. If you do want a fresh acceptance on every reassignment, that's a project-level pointer to the current agreement, not a mutation of a historical one.

### 5. Four call sites, three spellings, one misleading name

```python
# project_files_panel
has_accepted_agreement = is_editor or project.upload_agreement_accepted()
# project_files            (unreachable, see #2)
has_accepted_agreement = is_editor or project.upload_agreement_accepted()
# process_files_post
if request.user != project.editor and not project.upload_agreement_accepted():
# console.copyedit_submission
(nothing, see #1)
```

`upload_agreement_accepted()` fixed the inner duplication, but the editor exemption around it is still copy-pasted, still spelled two different ways, and still missing from the fourth site. The three that exist also disagree on stage: the views require `copyeditable()`, `process_files_post` exempts the editor at any stage.

One method makes #1 and #2 both unrepresentable:

```python
# ActiveProject
def can_upload_files(self, user):
    if user == self.editor:
        return self.copyeditable()
    return self.upload_agreement_accepted()
```

`process_files_post` becomes `if not project.can_upload_files(request.user):`, all three view contexts pass `'can_upload_files': project.can_upload_files(request.user)`, and `project_files_form.html` tests that. The name stops lying, too: `has_accepted_agreement` is currently `True` for an editor who has accepted nothing, which is what makes those templates hard to read.

### 6. The two riskiest new behaviours have no test

`TestUploadAgreement` passes and covers the happy and blocked paths, which is a real improvement. It doesn't cover:

- **the editor exemption**: wrong twice now, and currently broken per #1. A GET of `copyedit_submission` as the handling editor asserting the upload button isn't disabled would have caught it.
- **`UPLOAD_AGREEMENT_START_DATE`**: the entire production migration story. `@override_settings` either side of a project's `creation_datetime` is cheap, and would also have caught the naive-datetime `TypeError` in #3.
- **`check_integrity`** picking up the new error.

Two existing tests also assert absence rather than cause:

- `test_upload_blocked_without_agreement` asserts 200 and that the file doesn't exist, which passes if the upload failed for any reason at all. Assert the message: `assertContains(response, 'You must accept the upload agreement')`.
- `test_agreement_form_requires_checkbox` asserts no row was created, which would still pass if the constraint were only enforced in the database and the user got a 500. I confirmed the friendly message does render, so assert it: `assertContains(response, 'Please select at least one option')`. Worth pinning because `CheckConstraint.validate()` swallows `FieldError`. If anyone later adds a field to that constraint that isn't on the form, form-level validation silently stops and users get an `IntegrityError`, with no test failure.

`test_non_submitting_author_no_warning` would pass with the feature deleted; pairing it with the positive case makes it mean something.

## Nits

- `activeproject.py:30` imports `UploadAgreement` and no longer uses it.
- `DERIVED_DATA_LABEL.rstrip('.') + f' published on {settings.SITE_NAME} or elsewhere.'`: the constant was introduced to stop the wording drift, and this reintroduces it by string surgery: `rstrip('.')` also eats a legitimate ellipsis, and the model `help_text` and the form label now differ again. Two constants, or one with a `{site_name}` placeholder.
- `submission_info_card.html` hardcodes a fourth set of wordings ("No human subjects data", "Derived from de-identified datasets", "Human subjects data (de-identified)"). Deliberate short forms are fine, but these are the strings an editor reads when deciding whether to trust the assertion.
- `project_step_header` returns the raw key on a miss, so `{% project_step_header 'fils' %}` renders `fils`. A dict built from `PROJECT_SETTINGS_STEPS` and a `KeyError` is the more useful failure.
- `Meta.db_table = 'project_uploadagreement'` is exactly Django's default. Drop it.
- The fixture gives all seven agreements `no_human_subjects: true`, including MIMIC-III, so the console card shows "No human subjects data" against a clinical database. `human_subjects_deidentified` would read correctly for those.
- The `alert-info` block asserts "These files do not contain information that could identify individual human subjects" unconditionally, directly above a checkbox saying the project has no human subjects data at all. The blanket claim and the options don't line up.
- On `accepted` vs `accepted_datetime`: happy to leave it, but #4 is the first place they've actually diverged. If both stay, a check constraint (`accepted=False OR accepted_datetime IS NOT NULL`) would stop that class of drift.

## What I didn't check

- **The GCP storage path.** I ran against local storage only. When `STORAGE_TYPE` is GCP the upload modal's body is a dropzone rather than the file input, but it's still inside the same modal behind the same `prepareUpload()` onclick, so I'd expect #1 to bite identically, but I haven't confirmed it.
- **How deployments are actually configured.** I suggested `.env.example` for `UPLOAD_AGREEMENT_START_DATE` because that's where the other settings are documented; I didn't read `deploy/`, so if config lands somewhere else in practice, put it there instead. The point is only that a setting with this failure mode shouldn't be discoverable solely by reading `base.py`.
- **Anything outside `project` and `console`.** Those two suites are 174 tests and all green on this branch.

tompollard:
Thanks for the re-review — all three blocking issues and the should-fix items are addressed. Summary of changes below:

#1 and #2 — Editor exemption in the wrong places

  You're right, the editor branch in project_files was unreachable dead code. Fixed in 326720f4

#3 — UPLOAD_AGREEMENT_START_DATE

  Fixed in 370707b1. Naive inputs now get timezone.utc applied rather than crashing.

#4 — Transfer hook

  Deleted in c4614d70. The OneToOneField on Author already handles the "new submitting author must accept" case.

  #5 — Four call sites, three spellings

  Added can_upload_files(self, user) on ActiveProject in f1c3a932. All four sites now go through it, has_accepted_agreement is gone from views and templates, and process_files_post is now consistent with the views on requiring copyeditable().

  #6 — Tests

  Addressed in 3132cc54. The suite now covers:
  - Editor exemption: GET of copyedit_submission as the handling editor, asserting the upload button isn't disabled.
  - UPLOAD_AGREEMENT_START_DATE: @override_settings with a future and past date, including the full upload POST path.
  - check_integrity picking up the missing agreement error.
  - test_upload_blocked_without_agreement now asserts the specific message.
  - test_agreement_form_requires_checkbox now asserts "Please select at least one option".
  - Added test_warning_shown_for_submitting_author as the positive pair for the non-submitting author test.

  Nits

  - Unused import: removed in c4614d70.
  - rstrip('.') drift: added DERIVED_DATA_FORM_LABEL with a {site_name} placeholder in 534db03d. The base constant stays clean.
  - submission_info_card.html: added selected_data_types() method on UploadAgreement so the card uses the canonical labels (same commit).
  - project_step_header: now raises KeyError on miss (fadf578e).
  - db_table: leaving as-is — removing it generates a no-op migration.
  - Fixture: clinical databases now use human_subjects_deidentified (77d510c5).
  - alert-info wording: changed to "do not contain any personally identifiable information" (01d496d0).
  - accepted vs accepted_datetime constraint: deferring for now, agreed it's worth adding if both fields stay.

lukepayyapilli (approved):
@tompollard: re-reviewed at `962ca281`, run locally again. All three blockers are fixed and I checked each end to end, not just the diff: the editor's upload POST through `copyedit_submission` actually lands the file, naive `UPLOAD_AGREEMENT_START_DATE` values now parse aware, and dropping the transfer hook is correct (verified on MIMIC-III: transfer flips `upload_agreement_accepted()` to False and `check_integrity` picks it up). Suite is 171 green. Ship it.

Two things worth doing, neither blocking:

1. **`test_editor_can_upload_during_copyedit` will go false-green.** It does catch the real regression (removing the `can_upload_files` context line fails it). But the assertion is a negative match on an exact HTML string, so reordering `disabled title="..."` to `title="..." disabled` in the template, behaviourally identical, makes it pass with the button still disabled. I ran that. Assert the positive marker instead: `assertContains(response, 'data-target="#upload-files-modal"')`.

2. **`UPLOAD_AGREEMENT_START_DATE` isn't in `.env.example`.** It's now the whole answer to "no backfill needed", and its default is the strict direction: unset means every in-flight project is blocked from submitting on deploy day. Right now that's safe only if an operator sets a variable no file in the repo mentions.

Nit while you're in there: `can_upload_files(user)` ignores `user` unless it's the editor, so it returns True for a stranger, and only the editor branch checks project state. Nothing exploitable, every call site is behind `project_auth`. It's the name that overshoots, and I'm the one who asked you to consolidate those four call sites, so this one's on me to flag rather than you to have foreseen.

tompollard:
@lukepayyapilli I addressed the following two points in the new commits:

> test_editor_can_upload_during_copyedit will go false-green. It does catch the real regression (removing the can_upload_files context line fails it). But the assertion is a negative match on an exact HTML string, so reordering disabled title="..." to title="..." disabled in the template, behaviourally identical, makes it pass with the button still disabled. I ran that. Assert the positive marker instead: assertContains(response, 'data-target="#upload-files-modal"').

> UPLOAD_AGREEMENT_START_DATE isn't in .env.example. It's now the whole answer to "no backfill needed", and its default is the strict direction: unset means every in-flight project is blocked from submitting on deploy day. Right now that's safe only if an operator sets a variable no file in the repo mentions.

lukepayyapilli (approved):

PR #2483, opened 2025-09-04
Updating README.md - python version, compile styling

emmyxth:
Updating README to include python 3.11 as the default virtual environment version, adding missing extra step to compile the styling files before running the Django server.


tompollard (approved):
Thanks @emmyxth, this looks good to me. I've added a few quick thoughts below:

- Personally I'm not keen on Conda so don't really like it being suggested as an example, but no harm in leaving in.
- You are applying style fixes to the entire file (perhaps automatically applied by your code editor on save). It's not a problem here, but it's best to avoid this because it increases risk of conflicts. If you'd like to make broad style fixes, these can be put into a separate PR.

PR #2508, opened 2025-10-09
V0: Annotation API Fine-Grained Tokens Permissions

emmyxth:
As per our conversations @tompollard @bemoody and referenced in this [discussion](https://github.com/MIT-LCP/physionet-build/issues/1047#issuecomment-3285413494), we want to add fine-grained token authorization to the Annotation API. 

Key features & changes: 
1. Annotator group created, which admin can manually add select people to the group
2. Anyone in the Annotator group has "write" access to `AnnotationCollection, AnnotationType, Annotation` classes via fine-grained token generation.

Changes to the original `/settings/tokens` UI:
1. Removed the `name` parameter originally required to create a token: this is because the `AccessToken` class used originally from the [Django OAuth Toolkit](https://django-oauth-toolkit.readthedocs.io/en/latest/models.html#oauth2_provider.models.AccessToken) does not take in a "name" and instead identifies each token via ID. The name was not getting reflected anywhere. UI now reflects the ID of the generated token.
Screenshot 2025-10-09 at 10 32 09 AM
2. Created a modal popup wherein users can select endpoints corresponding to different annotation classes, each with options for read/write access.
Screenshot 2025-10-09 at 10 32 28 AM


physionet-django/user/templates/user/edit_tokens.html
```diff
@@ -40,6 +47,114 @@ <h3 class="mt-4">Your Tokens</h3>
...
+        });
+        
+        const checkboxes = scopeForm.querySelectorAll('input[type="checkbox"]');
+        checkboxes.forEach(function(checkbox) {
+            checkbox.checked = false;
+        });
+    });
+});
+</script>
+{% endblock %}
```
tompollard, line 160: Missing newline at end of file.

physionet-django/user/test_tokens.py
```diff
@@ -96,6 +108,7 @@ def test_token_limit_enforced(self):
                 scope="data:download",
             )
 
-        response = self.client.post(reverse("edit_tokens"),
-                                    data={"name": "Should Fail"}, follow=True)
+        response = self.client.post(
```
tompollard, line 111: According to the PR comment: "Removed the name parameter originally required to create a token", so name shouldn't be passed here? 

The `edit_tokens` view is looking for `request.POST.getlist("annotation_endpoints")`. Will this be available in the test?

physionet-django/user/views.py
```diff
@@ -452,28 +482,97 @@ def edit_tokens(request):
...
+        if (
+            AccessToken.objects.filter(
+                user=request.user, application=application
+            ).count()
+            >= 3
+        ):
+            messages.error(
+                request, "You can only have up to 3 tokens. Please delete one first."
+            )
+            return redirect("edit_tokens")
```
tompollard, line 494: The lines after `return redirect("edit_tokens")` are unreachable.

physionet-django/user/templates/user/edit_tokens.html
```diff
@@ -40,6 +47,114 @@ <h3 class="mt-4">Your Tokens</h3>
...
+            </div>
+            {% for endpoint, scopes_info in available_endpoints.items %}
+            <div class="card-body">
+              <div class="ml-4">                    
+                <div class="form-group">
+                  <label>{{ scopes_info.name }}:</label>
+                  <div class="endpoint-label">{{ endpoint }}</div>
+                  <div class="form-check form-check-inline gap-2">
+                    {% for scope in scopes_info.scopes %}
+                      <input class="form-check-input annotation-endpoint" type="checkbox" name="annotation_endpoints" value="{{ endpoint }}{{ scope }}" id="{{ scope }}">
```
tompollard, line 78: Building the value by concatenating the endpoint and scope in the template feels fragile. Could we move this logic to the view and pass the complete keys directly?

physionet-django/user/views.py
```diff
@@ -452,28 +482,97 @@ def edit_tokens(request):
         raise Exception(f"OAuth application named '{app_name}' not found.")
 
     if request.method == "POST":
-
-        if AccessToken.objects.filter(user=request.user, application=application).count() >= 3:
-            messages.error(request, "You can only have up to 3 tokens. Please delete one first.")
+        if (
```
tompollard, line 485: The user should be part of the annotator group before they can create annotation tokens. Do we check this?

physionet-django/user/views.py
```diff
@@ -71,6 +72,35 @@
...
+        "annotations/collections/<collection_slug>/",
+        "Annotations",
+    ],
+    "annotations:annotations:write": [
+        "annotations/collections/<collection_slug>/",
+        "Annotations",
+    ],
+}
+
+ENDPOINT_TO_SCOPE = {
```
tompollard, line 96: This seems just to be the inverse of `SCOPE_TO_ENDPOINT`. Could it be made unnecessary?

tompollard (commented):
The PR mentions a "Annotator group" being created, but I don't see that anywhere?

PR #2511, opened 2025-10-16
V0: Annotation API - Added annotation group fixture

emmyxth:
As per conversation [here](https://github.com/MIT-LCP/physionet-build/issues/1047#issuecomment-3285413494), we want to add fine-grained token authorization. This is the first step, wherein we've created the Annotator group fixture, which admin can manually add people to the group.

Members of the Annotator group have Write access to `AnnotationCollection, AnnotationType, and Annotation`s

### DIFF
--- a/physionet-django/user/fixtures/demo-user.json
+++ b/physionet-django/user/fixtures/demo-user.json
@@ -14669,6 +14669,49 @@
   }
 },
 {
+  "model": "auth.group",
+  "fields": {
+    "name": "Annotator",
+    "permissions": [
+      [
+        "add_annotation",
+        "annotation",
+        "annotation"
+      ],
+      [
+        "change_annotation",
+        "annotation",
+        "annotation"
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks Emmy. I think this is fine for now, but we will want to make changes to the fixture later. Specifically:

- We should move the fixture to `physionet-django/annotation/fixtures/site-data.json` because we will want the group to be loaded as part of the deployment process described at: https://github.com/MIT-LCP/physionet-build/blob/dev/deploy/README.md

> "There is initial data for the site (licenses, tags, etc) in site-data.json fixture files in the fixtures directories of certain apps. Load them before deploying the site for the first time with: ..."

Also, I think these permissions are too broad for the `annotator` group. I think we will want something like an "Annotator Manager" to manage collections etc.

tompollard (approved):

Issue #2516, opened 2025-10-28
Bogus link in "PhysioNet New Data Access Request" email


bemoody:
notify_owner_data_access_request.html says:

```
Dear {{ user.get_full_name }},

There is a new data access request for the project
{{ data_access_request.project }}.

You can view and reply to the request following this link:
{{ request_protocol }}://{{ request_host}}{% url 'data_access_request_view' data_access_request.project.slug data_access_request.project.version data_access_request.requester.id %}

{{ signature }}
```

This URL is wrong.  The final parameter for `data_access_request_view` is the ID of the DataAccessRequest, not the ID of the user.

tompollard:
Should be addressed by https://github.com/MIT-LCP/physionet-build/commit/1d52327911d5e15a89a808f95552087bd3ef20d1. 

@bemoody do we need to tell the user who reported this issue that it is now fixed?

tompollard:
We have followed up with the user. They can access the list of outstanding access requests at:
https://physionet.org/access-requests/SLUG/VERSION/

PR #2517, opened 2025-10-29
Fix incorrect URL in data access request notification email. Ref #2516

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/2516, the email notification sent to project owners when they receive a new data access request contains an incorrect URL.  

The [notify_owner_data_access_request.html](https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/notification/templates/notification/email/notify_owner_data_access_request.html) template currently uses the `data_access_request.requester.id` (the user's ID) instead of `data_access_request.id` (the request's ID) as the final parameter. This causes the URL to point to the wrong data access request, or potentially a non-existent one. 

This pull request fixes the issue by changing the URL in the template from:

```
{% url 'data_access_request_view' data_access_request.project.slug data_access_request.project.version data_access_request.requester.id %}
```

to:

```
{% url 'data_access_request_view' data_access_request.project.slug data_access_request.project.version data_access_request.id %}
```

I also added a new test that:

  - Uses the existing "Self Managed Access Database Demo" project from fixtures
  - Creates a DataAccessRequest with a different ID than the requester's ID
  - Renders the email template and verifies it contains the correct URL

### DIFF
--- a/physionet-django/notification/templates/notification/email/notify_owner_data_access_request.html
+++ b/physionet-django/notification/templates/notification/email/notify_owner_data_access_request.html
@@ -4,6 +4,6 @@ There is a new data access request for the project
 {{ data_access_request.project }}.
 
 You can view and reply to the request following this link:
-{{ request_protocol }}://{{ request_host}}{% url 'data_access_request_view' data_access_request.project.slug data_access_request.project.version data_access_request.requester.id %}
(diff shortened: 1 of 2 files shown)


Chrystinne (approved):
Looks good, @tompollard!

PR #2518, opened 2025-10-30
Feature: Add accessibility link to footer

rafalkosla101:
As far as i can see it is not used for physionet instance - could we move it to other place for hdn?

### DIFF
--- a/physionet-django/templates/footer.html
+++ b/physionet-django/templates/footer.html
@@ -17,9 +17,6 @@
           {% if FOOTER_SUPPORTED_BY %}
             <p class="dark-light-text mb-2">{{ FOOTER_SUPPORTED_BY }}</p>
           {% endif %}
-          {% if FOOTER_ACCESSIBILITY_PAGE_URL %}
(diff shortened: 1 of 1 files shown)


tompollard:
@Chrystinne are you able to look at this today?

> As far as i can see it is not used for physionet instance

This is a bug! It is important to display the link somewhere on the page, but for some reason it isn't being rendered at the moment (possibly because we haven't configured the .env file correctly?).

tompollard:
> As far as i can see it is not used for physionet instance

It looks like it currently isn't possible for the accessibility link to be displayed in the footer because `FOOTER_ACCESSIBILITY_PAGE_URL` can never be true. I assume this is fixed in the PR.

(FOOTER_ACCESSIBILITY_PAGE_URL isn't defined in settings/base.py.)

tompollard:
@rafalkosla101 I'm fine moving the accessibility link as suggested here, but the placement on the full width page seem seems awkward to me. Is this how you'd like it to look?

Screenshot 2025-10-31 at 3 53 42 PM

Centered might be better?

tompollard (approved):

PR #2519, opened 2025-11-04
Fix: Fix wrongly displayed main page button

rafalkosla101:
image

the main page button should be displayed as in the image instead of disappearing


tompollard (approved):
thanks!

Issue #2520, opened 2025-11-05
Improve clarity of credentialing application email with respect to training


tompollard:
Some datasets require identity check but **do not** require training. We have had reports to say that the email sent to applicants can be confusing where this is the case. An example email is below:

```
Subject: PhysioNet credentialing application notification
 

Dear X,

Thank you for submitting your request for credentialed access to
PhysioNet. If your request is approved, you will be authorized to use
protected databases upon signing Data Use Agreements and completing
any required training.

To submit your training documents for review, please visit:
XXX

Please note that each dataset may have its own data usage terms and/or
training requirements. The specific training requirements and data
usage terms for each dataset are listed in the project description.

Please allow one week for your application to be reviewed and
processed. Thank you for your understanding and patience. You can
follow the status of your application at:
XXX

Regards, The PhysioNet Team, XXX
```

tompollard:
My suggestion is to reword/restructure the letter to make it clear that training requirements vary by dataset:

```
  Thank you for submitting your request for credentialed access to PhysioNet.
  If your request is approved, you will be authorized to use protected databases
  upon signing Data Use Agreements.

  Please note that each dataset may have its own data usage terms and training
  requirements. Some datasets require completion of training courses before access
  is granted. The specific training requirements and data usage terms for each
  dataset are listed in the project description. If training is required for a
  dataset you wish to access, you can submit your training documents for review
  at: [link]
```

PR #2521, opened 2025-11-05
Tweak wording of credentialing application email. Ref #2520

tompollard:
As discussed in https://github.com/MIT-LCP/physionet-build/issues/2520, some datasets require an identity check but do not require training. We have had reports to say that the email sent to applicants can be confusing where this is the case. 

This is a minor change to the email content to (I hope) make it clear that the training requirement depends on the dataset.


Chrystinne (approved):
Looks good, @tompollard!

PR #2522, opened 2025-11-05
Rm unused FOOTER_ACCESSIBILITY_PAGE_URL variable.

tompollard:
Our example .env file includes a variable called `FOOTER_ACCESSIBILITY_PAGE_URL=`. This variable isn't used in the repository, so it can be removed.


Chrystinne (approved):
Looks good, @tompollard!

PR #2525, opened 2025-11-07
Move hardcoded PhysioNet content to dynamic Page

tompollard:
This page includes hardcoded content that is only relevant to PhysioNet. We now support dynamic pages via the console, so the content can be removed here (and served via a new page created in the console).


Chrystinne (approved):
Looks good, @tompollard!

PR #2527, opened 2025-11-10
Fix `related_name` values in `Annotation` model ForeignKey fields.

tompollard:
The current `related_name` values on the `Annotation` model (e.g., `collection_slug`, `annotation_type_slug`, `project_slug`) lead to confusing reverse lookups, e.g. 

- `collection = models.ForeignKey(AnnotationCollection, on_delete=models.CASCADE, related_name='collection_slug')`
  - To look up all annotations in a collection, you'd need to use: `collection.collection_slug.all()`
  - `collection.annotations.all()` would be clearer.

- `annotation_type = models.ForeignKey(AnnotationType, on_delete=models.PROTECT, related_name='annotation_type_slug')`
  - To look up all annotations of a specific type, you'd need to use: `annotation_type.annotation_type_slug.all()`
  - `annotation_type.annotations.all()` would be clearer.

- `project = models.ForeignKey('project.PublishedProject', on_delete=models.CASCADE, null=True, blank=True,
  related_name='project_slug')`
  - To look up all annotations for a project, you'd need to run: `project.project_slug.all()`
  - `project.annotations.all()` would be clearer.

- `location = models.OneToOneField(BaseLocation, on_delete=models.CASCADE, related_name='location')`
  - To look up all annotations for a location, you'd need to run: `base_location.location.all()`
  - `base_location.annotation.all()` would be clearer.

## Changes

  - Change `collection.related_name` from 'collection_slug' → 'annotations'
  - Change `annotation_type.related_name` from 'annotation_type_slug' → 'annotations'
  - Change `project.related_name` from 'project_slug' → 'project_annotations'
  - Change `location.related_name` from 'location' → 'annotation'

## Note

The `location` field currently has a one-to-one relationship with `Annotation` (hence `annotation` not `annotations` for `related_name`). This means each location can only be associated with a single annotation. We should consider whether we want to change this to a one location to many annotation relationship.


emmyxth (approved):
Need to be mindful of models migrations

PR #2530, opened 2025-11-19
Move rgmark's training forward 100 years

bemoody:
The rgmark user in the demo environment is useful for manually testing features that require a user to have completed training.  This user is also used by the test suite.

Set this user's training report to have been submitted and approved in 2122 rather than 2022, so that it will not be considered "expired" until 2125.


tompollard (approved):
Thanks!

PR #2538, opened 2025-12-03
Add warning to create project page

briangow:
This PR adds an environment variable which can be used to set an alert/warning on the create project page. The warning is intended to convey information about how access restrictions impact whether we can accept a submitted project. The example I provide shows how to reference a link since we will likely want to point to more detailed information (in a News article, etc). If the environment variable is left empty, no warning will be displayed.

### DIFF
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -60,7 +60,7 @@ BLOCKED_REGIONS = set(
 
 # Create project warning message
 # CREATE_PROJECT_WARNING: Custom message to display on create project page
-# Example: CREATE_PROJECT_WARNING=strong>TITLE:</strong> MESSAGE LINK<a href="LINK URL">LINK TEXT</a>.
+# Example: CREATE_PROJECT_WARNING=<strong>TITLE:</strong> MESSAGE LINK<a href="LINK URL">LINK TEXT</a>.
 CREATE_PROJECT_WARNING = config('CREATE_PROJECT_WARNING', default='')
 
 # Installed apps
--- a/physionet-django/physionet/settings/base.py
+++ b/physionet-django/physionet/settings/base.py
@@ -58,10 +58,10 @@ BLOCKED_REGIONS = set(
     if region.strip()
 )
 
-# Create project georestriction warning message
-# CREATE_PROJECT_GEORESTRICTION: Custom message to display on create project page
-# Example: CREATE_PROJECT_GEORESTRICTION=strong>TITLE:</strong> MESSAGE LINK<a href="LINK URL">LINK TEXT</a>.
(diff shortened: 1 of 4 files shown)


physionet-django/project/views.py
```diff
@@ -339,7 +339,10 @@ def create_project(request):
     else:
         form = forms.CreateProjectForm(user=user)
 
-    return render(request, 'project/create_project.html', {'form':form})
+    return render(request, 'project/create_project.html', {
+        'form': form,
+        'georestriction_message': getattr(settings, 'CREATE_PROJECT_GEORESTRICTION', ''),
+    })
```
tompollard: I think this is cleaner:

`

settings.CREATE_PROJECT_GEORESTRICTION
`

The default is already set in the config: `config('CREATE_PROJECT_GEORESTRICTION', default='')`

physionet-django/physionet/settings/base.py
```diff
@@ -58,6 +58,11 @@
     if region.strip()
 )
 
+# Create project georestriction warning message
+# CREATE_PROJECT_GEORESTRICTION: Custom message to display on create project page
+# Example: CREATE_PROJECT_GEORESTRICTION=strong>TITLE:</strong> MESSAGE LINK<a href="LINK URL">LINK TEXT</a>.
+CREATE_PROJECT_GEORESTRICTION = config('CREATE_PROJECT_GEORESTRICTION', default='')
```
tompollard: As far as I can see, there is no georestriction-specific logic here, so perhaps choose more general variable names?

tompollard (changes requested):

physionet-django/project/views.py
briangow: I agree - this has been updated.

physionet-django/physionet/settings/base.py
briangow: I've changed the variable names to be more generic.

physionet-django/physionet/settings/base.py
```diff
@@ -58,6 +58,11 @@
     if region.strip()
 )
 
+# Create project warning message
+# CREATE_PROJECT_WARNING: Custom message to display on create project page
+# Example: CREATE_PROJECT_WARNING=strong>TITLE:</strong> MESSAGE LINK<a href="LINK URL">LINK TEXT</a>.
```
tompollard: minor thing, but please could you add the missing `<` in the example here, and where it appears in .env.example too?

tompollard (approved):

physionet-django/physionet/settings/base.py
briangow: Good catch, thanks for pointing this out. It has been fixed.

PR #2539, opened 2025-12-04
Always show version list for published projects

### LINKED ISSUE #2532: Versions box on project landing pages should always be displayed
It looks like the following "Versions" box is only displayed if there are multiple versions of a project. It would be clearer to always display the box, regardless of the number of versions:

Image

bemoody:
The list of versions (shown in the sidebar of each published project page) is currently only shown when the project has multiple versions.

For the sake of clarity, show the list for every published project, even if there's only one version.  The list still is not shown for *unpublished* (active) projects.

Also, improve the styling and accessibility a little.

Fixes #2532

### DIFF
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -255,10 +255,8 @@
             {% endif %}
           </div>
         </div>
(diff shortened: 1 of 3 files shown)


tompollard (approved):
Looks great, thanks!

PR #2541, opened 2025-12-06
Add KHDP account linking via OAuth

### LINKED ISSUE #2407: Add "KHDP" login to PhysioNet
As part of the Korean collaboration, we should add KHDP login to PhysioNet. As a first step, this would allow users to link their KHDP account to PhysioNet on the /settings/profile/ page.

@westernsalt

Chrystinne:
This PR adds support for linking PhysioNet user accounts with the Korea Health Data Platform (KHDP):

- Implement KHDP account linking using the KHDP OAuth-style flow (code → token → userinfo).
- Add a KHDP settings page to explain the integration and display the “Link KHDP Account” action, in a style consistent with the existing ORCID linking page.
- Store KHDP user metadata and access token details in the linked account.


bemoody:
There is a lot about this that is really unclear to me.  I don't like trying to think about OAuth because it is really complicated and the consequences of getting it wrong are disastrous.

(To put it simply, what we should be doing is using *OpenID Connect*.  And if we can't use OIDC then we have to reinvent it, which means understanding exactly what OIDC does and why.)

How can we verify the response - to know whether the authorization code was legitimately generated by the same person who is submitting it to us?  I think that this is why OIDC has a "nonce" field.

There is also the "state" field and if I understand correctly, that's needed to ensure the person intended to submit the code to PhysioNet (i.e. to prevent CSRF.)  https://stackoverflow.com/questions/46844285/

----

I can try to test things myself, but it would help if you would provide example data with the secret information removed.
- What does the URL returned by "edit_khdp" look like?
- What does the URL passed to "auth_khdp" look like?
- What does the *token* JSON response look like?
- What does the *userinfo* JSON response look like?
We need to have some test cases, so we need mock data for all of the above.

----

Some things that jump out:
- settings tabs shouldn't include "KHDP" if KHDP is not configured
- auth_url should be generated by `reverse` + `build_absolute_uri`, not a config variable
- what is "legacy KHDP callback path"?  This doesn't exist today so surely there is no "legacy URL" that needs to be maintained
- don't use `except Exception`
- need `@disallow_during_maintenance` (probably for both views?)

Chrystinne:
> I can try to test things myself, but it would help if you would provide example data with the secret information removed.
> 
> * What does the URL returned by "edit_khdp" look like?
> * What does the URL passed to "auth_khdp" look like?
> * What does the _token_ JSON response look like?
> * What does the _userinfo_ JSON response look like?
>   We need to have some test cases, so we need mock data for all of the above.

Mock data for KHDP OAuth testing

Here's an example data representing the various requests and responses in the KHDP OAuth flow, with all sensitive information removed.

1. URL returned by edit_khdp() - redirect to KHDP authorization
When a user clicks "Link KHDP Account" in PhysioNet->Settings, they are redirected to:

KHDP_AUTH_URL_EXAMPLE = (
    "https://khdp.example.com/oauth/authorize?"
    "appId=test_client_id_12345&"
    "redirectUrl=http%3A%2F%2Flocalhost%3A8000%2Fuser%2Fauth%2Fkhdp%2F&"
    "nonce=a1b2c3d4e5f6g7h8i9j0k1l2m3n4o5p6"
)

Components:
- appId: KHDP_CLIENT_ID (e.g., "test_client_id_12345")
- redirectUrl: KHDP_LINK_REDIRECT_URI (e.g., "http://localhost:8000/user/auth/khdp/")
- nonce: 32-character random string (e.g., "a1b2c3d4e5f6g7h8i9j0k1l2m3n4o5p6")

2. URL passed to auth_khdp() - callback from KHDP
KHDP redirects the user back to auth_khdp() with:

KHDP_CALLBACK_URL_EXAMPLE = (
    "http://localhost:8000/user/auth/khdp/"
    "?code=auth_code_xyz789abc123def456ghi789"
)

Components:
- code: Authorization code from KHDP (e.g., "auth_code_xyz789abc123def456ghi789")
(Note: KHDP does NOT return 'state' or 'nonce' parameters. The nonce is validated against request.session['khdp_nonce'] to prevent CSRF.)

3. Token JSON response - response from KHDP token endpoint

POST request to KHDP_TOKEN_URL with:
{
  "code": "auth_code_xyz789abc123def456ghi789",
  "appId": "test_client_id_12345",
  "secretKey": "REDACTED_SECRET"
}

KHDP_TOKEN_RESPONSE_EXAMPLE = {
    "accessToken": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9...",
    "tokenType": "Bearer",
    "expiresIn": 3600,
    "refreshToken": "refresh_token_abc123def456...",
    "idToken": (
        "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9."
        "eyJzdWIiOiJ1c2VyMTIzIiwibm9uY2UiOiJhMWIyYzNkNGU1ZjZnN2g4aTlqMGsx..."
        ".signature_part_..."
    ),
}

Components:
- accessToken: Bearer token for calling userinfo endpoint
- tokenType: "Bearer"
- expiresIn: Token lifetime in seconds (e.g., 3600 = 1 hour)
- refreshToken: Optional refresh token (may not be used for account linking)
- idToken: **Optional** - Not confirmed if KHDP returns this. 
  (If present, contains nonce for additional validation. Code handles both cases.)

4. Userinfo JSON response - response from KHDP userinfo endpoint
Extract userId and create/update KhdpAccount in the database

GET request to KHDP_USERINFO_URL with:
Headers: { "Authorization": "Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9..." }

KHDP_USERINFO_RESPONSE_EXAMPLE = {
    "userId": "chrystinne9999",
    "userName": "Chrystinne",
    "mail": "chrystinne@example.com",
    "affiliation": "MIT",
    "orcid": "0000-0000-0000-0000",
    "physionetId": "chrystinne",
}

Components:
- userId: Unique KHDP user identifier (REQUIRED - used as khdp_user_id)
- userName: User's display name
- mail: User's email address
- affiliation: User's organization/affiliation
- orcid: ORCID identifier (if available)
- physionetId: PhysioNet username (if KHDP has it)

Chrystinne:
@bemoody Here's a summary for testing

Test Case Flow:

1. User clicks "Link KHDP Account" button in settings (POST to edit_khdp)
   - Session gets: request.session['khdp_nonce'] = "a1b2c3d4e5f6g7h8i9j0k1l2m3n4o5p6"
   - User redirected to KHDP_AUTH_URL_EXAMPLE

2. KHDP authenticates user and redirects back (callback to auth_khdp)
   - Request URL: KHDP_CALLBACK_URL_EXAMPLE
   - code extracted from query parameter

3. Code is exchanged for token (POST to token_url)
   - Request body: {"code": "...", "appId": "...", "secretKey": "..."}
   - Response: KHDP_TOKEN_RESPONSE_EXAMPLE

4. Call userinfo endpoint (GET to userinfo_url)
   - Request headers: {"Authorization": "Bearer <accessToken>"}
   - Response: KHDP_USERINFO_RESPONSE_EXAMPLE
   - Extract userId and create/update KhdpAccount

Chrystinne:
> There is a lot about this that is really unclear to me. I don't like trying to think about OAuth because it is really complicated and the consequences of getting it wrong are disastrous.
> 
> (To put it simply, what we should be doing is using _OpenID Connect_. And if we can't use OIDC then we have to reinvent it, which means understanding exactly what OIDC does and why.)
> 
> How can we verify the response - to know whether the authorization code was legitimately generated by the same person who is submitting it to us? I think that this is why OIDC has a "nonce" field.
> 
> There is also the "state" field and if I understand correctly, that's needed to ensure the person intended to submit the code to PhysioNet (i.e. to prevent CSRF.) https://stackoverflow.com/questions/46844285/
> 

Regarding the verification of the authorization code's authenticity (nonce), I've implemented nonce validation as follows: 

- We generate a random 32-character nonce (in edit_khdp method); 
- We store it in the Django session (server-side), and in auth_khdp(); 
- We verify that the nonce exists in the session (i.e., proves the user initiated from within PhysioNet).

If KHDP returns an ID token, we decode and compare the nonce inside it as an additional validation layer. This ensures the authorization code came from the legitimate authentication flow that PhysioNet initiated.

Regarding CSRF protection (state), KHDP doesn't return the state parameter in the callback either. So I've implemented CSRF protection through the session-stored nonce. We now use the nonce stored in the user's authenticated session as proof against CSRF.

Chrystinne:
> Some things that jump out:
> 
> * settings tabs shouldn't include "KHDP" if KHDP is not configured
> * auth_url should be generated by `reverse` + `build_absolute_uri`, not a config variable
> * what is "legacy KHDP callback path"?  This doesn't exist today so surely there is no "legacy URL" that needs to be maintained
> * don't use `except Exception`
> * need `@disallow_during_maintenance` (probably for both views?)

Thanks for your comments, @bemoody. Here are the final changes I'm pushing:

- KHDP tab only appears when properly configured;
- Callback URL generated at runtime with proper URL using reverse();
- Specific exception handling;
- Views protected during system maintenance.
- CSRF Protection added: Nonce-based validation prevents CSRF.

bemoody:
You're handling a null value for `physionetId` but not for `orcid`.  If `orcid` is null:
```
ERROR 2026-01-14 16:00:25,591 KHDP account save failed: NOT NULL constraint failed: user_khdpaccount.orcid
  File "/home/benjamin/ve-pndjangoplus/lib/python3.11/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benjamin/ve-pndjangoplus/lib/python3.11/site-packages/django/db/backends/sqlite3/base.py", line 328, in execute
    return super().execute(query, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sqlite3.IntegrityError: NOT NULL constraint failed: user_khdpaccount.orcid

```

bemoody:
Thanks for adding tests; I'm guessing they would work if you put the `override_settings` on the class rather than the individual functions.

So here's the issue I mentioned above.  Your code works if it receives a response that looks like this:
```
        mock_userinfo = {
            'publicUuid': 'khdp-public-uuid-456',
            'userId': 'khdp-user-456',
            'userName': 'User Without ORCID',
            'affiliation': 'Another University',
            'mail': 'noorcid@khdp.example.com',
            # No orcid field
            # No physionetId field
        }
```

But in reality, the userinfo response looks like this:
```
        mock_userinfo = {
            'publicUuid': 'khdp-public-uuid-456',
            'userId': 'khdp-user-456',
            'userName': 'User Without ORCID',
            'affiliation': 'Another University',
            'mail': 'noorcid@khdp.example.com',
            'orcid': None,
            'physionetId': None,
        }
```

Change that in your test code and you should see what I mean.

Chrystinne:
> Thanks for adding tests; I'm guessing they would work if you put the `override_settings` on the class rather than the individual functions.
> 
> So here's the issue I mentioned above. Your code works if it receives a response that looks like this:
> 
> ```
>         mock_userinfo = {
>             'publicUuid': 'khdp-public-uuid-456',
>             'userId': 'khdp-user-456',
>             'userName': 'User Without ORCID',
>             'affiliation': 'Another University',
>             'mail': 'noorcid@khdp.example.com',
>             # No orcid field
>             # No physionetId field
>         }
> ```
> 
> But in reality, the userinfo response looks like this:
> 
> ```
>         mock_userinfo = {
>             'publicUuid': 'khdp-public-uuid-456',
>             'userId': 'khdp-user-456',
>             'userName': 'User Without ORCID',
>             'affiliation': 'Another University',
>             'mail': 'noorcid@khdp.example.com',
>             'orcid': None,
>             'physionetId': None,
>         }
> ```
> 
> Change that in your test code and you should see what I mean.

Good catch, @bemoody! I’m doing that, thanks!

bemoody (approved):

PR #2543, opened 2025-12-10
Display correct DOI prefix in project previews

bemoody:
The project preview page shows a placeholder citation for the project, as well as placeholder DOIs in the sidebar.

Instead of hardcoding PhysioNet's DOI prefix (10.13026), this should use the configured DATACITE_PREFIX (or the actual DOI prefix, if a DOI has already been assigned.)

If the project has no DOI and there is no configured DATACITE_PREFIX, then the preview shouldn't show DOIs at all.

Also fix the formatting of DOI-less citations which was broken by pull #2409.

### DIFF
--- a/physionet-django/project/modelcomponents/coreproject.py
+++ b/physionet-django/project/modelcomponents/coreproject.py
@@ -1,5 +1,6 @@
 import uuid
 
+from django.conf import settings
 from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models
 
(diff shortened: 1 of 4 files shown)


tompollard (approved):
Looks good to me, thanks. Seems odd to define `future_doi_prefix` twice like that, but I guess that's easiest.

bemoody:
You're right.  I don't have a better idea offhand, I suppose it could be a mixin class, but the classes are complicated enough already...

PR #2545, opened 2025-12-18
Add resource_type, access_policy, and topics to API responses

Rutvikrj26:
Enhance API responses by including resource type, access policy, and topics for published projects. This change improves the metadata provided to users, making the API more informative and user-friendly.


bemoody (approved):
Looks good, thanks!

PR #2546, opened 2025-12-19
Add federated search and enhance API responses

Rutvikrj26:
Enhance API responses by including resource type, access policy, and topics. Implement federated search functionality with UUID support for project models.

## API Updates
- Expose public_project_uuid in PublishedProjectSerializer and PublishedProjectDetailSerializer
- Update API documentation in export/views.py to include UUID field

## Federation Models Migration
- Move FederatedSite, FederatedProject, and FederationSyncLog models from project app to search app
- Update FederatedProject model:
  * Add public_project_uuid field for stable identification
  * Change resource_type and access_policy to string fields (from integer codes)
  * Remove is_stale field (using full refresh strategy instead)
  * Update unique_together to use public_project_uuid instead of slug+version

## Search Functionality
- Create search/federation.py with federated search logic
- Update resource_type filtering to work with string values
- Remove is_stale filtering (full refresh approach)

## Management Commands
- Create sync_federated_sites command in search app
- Update sync logic to:
  * Use public_project_uuid for identification
  * Validate presence of UUID in API responses
  * Use full refresh (delete + recreate) instead of stale marking
  * Support string values for resource_type and access_policy

## Admin Interface
- Register federation models in search/admin.py with comprehensive admin classes
- Add list displays, filters, and fieldsets for all federation models

Based on PR #2534 feedback from @bemoody:
- UUID replaces slug+version as stable identifier
- String values for resource_type and access_policy for API extensibility
- Models moved to search app (proper separation of concerns)
- Removed stale objects logic in favor of batch delete+add


Rutvikrj26:
@tompollard, @bemoody  Please take a look - I think the reverse migrations is failing because of the retroactive uuid addition for published projects. Apart from that, things seems to be good to go.

bemoody:
The `0090` migration needs to have `MIGRATE_AFTER_INSTALL = True`.  That's how the two-step upgrade works: the first step (pre-install) adds a new column that is nullable and non-unique, then the second step (post-install) changes it to be unique and non-nullable.

physionet-django/search/templates/search/content_list.html
```diff
@@ -2,31 +2,73 @@
...
+        <a href="{{ published_project.source_url }}" target="_blank" rel="noopener">
+          {{ published_project.title }}
+          <i class="fa fa-external-link-alt fa-sm"></i>
+        </a>
+      </{% firstof content_list_header 'h2' %}>
+      <p class="text-muted card-authors">
+        <small>Source: {{ published_project.source_site.site_name }}</small>
+      </p>
+      <div style="margin-bottom: 1rem;">
         {{ published_project.abstract|safe|truncatechars_html:250 }}
```
tompollard, line 28: Should we remove "safe" here? I don't think we want unescaped HTML from external sites?

physionet-django/search/migrations/0001_initial.py
```diff
@@ -0,0 +1,322 @@
...
+                            ("failed", "Failed"),
+                            ("in_progress", "In progress"),
+                        ],
+                        default="never",
+                        help_text="Status of most recent sync attempt",
+                        max_length=20,
+                    ),
+                ),
+                (
+                    "api_key",
```
tompollard: is this being stored as plain text? is that okay?

physionet-django/search/management/commands/sync_federated_sites.py
```diff
@@ -0,0 +1,290 @@
...
+            site: FederatedSite instance
+
+        Returns:
+            list: All project metadata dicts with source_url and public_project_uuid
+
+        Raises:
+            Exception: If API request fails or required fields are missing
+        """
+        all_projects = []
+        next_url = site.get_api_endpoint()
```
tompollard, line 195: do we do anything to check that the next URL belongs to the same site? if not, should we add check?

physionet-django/console/forms.py
```diff
@@ -1037,3 +1037,63 @@ class Meta:
...
+        if self.site_instance:
+            self.fields['site_identifier'].initial = self.site_instance.site_identifier
+            self.fields['site_identifier'].widget.attrs['readonly'] = True
+            self.fields['site_name'].initial = self.site_instance.site_name
+            self.fields['api_base_url'].initial = self.site_instance.api_base_url
+            self.fields['is_active'].initial = self.site_instance.is_active
+            self.fields['api_key'].initial = self.site_instance.api_key
+
+    def clean_site_identifier(self):
+        from search.models import FederatedSite
```
tompollard: why is this import here?

physionet-django/console/forms.py
```diff
@@ -1037,3 +1037,63 @@ class Meta:
         model = CodeOfConduct
         fields = ('name', 'version', 'slug', 'html_content')
         labels = {'html_content': 'Content'}
+
+
+class FederatedSiteForm(forms.Form):
+    """
+    Form for adding or editing a federated site
+    """
+    from search.models import FederatedSite
```
tompollard: why is this import here?

physionet-django/console/views.py
```diff
@@ -3706,3 +3706,199 @@ def event_agreement_delete(request, pk):
...
+
+# ------------------------- Federated Sites Views ------------------------- #
+
+
+@console_permission_required('user.can_view_admin_console')
+def federated_sites(request):
+    """
+    List all federated sites with their sync status
+    """
+    from search.models import FederatedSite, FederationSyncLog
```
tompollard: as for other imports, why not import them at a module level?

physionet-django/console/views.py
```diff
@@ -3706,3 +3706,199 @@ def event_agreement_delete(request, pk):
...
+    """
+    from search.models import FederatedSite
+    from django.core.management import call_command
+
+    site = get_object_or_404(FederatedSite, id=site_id)
+
+    if request.method == 'POST':
+        try:
+            # Call the sync management command for this specific site
+            call_command('sync_federated_sites', site=site.site_identifier, verbosity=0)
```
tompollard, line 3870: Maybe deal with later, but can we handle this in a background task?

physionet-django/search/views.py
```diff
@@ -11,6 +11,51 @@
...
+
+
+def get_federated_projects(resource_type, search_term):
+    """
+    Search federated projects by resource type and search term.
+
+    Returns a queryset of FederatedProject objects matching the criteria.
+    """
+    # Map resource_type IDs to string values
+    RESOURCE_TYPE_MAP = {
```
tompollard: Can we avoid hardcoding this? Isn't the resource list stored elsewhere (e.g. in database or enum)?

physionet-django/search/views.py
```diff
@@ -11,6 +11,51 @@
...
+
+    # Filter by resource type if specified
+    if resource_type_strings:
+        federated_projects = federated_projects.filter(
+            resource_type__in=resource_type_strings
+        )
+
+    # Apply search term filtering if provided
+    if search_term:
+        search_terms = re.split(r'\s*[\;\,\s]\s*', re.escape(search_term))
```
tompollard: Not 100% sure but I think this isn't quite right because isn't the regex being applied to the escaped term, but written to be applied to the unescaped term? 

e.g. doesn't `heart rate, blood pressure` get escaped to something like `heart\ rate\,\ blood\ pressure` before matching?

physionet-django/project/test_views.py
```diff
@@ -1739,7 +1739,7 @@ def test_bibtex_doi_and_url(self):
 
         if project.doi:
             self.assertIn(f'doi = {{{project.doi}}}', bibtex)
-            self.assertIn(f'url = {{https://doi.org/{project.doi}}}', bibtex)
+            self.assertIn(f'url = {{https://doi.org/{project.doi}}}', bibtex)  # noqa: E231
```
tompollard: is this necessary? the only change seems to be `noqa: E231`

physionet-django/project/test_views.py
```diff
@@ -1756,7 +1756,7 @@ def test_bibtex_citation_key(self):
 
         version_str = project.version if project.version else ''
         expected_key = f'{settings.SITE_NAME}-{project.slug}-{version_str}'
-        self.assertIn(f'@article{{{expected_key},', bibtex)
+        self.assertIn(f'@article{{{expected_key},', bibtex)  # noqa: E231
```
tompollard: As above (is this necessary? the only change seems to be `noqa: E231`)

physionet-django/search/views.py
```diff
@@ -199,14 +244,52 @@ def content_index(request, resource_type=None):
...
 
-    # BUILD
+    # BUILD - Get local published projects
     published_projects = get_content(resource_type=resource_type,
                                      orderby=orderby,
                                      direction=direction,
                                      search_term=topic)
 
+    # Get federated projects
+    federated_projects = get_federated_projects(resource_type=resource_type,
```
tompollard, line 256: Should we add upper bound to the results to avoid loading absurdly long lists? e.g. limit to 100?

physionet-django/console/forms.py
```diff
@@ -1037,3 +1037,63 @@ class Meta:
...
+    )
+    site_name = forms.CharField(
+        max_length=200,
+        label='Site Name',
+        help_text='Display name for this site (e.g., "PhysioNet", "HealthDataNexus")'
+    )
+    api_base_url = forms.URLField(
+        max_length=500,
+        label='API Base URL',
+        help_text='Base URL for the federation API (e.g., "https://healthdatanexus.ca")'
```
tompollard: Your URL is https://healthdatanexus.ai!

tompollard (changes requested):
Very cool @Rutvikrj26. I successfully implemented federated search locally.

Image

I added some comments inline.

bemoody (dismissed):
Haven't fully reviewed but here are some style things:

Put import statements at the top level; don't put them inside functions or classes.  (console/forms.py, console/views.py)

Please don't do unrelated reformatting (project/modelcomponents/metadata.py, project/test_views.py, project/views.py, maybe other places?).  Make those changes, if needed, in a separate pull request.

Don't use integer keys in URLs (console/urls.py).  FederatedSite has a "site_identifier"; use that in the URL instead.

Rutvikrj26:
Unfortunaltely, I asked an LLM to do the style changes - I'll revert those and manually fix them. That should remove almost all of the styling comments.

physionet-django/console/views.py
Rutvikrj26: I wanted this command to be synchronous, rather than a background task. This is for the manual sync button.

physionet-django/search/management/commands/sync_federated_sites.py
Rutvikrj26: The reason I have not implemented it is - the next site is provided by the one from where the paginated results are coming in, which essentially with the current design has to be a physionet instance. Furthermore, the physionet instance needs to be registered by the admin and hence there's almost no posisbility of a rogure instance being registered. 

I can implement that check, but it's just more code without much value in my opinion.

let me know your thoughts!

physionet-django/search/migrations/0001_initial.py
Rutvikrj26: I'm completely removing the api key altogether since we're not using it. The current pulling mechanisms use physionet's public facing api endpoint, which essentially renders the api key field useless, and to some extent misleading.

physionet-django/search/templates/search/content_list.html
Rutvikrj26: I don't think it is a good idea.
- This might break the visual hierarchy since the local projects will still have the safe tag, but not the federated projects.
- The federated projects are trusted physionet instances. The idea is that those projects seamlessly blend in with the local projects, reducing the difference, making the search truly federated.

physionet-django/search/views.py
Rutvikrj26: Fully agreed! Great catch!
This also seems to be a general problem elsewhere - will fix those as well.

physionet-django/search/views.py
Rutvikrj26: Perfect, I've put in a 500 project limit for local and federated projects each. Essntially that leads to 1000 total (essentially 100 research pages at 10 results per page). This limit requires 5 MB of memory - hence generous enough but still not a hit to the system. 
let me know if you want to change the limit to something lower.

physionet-django/search/templates/search/content_list.html
Rutvikrj26: Sanitize the incoming API response HTML content with bleach.

physionet-django/search/management/commands/sync_federated_sites.py
Rutvikrj26: we want to be double sure.

physionet-django/search/templates/search/content_list.html
bemoody: The abstract is defined in FederatedProject as a `TextField`.  One approach to sanitizing would be:

(a) change `TextField` to `SafeHTMLField`, **and**

(b) invoke full_clean before creating FederatedProject.

`FederatedProject.objects.create` does not invoke full_clean; it bypasses all model/field validation.

bemoody:
Since pull #2574 was merged, there will be a migration conflict.  You should be able to fix this by editing the dependencies in `0088_publishedproject_public_project_uuid.py`:

```
--- a/physionet-django/project/migrations/0088_publishedproject_public_project_uuid.py
+++ b/physionet-django/project/migrations/0088_publishedproject_public_project_uuid.py
@@ -5,7 +5,7 @@
 
     dependencies = [
-        ("project", "0087_anonymousaccess_hide_authors"),
+        ("project", "0089_alter_activeproject_is_on_hold"),
     ]
 
     operations = [
```

tompollard:
Hey @Rutvikrj26 please could you update the migrations when you get a chance?

Rutvikrj26:
@tompollard I've recreated migrations.

tompollard:
@Rutvikrj26 Can you check the migrations? Looks like the upgrade test is failing.

Rutvikrj26:
@tompollard just fixed it.

tompollard (approved):
Thanks @Rutvikrj26!

tompollard (changes requested):
Sorry, I notice a couple of issues.

physionet-django/console/views.py
```diff
@@ -3743,3 +3744,191 @@ def event_agreement_delete(request, pk):
...
+        'cached_projects_count': cached_projects_count,
+    })
+
+
+@console_permission_required('user.can_view_admin_console')
+def federated_site_sync(request, site_id):
+    """
+    Trigger a manual sync for a specific site
+    """
+    from search.models import FederatedSite
```
tompollard: Could you move these imports to top of module? (or explain why they need to be here)?

physionet-django/console/views.py
```diff
@@ -3743,3 +3744,191 @@ def event_agreement_delete(request, pk):
...
+        'site': site,
+    })
+
+
+@console_permission_required('user.can_view_admin_console')
+def federated_site_delete(request, site_id):
+    """
+    Delete a federated site and all its cached projects
+    """
+    from search.models import FederatedSite
```
tompollard: Move import to top of module

tompollard (changes requested):

Rutvikrj26:
@tompollard  Thanks for catching these.

tompollard (approved):

bemoody:
My comment here (and Tom's) was not addressed.  https://github.com/MIT-LCP/physionet-build/pull/2546#discussion_r2770986940

tompollard:
Good catch @bemoody, sorry for missing this. @Rutvikrj26 please could you raise an issue and add a fix in a new PR?

physionet-django/search/templates/search/content_list.html
Rutvikrj26: Good Catch, Thanks! 
Will open a quick pr for this.

Issue #2201, opened 2024-03-05
Allow author names to be anonymized on "Anonymous link" for projects


tompollard:
We allow projects to be shared using an "anonymous link". 

We have had a couple of requests for author names to be concealed on the project page that is accessible via these links. This would allow the links to be used where double blind peer review is used.

Double blind peer review (i.e. both authors concealed from reviewers, and reviewers concealed from authors) seems relatively common for ML conferences.

PR #2549, opened 2025-12-26
Add author-masked option for anonymous links

tompollard:
Many machine learning conferences apply mutually anonymous (aka "double blind") reviewing for submissions, where reviewers do not know the author identity. For example:

- https://www.mlforhc.org/paper-submission: "Reviewing for MLHC is double-blind: the reviewers will not know the authors' identity and the authors will not know the reviewers' identity."
- https://chil.ahli.cc/submit/call-for-papers/: "The review process is mutually anonymous (aka "double blind"). Your submitted paper, as well as any supporting text or revisions provided during the discussion period, should be completely anonymized (including links to code repositories such as Github)."

We currently provide functionality to share active projects using "anonymous links" (see #2201 for background). This pull request extends that functionality by allowing authors to hide their identity.

This pull request:

  - Adds a `hide_authors` boolean field to the `AnonymousAccess` model to control whether author information is displayed
  - Provides two types of anonymous links:
    - **Standard Link**: Shows author names, affiliations, and corresponding author information (existing behavior)
    - **Author-Masked Link**: Hides all author information, citation box, and corresponding author details for anonymous peer review
  - Updates the submission info page with separate buttons to generate each link type
  - Updates the project preview page to conditionally hide author information based on the link type
  - Allows toggling between link types by regenerating the anonymous access link
  - Includes tests covering link generation, toggling, and preview display behavior

### Console

Image

### Project preview

Image


tompollard:
@lukepayyapilli to test:

- Login as admin
- Navigate to a submitted project in the console (e.g. http://localhost:8000/console/submitted-projects/p7TCIMkltNswuOB9FZH1/info/)
- Click the "Anonymous access" tab and then generate a link
- Check whether the information rendered with the private link is correct.

lukepayyapilli:
Looks good Tom - adding some suggestions for the future:

### 1. DRY-er
This pattern appears 3 times across `views.py` and `console/views.py`:
```python
anonymous = project.anonymous.first()
hide_authors = anonymous.hide_authors if anonymous else False
```

Could be extracted to a model method in `metadata.py`:
```python
def should_hide_authors(self):
    """Check if anonymous access is configured to hide author information."""
    anonymous = self.anonymous.first()
    return anonymous.hide_authors if anonymous else False
```

Then usage becomes: `hide_authors = project.should_hide_authors()`

### 2. Model field help_text
```python
# Current
hide_authors = models.BooleanField(default=False)

# Suggested
hide_authors = models.BooleanField(
    default=False,
    help_text="When True, author information is hidden for anonymous peer review"
)
```

### 3. Optional: Enum for access types
If more link types are added in the future, an enum would make the intent clearer:
```python
class AnonymousAccessType(models.TextChoices):
    STANDARD = "standard", "Standard Link"
    AUTHOR_MASKED = "author_masked", "Author-Masked Link"
```

These are minor nits so I'm happy to approve now.

lukepayyapilli (approved):

tompollard:
Thanks again Luke, great suggestions. I added the help_text, but decided not to make the other changes just because for (1) I felt it would be slightly confusing to have a "should_hide_authors" method on the project itself, because this mostly just applies to the case of anonymous links during review and for (3) I don't see the need for additional access types (though may be wrong!).

PR #2550, opened 2025-12-28
Truncate long author lists on front page cards

### LINKED ISSUE #2462: Author list on front page should be truncated
We should truncate long author lists on the front page:

Image

lukepayyapilli:
## Summary
- [x] Add CSS `line-clamp` truncation for author lists on front page and search results.
- [x] Apply `.card-authors` class to both `homepage_content_list.html` and `content_list.html`.
- [x] Update demo fixtures to include a project with 14 authors to demonstrate truncation.

## Approach
Uses CSS-only truncation with `line-clamp` to limit author lists to a single line with ellipsis. This is simpler than a Django filter approach and provides consistent rendering across all cards.

## Screenshots
author-trunc
content-list-trunc

Fixes #2462

### DIFF
--- a/physionet-django/project/fixtures/demo-project.json
+++ b/physionet-django/project/fixtures/demo-project.json
@@ -496,6 +496,201 @@
   }
 },
 {
+  "model": "project.publishedauthor",
+  "pk": 9,
+  "fields": {
+    "user": ["george"],
+    "display_order": 2,
+    "is_submitting": false,
+    "is_corresponding": false,
+    "approval_datetime": "2018-11-14T21:11:11.781Z",
+    "first_names": "George",
+    "last_name": "Moody",
+    "corresponding_email": null,
+    "project": 1
+  }
+},
+{
+  "model": "project.publishedauthor",
+  "pk": 10,
+  "fields": {
+    "user": ["aewj"],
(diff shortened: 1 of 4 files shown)


lukepayyapilli:
@tompollard - let me know if this works for the issue. Thanks!

tompollard (changes requested):
Thanks Luke!  Right now it looks like we are applying two independent filters (CSS line clamp and Django filter). This could lead to inconsistent rendering (sometimes et al, sometimes not). I think we should probably just apply one.

I'm okay with either, but I think my preference is the CSS method with the ellipsis, for simplicity. The "et al" is good for formal citations, but perhaps more than we need here.

Side note and we can do in a separate PR if needed, but we should probably apply the filtering method to the search results too, to avoid this: https://physionet.org/content/. It's in the template here:

https://github.com/MIT-LCP/physionet-build/blob/7a5ab9f24cb1e39432cb7d53f773d779f9b0fe47/physionet-django/search/templates/search/content_list.html#L10-L13

lukepayyapilli:
updated @tompollard!

tompollard (approved):
Looks great, thanks Luke.

PR #2551, opened 2025-12-30
Fix card-authors flex overflow

lukepayyapilli:
## Summary

- Prevents `.card-authors` from inheriting `flex-grow: 1` from the `.card--flat > p:last-of-type` rule
- Adds `flex-grow: 0` and `flex-shrink: 0` to keep the author list at its natural size

## Problem

When `.card-authors` is the last `<p>` element in a card (e.g., when the abstract doesn't contain `<p>` tags), it inherits `flex-grow: 1` which causes it to stretch and overflow.

## Screenshot

Before
Screenshot 2025-12-30 at 2 50 11 PM

After
Screenshot 2025-12-30 at 2 51 25 PM

## Test plan

- [x] Verified locally that card authors no longer overflow
- [x] Confirmed card spacing remains consistent with other cards

### DIFF
--- a/physionet-django/static/custom/css/style.css
+++ b/physionet-django/static/custom/css/style.css
@@ -245,6 +245,8 @@ body .main {
   -webkit-line-clamp: 1;
   -webkit-box-orient: vertical;
   overflow: hidden;
+  flex-grow: 0;
+  flex-shrink: 0;
 }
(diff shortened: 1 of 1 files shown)


tompollard (approved):
Good catch, thanks!

PR #2552, opened 2025-12-30
Truncate author lists using model method

lukepayyapilli:
## Summary

Replaces CSS-based author truncation with a model method approach for more reliable and consistent behavior.

- Adds `display_authors(max_authors=3)` method to project metadata mixin
- Shows first 3 authors followed by "et al." when there are more
- Simplifies templates to use `{{ published_project.display_authors }}`
- Includes unit tests for the new method

## Why

The previous CSS approach (`line-clamp`, `flex-grow`) had layout issues where `flex-grow: 1` from parent rules caused overflow. A model-based approach gives predictable output regardless of CSS context.

## Test plan

- [x] Unit tests pass (`python manage.py test project.test_views.TestDisplayAuthors`)
- [x] Verified locally on homepage and search results

### DIFF
--- a/physionet-django/project/modelcomponents/metadata.py
+++ b/physionet-django/project/modelcomponents/metadata.py
@@ -115,6 +115,15 @@ class Metadata(models.Model):
         """
         return self.authors.all().order_by('display_order')
 
+    def display_authors(self, max_authors=3):
+        """
+        Get a truncated display string of author names.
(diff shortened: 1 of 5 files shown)


tompollard (approved):
Looks good, thanks. The author list can be important for people, so displaying more names might be nice, but this works for now.

Issue #2336, opened 2025-01-29
Metrics for downloads - B2AI Voice


satra:
we would like to be able to download metrics (over some period). here are some, i'll let @alistairewj chime in with anything else:

- number of registrations
- number of downloads (by version)
- institutions of the registrants

if a download button is available for project contributors/owners that would be sufficient for the time being. these could be cumulative metrics or per period (e.g., month).

tompollard:
Related: https://github.com/MIT-LCP/physionet-build/issues/2168

tompollard:
@lukepayyapilli For general metrics displayed on the project landing page, some examples are:

- Zenodo: https://zenodo.org/records/18212661 (right side has views and downloads, alongside "show more details".
- Figshare: https://figshare.com/articles/dataset/COVID-19_ICU_treated_patients/23266205?file=41006783 (right side has a Usage metrics section".
- Dryad: https://datadryad.org/dataset/doi:10.5061/dryad.pg4f4qrn3 (right side has views, downloads, citations).

In the request above, @satra sounds like he may be asking for metrics that might be less broadly available - e.g. just displayed to the data contributor. We'd need to look into the ethics of revealing information like institutions - for this we might first need to first get approval from the user?

tompollard:
@lukepayyapilli a good starting point might be displaying "views" (like those shown at console/project-access-logs/).

lukepayyapilli:
Thanks for the guidance @tompollard. Before implementing, I wanted to clarify a few things and propose some approaches based on the examples you provided.

## Clarifying Questions

1. **Visibility**: Should metrics be displayed publicly (like Zenodo/Figshare/Dryad) or only to project authors/contributors (as mentioned in #2168)?

2. **Metric definition**: The existing `AccessLog` tracks unique users who accessed project files. Should we label this as "Views" or "Accesses" to be clear about what's being counted?

3. **Version scope**: Should we show metrics for the specific version being viewed, or aggregate across all versions of a project?

## Proposed Approaches

Based on the examples you shared:

**Option A: Public metrics in sidebar (like Zenodo/Dryad)**
- Add a "Metrics" card in the right sidebar
- Display access count with an icon (e.g., 👁 1,234 views)
- Simple, visible to everyone
- Similar to: [Zenodo](https://zenodo.org/records/18212661), [Dryad](https://datadryad.org/dataset/doi:10.5061/dryad.pg4f4qrn3)

**Option B: Author-only metrics page (per #2168)**
[...]

PR #2556, opened 2026-01-14
Add Project Views metric to sidebar

lukepayyapilli:
## Summary
- [x] Adds a "Project Views" card to the published project sidebar displaying unique file views count.
- [x] Counts unique authenticated users who have viewed the project's files section.
- [x] Includes tests for the new feature.

Part of #2336 (Phase 1a)

## Screenshot
Screenshot 2026-01-14 at 5 29 48 PM

## Future Optimizations
- Add composite index on `(content_type_id, object_id)` to the Log model for faster lookups.
- Consider denormalizing the count on the project model for high-traffic projects.

### DIFF
--- a/physionet-django/project/templates/project/published_project.html
+++ b/physionet-django/project/templates/project/published_project.html
@@ -243,6 +243,23 @@
         </div>
 
         <div class="card my-4">
+          <h5 class="card-header">Project Views</h5>
+          <div class="card-body text-center">
+            <div class="d-flex justify-content-around mb-2">
+              <div>
+                <h3 class="mb-0">{{ project_views_count }}</h3>
+                <small class="text-muted">Current Version</small>
+              </div>
+              <div>
+                <h3 class="mb-0">{{ all_versions_views_count }}</h3>
+                <small class="text-muted">All Versions</small>
+              </div>
+            </div>
+            <small class="text-muted mb-0"><em>Project Views by Unique Registered Users</em></small>
+          </div>
+        </div>
+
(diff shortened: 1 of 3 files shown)


tompollard (changes requested):
Thanks Luke! I have added a couple of quick thoughts below:

- I hadn't realized that these logs only count logged-in users. I think this is okay, but we should make it very clear that this is the case (i.e. somehow explain that we are only counting registered users and that we are significantly undercounting true views).
- I wonder whether we should aggregate the count for all versions of a project (i.e. if a project has v1 and v2 then we combine the unique counts)? We could consider counting two counts: this version and all versions.
- Instead of "File views" we should say "Views" or "Project Views" (project is the term we use to describe any published resource).

tompollard:
Thanks! Instead of "Count of Registered Users", how about "Project Views by Unique Registered Users"?

lukepayyapilli:
@tompollard - Updated to address your feedback:

- [x] Renamed from "File Views" to "Project Views".
- [x] Shows Project Views for both Current Version and All Versions.
- [x] Clarifies counts are for registered users only.

lukepayyapilli:
> Thanks! Instead of "Count of Registered Users", how about "Project Views by Unique Registered Users"?

Done @tompollard!

tompollard (approved):
Thanks Luke, looks good to me. We will want to make some tweaks later (not completely happy with "Project Views by Unique Registered Users" and want to check exactly what we capture) but the addition of the number is step forward!

PR #2557, opened 2026-01-15
Add detailed public metrics view for published projects

lukepayyapilli:
## Summary
- [x] Add `view_count()`, `views_by_version()`, and `views_over_time()` methods to PublishedProject model.
- [x] Add `unique_viewers_count()` and `unique_viewers_by_month()` methods to AccessLogQuerySet.
- [x] Add new `published_project_metrics` view using model methods.
- [x] Create metrics template showing Unique Registered Project Views (current + all versions), views by version table, and views over time.
- [x] Add URL route at `/content/<slug>/metrics/<version>/`.
- [x] Add "View Details" button to Project Views sidebar card.
- [x] Add fixture data with AccessLog entries for local testing.
- [x] Add tests for metrics view and manager methods.

## Approach
Created a dedicated metrics page accessible from the Project Views card in the sidebar. View counting logic is encapsulated in PublishedProject model methods (`view_count()`, `views_by_version()`, `views_over_time()`), which delegate to AccessLogQuerySet methods for the actual queries.

## Reasoning
- Model methods make the logic reusable and testable.
- Using `unique_viewers_by_month()` counts unique viewers per month (same user viewing in multiple months is counted once per month).
- Fixture data uses "LogCategory.ACCESS" format to match Django's enum string representation.

## Screenshots

Screenshot 2026-01-19 at 10 25 03 AM
Screenshot 2026-01-19 at 10 25 23 AM

Part of #2336

### DIFF
--- a/physionet-django/project/managers/log.py
+++ b/physionet-django/project/managers/log.py
@@ -1,9 +1,9 @@
 import datetime as dt
-from collections import defaultdict
 
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
-from django.db.models import QuerySet, Manager, Min
+from django.db.models import QuerySet, Manager, Count
+from django.db.models.functions import TruncMonth
 from django.utils import timezone
 
 from physionet.enums import LogCategory
@@ -28,19 +28,17 @@ class AccessLogQuerySet(QuerySet):
         """Count unique users who have viewed."""
         return self.values('user').distinct().count()
 
-    def first_views_by_month(self):
+    def unique_viewers_by_month(self):
         """
-        Count first-time viewers per month.
-        Each user is only counted in the month of their first view.
+        Count unique viewers per month.
+        A user viewing in multiple months is counted once per month.
         """
-        first_views = self.values('user').annotate(first_view=Min('creation_datetime'))
-
-        monthly_counts = defaultdict(int)
-        for fv in first_views:
-            month = fv['first_view'].replace(day=1, hour=0, minute=0, second=0, microsecond=0)
-            monthly_counts[month] += 1
-
-        return [{'month': m, 'count': c} for m, c in sorted(monthly_counts.items())]
+        return list(
+            self.annotate(month=TruncMonth('creation_datetime'))
+            .values('month')
+            .annotate(count=Count('user', distinct=True))
+            .order_by('month')
+        )
 
     def update_or_create(self, defaults=None, **kwargs):
         user = kwargs.get('user')
--- a/physionet-django/project/managers/log.py
+++ b/physionet-django/project/managers/log.py
@@ -1,8 +1,9 @@
 import datetime as dt
+from collections import defaultdict
 
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
-from django.db.models import QuerySet, Manager
+from django.db.models import QuerySet, Manager, Min
 from django.utils import timezone
 
 from physionet.enums import LogCategory
@@ -22,6 +23,24 @@ class AccessLogQuerySet(QuerySet):
     def create(self, **kwargs):
         kwargs['category'] = LogCategory.ACCESS
         return super().create(**kwargs)
+
+    def unique_viewers_count(self):
+        """Count unique users who have viewed."""
+        return self.values('user').distinct().count()
+
+    def first_views_by_month(self):
+        """
+        Count first-time viewers per month.
+        Each user is only counted in the month of their first view.
+        """
+        first_views = self.values('user').annotate(first_view=Min('creation_datetime'))
+
+        monthly_counts = defaultdict(int)
+        for fv in first_views:
+            month = fv['first_view'].replace(day=1, hour=0, minute=0, second=0, microsecond=0)
+            monthly_counts[month] += 1
+
+        return [{'month': m, 'count': c} for m, c in sorted(monthly_counts.items())]
 
     def update_or_create(self, defaults=None, **kwargs):
         user = kwargs.get('user')
--- a/physionet-django/project/modelcomponents/publishedproject.py
+++ b/physionet-django/project/modelcomponents/publishedproject.py
@@ -430,11 +430,11 @@ class PublishedProject(Metadata, SubmissionInfo):
 
     def views_over_time(self):
         """
-        Return monthly view counts for this project version.
-        Each user is only counted in the month of their first view.
+        Return monthly unique viewer counts for this project version.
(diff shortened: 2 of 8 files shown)


physionet-django/project/views.py
```diff
@@ -2108,6 +2108,45 @@ def published_project(request, project_slug, version, subdir=''):
...
 
+def published_project_metrics(request, project_slug, version):
+    """
+    Public metrics page for a published project.
+    """
+    from django.db.models import Count
+    from django.db.models.functions import TruncMonth
+
+    try:
+        project = PublishedProject.objects.get(slug=project_slug, version=version)
```
tompollard: can you use `get_object_or_404()` here?

physionet-django/project/views.py
```diff
@@ -2108,6 +2108,45 @@ def published_project(request, project_slug, version, subdir=''):
                   status=status)
 
 
+def published_project_metrics(request, project_slug, version):
+    """
+    Public metrics page for a published project.
+    """
+    from django.db.models import Count
```
tompollard: please could you move to top of module?

tompollard (commented):
Thanks again Luke. I added a couple of quick comments but would like to take a deeper look later!

physionet-django/project/views.py
```diff
@@ -2108,6 +2108,45 @@ def published_project(request, project_slug, version, subdir=''):
...
+    except ObjectDoesNotExist:
+        raise Http404()
+
+    content_type = ContentType.objects.get_for_model(project)
+
+    project_logs = AccessLog.objects.filter(object_id=project.id, content_type=content_type)
+    project_views_count = project_logs.unique_viewers_count()
+    views_over_time = project_logs.first_views_by_month()
+
+    all_versions = PublishedProject.objects.filter(slug=project_slug).order_by('version_order')
```
tompollard: Feels like it would be cleaner to put this logic on the `PublishedProject` object. e.g. we could have a "count_views" method with an optional "version" arg or something?

lukepayyapilli:
Thanks for the review @tompollard! These are great improvements. I've addressed your feedback:

- **Use `get_object_or_404()`**: Updated to use `get_object_or_404()` instead of try/except.
- **Move imports to top of module**: Removed the unused imports (`Count`, `TruncMonth`) that were inside the function.
- **Move logic to PublishedProject model**: Added three new methods to `PublishedProject`:
  - `view_count(all_versions=False)` - returns unique viewer count for this version or all versions
  - `views_by_version()` - returns list of view counts per version
  - `views_over_time()` - returns monthly view counts

The view is now much cleaner and delegates to these model methods. Let me know if there are other improvements you'd like me to make (e.g. add additional visualizations or aggregated metrics etc.) - thanks!

physionet-django/project/modelcomponents/publishedproject.py
```diff
@@ -385,3 +386,60 @@ def get_all_news(self):
...
+
+    def view_count(self, all_versions=False):
+        """
+        Return the count of unique registered users who have viewed this project.
+
+        Args:
+            all_versions: If True, count views across all versions of this project.
+                         If False (default), count only views of this specific version.
+        """
+        from project.models import AccessLog
```
tompollard: Sorry, please could you move import to top of module?

physionet-django/project/modelcomponents/publishedproject.py
```diff
@@ -385,3 +386,60 @@ def get_all_news(self):
...
+            return AccessLog.objects.filter(
+                object_id=self.id,
+                content_type=content_type
+            ).unique_viewers_count()
+
+    def views_by_version(self):
+        """
+        Return a list of view counts for each version of this project.
+        """
+        from project.models import AccessLog
```
tompollard: please could you move import to top of module?

physionet-django/project/modelcomponents/publishedproject.py
```diff
@@ -385,3 +386,60 @@ def get_all_news(self):
...
+            ).unique_viewers_count()
+            result.append({'version': v.version, 'count': count})
+        return result
+
+    def views_over_time(self):
+        """
+        Return monthly view counts for this project version.
+        Each user is only counted in the month of their first view.
+        """
+        from project.models import AccessLog
```
tompollard: please could you move import to top of module?

tompollard (changes requested):

lukepayyapilli:
Apologies @tompollard - I missed some inline imports in the model methods. Fixed now by importing `AccessLog` at the top of the file from `project.modelcomponents.log` to avoid circular import issues.

physionet-django/project/managers/log.py
```diff
@@ -23,6 +24,24 @@ def create(self, **kwargs):
         kwargs['category'] = LogCategory.ACCESS
         return super().create(**kwargs)
 
+    def unique_viewers_count(self):
+        """Count unique users who have viewed."""
+        return self.values('user').distinct().count()
+
+    def first_views_by_month(self):
```
tompollard: I'm not sure that I like the idea of tracking just the first-time views. My preference would be to count the number of unique viewers per month (i.e. if the same person looks at the project in March and April, they would be counted once in March and once in April).

lukepayyapilli: ok I can make that change - I wasn't sure what we wanted the desired behavior to be. Also, if that's counted then should the total views count that as 2 instead of 1? I initially made it first view because the unique views didn't add up to the total views.

tompollard: Hmm, my thought is that for the "Views over time", the "Unique viewers" field shows the unique viewers for that month. For the total views, I think counting unique views over the entire period makes sense (perhaps "Unique registered viewers" is clearer than "Project Views by Unique Registered Users"?).

I think the important thing is that we are clear about what we are displaying (so we should consider some explanatory text somewhere).

lukepayyapilli: @tompollard Thanks for the feedback! I've updated the implementation:

- [x] Changed `first_views_by_month()` to `unique_viewers_by_month()` - now counts unique viewers per month (same user viewing in multiple months is counted once per month).
- [x] Updated the label to "Unique Registered Project Views" to be clearer about what's being measured.
- [x] The total count still shows unique viewers across the entire period.

The distinction is that the monthly breakdown may sum to more than the total if a user viewed in multiple months, which I think is the expected behavior you've outlined for understanding engagement over time.

tompollard (approved):
Thanks Luke!

PR #2558, opened 2026-01-16
Require name initials for DUA signing (#2106)

### LINKED ISSUE #2106: Agreements should be signed by entering names, rather than clicking "I agree"
Currently, signing a Data Use Agreement means clicking "I agree". We should make the process of signing a Data Use Agreement more significant by requiring the user to type their full name.

lukepayyapilli:
## Summary

   - [x] Users must now type their initials to sign a Data Use Agreement instead of just clicking "I agree".
   - [x] Initials are validated against the user's profile name (case-insensitive with unicode normalization).
   - [x] Provides stronger confirmation that users have read and agreed to the terms.

## Approach

   1. **Business logic in the model**: The initials validation rule lives in `DUASignature.validate_signature_initials()` as a static method. This keeps the business rule testable and reusable.

   2. **Unicode-aware comparison**: Uses `unicodedata.normalize('NFKD', text.casefold())` for proper unicode handling per review feedback.

   3. **Simple Form**: `DUASignatureForm` is a plain Django Form that only collects data. It delegates validation to the model's static method.

   4. **Explicit view logic**: The view explicitly creates the `DUASignature` object, making it clear what's being persisted.

## Screenshots
Screenshot 2026-01-16 at 1 42 36 PM

Closes #2106

### DIFF
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -1116,9 +1116,9 @@ class DUASignatureForm(forms.Form):
     full_name = forms.CharField(
         widget=forms.TextInput(attrs={
             'class': 'form-control',
-            'placeholder': 'Type your full name to sign',
+            'placeholder': 'Type your full name',
         }),
-        label='Full Name',
+        label='Please sign your full name to indicate your consent',
     )
 
     def __init__(self, user, *args, **kwargs):
--- a/physionet-django/project/forms.py
+++ b/physionet-django/project/forms.py
@@ -1111,27 +1111,28 @@ class AnonymousAccessLoginForm(forms.ModelForm):
 
 class DUASignatureForm(forms.Form):
     """
-    Form for signing Data Use Agreement by typing full name.
+    Form for signing Data Use Agreement by typing initials.
     """
-    full_name = forms.CharField(
+    initials = forms.CharField(
         widget=forms.TextInput(attrs={
             'class': 'form-control',
-            'placeholder': 'Type your full name',
+            'placeholder': 'Type your initials',
         }),
-        label='Please sign your full name to indicate your consent',
+        label='Please type your initials to indicate your consent',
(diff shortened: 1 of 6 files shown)


bemoody:
> Name must match the user's profile name (case-insensitive).

Oof.  If you want to open that can of worms, it's gotta be at least
```
unicodedata.normalize("NFKD", a.casefold()) == \
unicodedata.normalize("NFKD", b.casefold())
```
not merely `a.lower() == b.lower()`.


This is a good thing to read if you haven't: https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

Frankly, I'd scrap the whole idea, or accept *anything that vaguely approximates* the given name and family name fields that appear in the profile.

tompollard:
Relaxing the rules is reasonable, but requiring something beyond a thoughtless click is important. This doesn't solve all our problems, but it moves us in a good direction. 

One option is just to tell people what to type. e.g. Please type your initials below, exactly as they appear here: "LP".

If we **really** don't like initials, then an alternative option is to type a statement. e.g. "I agree"

lukepayyapilli:
Thats a good point @bemoody - thanks! I can do as @tompollard suggests and change it to a phrase but a phrase would imply the language is english. I think that's a fair assumption given the agreement is in english also but happy to modify to whatever phrase we'd like to. Let me know what you both prefer?

bemoody:
Hmm.  Initials might not be a bad idea.  Not just less to get wrong, but less annoyin for the person to type.

tompollard:
Adding some friction to the process isn't necessarily a bad thing. @lukepayyapilli Let's go with initials then!

lukepayyapilli:
ready @tompollard!

tompollard (approved):
Looks great Luke, thanks!

PR #2569, opened 2026-01-29
Add KHDP account unlink functionality

### LINKED ISSUE #2568: KHDP: Add KHDP account unlink functionality
We need to implement the KHDP account unlinking feature by simply adding an UI button on the http://localhost:8000/settings/khdp/ page.

Chrystinne:
This PR implements the possibility to unlink a KHDP account from the user's PhysioNet profile by adding an "Unlink KHDP Account" button on the settings page. 

The button is displayed conditionally when an account is already linked, mirroring the pattern used for ORCID account management.

### DIFF
--- a/physionet-django/user/templates/user/edit_khdp.html
+++ b/physionet-django/user/templates/user/edit_khdp.html
@@ -21,12 +21,12 @@
     </p>
 
     <ul>
-        <li>userId</li>
-        <li>userName</li>
-        <li>affiliation</li>
-        <li>mail</li>
-        <li>orcid</li>
-        <li>physionetId</li>
+        <li>User ID</li>
+        <li>Username</li>
+        <li>Affiliation</li>
+        <li>Email</li>
+        <li>ORCID</li>
+        <li>PhysioNet ID</li>
     </ul>
 
     <p>
--- a/physionet-django/user/templates/user/edit_khdp.html
+++ b/physionet-django/user/templates/user/edit_khdp.html
@@ -42,7 +42,7 @@
             <li>Affiliation: {{ request.user.khdp.affiliation }}</li>
             <li>Email: {{ request.user.khdp.email }}</li>
         </ul>
-        <button class="btn btn-primary btn-rsp" type="submit" name="remove_khdp" value="make_request">
+        <button class="btn btn-primary btn-rsp" type="submit" name="remove_khdp" value="make_request" onclick="return confirm('Are you sure you want to unlink your KHDP account?');">
             Unlink KHDP Account
         </button>
     {% else %}
--- a/physionet-django/user/templates/user/edit_khdp.html
+++ b/physionet-django/user/templates/user/edit_khdp.html
@@ -42,12 +42,14 @@
             <li>Affiliation: {{ request.user.khdp.affiliation }}</li>
             <li>Email: {{ request.user.khdp.email }}</li>
         </ul>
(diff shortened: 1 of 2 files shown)


physionet-django/user/views.py
```diff
@@ -1422,40 +1422,51 @@ def edit_khdp(request):
...
+            return redirect(final_url)
+
+        if request.POST.get('remove_khdp'):
+            try:
+                KhdpAccount.objects.get(user=request.user).delete()
+                messages.success(request, 'Your KHDP account has been unlinked.')
+            except KhdpAccount.DoesNotExist:
+                messages.error(
+                    request,
+                    'Object Does Not Exist Error: tried to unlink an object which does not exist.',
```
tompollard: This message reads like an error message for a developer. Perhaps better to give a user friendly message here (e.g. "No KHDP currently linked").

physionet-django/user/views.py
```diff
@@ -1422,40 +1422,51 @@ def edit_khdp(request):
...
+                )
```
tompollard, line 1469: I think you might want another return redirect here (consistent with previous if clause)?

physionet-django/user/templates/user/edit_khdp.html
tompollard: Not directly relevant to this PR, but these should probably be human readable fields rather than variable names. e.g. "User ID" instead of "userID"; "Email addresses" instead of "mail".

physionet-django/user/templates/user/edit_khdp.html
```diff
@@ -42,12 +42,13 @@ <h1>Edit KHDP</h1>
             <li>Affiliation: {{ request.user.khdp.affiliation }}</li>
             <li>Email: {{ request.user.khdp.email }}</li>
         </ul>
+        <button class="btn btn-primary btn-rsp" type="submit" name="remove_khdp" value="make_request">
+            Unlink KHDP Account
```
tompollard, line 46: Should we have an "Are you sure?" when unlinking?

tompollard (changes requested):
Thanks Chrystinne - added some comments inline.

physionet-django/user/templates/user/edit_khdp.html
Chrystinne: @tompollardI I followed what we currently have for the ORCID implementation, which doesn’t have a confirmation. But I’ve added one now because I agree it works better this way

Chrystinne:
> Thanks Chrystinne - added some comments inline.

Thanks for your comments, @tompollard. I've made the changes. That actually makes it more user-friendly.

physionet-django/user/views.py
```diff
@@ -1422,40 +1422,52 @@ def edit_khdp(request):
...
+            # KHDP expects appId and redirectUrl instead of standard OAuth2 params
+            params = {
+                'appId': client_id,
+                'redirectUrl': redirect_uri,
+                'nonce': nonce,
+            }
+            final_url = f"{khdp_auth_url}?{urlencode(params)}"
+            return redirect(final_url)
+
+        if request.POST.get('remove_khdp'):
```
tompollard: sorry, i should have picked this up before, but i think elif is better here. we only ever want request_khdp or remove_khdp so coupling them makes sense.

Chrystinne: @tompollard I initially did it the way you suggested, but later changed it to match the orcid implementation. No worries, I’ll make that change

tompollard (approved):
Thanks @Chrystinne, looks good!

PR #2574, opened 2026-02-04
Add "On Hold" status to editorial workflow

tompollard:
This pull request add an `is_on_hold` flag to ActiveProject that allows editors to temporarily park submissions without changing their workflow stage (requested by @lilehman):

- Projects placed on hold are hidden from active tabs and shown in a dedicated "On Hold" tab on both the submitted projects and editor home pages
- The submission info page displays the current status including an "On Hold" badge when applicable
- Permission checks ensure only the assigned editor or users with can_assign_editor permission can toggle hold status

Notes for testing the new functionality:

- Place a project on hold from each tab (assignment, decision, revision, copyedit, approval, publish) in the editorial workflow and verify it moves to the "On Hold" tab
- Reactivate a project from the "On Hold" tab and verify it returns to the correct stage tab
- Verify the "On Hold" badge appears on the submission info page for held projects
- Verify the editor home page also shows/hides on-hold projects correctly
- Confirm that non-editors cannot place projects on hold
- Verify the "Assign Editor" form still works correctly on the submitted projects page


bemoody:
Not great to implement the form submission in multiple separate places (submitted_projects() and editor_home().)  Instead, how about:
```
<form action="{% url 'park_project' project.slug %}" method="POST">
  {% csrf_token %}
  <input type="hidden" name="redirect" value="{{ request.get_full_path }}">
  <button type="submit" name="park_project">Place on Hold</button>
</form>
```

Please use consistent terminology; don't use "activate" as an antonym for "park".  "park" and "unpark" would be fine terms to use internally if you don't want to use "on-hold" and "not-on-hold".  "activate" is I think already badly overloaded and should be avoided.

bemoody:
Although maybe if "park" is the verb used in the console then "parked" should be the name of the attribute in ActiveProject.

tompollard:
Thanks @bemoody, good points! I think it would be easier to switch all of the "park" terminology to "on hold". Are you okay with that?

bemoody:
> I think it would be easier to switch all of the "park" terminology to "on hold".

Sure, that sounds good.

tompollard:
Terminology should be fixed now. I added a single view for handling the on-hold switching, and moved to the two step migration.

bemoody (approved):
This looks good.  An improvement would be to disable or hide the button if the user does not have permission (not a managing editor and not assigned editor of this project.)

tompollard:
> An improvement would be to disable or hide the button if the user does not have permission (not a managing editor and not assigned editor of this project.)

Thanks @bemoody, I'll implement this fix (tracked in https://github.com/MIT-LCP/physionet-build/issues/2587).

PR #2578, opened 2026-02-09
Rename project.modelcomponents.fields to project.fields

bemoody:
The SafeHTMLField class was originally defined in project/models.py, later moved to project/modelcomponents/fields.py.

Nowadays, however, this field is used in a number of places outside the project app - it shouldn't be considered internal to project.models anymore.  Following Django API conventions, it should be defined in a module called "project.fields".

Moving this out of the project.modelcomponents hierarchy is one step towards fixing the convoluted import dependencies (and the associated flake8/pyflakes warnings.)


tompollard (approved):
Looks good, thanks

PR #2581, opened 2026-02-10
Add pagination and reverse ordering to metrics views over time

### LINKED ISSUE #2561: Metrics UI: Add pagination and reverse ordering for views over time
## Context

Related to #2336 (Metrics for downloads). The current metrics page displays views over time in a simple table showing all months since tracking began. For projects with long histories, this table can become unwieldy.

## Problem

1. **No pagination**: All monthly data is displayed at once, which can be overwhelming for datasets with years of history
2. **Chronological ordering only**: Data is shown oldest-first, but users typically want to see recent activity first

### Pagination
- Add pagination to the "Views Over Time" table with a reasonable page size (e.g., 12 months per page)
- Use Django's built-in `Paginator` class for server-side pagination
- Add page navigation controls below the table

### Reverse Ordering
- Default to newest-first (reverse chronological) ordering
- Optionally add a toggle to switch between newest-first and oldest-first

### Implementation Details
- Modify `views_over_time()` method or the view to accept ordering parameter
- Add pagination parameters to the view context
- Update the template with pagination controls

## Alternatives Considered

1. **Client-side pagination with JavaScript**: Could use DataTables or similar library for sorting/pagination, but adds JS dependency and may not work well with large datasets
2. **Infinite scroll**: More modern UX but more complex to implement and may not fit the existing UI patterns
3. **Date range filter**: Allow users to select a specific date range instead of pagination - could be added as a future enhancement

## Acceptance Criteria

- [ ] Views over time table is paginated (default 12 months per page)
- [ ] Data displays newest-first by default
- [ ] Pagination controls are intuitive and accessible
- [ ] Works correctly for projects with no view data

lukepayyapilli:
## Summary
- [x] Display newest months first and paginate at 12 items per page on the project metrics page.
- [x] Update shared pagination template (`console/pagination.html`) to use Bootstrap 4 `page-item`/`page-link` classes for proper styling across all 29 paginated views.
- [x] Add 15 months of demo AccessLog fixture data so pagination is visible out of the box.

## Approach
- **Ordering in the view, not the model.** The queryset method `unique_viewers_by_month()` returns data in chronological order which is correct for the data layer. Newest-first is a presentation concern, so we reverse in the view to keep the model reusable for other contexts (charts, exports, API).
- **Reuse existing pagination infrastructure.** The codebase already has `paginate()` in `physionet/utility.py` and `console/pagination.html`, so no new utilities were needed.
- **12 months per page.** One year of data per page is a natural unit and matches the suggestion in #2561.
- **Bootstrap 4 pagination classes.** The shared template was using Bootstrap 3-style markup (bare `<li>` and `<a>` tags). Adding `page-item` and `page-link` classes gives proper spacing, borders, and active/disabled states site-wide.

## Screenshot
Screenshot 2026-02-10 at 2 17 31 PM

Closes #2561

### DIFF
--- a/physionet-django/console/templates/console/pagination.html
+++ b/physionet-django/console/templates/console/pagination.html
@@ -1,29 +1,29 @@
 {% if pagination.has_other_pages %}
-  <div style="text-align: center;">
+  <div class="d-flex justify-content-center">
     <ul class="pagination">
-      <li><a href="?{{ querystring }}&page=1">&laquo;</a></li>
+      <li class="page-item"><a class="page-link" href="?{{ querystring }}&page=1">&laquo;</a></li>
       {% if pagination.has_previous %}
-        <li><a href="?{{ querystring }}&page={{ pagination.previous_page_number }}">&lsaquo;</a></li>
+        <li class="page-item"><a class="page-link" href="?{{ querystring }}&page={{ pagination.previous_page_number }}">&lsaquo;</a></li>
       {% else %}
-        <li class="disabled"><span>&lsaquo;</span></li>
+        <li class="page-item disabled"><span class="page-link">&lsaquo;</span></li>
       {% endif %}
       {% for i in i|ljust:10 %}
         {% with num=forloop.counter|add:pagination.number|add:"-4" %}
         {% if num > 0 and num <= pagination.paginator.num_pages %}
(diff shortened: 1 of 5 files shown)


tompollard (approved):
thanks luke, looks good to me

PR #2588, opened 2026-02-20
Add georestricted user information to admin console

briangow:
This PR add a method for checking if a user is registered with an IP from a georestricted region.

It also uses that method to display whether the user will be restricted (for applicable datasets) on the `/user/manage` page in the admin console. In the new `Georestriction` section there is a field called `Restricted` which will display `True`, `False`, or `Unknown (no registration IP)`.


tompollard (approved):
Thanks Brian, looks good to me.

Issue #66, opened 2018-01-01
User Notifications


cx1111:
I believe user notifications are a necessary feature to implement. Some uses of a notifications box are for:
- Project review status update
- Being invited to be a project collaborator
- Having storage requests approved/denied

If we're all happy with this idea, I will start a new branch to implement notifications.

superryeti:
Do you think something like this will be a good idea?

1. Create a new model on the existing app notification(for example)
```
class Notification(models.Model):
    is_read = models.BooleanField(default=False)
    user = models.ForeignKey(User, on_delete=models.CASCADE, related_name='notifications')
    subject = models.CharField(max_length=255)
    message = models.TextField()
    notification_date = models.DateTimeField(auto_now_add=True)
    redirect_url = models.URLField(blank=True, null=True)

    def __str__(self):
        return self.subject
```
3. Add a notification icon on the navbar that shows the number of unread notifications.


![image](https://user-images.githubusercontent.com/24412619/232115424-2585aba4-2e96-44d0-9079-86388198f3c0.png)

4.When the user clicks on the icon, display a modal with the latest notifications, sorted by date.

5. On the modal, add a link to view all notifications, which takes the user to a page where they can see their notification history, mark notifications as read, etc.



This way, we can send any type of notification and use the redirect_url field to direct the user to the relevant page if needed(like sending them to project page). or we could also add the urls in the message or the notification body

superryeti:
@tompollard  forgot to tag you earlier.

alistairewj:
Our navbar is pretty cluttered. Maybe a notification icon on the right side to save space? We do have an implicit separate of general pages on the left (about, share) and user-specific pages on the right (login, settings, etc).

PR #2589, opened 2026-02-23
Add in-app notifications.

tompollard:
As discussed in #66, it would be helpful for users to have a notification panel to highlight updates to projects, outstanding tasks, etc. This pull request adds notifications for project updates.

## Summary                                               
                                                                                                                                                                        
  - Adds a notification system that complements the existing email notifications                                                                     
  - When key actions occur (author invitations, editor decisions, project publications, etc.), a notification is now created alongside the email
  - A bell icon in the navbar shows the unread count and links to a paginated notifications inbox
  - A cron job cleans up read notifications older than 90 days

## Details

- Adds a `Notification` Model: Includes recipient (person receiving the notification), actor (person triggering the notification), type, message, URL, read status. `NotificationType` enum covers 12 event types.
- Hooks into existing email paths: 
  - `create_notification()` is called after send_mail in 7 functions: `invitation_notify`, `invitation_response_notify`, `submit_notify`, `assign_editor_notify`, `edit_decision_notify`, `copyedit_complete_notify`, and `publish_notify`. 
  - There are probably other actions that should trigger a notification, but these are a start.

There are no new dependencies and email notifications continue unchanged.

## To test

  - Log in, trigger an invitation or status change, verify bell icon shows count
  - Click a notification. Verify it marks as read and redirects
  - Click the checkmark button on a URL-less notification. Verify it marks as read
  - Click "Mark all as read". Verify all notifications clear


physionet-django/notification/utility.py
```diff
@@ -237,6 +285,14 @@ def submit_notify(project):
         send_mail(subject, body, settings.DEFAULT_FROM_EMAIL,
                   [email], fail_silently=False)
 
+    for author in project.authors.all():
+        create_notification(
+            recipient=author.user,
+            notification_type=NotificationType.PROJECT_SUBMISSION,
+            message='Project submitted: {}'.format(project.title),
+            url=reverse('project_home'),
+        )
```
briangow, line 297: In a number of cases `actor` is not being set but is isn't clear to me why. Are "system" notifications being set without `actor`. If so, what constitutes a "system" notification?

physionet-django/physionet/cron.py
```diff
@@ -71,3 +72,27 @@ def do(self):
...
                 invitation.project.title, invitation.inviter.email,
                 invitation.email))
+
+
+READ_NOTIFICATION_DAY_LIMIT = 90
+
+
+class CleanOldNotifications(CronJobBase):
+    RUN_EVERY_MINS = 60
+    RETRY_AFTER_FAILURE_MINS = 10
```
briangow, line 82: Should we consider moving these hard-coded variable to a more central place such as https://github.com/MIT-LCP/physionet-build/blob/dev/physionet-django/physionet/settings/settings.py ?

briangow (commented):
@tompollard , this looks good to me! I looked over the code and did the tests you suggested - the notifications all seem to generate a URL though, so I didn't do "Click the checkmark button on a URL-less notification."

briangow:
Also, during testing I found that when inviting an author to a project that the invited author could not successfully use the "Submit Response" button in the Respond to invitation pop-up - the button is unresponsive.

tompollard:
> Also, during testing I found that when inviting an author to a project that the invited author could not successfully use the "Submit Response" button in the Respond to invitation pop-up - the button is unresponsive.

Good catch! This should be fixed now. It looks like the `<form>` field was being closed prematurely (i.e. before `</form>` was reached).

physionet-django/physionet/cron.py
tompollard: Yes, this might be worthwhile. If it is okay, I will do this in a later PR (and raise an issue for the meantime).

physionet-django/notification/utility.py
tompollard: Good point, the difference wasn't clear. I have now set an actor for all notifications. The new ones are:

- Invitation response: Actor is the author who responds to the invitation
- Project submission: Actor is the submitting author
- Project published: Actor is the editor who clicks "Publish"

briangow: Looks good, thanks!

briangow (approved):
Yes, the bug seems to be fixed and everything else looks good to me.

PR #2592, opened 2026-03-04
Log URLs and responses for AWS verification attempts

bemoody:
Sometimes people are not able to complete AWS verification and it's difficult for us to tell why it didn't work.

To potentially aid administrators in troubleshooting, log these URLs and responses (in physionet_uwsgi.log).  I don't expect this will give us a lot of useful information, but it should give us something: for example, we can see whether the error was "SignatureDoesNotMatch" or "AccessDenied".

### DIFF
--- a/physionet-django/user/awsverification.py
+++ b/physionet-django/user/awsverification.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import re
 import urllib.parse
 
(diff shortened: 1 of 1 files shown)


tompollard (approved):
looks good to me, thanks.

Chrystinne:
That'd be good, @bemoody. It's currently a bit difficult to identify errors and provide support for the identity verification process/AccessDenied error. We had a user contacting us via email yesterday saying he's getting an AccessDenied error when trying to download a dataset from AWS. He mentioned that he changed his AWS info on PhysioNet and redid the identity verification. I needed to ask for more details so we can check whether he had any error during the identity verification and also whether he signed the DUA before or after the verification process.

PR #2597, opened 2026-03-12
Remove csrfmiddlewaretoken where it is inappropriate

bemoody:
The /content/ page and a few other pages currently contain an `<input type="hidden" name="csrfmiddlewaretoken">` in a GET form.

This is dangerous for various reasons (CSRF tokens should *not* be transmitted in URLs), as well as being useless (GET forms do not use CSRF tokens) and making the URLs ugly.

I think at some point somebody thought it would be convenient to embed the Django csrf_token tag in the "form snippet" templates so that you wouldn't have to include it in the parent template.  But in addition to being dangerous for GET forms, this wasn't consistently used, and leads to confusion and redundancy.

Better to keep the `{% csrf_token %}` together with the `<form>` tag, so it is clear that there is one token for each POST form and vice versa.

### DIFF
--- a/physionet-django/console/templates/console/code_of_conduct_detail.html
+++ b/physionet-django/console/templates/console/code_of_conduct_detail.html
@@ -11,6 +11,7 @@
     </div>
     <div class="card-body">
         <form method="post" action="{% url 'code_of_conduct_detail' code_of_conduct.pk %}">
+            {% csrf_token %}
             {% include "inline_form_snippet.html" with form=code_of_conduct_form %}
           <button class="btn btn-primary btn-fixed" type="submit">Save</button>
         </form>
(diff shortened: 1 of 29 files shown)


bemoody:
Searching for bad forms
```
find . -name \*.html | while read filename; do F=$filename perl -ne 'if(/<form/){$form="\n\033[32;1m$ENV{F}\033[m\n"; } if(/"inline_form_snippet/){print "$form$_"; $form="";} if(/<\/form/){$form="";} s/(method="*post"*)/\033[31;1m\1\033[m/ig; s/(\{% *csrf_token *%\})/\033[35;1m\1\033[m/; if($form ne ""){$form .= $_}' <$filename; done | less -R
```

tompollard (approved):
looks good to me!

PR #2603, opened 2026-03-27
Save submission slug for published projects

bemoody:
Active projects have a (supposedly) unique slug that is used to identify them.  In official emails this is referred to as the "Submission ID".

Currently, this is forgotten after the project is published: there's no way to locate the project formerly known as `/projects/AkMgLEcVaSpJZH7oFQmp/`.  This information should be preserved.

For historical projects, the submission slugs can mostly be extracted from the background_task data.


tompollard (approved):
Good catch!

bemoody:
Worth noting that this change broke running background-tasks.

The use of `save()` in set_storage_info and in make_zip is problematic.  I knew that was a problem, but didn't consider how it would break in this case.

Interestingly, I think pull #2546 managed to avoid this problem because it created the UUIDs as an early migration.  In contrast, pull #2546 would have broken if somebody tried to publish a project while the deployment was happening.  (So although it worked out fine for PublishedProject, it wouldn't be a good strategy for "hot" tables.)

It would probably be wiser for the deployment scripts to fully *stop* background-tasks before running any migrations, and then *restart* after running all migrations.

Issue #663, opened 2019-08-07
Search is duplicating content


lbulgarelli:
The search is currently duplicating content if a keyword is contained more than once on one of the fields we use for filtering.

A simple `distinct()` would solve the problem, but it is returning an error for some reason.

I don't know when this started happening but it does not seem to be affecting production somehow.

bemoody:
"challenge" returns demobsn once

"eicu" returns demoeicu three times

"database" returns demoeicu three times and demobsn once

So is there something strange about the demoeicu project?

elfeto:
This seems fixed. Maybe #779 ?

bemoody:
No, it's not fixed.  For me, http://localhost:8000/content/?topic=eicu
returns the same result three times.

elfeto:
Maybe is database related, for example https://physionet.org/content/?topic=eicu returns only one.

PR #2606, opened 2026-04-02
Fix search returning multiple results

alistairewj:
This PR fixes the current search returning more than one result for the same object - raised in #663 

Two fixes:

1. Deduplicate federated projects if they also occur locally; I thought this was the issue originally but it wasn't. Still I think it's correct to not include federated projects if they are local to the current platform (not that we have any).
2. The issue was the search vector doing an M2M join with the topic. Fixed it by using a separate vector for ranking. Details in [081faaa4](https://github.com/MIT-LCP/physionet-build/commit/081faaa474f37d93f256c620b529ac5807ca5c24).

I also added a fixture for the federated project to demonstrate that fix.

### DIFF
--- a/physionet-django/search/views.py
+++ b/physionet-django/search/views.py
@@ -1,5 +1,4 @@
 import operator
-import pdb
 import re
 from functools import reduce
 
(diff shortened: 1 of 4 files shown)


physionet-django/search/views.py
```diff
@@ -120,15 +119,16 @@ def get_content_postgres_full_text_search(resource_type, orderby, direction, sea
...
-    # Filter projects by latest version and annotate relevance field
-    published_projects = PublishedProject.objects.annotate(search=vector).filter(query, is_latest_version=True)
+    # Create a vector without the topics to avoid row multiplication from the M2M join when ranking
+    rank_vector = SearchVector('title', weight='A') + SearchVector('abstract', weight='B')
+    published_projects = PublishedProject.objects.annotate(search=match_vector).filter(query, is_latest_version=True)
 
-    # get distinct projects with subquery and also include relevance from published_projects
+    # Get distinct projects including relevance annotation
     published_projects = PublishedProject.objects.filter(id__in=published_projects.values('id')).annotate(
-        relevance=SearchRank(vector, search_query)).distinct()
```
tompollard, line 131: I remember you saying you had to be suspicious of anyone using "distinct" in their queries, so happy to see it being removed here.

tompollard (approved):

PR #2608, opened 2026-04-08
Do not assume background tasks are dead after one hour

bemoody:
The django-background-tasks daemon allows a task to be started if it is not currently "locked", or if it was "locked" more than MAX_RUN_TIME seconds ago (presumably as a safety valve to handle the daemon being killed.)

Since MAX_RUN_TIME is 3600 by default, this means that a new copy of every task is getting spawned once per hour (assuming there are threads available - this was not a problem prior to #699 because there was only one thread running.)

We need to make MAX_RUN_TIME much larger to avoid spawning multiple instances of the same task.

But we also don't want to wait for a long time to restart tasks when the daemon is restarted.

So, instead, let's make the assumption that there is only one django-background-tasks service, and therefore if we are restarting the service then we can assume any existing locks are stale.

### DIFF
--- a/deploy/production/etc/systemd/system/django-background-tasks.service
+++ b/deploy/production/etc/systemd/system/django-background-tasks.service
@@ -4,6 +4,7 @@ After=emperor.uwsgi.service
 
 [Service]
 Environment=DJANGO_SETTINGS_MODULE=physionet.settings.production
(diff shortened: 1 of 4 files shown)


bemoody:
I haven't tested the systemd side of things, but this issue is a bit urgent so we should deploy the MAX_RUN_TIME fix as soon as possible.

tompollard (approved):

PR #2609, opened 2026-04-09
Refactor access.py into separate modules

bemoody:
The `project.modelcomponents` module is divided into a dozen parts with no rhyme or reason.  I've gotten tired of not knowing where to find the thing I'm looking for.

So I'm intending to move each model class into a file of its own, where the name of the module is the same as the lowercase name of the class.  Modules that don't correspond to a model class will go away.

I'll do this in pieces, because it's a large task, and try to avoid conflicting with stuff other people are working on.  This pull request splits up `project.modelcomponents.access`, which is alphabetically first and also possibly the most in need of cleanup.


tompollard:
Not convinced this will make things easier to find but also I trust you to make good decisions Benjamin. 

@Chrystinne please could you review? It would be good to review/merge quickly so Benjamin can keep momentum on the refactor.

Chrystinne:
@tompollard I'm reviewing this now

Chrystinne (approved):
@bemoody Looks good!

PR #2611, opened 2026-04-10
Generate SHA256SUMS.txt before publication

### LINKED ISSUE #519: Generating SHA256SUMS.txt

### LINKED ISSUE #2591: Make SHA256 available before publishing
In a recent upload for [Bridge2AI-Voice](https://physionet.org/content/b2ai-voice/3.0.0/) we found that some of our SHA256 values were incorrect. While it is possible this happened in between downloading locally and then moving them to Physionet, we are pretty certain that the error happened in our upload to Physionet. The only reason we found this was that the bits that were malformed are related to the header for the file and so the file can't load properly. Generally, it would be quite useful to have the SHA256 available prior to publishing the dataset or have some way to verify that the uploaded files are intact and correct (we're currently planning to make part of our procedure uploading the files and then download them from staging and check the hashes).

bemoody:
PhysioNet creates a `SHA256SUMS.txt` file for every published project, containing the hashes of all other files.

Currently, this file is generated after the project is officially published, which means:
- authors and editors cannot see the hashes before publication, to verify that everything was uploaded correctly
- visitors who try to download the project immediately after publication may not receive the file at all, or may receive an incomplete copy

Instead, this file should be generated after the project is copyedited and before it is published.

It is also useful to calculate the project's `main_storage_size` at the same time.  This means that the "Total uncompressed size" will be displayed as soon as the project is published.

This change does not require SHA256SUMS.txt to be generated before authors can approve the project.  Authors will still be notified and asked to approve as soon as the editor clicks "complete copyedit".  However, the editor will see that even if all authors have approved, they won't be able to click "publish" until the background task has finished.

For existing projects that have been copyedited, the editor will need to click a button "Update checksums", which only appears when the project is otherwise ready for publication.

For existing projects that have been published (but the background task hasn't completed), there are no changes.  The old background task will still run.

Fixes #2591
Fixes #519


tompollard (approved):
Very cool, looks good to me!

Issue #2620, opened 2026-05-13
Improve editorial, credentialing, and dataset metrics for PhysioNet impact showcase


Chrystinne:
@lilehman and I are working on milestone 1.7 "Showcase Impact of PhysioNet" and she has made some suggestions on how to improve our current editorial, credentialing, and dataset metrics:

**Metrics - Credentialing**

- [x] Review how the metrics are calculated;
- [x] Decision time (days, median) for credentialing: Why are Reference response times (days, median) zero for all years?

**Metrics - Editorial**

- [x] Add Rejected column in addition to (Created, Submitted, Resubmitted, Published);
- [ ] Add Reviewed column;
- [ ] Tracking over multiple years (instead of only the past 18 months)tracking over multiple years (instead of only the past 18 months)
- [x] In addition to time to publication, create two new fields (time to review/first feedback, time to decision);
- [x] Add counts per year;

**Metrics - Dataset Views**

- [x] Export a CSV file with Unique Registered Project Views (Current Version and All Versions), including project slug, URL, title, and project type (dataset, model, challenge, etc.)

**Metrics - Citations**

- [x] Use Dimensions.ai to extract citation-related data

**Metrics - Operational Efficiency**

- [ ] Extract operational efficiency measures (e.g., time to complete credentialing and time to review projects)
- [ ] Extract operational efficiency measures related to project rejection rates, excluding projects currently on hold

This mainly affects the metrics pages. I've created separate small PRs to address each change.

PR #2625, opened 2026-05-14
Add project view counts to Export data page

Chrystinne:
Part of #2620

Adds a new button on the Export data page under "Project view counts" to download unique registered project view counts per published project (current version and all versions), including the following fields:

- title;
- project slug;
- resource type;
- DOI URL;
- current version;
- unique view count for current version;
- unique view count across all versions.

---
CSV exported

Example output using local demo data:

image

### DIFF
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -2663,6 +2663,40 @@ def download_projects(request):
 
 
 @console_permission_required('user.change_credentialapplication')
+def download_projects_view_counts(request):
+    """
+    Delivers a CSV file containing unique registered project views per published project.
+    """
+    response = HttpResponse(content_type='text/csv')
+    response['Content-Disposition'] = 'attachment; filename="projects_views_counts.csv"'
+
+    writer = csv.writer(response, quoting=csv.QUOTE_ALL)
+    writer.writerow(["title",
+                     "project_slug",
+                     "resource_type",
+                     "DOI (latest version)",
+                     "current_version",
+                     "count_current_version",
+                     "count_all_versions",
+                     ])
+
+    projects = PublishedProject.objects.filter(is_latest_version=True)
+
+    for project in projects:
+        project_data = [project.title,
+                        project.slug,
+                        project.resource_type.name,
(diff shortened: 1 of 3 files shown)


physionet-django/console/views.py
```diff
@@ -2662,6 +2662,40 @@ def clean_html(html_content):
...
+    projects = PublishedProject.objects.filter(is_latest_version=True)
+
+    for project in projects:
+        project_data = [project.title,
+                        project.slug,
+                        project.resource_type.name,
+                        f"https://doi.org/{project.core_project.doi}" if project.core_project.doi else None,
+                        project.version,
+                        project.view_count(),
+                        project.view_count(all_versions=True),
```
tompollard, line 2692: Pre-existing bug but this will double count people across versions.

physionet-django/console/views.py
```diff
@@ -2662,6 +2662,40 @@ def clean_html(html_content):
     return response
 
 
+@console_permission_required('user.change_credentialapplication')
```
tompollard, line 2665: also pre-existing issue, but this is a slightly odd permission for us to be using.

physionet-django/console/views.py
```diff
@@ -2662,6 +2662,40 @@ def clean_html(html_content):
...
+
+    projects = PublishedProject.objects.filter(is_latest_version=True)
+
+    for project in projects:
+        project_data = [project.title,
+                        project.slug,
+                        project.resource_type.name,
+                        f"https://doi.org/{project.core_project.doi}" if project.core_project.doi else None,
+                        project.version,
+                        project.view_count(),
```
tompollard, line 2691: for each project in the loop (~1000?) we call `project.view_count()` + `project.view_count(all_versions=True)` (same query for each version of the project), so for ~1000 latest version projects we are making ~3 * 1000 queries per download. probably already an existing issue for other exports, but we should optimise at some point.

tompollard (approved):
i'm okay if this is merged, but i've noted some issues in the review. either way, we should do a refactor/rethink of the metrics pages as a whole.

PR #2621, opened 2026-05-13
Add rejection metrics to submission statistics page

Chrystinne:
Part of #2620 

This PR adds metrics for rejected projects to the submission statistics page. 

It tracks project rejection dates based on `EditLog.decision_datetime` when `decision == 0` and aggregates the counts alongside the existing project creation, submission, resubmission, and publication metrics.

Screen Shot 2026-05-13 at 11 25 41 AM

PR #2626, opened 2026-05-14
Add yearly project counts to submission statistics

Chrystinne:
> **Note:** This PR should be merged after #2621 (Add rejection metrics to submission statistics page), as it depends on the Rejected field added there.

Part of #2620

Adds a "Projects (past 6 years)" table to the submission statistics page, showing yearly counts for Created, Submitted, Resubmitted, Published, and Rejected projects.

Screen Shot 2026-05-14 at 12 07 35 PM

### DIFF
--- a/physionet-django/console/templates/console/submission_stats.html
+++ b/physionet-django/console/templates/console/submission_stats.html
@@ -39,7 +39,7 @@
     </div>
   </div>
   <div class="card-header">
-    New projects (past 6 years)
+    Projects (past 6 years)
   </div>
   <div class="card-body">
     <div class="table-responsive">
--- a/physionet-django/console/templates/console/submission_stats.html
+++ b/physionet-django/console/templates/console/submission_stats.html
@@ -38,5 +38,34 @@
 
     </div>
   </div>
+  <div class="card-header">
+    New projects (past 6 years)
+  </div>
+  <div class="card-body">
+    <div class="table-responsive">
+
+      <table class="table table-bordered">
+        <tr>
+          <th>Year</th>
+          <th>Created</th>
+          <th>Submitted</th>
+          <th>Resubmitted</th>
+          <th>Published</th>
+          <th>Rejected</th>
+        </tr>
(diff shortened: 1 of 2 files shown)


physionet-django/console/views.py
```diff
@@ -2445,8 +2445,42 @@ def submission_stats(request):
             except AttributeError:
                 pass
 
+    # Aggregate counts per year (last 6 years)
```
tompollard, line 2448: this looks like mostly a copy of the previous loop? for this pr, can you modify the previous loop rather than copying again? 

we should rethink and refactor all logic at some point.

physionet-django/console/templates/console/submission_stats.html
```diff
@@ -38,5 +38,34 @@
...
+          <th>Year</th>
+          <th>Created</th>
+          <th>Submitted</th>
+          <th>Resubmitted</th>
+          <th>Published</th>
+          <th>Rejected</th>
+        </tr>
+        {% for year, val in yearly_stats.items %}
+          <tr>
+            <td>{{ year }}</td>
```
tompollard, line 58: an issue with existing code, but we should switch to a clearer way of indexing the information at some point. unclear from looking at the code here what val.0, val.1 ec represents.

physionet-django/console/views.py
```diff
@@ -2445,8 +2445,42 @@ def submission_stats(request):
...
+                yearly_stats[create_yr][0] += 1
+
+            for log in project.edit_log_history():
+                sub_yr = log.submission_datetime.year
+                if sub_yr in yearly_stats:
+                    if log.is_resubmission:
+                        yearly_stats[sub_yr][2] += 1
+                    else:
+                        yearly_stats[sub_yr][1] += 1
+                try:
```
tompollard, line 2467: what is is try/except trying to catch? not sure it is needed here.

tompollard (changes requested):
Please could you try to merge the loops? The other issues are pre-existing and we can refactor out later.

Issue #2348, opened 2025-02-18
Administrator controls for cloud mirrors


bemoody:
Project cloud mirrors (GCP and AWS buckets) are created via the published project management page (https://physionet.org/console/published-projects/mitdb/1.0.0/)

Some **additional controls** that should be available to storage administrators:

- Buttons to *temporarily enable/disable access* to the cloud mirror.
- Buttons to *temporarily show/hide* the cloud mirror from the published project page.
- A button to *disable all access and delete* the cloud mirror.

These will require adding fields to the GCP and AWS models to indicate whether the mirror is "visible" and/or "accessible".

We will also need to implement:
- Disabling access to public GCP buckets: probably this just means toggling the bucket to "private".
- Disabling access to controlled GCP buckets: probably we want to *leave the fake email list as-is*, but remove its access privileges.
- Disabling access to AWS prefixes: probably we want to add a "deny" policy rule to the bucket.  Note that this means there is an upper limit on the number of disabled projects.

Chrystinne:
Thanks, @bemoody. We really need to add those controls to handle enabling/disabling/deletion of cloud mirrors through the platform. I'm taking care of some of the functionalities you listed here:

- Buttons to temporarily enable/disable access to the cloud mirror (AWS).
- Buttons to temporarily show/hide the cloud mirror from the published project page (AWS).
- A button to disable all access and delete the cloud mirror (AWS).

PR #2627, opened 2026-05-15
Add support for deleting project files and access points from S3

Chrystinne:
Part of #2348

This PR adds support for deleting project files from AWS S3 through the console. It includes deletion of the zip file and, for private projects, the associated access points.

A "Delete files" button has been added to the AWS cloud management section of the manage published project page, visible only when `sent_files` is `True`.

For public projects, the function removes all files and the zip file from S3 and deletes the AWS record from the database. For private projects, it also deletes all associated access points from S3 and removes the AWS, AWSAccessPoint, and AWSAccessPointUser records via cascade.

>Note: Despite the branch name, this PR covers both public and private projects.

>Note: the "Delete project files" button has been renamed to "Delete files" in #2628 for consistency with the other buttons.

### DIFF
--- a/physionet-django/console/templates/console/manage_published_project.html
+++ b/physionet-django/console/templates/console/manage_published_project.html
@@ -447,6 +447,17 @@
                   Upload files
                 {% endif %}
               </button>
+              {% if project.aws.sent_files %}
+                <button class="btn btn-primary"
+                        name="aws-delete-project-files" value="{{ project.slug }}"
+                        {% if aws_upload_pending or rw_tasks %}
+                          disabled="disabled"
+                        {% endif %}
+                        type="submit">
+                    <span class="fa fa-trash"></span>
+                    Delete project files
+                </button>
+              {% endif %}
             </p>
           </form>
         {% endif %}
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -1197,7 +1197,7 @@ def manage_published_project(request, project_slug, version):
                 messages.error(request, 'Project has tasks pending.')
             else:
                 delete_project_files_from_s3(project)
-                messages.success(request, 'The project files have been deleted from S3.')
+                messages.success(request, 'The project files are being deleted from S3')
                 # Redirect is required after deletion to avoid rendering the page with
                 # a stale AWS instance that no longer has a primary key in the database.
                 return redirect('manage_published_project', project_slug=project_slug, version=version)
--- a/physionet-django/console/views.py
+++ b/physionet-django/console/views.py
@@ -87,6 +87,7 @@ from project.cloud.s3 import (
     get_bucket_name,
     check_s3_bucket_exists,
     has_s3_credentials,
+    delete_project_files_from_s3,
 )
 
 from django.core.management import call_command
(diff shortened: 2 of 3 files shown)


tompollard (approved):

tompollard:
Looks good to me, thanks!

bemoody (changes requested):
## What does this do?

```
bucket.objects.filter(Prefix=prefix).delete()
```

Ultimately, we may have millions of individual files/objects to delete.  Does this line perform a *single API request* to S3 that deletes every file at once?  Or as I suspect, does this require making millions of requests to delete the files one at a time?

Even if you can delete multiple files with one request, it might have limits (maximum of 1000 files?) or it might take a very long time to complete.  So I strongly suspect that this needs to be implemented as a background task.

## Order of operations

Access points should be deleted from S3 before starting to delete files.  From a security standpoint: access must be disabled immediately, not hours later.  From a data integrity standpoint: people should not see that their `aws s3 sync` command succeeded if they received an incomplete copy.

Before starting to delete files, the AWS object should also have `sent_files` set to false.

I think we probably don't want to delete the AWS object until after the files are deleted, but I'm not sure about that.

Chrystinne:
> ## What does this do?
> ```
> bucket.objects.filter(Prefix=prefix).delete()
> ```
> 
> Ultimately, we may have millions of individual files/objects to delete. Does this line perform a _single API request_ to S3 that deletes every file at once? Or as I suspect, does this require making millions of requests to delete the files one at a time?
> 
> Even if you can delete multiple files with one request, it might have limits (maximum of 1000 files?) or it might take a very long time to complete. So I strongly suspect that this needs to be implemented as a background task.
> 
> ## Order of operations
> Access points should be deleted from S3 before starting to delete files. From a security standpoint: access must be disabled immediately, not hours later. From a data integrity standpoint: people should not see that their `aws s3 sync` command succeeded if they received an incomplete copy.
> 
> Before starting to delete files, the AWS object should also have `sent_files` set to false.
> 
> I think we probably don't want to delete the AWS object until after the files are deleted, but I'm not sure about that.

Thanks for your comments, @bemoody. These are all good points.

Most importantly here: this method does not delete the files/objects one at a time. The operation deletes 1,000 objects in a single batch (and uses only two API requests per batch, one for listing key objects and another for deleting them). 
(boto3 documentation: https://docs.aws.amazon.com/boto3/latest/reference/services/s3/bucket/objects.html)

So when we call `bucket.objects.filter(Prefix=prefix).delete()`, what boto3 does is the following (pagination):

- List the first 1,000 objects under the prefix (1 listing request);
- Delete those 1,000 objects in a single batch (1 deletion request);
- List the next 1,000 objects;
- Delete those 1,000 objects;
- And so on until all objects/files are deleted.

So, for example, for a project with 50,000 files, this would result in 100 requests total.

Although this can be unpredictable, this operation should be executed very rarely when authors spot some problems in the dataset. It took us quite some time of managing projects on AWS before we first needed to use such a function. 

I am unsure about the amount of time it would take for very large datasets with millions of files as I have only tested the function with small projects, but I agree that we might need to implement this as a background task considering cases where we have millions of files. 
 
Regarding the order of operations, that makes sense also when thinking about large datasets. I will update the code.

Chrystinne:
@bemoody I've made the updates. Could you please have another look?

bemoody (changes requested):
Some small things:

```
messages.success(request, 'The project files have been deleted from S3.')
```
Please change this to say "The project files are being deleted from S3".

```
def delete_project_files_from_s3(project):
    # Import here to avoid circular import
    from console.views import delete_project_files_task
    """
    Immediately revoke access by deleting access points,
    set project.aws.sent_files = False, then schedule
    file deletion as a background task.
    """
```
Please put the doc string before the "import" line.  Otherwise it isn't recognized as a doc string.

Chrystinne:
> Chrystinne

Good catch, @bemoody! thanks!

bemoody (approved):

PR #2631, opened 2026-05-22
Add OpenID Connect (OIDC) provider support

p-hoffmann:
Minimal config-driven OIDC IdP on top of django-oauth-toolkit:
- RSA signing key + issuer configured via env (OIDC_RSA_KEY_FILE, OIDC_RSA_PRIVATE_KEY, OIDC_RSA_INACTIVE_KEY_FILES, OIDC_ISS_ENDPOINT)
- Discovery (root + /oauth scoped), JWKS, and OIDC userinfo endpoints
- CustomOAuth2Validator maps PhysioNet scopes (profile, email, institution, credentialing, orcid, public_id) into ID token / userinfo claims
- generate_oidc_rsa_key management command
- Restrict response_types to "code" (RFC 9700)


Chrystinne (changes requested):
Thanks for working on this, @p-hoffmann. Overall the OIDC flow looks good! 

A few things that might be worth adding:

- Document that OAuth applications used with the `openid` scope require `algorithm=RS256`, and improve the error message from "This application does not support signed tokens" to something like "set the application algorithm to RS256 for OIDC."

- Override CustomOAuth2Validator.get_discovery_claims() so claims_supported matches what ID tokens and /oauth/oidc/userinfo actually return. (We might want to include the full list: sub, name, given_name, family_name, preferred_username, website, email, email_verified, affiliation, is_credentialed, orcid, and public_user_uuid).

- Add setup documentation clarifying that the RSA key must be generated before setting OIDC_RSA_KEY_FILE.

Chrystinne (changes requested):
Thanks for the updates, @p-hoffmann! 

It seems that the custom claims now appear in `claims_supported` and `/oauth/oidc/userinfo`, but they still do not appear in the ID token. For example, `affiliation`, `is_credentialed`, `orcid`, and `public_user_uuid` are missing from the ID token even when the corresponding scopes are granted.

It looks like DOT’s default `oidc_claim_scope` filter is still excluding non-standard claims there. Adding something like this to `validators.py` fixes it:

```python
def get_oidc_claims(self, token, token_handler, request):
    """
    Bypass DOT's default oidc_claim_scope filter for ID tokens, too.
    _build_claims() already limits each claim to its granted scope, including
    PhysioNet-specific scopes that DOT does not know about.
    """
    if not (hasattr(request, 'user') and request.user):
        return super().get_oidc_claims(token, token_handler, request)
    scopes = set(getattr(request, 'scopes', []) or [])
    return self._build_claims(request.user, scopes)

Chrystinne:
@p-hoffmann I tested this locally with an RS256 OAuth application and the scopes openid profile email public_id:read institution:read credentialing:read orcid:read.

Without overriding get_oidc_claims(), the custom claims appeared in /oauth/oidc/userinfo but were missing from the ID token:

> ID TOKEN:
> affiliation=None, is_credentialed=None, orcid=None, public_user_uuid=None
> 
> USERINFO:
> affiliation="MIT", is_credentialed=False, orcid=None, public_user_uuid="<uuid>"

After adding the override, affiliation, is_credentialed, orcid, and public_user_uuid appear in the ID token when the corresponding scopes are granted.

Chrystinne:
Thanks, @p-hoffmann! I'm having another look at the code

Chrystinne (approved):
Thanks, @p-hoffmann! Looks good to me!

PR #2634, opened 2026-05-29
updating pyproject.toml, poetry lock & requirements for 3.2.20

Rutvikrj26:
This pull request updates the version of the `hdn-research-environment` dependency in the `pyproject.toml` file. The version is incremented from `3.2.13` to `3.2.20`, which will bring in any new features, bug fixes, or security updates included in the newer release. 

Dependency update:

* Bumped `hdn-research-environment` from version `3.2.13` to `3.2.20` in `pyproject.toml`.


requirements.txt
```diff
@@ -470,7 +470,7 @@ frozenlist==1.7.0 ; python_version >= "3.11" and python_version < "4.0" \
 geoip2==5.1.0 ; python_version >= "3.11" and python_version < "4.0" \
     --hash=sha256:445a058995ad5bb3e665ae716413298d4383b1fb38d372ad59b9b405f6b0ca19 \
     --hash=sha256:ee3f87f0ce9325eb6484fe18cbd9771a03d0a2bad1dd156fa3584fafa562d39a
-google-api-core[grpc]==2.25.1 ; python_version >= "3.11" and python_version < "4.0" \
```
tompollard, line 473: Please could you add back the `[grpc]` here? The line should not have been modified.

Rutvikrj26: Is Physionet using this?

tompollard: Good question, I'm not entirely sure how this dependency is being used. I'd need to look at it. If not needed, it would be great to remove.

Rutvikrj26: Based on my understanding - this is added for the hdn-research-environment to function & the current version doesn't use the grpc protocol. I have however added it back rn, but we can always remove it in an upcoming version bump if not needed.

Rutvikrj26: @copilot check if the grpc extra is being used - point to the exact code & the potential user workflow if yes.

tompollard (approved):

PR #2635, opened 2026-05-29
Fix purgeaccounts crash on accounts with no Profile

Rutvikrj26:
purgeaccounts builds its log line via User.get_full_name(), which delegates to self.profile.get_full_name(). Accounts created by UserManager.create_user save the User and the Profile in two separate, non-atomic steps, so an interrupted registration can leave an inactive User with no Profile row. When the command reaches such an account, accessing user.profile raises Profile.DoesNotExist, which propagates out of handle() and aborts the entire run -- so no stale accounts get purged and the orphan is never deleted, causing the hourly CronJob to crash on the same account every run.

Guard the display-name lookup with a _display_name() helper that falls back to "(no profile)" when the Profile is missing, in both the user and associated-email loops. Orphaned accounts are now purged instead of blocking the run. Add a regression test covering a profile-less account.

This is a recurring failure since HDN's Production Database has users without profiles.

Follow-up (not in this change): make UserManager.create_user atomic so profile-less users are never created in the first place.

### DIFF
--- a/physionet-django/user/management/commands/purgeaccounts.py
+++ b/physionet-django/user/management/commands/purgeaccounts.py
@@ -3,17 +3,9 @@ import logging
 from django.core.management.base import BaseCommand
 from django.utils import timezone
 
-from user.models import User, AssociatedEmail, Profile
+from user.models import User, AssociatedEmail
 
 LOGGER = logging.getLogger(__name__)
-
-
-def _display_name(user):
-    # Accounts created without a Profile must stay purgeable, not crash the run.
-    try:
-        return user.get_full_name()
-    except Profile.DoesNotExist:
-        return '(no profile)'
 
 
 class Command(BaseCommand):
@@ -30,7 +22,7 @@ class Command(BaseCommand):
         for user in user_list:
             deleted.append("\n - Username: {0}\n   Email: {1}\n   Full Name: "
                            "{2}".format(user.username, user.email,
-                                        _display_name(user)))
+                                        user.get_full_name(ignore_missing_profile=True)))
             user.delete()
 
         if deleted:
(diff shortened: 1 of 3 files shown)


physionet-django/user/management/commands/purgeaccounts.py
```diff
@@ -3,11 +3,19 @@
 from django.core.management.base import BaseCommand
 from django.utils import timezone
 
-from user.models import User, AssociatedEmail
+from user.models import User, AssociatedEmail, Profile
 
 LOGGER = logging.getLogger(__name__)
 
 
+def _display_name(user):
```
tompollard: Wouldn't it be cleaner to update the `get_full_name` method? Perhaps handle the return empty string (or no profile) behavior with an argument?

Rutvikrj26: That works too! The only reason I avoided it was that this is a core method that has a wide usage. I don't want to accidentally enable something that was supposed to fail.
This way, the blast radius stays very minimal & contained only to this very specific management command.

Rutvikrj26: Let me know your final thoughts on the matter - I can choose accordingly. I agree the one you mentioned is cleaner, but I am not sure the best idea.

tompollard: I could see this behavior being useful elsewhere. I'd add an `ignore_missing_profile` argument to `get_full_name ` that defaults to False.

Rutvikrj26: Sounds good!

Rutvikrj26:
Updated the Implementation.

tompollard (approved):
looks good to me, thanks!

PR #2643, opened 2026-06-17
Bump hdn-research-environment to 3.2.21 + add RESEARCH_ENVIRONMENTS_POLL_INTERVAL_MS setting

Rutvikrj26:
## Summary

- Bumps `hdn-research-environment` `3.2.20` → `3.2.21`.
- Adds `RESEARCH_ENVIRONMENTS_POLL_INTERVAL_MS` (env-configurable, default `30000` ms) inside the `ENABLE_CLOUD_RESEARCH_ENVIRONMENTS` block.

## Why

`hdn-research-environment` 3.2.21 drives the research environments page with HTTP polling for live status updates (instead of websockets) and reads the poll cadence from Django settings. This adds the corresponding setting so the interval is configurable per deployment — a new option exposed by 3.2.21. Defaults to 30s; the app also falls back to 30s if the setting is unset, so this is backwards-compatible.

## Changes
- `pyproject.toml` / `poetry.lock`: dependency pin `3.2.20` → `3.2.21`
- `physionet-django/physionet/settings/base.py`: `RESEARCH_ENVIRONMENTS_POLL_INTERVAL_MS = config('RESEARCH_ENVIRONMENTS_POLL_INTERVAL_MS', default=30000, cast=int)`


tompollard:
Does the requirements file need bumping too?

```
poetry export -f requirements.txt --output requirements.txt --with dev
```

Rutvikrj26:
Done.

tompollard (approved):
thanks rutvik, looks good to me

PR #2646, opened 2026-07-09
Add country to project author affiliations

### LINKED ISSUE #2645: Add "Country" to author affiliations
We would like to add support in the PhysioNet submission pipeline for capturing and preserving project-specific author affiliation metadata.

Requested changes
1. Capture country for each author affiliation
For each submitted project, add a field to record the country associated with each author’s affiliation.

This should be stored as part of the author affiliation information for that specific project.

2. Freeze project-specific affiliation metadata at publication  (if we are not already doing this)
Once a project is published, the affiliation information associated with each author on that project should be preserved as published and should not change retroactively.  In particular, each project should retain the author affiliation(s) that were submitted and published for that project. If an author later updates their default affiliation in their User Profile, that change should not alter the affiliation displayed or stored for previously published projects.

Chrystinne:
This PR adds country metadata to project author affiliations.

Changes:
- Adds a country field to active and published author affiliations;
- Captures affiliation country in the author affiliation form and invitation acceptance flow;
- Preserves affiliation country when publishing a project;
- Displays affiliation country on published project author info;
- Adds tests for affiliation country capture and publication.

New country selection field in the project author affiliation form:

image

How it is displayed on projects' landing page:

image

Fixes #2645

### DIFF
--- a/physionet-django/console/test_views.py
+++ b/physionet-django/console/test_views.py
@@ -62,7 +62,7 @@ class TestState(TestMixin):
             user=editor,
             display_order=project.authors.count() + 1,
         )
-        temp_author.affiliations.create(name='MIT')
+        temp_author.affiliations.create(name='MIT', country='US')
 
         # Submit project
         self.assertTrue(project.is_submittable())
@@ -107,7 +107,7 @@ class TestState(TestMixin):
             user=editor2,
             display_order=project.authors.count() + 1,
         )
-        temp_author.affiliations.create(name='MIT')
+        temp_author.affiliations.create(name='MIT', country='US')
 
         # Submit project
         self.assertTrue(project.is_submittable())
@@ -322,6 +322,8 @@ class TestState(TestMixin):
         Author approves publication
         """
         project = ActiveProject.objects.get(title='MIT-BIH Arrhythmia Database')
+        for author in project.authors.all():
+            author.affiliations.filter(country='').update(country='US')
 
         def get_project():
             return ActiveProject.objects.get(id=project.id)
@@ -461,6 +463,10 @@ class TestState(TestMixin):
         project = PublishedProject.objects.get(slug=custom_slug,
                                                version=project.version)
         self.assertEqual(project.submission_slug, project_slug)
+        published_affiliation = project.authors.get(
+            user__username='rgmark'
+        ).affiliations.first()
+        self.assertEqual(published_affiliation.country, 'US')
         # Access the published project's page and its (open) files
         response = self.client.get(reverse('published_project',
             args=(project.slug, project.version)))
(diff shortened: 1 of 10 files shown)


physionet-django/console/test_views.py
```diff
@@ -35,6 +36,13 @@
 LOGGER = logging.getLogger(__name__)
 
 
+def _fill_missing_affiliation_countries(project, country='US'):
```
tompollard: Instead of a helper function to populate country, we should populate the fixture data. The Q import and patches can then be removed.

physionet-django/project/test_views.py
```diff
@@ -74,6 +75,13 @@ def handle_starttag(self, tag, attrs):
     return fields
 
 
+def _fill_missing_affiliation_countries(project, country='US'):
```
tompollard: As before, can we populate countries in the fixtures?

tompollard (changes requested):

physionet-django/console/test_views.py
Chrystinne: Thanks, @tompollard. I populated country in the fixtures and removed the helpers/imports.

tompollard:
Thanks Chrystinne, this looks good to me. One other question - now that we are capturing the affiliation in multiple fields, should we be asking people for "Institution" and "Country" rather than "Affiliation" and "Country"? I think we might want to change this on the forms, and perhaps on the model too (e.g. Rename "Affiliation.name" to "Affiliation.institution").

physionet-django/project/forms.py
```diff
@@ -1092,9 +1118,14 @@ def clean(self):
                 raise forms.ValidationError(
                     'You must specify your affiliation.'
                 )
+            if not cleaned_data.get('affiliation_country'):
+                raise forms.ValidationError(
+                    'You must specify your affiliation country.'
```
tompollard: "Affiliation country" is slightly awkward wording. I think this content would sound more natural if "Affiliation" was changed to "Institution", as mentioned in my earlier comment. Then we just ask for "Institution" and "Country".

bemoody:
I don't see that it makes sense to require this for publishability.  That's creating a huge nuisance for projects that are already submitted and cannot now be edited.

Instead, only require the country to be supplied in the forms where author affiliation is set.  Making the field *mandatory for project submission* and/or *mandatory for publication* are a separate issue that can be addressed in a separate pull request.

physionet-django/project/forms.py
Chrystinne: Thanks, @tompollard. That sounds good. I updated the form to ask for “Institution” and “Country”. However, I left the model unchanged to avoid a broader refactor in this PR. What do you think?

tompollard (approved):
Looks good to me, thanks

PR #2653, opened 2026-07-16
Limit institution country check to unsubmitted projects

### LINKED ISSUE #2645: Add "Country" to author affiliations

Chrystinne:
This PR adjusts the institution country integrity check so that country is still required for new/unsubmitted projects, but projects that were already submitted or accepted before the country field was introduced are not blocked from publication.


bemoody (approved):
Makes sense.  Thanks!

PR #2662, opened 2026-07-22
Upgrade uwsgi from 2.0.22 to 2.0.31

tompollard:
uwsgi 2.0.22 fails to compile under Poetry 2.4.x's PEP 517 build isolation, causing upgrade test failures (OSError in threaded compilation leaves object files missing, breaking the link step).

e.g. see: https://github.com/MIT-LCP/physionet-build/actions/runs/29822812560/job/88609145846

This PR upgrades uwsgi from 2.0.22 to 2.0.31, which should fix Docker build failures.


Chrystinne (approved):
Looks good to me, @tompollard!

PR #2669, opened 2026-08-10
Fix/primary email integrity

Rutvikrj26:
This pull request addresses several issues related to the management and integrity of users' associated email addresses, particularly focusing on ensuring that each user always has a single, valid primary email. It introduces a new database constraint, migration, and logic updates to maintain this invariant, and fixes potential edge cases where users could be left without a primary email. Additionally, it includes a minor CSS update.

**Email integrity and primary email handling:**

* Added a partial unique constraint to the `AssociatedEmail` model to enforce that each user can have at most one primary associated email (`is_primary_email=True`). This is implemented both in the model definition and as a database migration. [[1]](diffhunk://#diff-e8178036221677f622720aa70ad8b6c37dcc82973d7901f556132b09bc540366R582-R588) [[2]](diffhunk://#diff-38665c983b1b6be9442b01ae97a94d65d5a453dbfce10206b337be4a68bb8400R1-R49)
* Introduced a data migration (`0072_primary_email_integrity.py`) that repairs users missing a primary associated email by promoting an appropriate associated email or realigning the user's main email field if needed.
* Refactored the logic for updating a user's primary email to use an atomic transaction, ensuring that demotion and promotion of primary emails is safe and respects the new unique constraint.
* Updated the method for retrieving a user's primary email to gracefully handle cases where the invariant is broken, logging an error and falling back to the user's main email if necessary.

**Account and email purging:**

* Modified the account purging command to avoid deleting unverified primary emails by filtering out primary emails from deletion, preventing users from being left without a primary email.

### DIFF
--- a/physionet-django/static/custom/css/style.css
+++ b/physionet-django/static/custom/css/style.css
@@ -336,6 +336,8 @@ body .main {
   border-radius: 0.5rem;
   position: absolute;
   top: var(--nav-height);
+  /* anchor() needs the invoker as anchor; position-anchor's initial value no longer opts in implicitly */
+  position-anchor: auto;
   left: anchor(left);
   background-color: white;
   padding: 0.75rem;
--- a/physionet-django/user/management/commands/purgeaccounts.py
+++ b/physionet-django/user/management/commands/purgeaccounts.py
@@ -32,7 +32,8 @@ class Command(BaseCommand):
         LOGGER.info("Total accounts removed {}".format(len(deleted)))
 
         associated_email_list = AssociatedEmail.objects.filter(
-            is_verified=False, added_date__lt=limit, user__is_active=True)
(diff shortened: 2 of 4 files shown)


bemoody:
Thanks.  I think that the better long-term solution is to remove the User.email field, but a data migration is a good immediate fix.

I think we need to see, with a specific anonymized example, what the database looks like on your side.  We may not know exactly how the database got into that state, but we want to know that this migration will fix the issue and won't break anything.

PhysioNet requires a way to contact people, so the system *shouldn't* be allowing a user to become active (User.is_active = true) without having a verified email address (AssociatedEmail.is_verified = true).  We also shouldn't be allowing people to set their primary email address to one that isn't verified, or to delete an email address that is primary.  I'm a little puzzled that your code isn't doing anything with is_verified.

Issue #1615, opened 2022-06-21
Add a test case for sandboxed-lightwave/FLAC


bemoody:
WFDB 10.7 uses libFLAC to read FLAC signals.

sandboxed-lightwave runs libwfdb inside a custom seccomp sandbox.

It's therefore possible (though unlikely) that some future version of libFLAC could fail to work inside the sandbox, so we should add a test case for it.

PR #2677, opened 2026-08-18
Bump lightwave version to 0.73 for tests

bemoody:
LightWAVE version 0.73 fixes the sandbox so that it should hopefully now work with glibc 2.41 (Debian 13 amd64).

It also allows clients to see the error message from WFDB in the event that a record is not readable (for debugging; lightwave.js doesn't make the error message visible anywhere, but it could in the future.)

### DIFF
--- a/.github/workflows/physionet-build-test.yml
+++ b/.github/workflows/physionet-build-test.yml
@@ -94,7 +94,7 @@ jobs:
 
       - name: Install and setup lightwave
         run: |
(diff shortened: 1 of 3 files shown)


bemoody:
While I don't have any reason to think this won't work, we *should* be testing lightwave more thoroughly; see issue #1615.

tompollard (approved):

PR #2684, opened 2026-08-31
Update django-autocomplete-light from 3.9.4 to 4.0

tompollard:
The 3.9.4 django-autocomplete-light sdist has broken tar metadata (group name is None) which crashes tarfile.chown during extraction on Debian 12, causing the Poetry install step to fail in CI:                                 

```                                                                                                                                                                                        
  TypeError                                                                                                                                                                                               
  getgrnam() argument 'name' must be str, not None                                                                                                                                                        
  at /usr/lib/python3.11/tarfile.py:2348 in chown
```

See for example: https://github.com/MIT-LCP/physionet-build/actions/runs/33286356667/job/99560210589

This pull request bumps django-autocomplete-light from 3.9.4 to 3.12.1


tompollard:
Hmm, it looks like it is still an issue in this version of django-autocomplete-light. Will need an alternative fix.

tompollard:
This is actually a broader issue than just this package. See https://github.com/MIT-LCP/physionet-build/issues/2685

PR #2688, opened 2026-09-01
Validate credential reviewer comments

Chrystinne:
Prevent credential application processing from saving reviewer comments longer than the 500-character database limit.

Add a test for form validation.


tompollard (approved):
thanks chrystinne, looks good to me

PR #2691, opened 2026-09-03
Remove duplicate DOI links in citation text

tompollard:
In https://github.com/MIT-LCP/physionet-build/pull/2681 we added the `urlize` filter when displaying citations, to display URLs as hyperlinks.

The `citation_text` method currently hardcodes links by wrapping DOI URLs in <a> tags, which is no longer necessary.

This change remove the <a> tags from `doi_format` in all five citation styles (MLA, APA, Chicago, Harvard, Vancouver) and lets `urlize` handle linking.

## Test plan         
- [ ] Verify citations display correctly on a published project page
- [ ] Confirm DOI links are clickable and not duplicated
- [ ] Check all citation styles in the citation modal"


lukepayyapilli (approved):

tompollard:
thanks @lukepayyapilli

PR #2692, opened 2026-09-03
Update approved credentialing email

### LINKED ISSUE #2658: Add standard reminders to credentialing approval email
As discussed in the PhysioNet meeting, reviewers frequently add the same two reminders to credentialing approval emails (individual access and LLM guidance). Including these reminders in the default approval email would reduce repetitive manual edits.

One possible location would be immediately after the approval paragraph ("We are pleased to say that your application for credentialed access has been approved.") and before the paragraph beginning "You are now able to access protected databases...".

Suggested introductory text:

Please note the following reminders regarding your credentialed access:

- Access to credentialed PhysioNet data is granted only to the approved applicant. The data may not be shared with others. Each person who will access the data must complete the PhysioNet credentialing process individually.
- Please ensure that any use of large language models (LLMs) with credentialed PhysioNet data complies with the PhysioNet Data Use Agreement and the guidance available here: https://physionet.org/news/post/llm-responsible-use/

briangow:
This PR updates the email sent once a credentialing application has been approved, as suggested in #2658 . The update adds reminders about each user needing to go through credentialing individually and to follow the guidance for appropriate use of LLMs with credentialed data. 

Fixes #2658.

### DIFF
--- a/physionet-django/notification/templates/notification/email/process_credential_complete.html
+++ b/physionet-django/notification/templates/notification/email/process_credential_complete.html
@@ -28,6 +28,10 @@ We also wanted to share these comments from the reviewer:
 {{ application.responder_comments }}
 {% endif %}
 {% if example_project %}
+Please note the following reminders regarding your credentialed access:
+
+- Access to credentialed PhysioNet data is granted only to the approved applicant. The data may not be shared with others. Each person who will access the data must complete the PhysioNet credentialing process individually.
+- Please ensure that any use of large language models (LLMs) with credentialed PhysioNet data complies with the PhysioNet Data Use Agreement and the guidance available here: https://physionet.org/news/post/llm-responsible-use/
 
 You are now able to access protected databases upon agreeing to the terms of usage and completing any required training. For example, you can access {{ example_project.title }} by following the steps below:
 


tompollard:
It looks like these changes hardcode PhysioNet specific content, which is something that we don't want to do. In many cases you can simply use a variable to display the platform name, but this doesn't solve the generalizability issue here.

briangow:
Thanks for the feedback @tompollard, that makes sense. I agree the issue here is broader than just replacing “PhysioNet” with `SITE_NAME` — this is really deployment-specific policy/guidance text.

I don’t see an existing first-class PhysioNet-specific override mechanism for this email, beyond the standard Django template lookup behavior. What are your thoughts on these options?

**Template-based site-specific override**
Keep the shared/default template generic, and allow the credentialing approval email template to be selected per deployment/site via configuration.

**Block-based site-specific override**
Keep the existing shared template, but make this reminder section configurable/injectable so deployments can provide their own policy text/links.

**Use more generic wording for now**
Remove the PhysioNet-specific references/URL and replace them with a generic reminder that applies across deployments,  at least as an interim step.

tompollard:
I think you could probably come up with some generic text that make the points you want to make. It would be great to have configurable email content (and webpage content) but finding a good way to do this would need some thought.

PR #2697, opened 2026-09-08
[WIP] S3 to S3 synchronization

bemoody:
When uploading to S3, attempt to copy each file from the previous project version rather than re-uploading.

This is a quick and untested implementation.  Some research is needed:
- Does "copy" generate the same ETag that "upload_file" does?
- What exactly is the 5 GB limit and what effect does that have?
- Any other metadata that would allow people to distinguish between newly-uploaded and cloud-duplicated files?

The approach used here should be adequate for common cases.  A better approach would be to identify matching files via SHA256SUMS.txt and then to use "If-Match", instead of listing the old bucket/prefix.

Something like `test_reupload_open_project` could serve as a basic test case.


bemoody:
> Does "copy" generate the same ETag that "upload_file" does?

`s3.copy` does appear to generate the same ETag that `s3.upload_file` does.

The similarly named function `s3.copy_object` does *not*; for files larger than 8 MB, it uses the real MD5 instead of the `s3_file_etag` algorithm.

The `aws s3 sync` command appears to do something similar to `s3.copy`, not `s3.copy_object`.

> Any other metadata that would allow people to distinguish between newly-uploaded and cloud-duplicated files?

The "ChecksumAlgorithm" and "ChecksumType" (shown when listing the bucket) appear to be different:

- upload_file (< 8 MB): ChecksumAlgorithm `CRC32`, ChecksumType `FULL_OBJECT`
- upload_file (> 8 MB): ChecksumAlgorithm `CRC32`, ChecksumType `COMPOSITE`
- copy (< 8 MB): ChecksumAlgorithm `CRC32`, ChecksumType `FULL_OBJECT`
- copy (> 8 MB): ChecksumAlgorithm `CRC64NVME`, ChecksumType `FULL_OBJECT`
- copy_object (< 8 MB): ChecksumAlgorithm `CRC32`, ChecksumType `FULL_OBJECT`
- copy_object (> 8 MB): ChecksumAlgorithm `CRC32`, ChecksumType `FULL_OBJECT`

It might be possible to choose a checksum algorithm explicitly.

PR #2700, opened 2026-09-09
Add CITI API verification as background check for training submissions

tompollard:
## Summary                                                                                                    

  - This is a pilot integration with the CITI API to retrieve training information for users.
  - When a user submits CITI training, we now query the CITI SOAP API via a background task to retrieve their completion records and member profile, storing the results in a new `CITIVerification`  model for admin review.                                                                          
  - Admins see a new "CITI API Verification" card on the training review page showing the member profile, all completions from the API, and warnings for name mismatches or unlinked emails.                   
  - A `CITIGroupMapping` model (configurable via Django admin) maps CITI group IDs to PhysioNet training types, determining which completions are relevant.

  ## Details

  - `citi_training_module.py`: Added `get_member_profile()`, `lookup_citi_completions_for_user()` (tries all verified emails), `find_matching_completion()`, `serialize_completions()`, and `verify_training_via_citi_api()`. Added request timeout (30s) and error handling. Note: `get_citiprogram_completion()` now returns a 3-tuple `(completions, member_id, member_profile)` instead of a plain list.
  - `CITIVerification` stores: lookup email, member ID, completion report ID, full API response (JSON), and any error message. Linked to `Training` via `OneToOneField`.
  - `CITIGroupMapping` includes "Data or Specimens Only Research" via data migration.
  - The background task only retrieves data for training types that have a `CITIGroupMapping` configured.
  - The user-facing training page suggests linking the CITI member profile email to speed up approval.

  ## Test plan

  - [ ] 24 tests in `user.test_citi_api` covering SOAP XML parsing, completion matching, user lookup, serialization, verification orchestration, and console template rendering
  - [ ] Submit CITI training, run `process_tasks`, verify the CITI API Verification card appears in console review
  - [ ] Verify training types without a `CITIGroupMapping` are unaffected


bemoody (changes requested):
Please add an example of the CITIVerification object in demo-user.json.

Don't try to parse the XML and then save some JSON transformed version of it in the database.  *Save the XML* so that we can audit it in the future.

tompollard:
Thanks @bemoody. 

> Please add an example of the CITIVerification object in demo-user.json.

I've added a CITIGroupMapping and CITIVerification entry to demo-user.json.

> Save the XML so that we can audit it in the future.

I replaced `completion_data` (JSONField) with `member_profile_xml` and `completions_xml` (TextFields) that store the original SOAP responses. Also did some cleaning for readability and tried to remove duplicated parsing logic.

bemoody (changes requested):
If `CITI_USERNAME`, `CITI_PASSWORD`, and `CITI_SOAP_URL` aren't set in the configuration, we shouldn't try to access the CITI API at all.

bemoody:
If I set CITI_USERNAME and CITI_PASSWORD to bogus values, `get_member_profile` doesn't raise an error.  Here's what the response looks like:
```
<?xml version="1.0" encoding="utf-8"?>
<soap:Envelope
    xmlns:soap="http://www.w3.org/2003/05/soap-envelope"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  <soap:Body>
    <GetInstMemberByEmailResponse
        xmlns="https://webservices.citiprogram.org/">
      <GetInstMemberByEmailResult>
        <xs:schema
            id="NewDataSet"
            xmlns=""
            xmlns:xs="http://www.w3.org/2001/XMLSchema"
            xmlns:msdata="urn:schemas-microsoft-com:xml-msdata">
          <xs:element
              name="NewDataSet"
              msdata:IsDataSet="true"
              msdata:UseCurrentLocale="true">
            <xs:complexType>
              <xs:choice
                  minOccurs="0"
                  maxOccurs="unbounded" />
            </xs:complexType>
          </xs:element>
        </xs:schema>
        <diffgr:diffgram
            xmlns:msdata="urn:schemas-microsoft-com:xml-msdata"
            xmlns:diffgr="urn:schemas-microsoft-com:xml-diffgram-v1" />
      </GetInstMemberByEmailResult>
    </GetInstMemberByEmailResponse>
  </soap:Body>
</soap:Envelope>
```
The examples in test_citi_api.py, in contrast, include a `NewDataSet` element.  This is a bit weak, but can we check for that element as a way to check that our credentials are good?

tompollard:
@bemoody, these points are now addressed:

> If CITI_USERNAME, CITI_PASSWORD, and CITI_SOAP_URL aren't set in the configuration, we shouldn't try to access the CITI API at all.

Added a check for missing credentials.

> If I set CITI_USERNAME and CITI_PASSWORD to bogus values, get_member_profile doesn't raise an error. Here's what the response looks like. The examples in test_citi_api.py, in contrast, include a NewDataSet element. This is a bit weak, but can we check for that element as a way to check that our credentials are good?

Updated to check for the element as you suggest.

bemoody (approved):

PR #2698, opened 2026-09-08
Fix query for latest published project version

bemoody:
In the published project page as well as various other places, we look up the "latest" version of the current project in order to provide a hint to the user.

This was relying on the default ordering of PublishedProject objects, which is not a great idea in general, and in this case would give the wrong result if the project title was changed from one version to the next.  (Of course, changing project titles isn't supposed to be possible, but can be done by administrators.)

Furthermore, avoid querying a bunch of stuff we don't need.

Fixes #2690

PR #2701, opened 2026-09-09
Prevent publishing multiple "latest versions" simultaneously

bemoody:
Published projects have a flag `is_latest_version` that marks which version is considered the "latest".  (The purpose of this flag is somewhat murky, and in practice it means having the largest version number.)

In pull #2698, @briangow pointed out that there isn't anything at the database level to prevent creating two versions that both have `is_latest_version = True`.

That shouldn't ever happen: currently, `is_latest_version` is set by `PublishedProject.set_version_order`, which always saves the latest version *last*, and is always invoked within a transaction.

However, some existing code will crash if there are two "latest" versions:
- `console.views.publish_submission` assumes that there is only one latest version for a given core project
- `project.views.publish_project_latest` assumes that there is only one latest version for a given *slug*

and the new code I proposed to add in pull #2698 would also assume there is only one latest version for a given core project.

(In general, we assume a one-to-one relationship between core projects and slugs, but again there is no hard restriction.)

Here, I've added constraints to preserve the invariants that the existing code requires.  It should be possible to change this in the future, if we decide to support changing slugs or taking over an existing slug.

The constraint logic is borrowed from @Rutvikrj26's pull #2669, which was attempting to fix a similar issue with users.


bemoody:
This should be working now and ready for review.  The existing publication logic breaks the expected invariants, so it needs to be reworked slightly and this needs to be a post-install migration.

bemoody:
Bumping this - @tompollard or @chrystinne, can you please take a look?  I think this change is quite safe, it will just guard against buggy code in the future.

Chrystinne:
@bemoody I'm taking a look.

tompollard (approved):
Looks good, thanks @bemoody

Chrystinne:
@bemoody I tested locally and the implementation looks correct to me.

PR #2702, opened 2026-09-10
Pin pyOpenSSL>=25.0.0 to fix cryptography compatibility.

tompollard:
pyOpenSSL 24.3.0 references _lib.GEN_EMAIL which was removed in cryptography 50.x, causing an AttributeError on startup. Pinning pyOpenSSL>=25.0.0 resolves the incompatibility.


bemoody:
What is referencing pyOpenSSL, then?  Nothing in this repository depends on it.

tompollard:
Sorry, you're right. I'd installed PyOpenSSL when testing Botocore. Doesn't need to be there.

PR #2706, opened 2026-09-12
Fix settings sidebar not highlighting the active page 

tompollard:
Currently, the active page is not highlighted in the settings menu:

Screenshot 2026-09-12 at 17 47 32

This PR fixes the issue, so the active page is highlighted.

### Summary

- The settings_tabs inclusion tag was not passing the template context through, so request.path was unavailable in the tab template and the active page highlight never applied                                                                                                              
- Add ``takes_context=True` to the inclusion tag and forward the request object so the existing `{% if settings_url == request.path %} active{% endif %}` check works as intended

### Test plan

- Visit any settings page (e.g. /settings/profile/, /settings/training/) — the corresponding sidebar tab should be highlighted
- Navigate between settings pages — the highlight should follow the active page


Chrystinne (approved):
@tompollard I've tested it locally, and it looks good.

Issue #2357, opened 2025-03-14
process_tasks RuntimeError: 'path' must be None or a list


bemoody:
In pull #2324, I added the django-static-mathjax package.  That package currently breaks django-background-tasks in a bad way:
```
RuntimeError: 'path' must be None or a list, not <class '_frozen_importlib_external._NamespacePath'>
```

**This hasn't caused a problem for PhysioNet yet.  It will cause a problem as soon as we try to deploy a migration, at which point the background-tasks daemon will be restarted.**

- Is it a bug in django-static-mathjax, that the package doesn't include an `__init__.py`?

- Is it a bug in django-background-tasks, that it doesn't handle `RuntimeError` along with `ImportError`?

- Is it a bug in Python, that `imp.find_module` can't cope with path objects produced by `importlib.import_module`?

Maybe all of the above, but the easiest solution is to add that __init__.py file.

However, it might be wise to add `manage.py process_tasks --duration 1` to physionet-build-test.yml as a smoke test.

tompollard:
I haven't investigated but I wouldn't be surprised if django-background-tasks is the problem (sounds like it is no longer maintained: https://github.com/MIT-LCP/physionet-build/issues/1647).

Chrystinne:
@bemoody and @tompollard In my local environment, Django fails to start when mathjax is added to INSTALLED_APPS.

This is what currently works for me:

```
INSTALLED_APPS = [
    ...
    'django_mathjax',
    ...
]
```

Issue #2601, opened 2026-03-24
GCP file upload timeout


briangow:
When attempting to upload the files from the MIMIC-IV-ECHO v1.0 project via the admin console (Google Cloud - Upload files), we are getting the timeout detailed below. Each timeout is happening after ~2 minutes. 

We should explore whether increasing the timeout will help for projects like this with very large file sizes. Additionally, we may consider reducing the chunk size, or using parallel composite uploads (doing so via `gsutil` may be more robust).  

```
Dear admins,

The task "GCP - MIMIC-IV-ECHO: Echocardiogram Matched Subset v1.0" has made 5 attempts and failed.

The task has been marked as completed and it will not be rescheduled.

The final attempt started at 2026-03-24 18:50:17.

The task name is: console.views.send_files_to_gcp and was called with the following parameters: [[826], {}].

The last error for this task is:

Traceback (most recent call last):
  File "/physionet/python-env/physionet/lib/python3.11/site-packages/urllib3/connectionpool.py", line 715, in urlopen
    httplib_response = self._make_request(
                       ^^^^^^^^^^^^^^^^^^^
  File "/physionet/python-env/physionet/lib/python3.11/site-packages/urllib3/connectionpool.py", line 416, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/physionet/python-env/physionet/lib/python3.11/site-packages/urllib3/connection.py", line 244, in request
    super(HTTPConnection, self).request(method, url, body=body, headers=headers)
  File "/usr/lib/python3.11/http/client.py", line 1282, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.11/http/client.py", line 1328, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.11/http/client.py", line 1277, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.11/http/client.py", line 1076, in _send_output
    self.send(chunk)
  File "/usr/lib/python3.11/http/client.py", line 998, in send
    self.sock.sendall(data)
  File "/usr/lib/python3.11/ssl.py", line 1274, in sendall
    v = self.send(byte_view[count:])
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/ssl.py", line 1243, in send
    return self._sslobj.write(data)
[...]

Issue #2408, opened 2025-05-28
Stalled pending tasks on Published Project (bigp3bci)


tompollard:
It looks like there are stalled pending tasks on the `bigp3bci` project. The published project is missing special files, zipfile, etc.

https://physionet.org/content/bigp3bci/1.0.0/

tompollard:
The Django background task handler (`django-background-tasks.service`) had stalled. As a result, tasks were not being processed. We resolved the issue by:

1. Stopping the background task handler service
2. Clearing task locks in the Django admin interface
3. Restarting the service

It looks like the cause was an AWS data transfer task. We should look into how we can improve these data transfers, especially for large files. We should think about:

We should look into improving monitoring and alerting for tasks (memory usage, task length and processing rates, and also improve recovery for stuck tasks.

PR #2708, opened 2026-09-14
Route all new background tasks through django-q2

tompollard:
This is a follow up to: https://github.com/MIT-LCP/physionet-build/pull/2207. django-background-tasks-updated was archived in June 2025 and does not support Django 5.2+, so tasks need to move to django-q2 which is actively maintained. See the archived versions of django-background-tasks at: https://pypi.org/project/django-background-tasks-updated/ and https://github.com/django-background-tasks/django-background-tasks

New tasks are now queued via `enqueue_task()` which dispatches to django-q2's `async_task()`. The legacy django-background-tasks system remains active so that any pending tasks in the old queue can drain through process_tasks. Both daemons (`process_tasks` and `qcluster`) should run simultaneously during the transition.

### Key changes:

- Add `enqueue_task()` helper and task_completion_hook to console/tasks
- `get_associated_tasks()` now checks both OrmQ and legacy Task tables
- All invocation sites updated to use enqueue_task()
- Update Q_CLUSTER timeout/retry to match legacy settings
- Add Q_CLUSTER sync=True for test suite
- Update assertBackgroundTasks for django-q2 sync mode
- Add qinfo management command test
- Exclude django_q permissions from admin forms

### Test plan

- Run the full test suite
- Verify `python manage.py qinfo` succeeds
- Verify `python manage.py process_tasks --duration=1` still succeeds
- After deploying to staging, confirm new tasks appear in the django-q2 admin and are processed by qcluster
- Confirm the legacy `process_tasks` daemon runs without error (it will idle once the old queue is empty)


lukepayyapilli:
Nice shape overall. Running both queues in parallel to drain is the right migration strategy, and the invocation sites all look converted (grepped the branch, no leftover `verbose_name=` task calls). A few things I'd want resolved before this goes out, mostly around django-q2 semantics differing from django-background-tasks in ways that bite at our task sizes.

### Blocking

**1. `Q_CLUSTER` timeout reintroduces the bug #2608 fixed.** `timeout: 120` / `retry: 600` / `max_attempts: 5` (settings/base.py:231). In django-q2 `timeout` is a hard kill of the worker, and `retry` re-queues regardless of whether the task is still running.

#2608 set `MAX_RUN_TIME = 9999999999  # 316 years` because a 3600s assumption was spawning a fresh copy of every long task once an hour (and noted that only became visible once #699 allowed parallel workers). That line is still in this file at :247, fourteen lines below a `Q_CLUSTER` that makes the same assumption with a 120-second window and four workers. Every task over two minutes gets SIGKILLed, retried five times, and doubled up at the 600s mark.

#2601 shows what that means concretely: `send_files_to_gcp` on MIMIC-IV-ECHO runs well past two minutes today. Same for `create_published_project_zip` and `prepare_active_project_files` on any large project.

The commit message says timeout/retry were updated to match legacy, but I don't see that in the diff. Those values are pre-existing from #2207 and predate #2608.

**2. `remove_existing=True` is too broad.** `Task.objects.filter(task_name=func_name).delete()` (tasks.py:303) and the OrmQ loop at :306 match on function name only. The legacy `remove_existing_tasks=True` matched on task *hash*, i.e. name **and** params (`background_task/models.py:95`). Since the call site is `CopyeditForm.save()`, two editors finishing copyedit around the same time means the second silently deletes the first project's pending task. Needs to filter on the instance.

**3. `get_associated_tasks` now fails open.** Three new swallow paths: `if model not in _model_tasks: return` (:207), `_model_tasks[model].get(task_name, [])`, and `except Exception: return None, (), {}` (:150). All feed the `if any(get_associated_tasks(project))` guards in `update_submission_checksums` / `gcp_bucket_management` / `aws_bucket_management` and the `disabled` logic in `manage_published_project.html`. "Couldn't tell" now reads as "nothing pending", which unlocks the destructive buttons. Given #2408 I'd rather this stayed loud.

**4. Long titles silently suppress failure notifications.** `django_q.models.Task.name` is `max_length=100`, but our labels are `'Making checksum file - {project}'` where `PublishedProject.__str__` is `'{title} v{version}'` and `title` is `max_length=200`. The resulting `DataError` is swallowed by `save_task` (`monitor.py:154`), so no Task row, so no `post_save`, so `call_hook` never fires, so `task_completion_hook` never runs and admins aren't told. The task in #2601 is already at 49 characters. Suggest truncating in `enqueue_task`.

**5. `task_completion_hook` fires per attempt.** It's wired to `Task.post_save`, and `save_task` re-saves on each retry incrementing `attempt_count` (`monitor.py:120-128`), so a permanently-failing task sends 5 admin emails. Compare the email in #2601: legacy sent one, after the final attempt. Needs an `attempt_count >= MAX_ATTEMPTS` guard.

Related to #2091: `enqueue_task` defaults `task_name=None`, and the `run_citi_api_verification` call in `user/views.py` passes none at all. django-q2 fills that with a humanhash, so instead of crashing the notification it sends admins "sixteen-mango-oranges has failed". Same class of problem #2091 was fixing.

**6. `assertBackgroundTasks` no longer asserts anything.** `DjangoQTask.objects.order_by('-started')[:N]` then `assertEqual(len(...), N)`: a slice of N can't exceed N, so this only fails if the whole table has fewer than N rows. The old `Task.objects.count() == N` caught *extra* tasks, which is the failure mode that matters when refactoring a queue. Rows also accumulate within a test method, so a second call re-inspects the first call's rows.

In sync mode the worker also re-raises (`worker.py:104`), so failures propagate out of `enqueue_task` at the call site and the `if not task.success: raise BackgroundTaskError(task)` branch, plus the rewritten `BackgroundTaskError.__init__`, are unreachable.

And removing `self.assertFalse(get_project().is_publishable())` in `console/test_views.py` is the giveaway: that was the only assertion verifying the work was *deferred*. Between that and the above, this PR changes the queueing system and ends up with no test coverage of queueing. `sync=True` is fine as a default, but I'd want one test with `sync=False` that asserts the OrmQ row and then drives the worker. Otherwise `_run_task`, `_unpack_ormq` and the whole django-q2 branch of `get_associated_tasks` never execute in CI.

### Worth considering

**Could `@background()` come off entirely?** `_run_task` and `importlib` exist only because the task functions are still `TaskProxy` objects, and the decorator no longer enqueues anything; it just holds `.name` and `.task_function`. Removing it lets `async_task` take the real callable, and `_run_task`, `_unwrap_run_task`, the duplicated `hasattr(func, 'task_function')` unwrap, and most of `_unpack_ormq` all go away. Would turn this into a net deletion.

**Smaller:**

- `_unpack_ormq` compares `func == _run_task` (object) while `_unwrap_run_task` compares against a string. Both correct (OrmQ pickles the callable, `Task.func` is `get_func_repr`), but nothing says why they differ.
- `_TaskInfo.__str__` renders `project.tasks.create_published_project_zip((42,), {})` into the admin UI. The human label is in the payload as `task_dict['name']`, which `_unpack_ormq` loads and discards. Relevant to #2210.
- `enqueue_task(**kwargs)` forwards straight to `async_task`, which pops `timeout`, `group`, `save`, `cluster`, `chain` etc. before passing the rest to the function, so a task param with one of those names silently becomes a queue option. `hook=` is passed unconditionally, so a caller passing `hook=` gets a duplicate-kwarg `TypeError`. An explicit `q_options` dict would be safer.
- `task_proxy.task_name = task_name` (:85): the comment says "consistent with django-q2 conventions", but nothing in django_q sets `task_name` on a function. `TaskProxy.name` already existed and is what the code used before, including in #2077's machinery. This makes `.task_name` work only when `@associated_task` was also applied.
- `_match_params` docstring says it yields `(task_info_or_task, ro_flag)`; it yields `ro_flag`. The new `matched` flag also dedupes where the old loop yielded twice, which changes `rw_tasks`/`ro_tasks` lengths.
- `OrmQ.objects.all()` is unpickled in full on every `get_associated_tasks` call, including every `manage_published_project` page load, and re-scanned per `task_name`. Putting the model key in the task's `group` would make it a SQL filter.
- `Q_CLUSTER['sync'] = True` sits inside the `if RUNNING_TEST_SUITE:` block that otherwise just opens a log file.
- `test_qinfo_command` asserts nothing and prints cluster info into test output.
- `from background_task.models import Task` and `from background_task.tasks import tasks` are now unused in `user/test_views.py` (`setup.cfg` has `ignore = F401`, so lint won't catch it).
- `cloud/s3.py` imports `enqueue_task` function-locally to avoid a circular import, while `console/forms.py`, `user/views.py` and `activeproject.py` import it at module scope. Worth confirming which is right.
- `creator=` was dropped from all call sites. Doesn't look like anything reads it, but flagging in case that's unintentional.
- #1647 is closed as completed but the dependency is obviously still here. Reopening it (or filing a successor) would give the removal step somewhere to live, since nothing currently stops a new call site being written in the old style while `@background()` is still importable.

<details>
<summary>References used</summary>

| Ref | What | Why it's relevant |
|---|---|---|
| #2608 | Do not assume background tasks are dead after one hour (merged) | Set `MAX_RUN_TIME` to 316 years for exactly the reason `Q_CLUSTER.timeout: 120` breaks |
| #699 | Allow multiple background tasks in parallel (merged) | Why duplicate-spawn became visible; `workers: 4` restores the same condition |
| #2207 | Add Django Q2 (merged) | Where the unchecked `Q_CLUSTER` values came from |
| #2091 | Fix notification for tasks with no verbose_name (merged) | Same failure class as the `task_name=None` default |
| #2077 | Mark move_files_as_readonly as associated task (merged) | Origin of the `.name` usage this PR swaps for `.task_name` |
| #2601 | GCP file upload timeout (open) | Evidence of real runtimes past 2 min |
| #2408 | Stalled pending tasks (closed) | Why fail-open on the pending check is risky |
| #2210 | Display background tasks in admin console (open) | `_TaskInfo.__str__` output |
| #1647 | Removing django-background-tasks (closed as completed) | Misleadingly closed; candidate home for the removal step |

Checked against django-q2 1.6.2 and django-background-tasks-updated 1.2.7 sources.

</details>

tompollard:
> Nice shape overall. Running both queues in parallel to drain is the right migration strategy, and the invocation sites all look converted (grepped the branch, no leftover verbose_name= task calls). A few things I'd want resolved before this goes out, mostly around django-q2 semantics differing from django-background-tasks in ways that bite at our task sizes.

Thanks Luke! I've responded to the blocking points below. I plan to address the "Worth considering" points in a follow up PR.

> Blocking
> 1. Q_CLUSTER timeout reintroduces the bug https://github.com/MIT-LCP/physionet-build/pull/2608 fixed. timeout: 120 / retry: 600 / max_attempts: 5 (settings/base.py:231). In django-q2 timeout is a hard kill of the worker, and retry re-queues regardless of whether the task is still running.

> https://github.com/MIT-LCP/physionet-build/pull/2608 set MAX_RUN_TIME = 9999999999  # 316 years because a 3600s assumption was spawning a fresh copy of every long task once an hour (and noted that only became visible once https://github.com/MIT-LCP/physionet-build/pull/699 allowed parallel workers). That line is still in this file at :247, fourteen lines below a Q_CLUSTER that makes the same assumption with a 120-second window and four workers. Every task over two minutes gets SIGKILLed, retried five times, and doubled up at the 600s mark.

> https://github.com/MIT-LCP/physionet-build/issues/2601 shows what that means concretely: send_files_to_gcp on MIMIC-IV-ECHO runs well past two minutes today. Same for create_published_project_zip and prepare_active_project_files on any large project.

> The commit message says timeout/retry were updated to match legacy, but I don't see that in the diff. Those values are pre-existing from https://github.com/MIT-LCP/physionet-build/pull/2207 and predate https://github.com/MIT-LCP/physionet-build/pull/2608.

Addressed in https://github.com/MIT-LCP/physionet-build/pull/2708/commits. Sorry about the earlier confusing commit message. I'd started with a larger PR and then split it into smaller PRs. This is the first of those.

> 2. remove_existing=True is too broad. Task.objects.filter(task_name=func_name).delete() (tasks.py:303) and the OrmQ loop at :306 match on function name only. The legacy remove_existing_tasks=True matched on task hash, i.e. name and params (background_task/models.py:95). Since the call site is CopyeditForm.save(), two editors finishing copyedit around the same time means the second silently deletes the first project's pending task. Needs to filter on the instance.

I don't think collisions are really a concern for us, but this point is addressed in https://github.com/MIT-LCP/physionet-build/pull/2708/commits/99fe51a181a76c5039e89717156ee3eb1489ee4b

> 3. get_associated_tasks now fails open. Three new swallow paths: if model not in _model_tasks: return (:207), _model_tasks[model].get(task_name, []), and except Exception: return None, (), {} (:150). All feed the if any(get_associated_tasks(project)) guards in update_submission_checksums / gcp_bucket_management / aws_bucket_management and the disabled logic in manage_published_project.html. "Couldn't tell" now reads as "nothing pending", which unlocks the destructive buttons. Given https://github.com/MIT-LCP/physionet-build/issues/2408 I'd rather this stayed loud.

Addressed in https://github.com/MIT-LCP/physionet-build/pull/2708/changes/854d5efe05a512e85b0ea15e25e6c144f05d0b8c

> 4. Long titles silently suppress failure notifications. django_q.models.Task.name is max_length=100, but our labels are 'Making checksum file - {project}' where PublishedProject.__str__ is '{title} v{version}' and title is max_length=200. The resulting DataError is swallowed by save_task (monitor.py:154), so no Task row, so no post_save, so call_hook never fires, so task_completion_hook never runs and admins aren't told. The task in https://github.com/MIT-LCP/physionet-build/issues/2601 is already at 49 characters. Suggest truncating in enqueue_task.

Titles are now truncated: https://github.com/MIT-LCP/physionet-build/pull/2708/changes/fdb7df732095c0a78c7011987febce6c7e7625ac

> 5. task_completion_hook fires per attempt. It's wired to Task.post_save, and save_task re-saves on each retry incrementing attempt_count (monitor.py:120-128), so a permanently-failing task sends 5 admin emails. Compare the email in https://github.com/MIT-LCP/physionet-build/issues/2601: legacy sent one, after the final attempt. Needs an attempt_count >= MAX_ATTEMPTS guard.

> Related to https://github.com/MIT-LCP/physionet-build/pull/2091: enqueue_task defaults task_name=None, and the run_citi_api_verification call in user/views.py passes none at all. django-q2 fills that with a humanhash, so instead of crashing the notification it sends admins "sixteen-mango-oranges has failed". Same class of problem https://github.com/MIT-LCP/physionet-build/pull/2091 was fixing.

Admins are now only notified once: https://github.com/MIT-LCP/physionet-build/pull/2708/changes/646391bf0e475cb3f4bbc4e00cef0b5cf677c54e

> 6. assertBackgroundTasks no longer asserts anything. DjangoQTask.objects.order_by('-started')[:N] then assertEqual(len(...), N): a slice of N can't exceed N, so this only fails if the whole table has fewer than N rows. The old Task.objects.count() == N caught extra tasks, which is the failure mode that matters when refactoring a queue. Rows also accumulate within a test method, so a second call re-inspects the first call's rows.

> In sync mode the worker also re-raises (worker.py:104), so failures propagate out of enqueue_task at the call site and the if not task.success: raise BackgroundTaskError(task) branch, plus the rewritten BackgroundTaskError.__init__, are unreachable.

> And removing self.assertFalse(get_project().is_publishable()) in console/test_views.py is the giveaway: that was the only assertion verifying the work was deferred. Between that and the above, this PR changes the queueing system and ends up with no test coverage of queueing. sync=True is fine as a default, but I'd want one test with sync=False that asserts the OrmQ row and then drives the worker. Otherwise _run_task, _unpack_ormq and the whole django-q2 branch of get_associated_tasks never execute in CI.

Addressed in https://github.com/MIT-LCP/physionet-build/pull/2708/changes/c45c7b6171e197b8c0561959698e741ec09aaa4a

tompollard:
Additional feedback tracked at: https://github.com/MIT-LCP/physionet-build/issues/2709

lukepayyapilli (commented):
Thanks for the fast turnaround, and for splitting out #2709. Fixes for 2 (hash formula matches `background_task/models.py:98-104` exactly), 3, 4 and 6 all look right to me. Two problems with the combination of 1 and 5, though, plus one smaller thing on 2.

### 1 + 5 together mean admins are never notified of a failure

`retry: 10000000000` makes the ORM broker set `lock = now + RETRY` when it dequeues (`brokers/orm.py:14`, `dequeue()` filters `lock__lt=timezone.now()`). On failure, `monitor` only acknowledges when `task["success"] or task.get("ack_failure")`, and `ack_failures` defaults to False (`conf.py:166`) and isn't set in `Q_CLUSTER`. So a failed task is never acknowledged, its lock sits 317 years out, and it is never re-dequeued.

`attempt_count` is only incremented in `save_task`'s `existing_task` branch (`monitor.py:120-128`), which is reached on a second attempt that now never happens. So `attempt_count` is permanently 1, and the new guard

```python
if task.attempt_count < max_attempts:   # 1 < 5
    return
```

returns every single time. No email is ever sent, for any failure. That's worse than the 5 duplicate emails it was fixing.

The failed OrmQ row also leaks permanently and is invisible to `queue_size()`, which filters `lock__lte=now`.

Simplest fix is probably `'ack_failures': True` plus notifying on the first failure, since with an effectively infinite retry there is only ever one attempt. If you'd rather keep real retries, a long-but-finite `retry` (a day, say) would let `max_attempts` work as intended.

### `retry: 10000000000` also needs #2608's other half

#2608 paired `MAX_RUN_TIME = 9999999999` with `ExecStartPre=... manage.py unlock_background_tasks` in the systemd unit, because a 316-year lock is only safe if something clears stale locks on restart. `django-q2-tasks.service` has no equivalent, django-q2 ships no unlock command (`qcluster`, `qinfo`, `qmemory`, `qmonitor`), and `cluster.py` does nothing with locks at startup.

So any task interrupted by a deploy, OOM kill or reboot has its lock set 317 years out and is never picked up again. The legacy daemon recovers from this; qcluster won't. A small command doing `OrmQ.objects.update(lock=timezone.now())`, wired as `ExecStartPre` the same way, would close it.

### Smaller, on 2

The hash is right, but the legacy branch can't match anything in practice. The hash covers `json.dumps((args, kwargs))`, and the call-site conversion changed `prepare_active_project_files(project_id=project.id, ...)` to `enqueue_task(..., project.id, ...)`. So the new hash is over `((5,), {})` while the rows actually sitting in the legacy queue were hashed over `((), {"project_id": 5})`. Since draining the old queue is the only reason that branch exists, it's currently dead code.

Also worth noting legacy filtered `locked_at__isnull=True` so it never deleted a running task, which the new version drops.

### Nits, happy for these to go to #2709

- The new tests cover `_run_task` and the `_unpack_ormq` round-trip nicely, but not the django-q2 branch of `get_associated_tasks`, which is the one feeding the destructive-action guards.
- Removing the bare except from `_unpack_ormq` is what I asked for, but note one corrupt row now 500s `manage_published_project` for every project, since the new `try/except` only wraps the call inside `enqueue_task`. Fine either way, just worth being deliberate.
- The watermark comment says it resets "by Django's TestCase transaction rollback". The reset actually happens because a fresh TestCase instance is constructed per test method; the rollback is what empties the table.

bemoody:
I have yet to be convinced that this is a positive change.  It appears that it wwould be a lot of work to try to reproduce the functionality we're relying on, in a new system that isn't designed for it.

I have not been impressed by what I've seen of django-q2, either in terms of robustness or in terms of the complexity of the implementation.

> django-background-tasks-updated was archived in June 2025 and does not support Django 5.2+, so tasks need to move to django-q2 which is actively maintained

django-background-tasks is under active development as of a couple months ago.  If it has issues on Django 5 then let's fix those issues.

tompollard:
> I have yet to be convinced that this is a positive change. I have not been impressed by what I've seen of django-q2, either in terms of robustness or in terms of the complexity of the implementation.

We're currently managing background tasks with a package that hasn't been updated in over two years: https://github.com/django-background-tasks/django-background-tasks. This dependency is preventing us from upgrading Django.

This is an intermediate step that moves us towards either adopting the built-in tasks framework in Django 6 (https://docs.djangoproject.com/en/6.1/topics/tasks/) or moving to an external service.

> django-background-tasks is under active development as of a couple months ago. If it has issues on Django 5 then let's fix those issues.

Can you point me to this? I believe you, but I can't find it. It's pretty confusing with all of the different "django-background-task", "django-background-tasks", django-background-tasks-updated" stuff going on.

tompollard:
Thanks @lukepayyapilli, those three points have been addressed:                                 

**1+5: Notifications never fire**. Added `ack_failures: True` and removed the guard.
                                                                                                                                       
**Stale locks on restart**. Added manage.py `unlock_q2_tasks` (resets lock to now for stuck rows) in both service files, same pattern as the legacy command.                                                                                           
                                                                            
``Legacy remove_existing``. Switched to value-based comparison of parsed `task_params` and added `locked_at__isnull=True`.

lukepayyapilli (commented):
Thanks, `ack_failures` and `unlock_q2_tasks` both do the job. One thing the first one changed that I don't think was intended, plus two small notes.

### `ack_failures: True` removes retries entirely

It fixes the notification, but with `ack_failures: True` monitor acknowledges on failure, and `acknowledge()` on the ORM broker is a delete (`brokers/orm.py:93`). So a failed task is removed from the queue immediately and never retried. `max_attempts: 5` is now dead config that misdescribes the behavior.

That's a change from legacy, which did retry. #2601 is a transient SSL write timeout, exactly the case retries exist for, and its email reads "has made 5 attempts and failed". This converts at-least-once into at-most-once.

Two ways out:

- Accept at-most-once: drop `max_attempts` from `Q_CLUSTER` so it doesn't imply retries, and note in the deploy docs that a failed task needs re-triggering by hand.
- Keep retries: now that `unlock_q2_tasks` exists, `retry` no longer has to be infinite for restart safety. A finite `retry` (86400, say) with `ack_failures: False` and the `attempt_count` guard restored gets legacy's behavior back. `timeout` can stay effectively infinite, it's independent.

I'd lean toward the second, since it matches what admins currently expect from the failure emails, but either is fine as long as it's deliberate.

### Value-set comparison is fragile

Works today because every production call site passes a single int, but:

```python
>>> set((1, True)) == set((1,))
True
```

`set()` collapses `1` and `True`, ignores order, and raises on an unhashable arg. Since `associated_task` already does the `inspect` work, binding both sides through `inspect.signature(real_func).bind(*args, **kwargs).arguments` and comparing those dicts would match exactly, regardless of positional vs keyword style.

### `unlock_q2_tasks` nits

Mechanically correct: `dequeue()` filters `lock__lt=now`, and `ExecStartPre` runs before `ExecStart`, so the clock advances enough between them.

- It unlocks rows that a *running* cluster is processing, so it bakes in #2608's stated "assume there is only one service" assumption. Worth saying so in the docstring, since it would be silent corruption if a second cluster were ever added.
- Legacy's equivalent lives in `user/management/commands/`, this one is in `console/`. Cosmetic, but they are a pair.

lukepayyapilli:
@tompollard @bemoody on the packaging question, the naming really is a mess. Here is what the metadata says.

We pin `django-background-tasks-updated = "=1.2.7"`, released 2022-03-31.

That fork (`iamjonmiller/django-background-tasks`) was merged back into upstream via django-background-tasks#300, "Support for Django 4.0+ and Python 3.12", merged 2024-08-16, and the fork was archived around 2024-08-01. So the "-updated" line did not die, it was absorbed.

The successor is `django-background-tasks==1.2.8`, released 2024-08-16. It declares `Framework :: Django :: 5.0` and `Requires-Dist: Django>=4.0`, and the deprecated APIs that prompted #1647 (`providing_args`, `ugettext`, `force_text`) are all gone from it.

The upstream repo is not archived and has 623 stars, but its last default-branch commit is 2024-08-28 and there has been no release since, with 99 open issues. The repo's `pushed_at` reads 2025-06-19, which is branch activity rather than commits, and I suspect that is what looked like recent development.

So the PR description's "archived in June 2025" refers to the fork (archived around Aug 2024), and "active development as of a couple of months ago" is not quite right either.

The practical upshot: we have never tested the current package. We are four years behind on a fork that was folded upstream two years ago. Bumping to `django-background-tasks==1.2.8` and running the suite against Django 5.2 is a cheap experiment, and it would settle whether this migration is needed at all before more time goes into it.

One roadmap note, since Django 6's tasks framework came up. Its reference implementation is `django-tasks` by RealOrangeOne (0.12.0, Feb 2026, still actively released). django-q2 is not that, so it is a detour relative to that endpoint. That does not make it the wrong call if 1.2.8 turns out not to work on Django 5.2, but it is worth being explicit that the intermediate step has its own cost.


---

## Update: I tested this rather than trusting the classifiers

I built clean venvs on Python 3.11 (the version we target) with Django 5.2.17, one per package version, and ran both.

| Test | 1.2.7 (what we pin) | 1.2.8 |
|---|---|---|
| Library's own 55-test suite | OK | OK |
| `RemovedInDjango` warnings under `-Wa` | 0 | 0 |
| `process_tasks` end to end (queue a task, drain it) | ran the task | ran the task |
| `autodiscover()` on Python 3.13 | `ModuleNotFoundError: No module named 'imp'` | OK |
| Django requirement | `>=3.2.12`, no upper bound | `>=4.0`, no upper bound |

The entire 1.2.7 to 1.2.8 diff is about fifteen lines, almost all of it whitespace and the version string. The single functional change is in `autodiscover()`:

```python
# 1.2.7
import imp
app_path = import_module(app).__path__
imp.find_module('tasks', app_path)
import_module("%s.tasks" % app)

# 1.2.8
import_module("%s.tasks" % app)
```

Three things follow:

1. **There is no Django 5.2 problem**, in 1.2.8 or in the 2022 release we currently pin. Both pass their full suite on 5.2.17 with zero deprecation warnings, and the daemon queues and runs a task end to end.
2. **The actual constraint is Python 3.12+**, where `imp` was removed. That is what 1.2.8 fixes, and it is a one-line bump.
3. Neither version caps Django, and `pyproject.toml` has no direct Django pin, so nothing blocks the resolver either.

Worth noting separately: that `imp.find_module` call is #2357. @bemoody diagnosed it correctly at the time ("Is it a bug in Python, that `imp.find_module` can't cope with path objects produced by `importlib.import_module`?"). 1.2.8 deletes the code outright, so the bump would likely close #2357 properly instead of relying on the `__init__.py` workaround.

tompollard:
It's pretty messy. 

https://pypi.org/project/django-background-tasks-updated/ say that: "This code has been merged into the original repository and the original package is now updated." 

But I'm not sure this is true. It looks like they opened a PR that was not merged: https://github.com/django-background-tasks/django-background-tasks/pull/308

https://pypi.org/project/django-background-tasks-updated/ does say that it supports Django 5, so we could stick with it for an upgrade to 5 I guess, and then revisit for move to 6.

tompollard:
Going to close this for now, try to get us to Django 5, then revisit.

PR #2710, opened 2026-09-15
Remove USE_L10N setting (default since Django 4.0)

tompollard:
`USE_L10N = True` has been the default since Django 4.0, and the setting is removed entirely in Django 5.0. Dropping it now is forward-compatible with the upcoming Django 5.2 upgrade.


tompollard:
Relates to https://github.com/MIT-LCP/physionet-build/pull/2667

bemoody (approved):

Issue #2722, opened 2026-09-16
Allow users to sign up as a peer reviewer


tompollard:
We have said that we will establish a pool of peer reviewers who can assist with review of project submissions.

We should allow users to sign up as a reviewer. As a first step, we could address this by (1) adding a reviewer flag to the user/profile model and (2) providing a sign up option in the user settings. 

I think it would be best to make credentialing a requirement for reviewers.

Issue #2338, opened 2025-01-31
Add communication/newsletter (or similar) page to settings


tompollard:
It would be good to add a page to the user settings that allows people to sign up to receive communication from us.

Needs some thought on how the page is organised and what kind of communication people can sign up for. Possibly just emails?

Some people might sign up for a regular newsletter. Some people might sign up for announcements about new datasets, etc.

PR #2723, opened 2026-09-16
Add newsletter and reviewer volunteer opt-in to Communications settings

tompollard:
As part of current funding, we have said that we will (1) create an opt-in newsletter and (2) create a pool of reviewers.

I played around with various options for addressing these points (initially dealing with them independently). In the end, the cleanest approach that I could was to combine both options in a new "Communications" tab.

This would address https://github.com/MIT-LCP/physionet-build/issues/2722 and https://github.com/MIT-LCP/physionet-build/issues/2338. Not perfect, but it moves us in the right direction. I'd be happy to receive better suggestions!


lukepayyapilli:
> I'd be happy to receive better suggestions!

Taking you up on that. The short version: I think this is two features, and one of them already exists on `dev`.

## The reviewer half

We already have a reviewer subsystem:

- `Reviewer` is an existing auth group (`user/fixtures/demo-user.json:14554`), alongside `Managing Editor`, `Handling Editor` and `Credentialer`
- `ReviewerInvitation` and `ExternalReview` in `project/modelcomponents/review.py` (migration `project/0098_add_review_models.py`)
- `SubmissionStatus.NEEDS_REVIEWER_ASSIGNMENT` and the reviewer-assignment queues in `console/views.py:201` and `:269`
- `InviteReviewerForm` in `console/forms.py:1124`, tested in `console/test_views.py:1695`

And here is the gap I think this PR is really aiming at. `InviteReviewerForm` takes a raw email address and rejects it unless an account already exists:

```python
reviewer_email = forms.EmailField(label='Reviewer email', ...)
...
user = User.objects.get(email=email)
except User.DoesNotExist:
    raise forms.ValidationError('No PhysioNet account found with this email address.')
```

So an editor has to already know who they want and which address they registered with. "Where do I find candidate reviewers?" is the unsolved problem, and it is unsolved right at this form. A flag on `Profile` that nothing queries does not reach it.

Suggestion: land the volunteer flag together with the thing that consumes it. Swap `reviewer_email` for a DAL autocomplete scoped to volunteers. We already do this twice in the same file (`console/forms.py:785` and `:1028` use `autocomplete.ModelSelect2(url='user-autocomplete')`, with `UserAutocomplete` at `console/urls.py:129`), so a `ReviewerAutocomplete` filtered on the flag is close to a copy of an existing pattern. That is the difference between collecting opt-ins and having a pool an editor can actually draw from.

Two things that follow from wiring it up:

- I would drop the credentialing lock on the checkbox and let anyone volunteer, then filter by credentialing in the autocomplete queryset. Credentialing is an editorial criterion rather than a reason to stop someone raising their hand. It also sidesteps a bug in the current form: `disabled` fields read from initial rather than POST, so a user whose credentialing is later revoked (`user/models.py:1096`) keeps `review_volunteer=True` and cannot uncheck it, while the view still reports "Your communication preferences have been updated."
- If volunteers can pick topics (`Topic` at `project/modelcomponents/metadata.py:708`), editors can filter by subject, which is what makes a pool usable once it is more than a handful of people.

## The newsletter half

I would hang the subscription off `AssociatedEmail` rather than `Profile`. That model already carries `email`, `is_primary_email`, `is_verified`, `verification_token`, `added_date` and `verification_date` (`user/models.py:550`), which answers three questions the current design leaves open:

- **Which address gets the mail?** The flagged one. On `Profile` it is ambiguous, since users can have several.
- **Can we mail it?** Filter on `is_verified`. Mailing unverified addresses is how we end up on a blocklist.
- **How do people unsubscribe?** `verify_email(request, uidb64, token)` (`user/views.py:936`) is an existing token-based flow that needs no login, and an unsubscribe link can mirror it. A newsletter you can only leave by logging in is a problem we would have to fix later anyway.

Worth adding `newsletter_subscribed_at` at the same time. One column now, versus reconstructing consent history later. Both points are cheapest to settle before `0073`/`0074` land, since moving the field afterwards means a data migration.

## What I would keep

The Communications tab is the right instinct, it just holds one thing. A newsletter preference is genuinely a communication preference. Volunteering to review is role membership, and its natural neighbours are Credentialing and Training.

For what it is worth, I think the awkwardness you hit combining them is the same awkwardness that pushed the `disabled` logic into a communications form. The two opt-ins are different kinds of object.

## Suggested sequencing

1. Newsletter on its own, on `AssociatedEmail`, with timestamp and token unsubscribe. Closes #2722.
2. Reviewer volunteering on its own, landing the flag and the `InviteReviewerForm` autocomplete together so the pool is usable when it ships. Closes #2338.
3. Both flags added to `generate_user_csv_data` (`console/views.py:2692`) and a `list_filter` on the Profile admin, so the data has a way out. Right now the only way to answer "send me the newsletter list" is a shell on production.

Happy to be argued out of any of this, particularly the `AssociatedEmail` move.

tompollard:
Thanks Luke, this is very helpful review. I'm going to split into separate PRs, which I should have done initially. 

Using a permission group to handle reviewers is an interesting suggestion. This will probably require a clean up of our permissions first.

PR #2726, opened 2026-09-17
Add host permission management to user console

Chrystinne:
Adds a Host checkbox to the user management page, allowing authorized users to grant or remove the host group.

The host group grants access to the Events menu and event creation. Includes coverage for granting and removing the permission.

This is what we will see when we try to grant a user Host permission (Users → Active Users):

image


tompollard:
Thanks Chrystinne, I think this works for now. We need to revisit permissions and their management later. Please could you fix the formatting so that the checkbox is aligned with the content?

tompollard:
@Chrystinne see this PR from Michael: https://github.com/MIT-LCP/physionet-build/pull/2123/changes. Which approach is best?

Chrystinne:
@tompollard thanks, Tom. I fixed the formatting. Could you please approve it?

image

tompollard:
We don't really want custom CSS just for this one checkbox if we can avoid it. I think you can remove the css changes and use a plain checkbox:

```                                                                                                    
  <div class="mb-2">                                                                                                                             
    <label><input type="checkbox" name="host" id="host-permission"{% if has_host_permission %} checked{% endif %}> Host</label>                  
  </div>
```

tompollard (approved):
Looks good, thanks.

tompollard (approved):

PR #2727, opened 2026-09-17
Add confirmation page for GET requests to /logout/

tompollard:
Add confirmation page for GET requests to /logout/
    
Django 5 removed GET-based logout, so visiting /logout/ directly would return a 405 error. Override LogoutView.get() to render a confirmation page with a POST form button instead.
    
This is a migration to help for the Django 4 to 5 upgrade. Users or external sites may have existing links or bookmarks pointing to /logout/ as a GET request. This page can be removed once the transition is complete.


tompollard:
This seems like a pretty narrow concern that we are addressing here. Unlikely anyone has bookmarked logout. 

I think we are only talking about someone who has a stale browser tab open before the deploy and then they hit logout without interacting elsewhere with the site (which would trigger a reload and the new navbar)?

bemoody (approved):
Looks good, thanks!

PR #2730, opened 2026-09-22
Allow GET requests in LogoutView for Django 5 compatibility

tompollard:
Django 5's LogoutView restricts http_method_names to ['post', 'options'], so dispatch() rejects GET with 405 before our get() override runs. Explicitly include 'get' in http_method_names.


briangow (approved):
Looks fine to me @tompollard.

